PR #204, opened 2015-01-27
Add About section to README.md

lyzidiamond:
@tmcw @morganherlocker I think this does a good job of introducing turf -- just the same text from [/static/docs](https://turfjs.org/static/docs). Merge if you approve.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,38 +10,7 @@
 
 - - -
 
-##About
-
-###What is Turf?
-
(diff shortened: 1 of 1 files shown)


tmcw:
:+1: looks good! Eventually we might want to make these the same document, but happy to punt on that because I expect the markdown parsing will be a lil different and it's good to have this inline right now.

lyzidiamond:
I see what you did here, @morganherlocker, and I like it!

morganherlocker:
Made some minor edits:

- removed "About" heading, since it seems redundant in this context
- moved installation above data section -- mostly just to follow standard readme ordering

lyzidiamond:
Awesome. Merging.

PR #210, opened 2015-02-04
update deps and harmonize turf-tin

morganherlocker:


lyzidiamond:
:+1: 1.4.0 ftw

morganherlocker:
@lyzidiamond fyi, ended up naming it 1.3.5

lyzidiamond:
@morganherlocker That is also good. :dancer:

PR #235, opened 2015-03-06
Wrong example due to API change

ThomasG77:
Due to API change, turf.point need as parameter an array with coordinates x, y and not two parameters x and y.


tmcw:
:+1: thanks!

PR #276, opened 2015-08-14
fix broken readme badges

ngoldman:
Hello Turf team!

Your version badge is broken!

screen shot 2015-08-14 at 10 45 46 am

I fixed it!

screen shot 2015-08-14 at 10 46 04 am

I've found http://shields.io to be pretty reliable.

Also noticed the build status for turf-erase was broken, and following the link it looks like that was renamed to turf-difference, so I fixed that too.

:grin:


tmcw:
:+1: awesome, thanks!

ngoldman:
:tophat: 
:panda_face: 
:necktie: :+1: 
:jeans:

PR #282, opened 2015-09-25
Update turf.js

Vic-thor:
fix compilation error (with YUI Compressor) with reserved word


tmcw:
turf.js is a generated file. please figure out which turf module the reserved word resides in and fix the problem at the source.

Issue #306, opened 2016-02-13
turf-next proposal


tmcw:
_opinion warning: this is opinion_

Turf 3.0 should

* Switch to a uni-repo method. All core modules will be in a single repo. Addons will be managed as such
* ~~Turf will have a global dependency to lodash 4.x, which will help with some of the fp stuff but more importantly with immutability~~
* Turf packages will still be bundled as split-out packages, like lodash modules or babel modules, and we'll take advantage of [lerna](https://github.com/kittens/lerna) to do that.
* ~~Turf support will be moved 100% to gis stackexchange or stackoverflow and off of github. This leaves GH as only the place for technical discussion~~

cc @Turfjs/owners

mourner:
Support everything except the lodash dependency, which I find unnecessary.

tmcw:
@mourner i'm fine with an alternative to lodash: mainly I'm excited by the [inclusion of lodash-fp](https://github.com/lodash/lodash/releases/tag/4.0.0), which provides immutable methods. I think turf's mutability problems need a generalized solution. lodash is one, there are surely others.

anandthakker:
Re: immutability -- I've been trying out [seamless-immutable](https://github.com/rtfeldman/seamless-immutable) in a recent project, and I've been 80% happy with it, but my guess is that there would be perf problems out of the box

mourner:
@tmcw although I don't have a lot of experience with immutability to have a weighted opinion here, I feel like immutability should be more of convention rather than a library you depend on (especially a bloated one like lodash) , at least in libraries like Turf as opposed to end-user applications. 

I like the approach of seamless-immutable — it basically throws helpful exceptions when you mutate objects/array that should be immutable, but you can switch to "production build" which doesn't do that and performs much better.

anandthakker:
> I like the approach of seamless-immutable — it basically throws helpful exceptions when you mutate objects/array that should be immutable, but you can switch to "production build" which doesn't do that and performs much better.

@mourner Yeah I like this approach too.  My concern using the actual lib with turf is that for each turf function you might have to do a deep traversal on input (to make input geojson immutable) and again on output (to make it back into a plain object), which I'd think would be noticeable overhead.

I guess that overhead could be avoided if turf were to return `seamless-immutable` objects, which look and act just like plain objects (especially using the production build), except for having a private, non-enumerable `__immutable_invariants_hold` property on them that's used internally by `seamless-immutable`.

tmcw:
[...]

PR #311, opened 2016-02-23
Unirepo

### LINKED ISSUE #162: master test runner
1. clone every submodule into a tmp directory
2. npm i
3. npm t
4. capture results so things blow up if any of the packages fail their tests

This will be kind of slow, but I think it makes sense to have this run from `npm test` every time. This will prevent subtle dependency update bugs that can happen when all the modules are working together.

### LINKED ISSUE #191: versioning for releases / development
thoughts on how to do this:

* do we want to pin dependencies for turf versions?
* how do we make sure that dependencies of modules don't require duplicate versions of the same module?

### LINKED ISSUE #232: sudo: false in all .travis.yml
Probably none of the turf repositories need to use sudo on Travis, so we should add `sudo: false` to all .travis.ymls for faster builds.

http://docs.travis-ci.com/user/workers/container-based-infrastructure/

tmcw:
* [x] Switch to a uni-repo method. All core modules will be in a single repo. Addons will be managed as such
* [x] Turf packages will still be bundled as split-out packages, like lodash modules or babel modules, and we'll take advantage of lerna to do that.
* [x] geometry-helpers these modules should be one module
* [x] `npm test` should run all package tests

Things this resolves

* Fixes #162 (makes tests easy to run)
* Fixes #232 (switches to circleci)
* Fixes #191 (lerna handles versioning)

Refs #306


package.json
```diff
@@ -128,6 +128,7 @@
     "eslint": "^0.15.0",
     "hcat": "0.0.5",
     "jsdoc": "^3.3.0-beta1",
+    "lerna": "https://github.com/kittens/lerna/archive/master.tar.gz",
```
morganherlocker: @tmcw Can we depend on lerna through npm? 

[babel reference](https://github.com/babel/babel/blob/master/package.json#L24)

tmcw: Yes, but the published version doesn't have `lerna run` yet

tmcw:
Okay all you @morganherlocker to push the button

migurski:
The internet salutes you.

maxogden:
was there a discussion on this? just curious to learn more about what drove the decision to go unirepo

tmcw:
Yep, discussion is at https://github.com/Turfjs/turf/issues/306 - coordinated multi-repo improvements like adding eslint and running tests were rough, and managing releases of all modules to coordinate with a release never found a solution - while the lerna/etc approach has been battle-tested

PR #341, opened 2016-03-31
index shifts with -1 for all the points after the first

DenislavD:


morganherlocker:
Thanks @DenislavD! This exact change was already merged to master in another PR.

Issue #365, opened 2016-04-26
turf-line-slice-along


mourner:
I've noticed that in a vast majority of cases, `turf-line-slice` is used in combination with `turf-along` to get a part of a line between specific distances, e.g. from 100 meters to 200. This is very inefficient though, because you have to do expensive intersection routines for both points while you already know that they're on the line.

So in `cheap-ruler`, I've implemented `lineSliceAlong` method that does this, and it's 268 times faster than the corresponding Turf calls combination. It's working great and I think we should make such a package in Turf as well, or make `turf-line-slice` accept distances in addition to points and take a different code path if so.

cc @morganherlocker @tmcw

morganherlocker:
:+1:  lineSliceAlong sounds like a great idea.

jvrousseau:
I've been playing around with implementation of lineSliceAlong. Before I incorporate it into the monorepo, I wanted to run the code by you all to ensure it was doing solving the needs of the `turf-line-slice/turf-along` combination use cases.

Here's the code: https://github.com/jvrousseau/turf-line-slice-along

If there are any changes I need to make, please let me know. Otherwise, I will add it to the turf monorepo and create a PR.

Thanks!

cc @morganherlocker @mourner

PR #376, opened 2016-05-17
Fix turf-along using raw geometry

jvrousseau:
I am looking at implementing #365 and found that turf-along throws an exception when passing in a raw LineString geometry.

The PR sets the coordinate array as line.coordinates (rather than line.geometry.coordiantes). The PR also updates the test suite to pepper in some raw geometry while testing turf-along.

If there are anything changes I need to make, please let me know.

Thanks!


tmcw:
👍 thanks!

tmcw:
Published in turf-along 3.0.4

PR #383, opened 2016-05-27
Detect straggling inter-dependencies

tmcw:


tmcw:
This is a pretty high priority; we have some modules that never made the jump from a turf interpedency from 1 to 3. Mainly turf-isolines. But I don't have the context for the changes behind the grids and thus going to punt on this and hope @morganherlocker gets the time to look at it at some point; the turf-grid to turf-point-grid api swap needs to happen in turf-isolines too

Issue #388, opened 2016-06-01
turf-point-on-line incorrect snapping


morganherlocker:
@camilleanne and I have isolated some failing cases for turf-point-on-line. The issue is that a point in the middle of two line vertices is snapping to the beginning of the line like so:

<img width="654" alt="screenshot 2016-05-31 17 47 04" src="https://cloud.githubusercontent.com/assets/814300/15695141/ff835ac4-2758-11e6-92fa-83e045a82ec1.png">

I'm not sure exactly what the cause is yet, but my guess is that the line intersection algorithm is flubbing when the point is exactly on the line, or possibly just to the right or left of the line due to float precision.

morganherlocker:
The [latest implementation](https://github.com/Turfjs/turf/issues/389) has been fixed for this edgecase and tested.

PR #389, opened 2016-06-01
better tests for turf-point-on-line

morganherlocker:
This PR updates test for turf-point-on-line to verify several bugs have been handled around [line snapping edge cases](https://github.com/Turfjs/turf/issues/388).


packages/turf-point-on-line/test/test.js
```diff
@@ -0,0 +1,102 @@
...
+var lineDistance = require('turf-line-distance');
+var along = require('turf-along');
+
+test('turf-point-on-line - first point', function (t) {
+    var line = linestring([[-122.45717525482178,37.72003306385638],[-122.45717525482178,37.718242366859215]]);
+    var pt = point([-122.45717525482178,37.72003306385638]);
+
+    var snapped = pointOnLine(line, pt);
+
+    t.equal(pt.geometry.coordinates.join(','), snapped.geometry.coordinates.join(','), 'pt on start does not move');
```
tmcw: why does this join the points instead of using t.deepEqual?

PR #394, opened 2016-06-02
expose all geometry and feature helpers - fixes #393

### LINKED ISSUE #393: add all helpers to turf.* ?
@tmcw We currently give access to some of the [turf feature geometry functions](https://github.com/Turfjs/turf/blob/master/packages/turf/index.js#L56-L61). Should we expose the rest of the turf-helper functions (or am I missing something and they are already exposed somewhere)?

morganherlocker:
cc @tmcw @tcql

### DIFF
--- a/packages/turf/index.js
+++ b/packages/turf/index.js
@@ -53,9 +53,14 @@ module.exports = {
     hexGrid: require('turf-hex-grid')
 };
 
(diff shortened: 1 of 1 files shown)


tmcw:
👍 lgtm

Issue #399, opened 2016-06-02
How to make a patch


morganherlocker:
The recommended publishing docs from [CONTRIBUTING.md](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#publishing) are a bit incomplete. Let's try to make a list of steps for publishing a patch.

Here's an incomplete starting point:
1. Make a PR with your changes (be sure to modify relevant jsdocs to reflect the change)
2. Merge PR after a code review from someone else in the Turfjs organization
3. decide which version increment to use (`major|minor|patch`) based on [semver rules](http://semver.org/#summary)
4. `lerna publish` (need details)
5. Add list of changes with the new version number to the top of CHANGELOG.md
6. generate new docs (need details)
7. publish docs (need details)
### ?s
- `lerna publish --canary` (suggested by our current docs) appears to be broken or deprecated by lerna (`"Too many arguments."`). 
- `lerna publish` asks for a VERSION file. It seems to default to `0.0.0`.
- `turf/package.json` has a version and `turf/packages/turf/package.json` has a version. Should we delete `turf/package.json` or modify it somehow?
- how are docs generated? how are they published?

cc @tmcw 

![](https://lesterandcharlie.files.wordpress.com/2012/02/im-just-a-bill.jpg)

tmcw:
(quick notes, gotta run in a moment)

> lerna --publish

This is still documented on their official repo.

https://github.com/lerna/lerna#--canary--c

Lerna has been a little weird about perferring global to local installs. That might be part of the cause.

turf/package.json is where i've been managing global dependencies and global tests: see the [stragglers pr for instance](https://github.com/Turfjs/turf/pull/383) - we can use it to run assertions over all the package.jsons

> how are docs generated? how are they published?

in turf-www, by running `npm run docs` https://github.com/Turfjs/turf-www/blob/gh-pages/package.json - this is still imperfect and we still need a fancy template

DenisCarriere:
I believe there's enough documentation on how to publish a new patch/minor release on [the official Lerna GitHub repo](https://github.com/lerna/lerna).

PR #408, opened 2016-06-09
Upgrade jsts dependency to 1.1.2

kozze89:
I need a fix that is in jsts `1.1.2` (https://github.com/bjornharrtell/jsts/commit/7ec80d5c0f7f0fa2c5ac0c8ca54ebfaa91e71180)

This patch upgrades the jsts dependency for the 4 turf-packages that use it. Also upgrading the turf version in lerna to `3.0.6`.

That said, I have no idea if I've done this patch the right way. Never used lerna before and the contributing guidelines don't say much, although they seem to be in the works (https://github.com/Turfjs/turf/issues/399).

Please let me know if you need me to do something differently.


tmcw:
👍 Thanks!

kozze89:
@tmcw Thank you!
When can I expect to see this in npmjs?

PR #411, opened 2016-06-13
fix documentation and examples of feature creators

tyrasd:
show the correct number of nested arrays in the polygon and multi* feature creator helper functions


tmcw:
👍 test failure is unrelated, I'll fix separately. Thanks for the fix!

Issue #415, opened 2016-06-15
use turf-point-on-line in turf-line-slice 


morganherlocker:
https://github.com/Turfjs/turf/blob/master/packages/turf-line-slice/index.js#L85-L145

Currently, turf-line-slice vendors its own version of turf-point-on-line, which is outdated. Instead, it should use the module'd version of this function.

cc @camilleanne

PR #417, opened 2016-06-16
Fix line slice

camilleanne:
Fixes https://github.com/Turfjs/turf/issues/415 by removing a version of `pointOnLine` and requiring the module'd version of the function instead.

Also fixes a small bug in `turf-point-on-line` where the coordinates were not being set correctly if the object was a `LineString` instead of a complete `Feature`.

cc @morganherlocker


morganherlocker:
👍 Looks great

morganherlocker:
Added test case:

screenshot 2016-06-17 11 49 17

PR #420, opened 2016-06-25
Turf Aim

mstdokumaci:
Calculates an impact point for a given projectile and target moving at different constant speeds.

### DIFF
--- a/packages/turf-aim/.npmignore
+++ b/packages/turf-aim/.npmignore
@@ -1,0 +1,2 @@
+test
+coverage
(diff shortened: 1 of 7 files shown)


tmcw:
@morganherlocker what do you think about the applicability of this for core?

morganherlocker:
Nice work @mstdokumaci. Unfortunately, I think this module may be a bit out of scope for the project. I could be convinced otherwise, but I'm hesitant to creep into "physics engine" territory here.

PR #431, opened 2016-07-07
[Fix] Fix a crash of turf-convex when the resulting hull has only 1 or 2 coordinates.

aurelien-defossez:
It can happen for instance when the input feature collection contains only a point, the resulting hull has only one point, which makes the turf-polygon algorithm crash because it needs at least 3 different vertices.


tmcw:
👍 awesome, thanks!

PR #432, opened 2016-07-08
Units parameter added on pointOnLine method.

aurelien-defossez:
The default value is "miles", which was the previous hard-coded value.

This is useful because pointOnLine returns the distance between the point and its projection on the line, but the distance is always in miles and is not configurable.


tmcw:
👍 thanks!

tmcw:
Published in v3.1.0

PR #433, opened 2016-07-12
meta coordEach method now handles nested GeometryCollections.

aurelien-defossez:
Example:
```
turf.explode(turf.geometryCollection([
	turf.geometryCollection([
		turf.point([0, 0]).geometry,
		turf.point([1, 1]).geometry
	]).geometry,
	turf.lineString([[2, 2], [3, 2]]).geometry
]));
```

Before, this code would have sent an error, but now the feature is correctly exploded.

### DIFF
--- a/packages/turf-meta/index.js
+++ b/packages/turf-meta/index.js
@@ -62,6 +62,9 @@ function coordEach(layer, callback, excludeWrapCoord) {
                     for (k = 0; k < coords[j].length; k++)
                         for (l = 0; l < coords[j][k].length - wrapShrink; l++)
                             callback(coords[j][k][l]);
+            } else if (geometry.type === 'GeometryCollection') {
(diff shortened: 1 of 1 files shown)


tmcw:
Nested geometry collections are not recommended in the [GeoJSON IETF draft](https://tools.ietf.org/html/draft-ietf-geojson-03#appendix-A.7), but they aren't forbidden, and the pre-IETF spec doesn't have any recommendation about nesting, so 👍 from me.

> To maximize interoperability implementations SHOULD avoid nested geometry collections.

aurelien-defossez:
Yeah, it's not supposed to exist, but I already saw this case in a geojson file I received.

Before really working on these data, I clean them and remove nested GCs and invalid geometries, but sometimes I need to work on the raw data.

Thanks!

PR #434, opened 2016-07-13
Modified the turf module.exports to fix the auto-completion of some IDEs (e.g. WebStorm)

aurelien-defossez:
The way the module.exports was built seemed to break the auto-completion of WebStorm.
By moving everything inside one declaration, it works like a charm.


tmcw:
👍 excellent, thanks!

PR #435, opened 2016-07-13
Union can now take a variable amount of polygons as arguments, instead of only 2.

aurelien-defossez:
It's now possible to do:
`turf.union(poly1, poly2, poly3, poly4);`


tmcw:
👍 awesome, can you add a testcase to confirm this functionality?

aurelien-defossez:
Done, I added a test with 3 polygons merged into a single one.

tmcw:
👍 awesome, thanks!

tmcw:
Published in turf@3.2.0

PR #438, opened 2016-07-15
Remove the Z coordinates in turf-convex because it breaks the convexHull algorithm

aurelien-defossez:
Now turf-convex ignores Z coordinates.

I added a test for this case.


tmcw:
👍 Thanks!

PR #439, opened 2016-07-15
New module: center-of-mass

aurelien-defossez:
Center of mass computes the center of mass of any feature.

The difference between centroid and center-of-mass is that centroid returns the average position of all the points, which can go towards where there are the most points, whereas center-of-mass ignores the number of points and computes a point more in the middle of the points.

Example:
```
 *********

 *       *

 *   *   *
```

Center of mass is in the middle of the points:
```
 *********

 *   c   *

 *   *   *
```

Centroid is a bit higher because there are more points to the north
```
 *********
     c
 *       *

 *   *   *
```


aurelien-defossez:
Please wait a bit before merging this one, I thought of more features to add to it (like supporting FCs) before merging it, it'll be done on monday.

aurelien-defossez:
Now the module is finished and you can merge it. :)

tmcw:
👍 awesome, thanks!

PR #442, opened 2016-07-21
Updated the turf-convex doc to tell that Features are accepted

aurelien-defossez:
turf.convex accepts FeatureCollections and any type of Feature, but the doc says that only FCs of Points are accepted.

Also added a test in which turf-convex works on a Polygon feature.


packages/turf-center-of-mass/README.md
```diff
@@ -9,7 +9,8 @@ using this formula: [Centroid of Polygon](https://en.wikipedia.org/wiki/Centroid
 
 **Parameters**
 
--   `feature` **[Feature](http://geojson.org/geojson-spec.html#feature-objects)** the feature
```
tmcw: Are these changes hand-written? The readmes are auto-generated with the generate-readmes script https://github.com/Turfjs/turf/blob/master/scripts/generate-readmes Running `./scripts/generate-readmes` may overwrite these cahnges

aurelien-defossez:
It was hand-written, I did not know this script. I ran it in the last commit to update the README.md files correctly.

Thanks!

tmcw:
👍 Thanks, merging!

PR #443, opened 2016-07-25
Expose turf.circle on index.

aurelien-defossez:
turf.circle is not exposed in index, and what's odd too is that turf.circle is that turf.circle is classified with the "assertions" in the [documentation](http://turfjs.org/docs/#circle).


tmcw:
👍 Thanks!

PR #447, opened 2016-08-03
Fix issues with turf-idw

tmcw:
Previously:

* Was not scoped
* Did not rely on scoped packages
* Had a logic bug with filtering
* Used deprecated filter module
* Wasn't included in package.json

### DIFF
--- a/packages/turf-idw/README.md
+++ b/packages/turf-idw/README.md
@@ -1,60 +1,39 @@
-# Inverse Distance Weighting (I.D.W.)
+# @turf/idw
 
-Inverse Distance Weighting (IDW) is a type of deterministic, nonlinear interpolation method over a set of points.
+# index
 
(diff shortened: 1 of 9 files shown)


miccferr:
Hi  @tmcw @morganherlocker ,
I believe commit fefceb75b0f65ea6ce391d83b2870cffd4050a7c introduced a small regression in the nested loop (should iterate on `controlPoints.features[j]` not on [`controlPoints[j]`](https://github.com/Turfjs/turf/blob/master/packages/turf-idw/index.js#L35) -- see also subsequent lines). As a result the interpolated value `z` is no longer computed. 
I can open a PR if you want.

morganherlocker:
@miccferr a PR would be appreciated 🙇

miccferr:
Sure thing!
see PR #461 👍

PR #450, opened 2016-08-09
Reword opening sentence

davidtheclark:
I had more difficulty understanding the opening sentence about this function than the rest of the paragraph. I tried rewording it — not perfect but maybe a little more clear? (Only the first sentence is changed: the rest of the changes are about line-length limiting.)


tmcw:
Can you edit at turf-collect/index.js? The readmes are generated.

PR #451, opened 2016-08-09
Typo fix in README

davidtheclark:


tmcw:
These readmes are autogenerated, can you fix at https://github.com/Turfjs/turf/blob/master/packages/turf-area/index.js#L4 ? Thanks

PR #453, opened 2016-08-09
Reword opening of doc-comment in turf-collect

davidtheclark:
Replacement for #450.

I found the opening sentence hard to understand. I don't feel 100% about this attempt at rewriting but I think it may be an improvement on the original?


tmcw:
👍

PR #458, opened 2016-08-11
Test-cases for intersect with weird outcomes

davidtheclark:
I am seeing some weird results from `turf.intersect`. I am not sure whether this is a bug in the algorithm or a shortcoming in my understanding of what it is and how to use it. So I created a few test cases and took some screenshots to illustrate what I'm seeing. I commented out the results that I'm getting from turf so they'd be in there for reference. If turf is wrong, I'm not sure how to get what the _accurate_ result should be, so there are no assertions in the tests yet.

Essentially, it looks like the intersection outcome contains inaccurate latitudes when features are large.

The Polygon test is how I first ran into this problem. I have a square and a triangle, and I want to create a new shape that is the portion of the original triangle contained within the square. In the screenshot you'll see that the new shape hits the square's edge at a different latitude than the original triangle.

screen shot 2016-08-11 at 10 55 22 am

I tried to simplify this case by using LineStrings instead of Polygons. I just took the relevant lines from the Polygons and found their intersection. The Point that `turf.intersect` returns is clearly not the actual intersection.

screen shot 2016-08-11 at 10 47 52 am

However, when I did the same thing _on a small scale_, the Point _appears_ to be an accurate intersection.

screen shot 2016-08-11 at 10 58 31 am

I am assuming that I can't just do simple Cartesian-coordinate math to figure out what the intersection points of these lines _should_ be, because projection needs to be accounted for in a way that I don't really understand yet.

What do you think of these cases and the apparent problem?

cc @samanpwbb @scothis

### DIFF
--- a/packages/turf-intersect/test/test.js
+++ b/packages/turf-intersect/test/test.js
@@ -31,3 +31,152 @@ test('intersect -- no overlap', function(t){
   t.deepEqual(output, undefined);
   t.end();
 });
+
+test('intersect -- polygons', function(t) {
+  var square = {
+    "type": "Feature",
+    "properties": {},
+    "geometry": {
+      "type": "Polygon",
+      "coordinates": [
+        [
+          [-10, 24],
+          [-10, 64],
+          [50, 64],
+          [50, 24],
+          [-10, 24]
+        ]
+      ]
+    }
+  };
+
+  var triangle = {
+    "type": "Feature",
+    "properties": {},
+    "geometry": {
+      "type": "Polygon",
+      "coordinates": [
+        [
+          [12, 40],
+          [38, 54],
+          [74, 36],
+          [12, 40]
+        ]
+      ]
+    }
+  };
+
+  var output = intersect(square, triangle);
+
+  /*
+  {
+    "type": "Feature",
+    "properties": {},
+    "geometry": {
+      "type": "Polygon",
+      "coordinates": [
+        [
+          [50, 48],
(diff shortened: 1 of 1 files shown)


davidtheclark:
cc @morganherlocker

morganherlocker:
> I am assuming that I can't just do simple Cartesian-coordinate math to figure out what the intersection points of these lines should be, because projection needs to be accounted for in a way that I don't really understand yet.

Yep, this is correct. The intersection algorithm assumes coordinates are on a flat cartesian plane, but in reality, polygons exist on a geodesic ellipsoid surface. It's a tradeoff between speed, algorithm complexity, and accuracy.

davidtheclark:
@morganherlocker Thanks for the response. Is the inaccuracy here — the tradeoff made for the sake of simplicity and speed — a long-established Turf strategy that is unlikely to change? Also: are you aware of any libraries that would make the *opposite* tradeoff — slower and more complex for the sake of accuracy?

DenisCarriere:
These test cases look good to be merged.
@davidtheclark

davidtheclark:
@DenisCarriere I'm not sure these cases should be merged — maybe that commit should be reverted. Are you ok with that? The test cases here were just to illustrate the point above; but they are logging things to the console instead of making assertions. We can close this PR anyway, since the question has been addressed.

DenisCarriere:
@davidtheclark Make sense, just pushed a revert for those changes.

PR #459, opened 2016-08-16
Update the intersect doc to add more return types

aurelien-defossez:
The doc says that it can only return Polygons or MultiLineStrings, but it can return any geometry type except GeometryCollection.

I added some tests to show this and updated the doc accordingly.


tmcw:
👍 thanks!

PR #467, opened 2016-09-08
Version bump to 3.5.2

DenisCarriere:
You're sub packages are getting updated by `lerna`, however shouldn't the root `package.json` also be updated?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "version": "3.0.0",
+  "version": "3.5.2",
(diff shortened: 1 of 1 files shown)


tmcw:
The root package's version doesn't really matter afaik, since it's never published.

DenisCarriere:
@tmcw That's what I thought when I found a `turf` package.

https://github.com/Turfjs/turf/tree/master/packages/turf

Thanks for letting me know 👍

PR #474, opened 2016-09-09
JSDocs hexGrid - add degrees & radians to units

DenisCarriere:
hexGrid module uses `@turf/distance` with 4 available options for units.

### DIFF
--- a/packages/turf-hex-grid/index.js
+++ b/packages/turf-hex-grid/index.js
@@ -21,7 +21,7 @@ for (var i = 0; i < 6; i++) {
  * @name hexGrid
  * @param {Array<number>} bbox bounding box in [minX, minY, maxX, maxY] order
  * @param {number} cellSize dimension of cell in specified units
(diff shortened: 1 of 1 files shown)


DenisCarriere:
Realized there's a lot more valid units of measurements, don't know if adding all of them is worth including in the JSDocs.
```javascript
'miles' | 'nauticalmiles' | 'degrees' | 'radians' | 'inches' | 'yards' | 'meters' | 'metres' | 'kilometers' | 'kilometres'
```
I think 4 of them gets the point across. 

I've added all of them into the Typescript definition.

```javascript
function hexGrid(
  bbox: Array<number>,
  cellSize: number,
  units?: 'miles' | 'nauticalmiles' | 'degrees' | 'radians' | 'inches' | 'yards' | 'meters' | 'metres' | 'kilometers' | 'kilometres',
  triangles?: boolean
): GeoJSON.FeatureCollection<GeoJSON.Polygon>;
```

PR #480, opened 2016-09-12
Simplify Turf `inside` example

DenisCarriere:
Uses the helper functions to create the point & polygon.


morganherlocker:
Nice; this is cleaner.

PR #483, opened 2016-09-13
Fixed bug in colinear triangle detection

### LINKED ISSUE #482: turf-tin collinear-ish triangles breaks algorithm
**Intro**

I recently ran into an issue trying to use turf-isolines (which uses turf-tin) on some fairly dense points. I tracked down the problem to an error in how the Bowyer-Watson algorithm is implemented.

**Example cases**

If I put in this small set of points, this valid TIN is created (Note the two very slim triangles on the upper left): 
good_tin

If we add just one more point, we get an invalid TIN - overlapping triangles and those triangles in the upper left have been erroneously deleted:
bad_tin

**Why does this happen?**

In the Bowyer-Watson algorithm, the TIN is created by (1) starting with a big enclosing triangle that will be removed later (the supertriangle) and (2) slowly adding vertices from the data one at a time, deleting any now-invalid triangles, and recreating Delaunay triangles by drawing edges from the new hole to the new vertex.

The important part is that the algorithm detects newly-invalid triangles by checking to see whether their circumcircles contain the new vertex, where the circumcircle is calculated from the three vertices of each triangle. Triangles that are very pointy and narrow have large circumcircles with centers far away from its vertices.

The error comes from [these lines](https://github.com/Turfjs/turf/blob/415d1105c23a4362c03c670fec047f1fd5a91695/packages/turf-tin/index.js#L81), which deals with the case where a given triangle has colinear vertices - not really a triangle at all. In the case that colinearity is detected, it calculates a circumcircle centered on the line with a radius on the order of the vertices.

So, what's happening is that those thin triangles in the 'good' example are being detected as colinear, their circumcircles are being placed too close to the triangle, and the algorithm is detecting them as containing the new vertex. This triggers their removal, which we don't want to happen.

**Potential fixes**

I solved my issue by making colinearity detected less frequently,  changing `Math.abs(G) < 0.000001` to `Math.abs(G) < 0.0000000001`, which makes more sense for lat-lon coordinates:
fixed_tin

That value should be tweaked to be much smaller, given how often lat-lon is used with Turf. But it still isn't a proper fix, because it's only avoiding the code block that calculates an erroneous circumcircle for colinear triangles. If one attempted to properly calculated circumcircles of colinear triangles, the circle center could be randomly placed on either side of the triangle due to floating point errors.

Colinearity could instead be detected at the time of triangle creation and used to prevent adding those triangles at all. I'm going to put in a pull request that prevents colinear triangles and reference this issue in it.

nbolten:
Fixes #482.

### DIFF
--- a/packages/turf-tin/index.js
+++ b/packages/turf-tin/index.js
@@ -74,7 +74,7 @@ function Triangle(a, b, c) {
         E = A * (a.x + b.x) + B * (a.y + b.y),
         F = C * (a.x + c.x) + D * (a.y + c.y),
         G = 2 * (A * (c.y - b.y) - B * (c.x - b.x)),
-        minx, miny, dx, dy;
+        dx, dy;
 
     // If the points of the triangle are collinear, then just find the
     // extremes and use the midpoint as the center of the circumcircle.
@@ -172,7 +172,7 @@ function triangulate(vertices) {
         )],
         closed = [],
         edges = [],
-        j, a, b;
+        j;
 
     // Incrementally add each vertex to the mesh.
     i = vertices.length;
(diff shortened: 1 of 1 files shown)


morganherlocker:
Thanks for the thorough explanation and fix!

PR #488, opened 2016-09-21
add propTypes fn to turf-meta

gmaclennan:
Would something like this be of interest/use?

### DIFF
--- a/packages/turf-meta/index.js
+++ b/packages/turf-meta/index.js
@@ -141,6 +141,24 @@ function propReduce(layer, callback, memo) {
 module.exports.propReduce = propReduce;
 
 /**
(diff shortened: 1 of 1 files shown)


tmcw:
I think this is great as a 'recipe' to share, but doesn't fit the turf-meta philosophy of only doing low-level, unopinionated tasks that would otherwise be made harder by GeoJSON's configurability. So certainly something that is great as an example of what to use turf-meta for, but doesn't seem general or essential enough for inclusion in core.

gmaclennan:
That makes sense to me.

tmcw:
Added this as an example for turf-meta in 446bab3f33576562168155feb70f4a14aabb869a

PR #489, opened 2016-09-26
Added `@turf/invariant` to package

DenisCarriere:
@tmcw I believe these should be added to the main package, please review this PR.

**Error**
```
turf.geojsonType(point1, 'point', 'Test')
     ^
TypeError: turf.geojsonType is not a function
    at Object.<anonymous> (../turf-tests.ts:515:6)
```

### DIFF
--- a/packages/turf/index.js
+++ b/packages/turf/index.js
@@ -8,6 +8,7 @@
  * @summary Geospatial analysis for JavaScript
  */
 var helpers = require('@turf/helpers');
+var invariant = require('@turf/invariant');
 
 var turf = {
     isolines: require('@turf/isolines'),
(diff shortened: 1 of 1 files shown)


tmcw:
I've been mulling whether to expose invariant or to continue using it only to make other turf methods more robust. What do you think?

DenisCarriere:
Well it's a bit deceiving when you have them listed in your docs and then you try to use them and it doesn't work.

http://turfjs.org/docs/#geojsontype

They are very specific functions, but I'm sure someone else would want to use this functionality.

You could remove these functions all together in the next major release, keep them for now.

I personally would want to use the `coordAll`, it plays well with [`concaveman`](https://github.com/mapbox/concaveman) which only inputs an Array of coordinates.

DenisCarriere:
@tmcw Ops, didn't mean to merge right away. Feel free to revert or keep it merged.

PR #497, opened 2016-10-15
'units' modules param defaults to 'kilometers'

ath0mas:


DenisCarriere:
Looks good!

ath0mas:
Every module with 'units' param defaults indirectly to kilometers, except pointOnLine

This one pass its 'units' to distance and destination
after reviewing my commit again, I finally think no default value should be set at all... so that only helpers.radiansToDistance/distanceToRadians set this default value

If OK for you, I'll update my commit to `return pointOnLine(pt, coords, units)`?

DenisCarriere:
:+1: Thanks

PR #502, opened 2016-10-24
Fix build on Windows

ath0mas:
Review 'install' and 'test' to make them pass on Windows.
 ~~Update auto-generated readmes and tests outputs.~~

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,7 +41,7 @@ turf-hello
 
 Install lerna:
 
-    npm install -g lerna@2.0.0-beta.9
+    npm install -g lerna@2.0.0-beta.30
 
 Publish a test release:
 
--- a/lerna.json
+++ b/lerna.json
@@ -1,4 +1,4 @@
 {
-  "lerna": "2.0.0-beta.23",
+  "lerna": "2.0.0-beta.30",
   "version": "3.5.2"
 }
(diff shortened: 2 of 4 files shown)


packages/turf-inside/README.md
```diff
@@ -13,53 +13,18 @@ be convex or concave. The function accounts for holes.
...
-};
-
-//=features
-
-var isInside1 = turf.inside(pt1, poly);
-//=isInside1
-
-var isInside2 = turf.inside(pt2, poly);
-//=isInside2
+var pt = point([-77, 44]);
```
tmcw: Could you separate these changes? The Windows support is positive, but the examples are intended _not_ to use `point` and `polygon` and other global turf methods.

DenisCarriere: Could we not exclude the `@turf/helpers` global methods?

Most of the README.md updates in this PR is from all the JSDocs updates.

ath0mas: Fine. I remove commit on "updating readmes" (put in in another PR to be discussed separately)? I keep the 2 other commits in this PR?

DenisCarriere: I'd suggest opening a new PR with only the windows fixes. Try to keep your PR related to your title and without mixing no-related commits.

ath0mas:
Kept only commit directly related to Windows build

DenisCarriere:
@tmcw those review comments you mentioned have been resolved in this PR.
:+1:

Issue #501, opened 2016-10-24
deprecation warnings still exist


contra:
Just letting you know, a fresh install of `@turf/turf` yields:

``` js
$ npm install @turf/turf
npm WARN deprecated turf-distance@3.0.12: Turf packages are now namespaced: please use @turf/distance instead
npm WARN deprecated turf-helpers@3.0.12: Turf packages are now namespaced: please use @turf/helpers instead
npm WARN deprecated turf-bearing@3.0.12: Turf packages are now namespaced: please use @turf/bearing instead
npm WARN deprecated turf-destination@3.0.12: Turf packages are now namespaced: please use @turf/destination instead
npm WARN deprecated turf-invariant@3.0.12: Turf packages are now namespaced: please use @turf/invariant instead
```

They all seem to be coming from one dep, `turf-line-slice-along`.

```
$ npm ls turf-helpers turf-distance turf-bearing turf-destination turf-invariant
staeco@0.0.1 /Users/contra/Projects/staeco
└─┬ @turf/turf@3.5.2
  └─┬ @turf/line-slice-along@3.5.2
    ├─┬ turf-bearing@3.0.12
    │ └── turf-invariant@3.0.12
    ├── turf-destination@3.0.12
    ├── turf-distance@3.0.12
    └── turf-helpers@3.0.12
```

Updating the package.json of that module should fix all of these errors and make installs clean.

ath0mas:
I have a PR to come that deal with those deps to 3.0.x (except maybe turf-isolines dependency to turf-grid 1.0.1)

contra:
@ath0mas Thanks!

DenisCarriere:
Re-opening Issue: Still might have a few more warnings from @ath0mas comment.

> Fix #501 removing deps to v3.0.x modules
> Still have a dep to "turf-grid" in turf-isolines though (vs turf-point-grid 3.0.x / @turf/point-grid 3.5.x ?)

morganherlocker:
The isolines warnings have been eliminated. Remaining warnings:

```
npm WARN deprecated turf-bearing@3.0.12: Turf packages are now namespaced: please use @turf/bearing instead
npm WARN deprecated turf-helpers@3.0.12: Turf packages are now namespaced: please use @turf/helpers instead
npm WARN deprecated turf-destination@3.0.12: Turf packages are now namespaced: please use @turf/destination instead
[...]

PR #506, opened 2016-10-29
Replaced turf-grid with @turf/point-grid

divingdeer:
This change fixes issue #501 

The @turf/point-grid package has the 'unit' input parameter, but is not defined as input in or used by the @turf/isolines package. The parameter is canceled out with the added code.


morganherlocker:
@divingdeer Thanks, this looks great!

PR #509, opened 2016-11-04
Keep module export consistent

DenisCarriere:
- Update Readme
- keep module.exports consistent


packages/turf-helpers/index.js
```diff
@@ -18,16 +18,14 @@
  *
  * //=feature
  */
-function feature(geometry, properties) {
```
DenisCarriere: This function is required, didn't have `eslint-config-mourner` installed.

PR #510, opened 2016-11-04
Add Typescript 2.0 definition

DenisCarriere:
Adding Typescript 2.0 definition inside the sub-modules will be a lot easier to maintain without any external dependencies for Typescript support.

Microsoft Typescript dev teams favours bundling the definitions into the npm package.

> Now that you have authored a declaration file following the steps of this guide, it is time to publish it to npm. There are two main ways you can publish your declaration files to npm:

> 1. bundling with your npm package, or
> 2. publishing to the @types organization on npm.

> If you control the npm package you are publishing declarations for, then the first approach is favored. That way, your declarations and JavaScript always travel together.

https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/Publishing.md

### DIFF
--- a/packages/turf-helpers/index.d.ts
+++ b/packages/turf-helpers/index.d.ts
@@ -1,0 +1,69 @@
+/// <reference types="geojson" />
+
+export type Point = Array<number>
+export type LineString = Array<Array<number>>
+export type Polygon = Array<Array<Array<number>>>
+export type MultiPoint = Array<Point>
+export type MultiLineString = Array<LineString>
+export type MultiPolygon = Array<Polygon>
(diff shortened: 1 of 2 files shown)


DenisCarriere:
@tmcw @morganherlocker Let me know you're opinion for this Typescript 2.0 support approach, I tried to keep it as lightweight as possible.

If this works, I'll continue to add the other sub-modules the same way.

tmcw:
👍 Looks great to me

DenisCarriere:
Great! Also I've excluded the JSDocs from the `index.d.ts` since they are already in the `index.js` & the online docs. We don't want to be maintaining two sets of JSDocs, one is hard enough ;)

PR #512, opened 2016-11-08
Fix use of wrong index variable in idw iteration

ubergesundheit:
This pull request corrects an error introduced by refactoring of for loops in the turf-idw package.


DenisCarriere:
Looks good 👍  `samplingGrid` should be using `i` instead of `j`.

PR #514, opened 2016-11-09
Add index to turf-meta callback arguments. Fixes #513

### LINKED ISSUE #513: Callback with index in propEach and featureEach
Hi,

Can you add index into callback in `turf-meta` functions?

Now you have:
```js
function propEach(layer, callback) {
    var i;
    switch (layer.type) {
    case 'FeatureCollection':
        for (i = 0; i < layer.features.length; i++) {
            callback(layer.features[i].properties);
        }
        break;
    case 'Feature':
        callback(layer.properties);
        break;
    }
}
```

But you can change it to:

```js
function propEach(layer, callback) {
    var i;
    switch (layer.type) {
    case 'FeatureCollection':
        for (i = 0; i < layer.features.length; i++) {
            callback(layer.features[i].properties, i);
        }
        break;
    case 'Feature':
        callback(layer.properties, 0);
        break;
    }
}
```

So using this function would be really **similar** (as you said in doc) to `Array.forEach` :)

tmcw:
cc @morganherlocker for the review

### DIFF
--- a/packages/turf-meta/index.js
+++ b/packages/turf-meta/index.js
@@ -114,11 +114,11 @@ function propEach(layer, callback) {
     switch (layer.type) {
     case 'FeatureCollection':
         for (i = 0; i < layer.features.length; i++) {
-            callback(layer.features[i].properties);
+            callback(layer.features[i].properties, i);
         }
         break;
     case 'Feature':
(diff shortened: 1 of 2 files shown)


morganherlocker:
👍 looks great

PR #516, opened 2016-11-10
Update README to warn about (long, lat) order.

wafisher:


README.md
```diff
@@ -77,13 +79,15 @@ Most Turf functions work with GeoJSON features. These are are pieces of data tha
 Turf provides a few geometry functions of its own. These are nothing more than simple (and optional) wrappers that output plain old GeoJSON. For example, these two methods of creating a point are functionally equivalent:
 
 ```js
-var point1 = turf.point([0, 0]);
+// Note order: longitude, latitude.
+var point1 = turf.point([-73.988214, 40.749128]);
```
DenisCarriere: I would even suggest having the longitude be greater than 90 degrees to make it even more obvious.

ex: `[120.45, 40.74]`

PR #522, opened 2016-11-25
Make units an optional param for `@turf/concave`

DenisCarriere:
Any objections to making `units` optional?

I believe this is the only library that has optional units.

The units param is only used by `@turf/distance`.

### DIFF
--- a/packages/turf-concave/index.js
+++ b/packages/turf-concave/index.js
@@ -15,7 +15,7 @@ var distance = require('@turf/distance');
  * @param {FeatureCollection<Point>} points input points
  * @param {number} maxEdge the size of an edge necessary for part of the
  * hull to become concave (in miles)
- * @param {String} [units=kilometers] can be degrees, radians, miles, or kilometers
+ * @param {string} [units=kilometers] can be degrees, radians, miles, or kilometers
  * @returns {Feature<Polygon>} a concave hull
  * @throws {Error} if maxEdge parameter is missing
  * @example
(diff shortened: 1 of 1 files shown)


packages/turf-concave/index.js
```diff
@@ -15,10 +15,9 @@ var distance = require('@turf/distance');
  * @param {FeatureCollection<Point>} points input points
  * @param {number} maxEdge the size of an edge necessary for part of the
  * hull to become concave (in miles)
- * @param {string} units used for maxEdge distance (miles or kilometers)
+ * @param {String} [units=kilometers] can be degrees, radians, miles, or kilometers
```
tmcw: string (rather than String) is canonical here, since primitive strings aren't actually full instances of the String class

DenisCarriere: Ops.. I just simply copied an example from TurfJS, just realized that `{string}` or `{String}` is not very consistent overall. I'll go fix the other JSDocs `units` that have `{String}`.

I normally put `{string}` as param, it's also explained in JSDocs
http://usejsdoc.org/tags-param.html

DenisCarriere:
More packages with `{String}` as param.
https://github.com/Turfjs/turf/blob/master/packages/turf-destination/index.js
https://github.com/Turfjs/turf/blob/master/packages/turf-along/index.js
https://github.com/Turfjs/turf/blob/master/packages/turf-line-distance/index.js
https://github.com/Turfjs/turf/blob/master/packages/turf-line-slice-along/index.js
https://github.com/Turfjs/turf/blob/master/packages/turf-helpers/index.js (`{string=kilometers}`)
https://github.com/Turfjs/turf/blob/master/packages/turf-point-on-line/index.js

DenisCarriere:
Fixed https://github.com/Turfjs/turf/commit/64aad55994cae707304295681ed61ac428f55483 more JSDocs params.

Used `documentation` to find warning messages:

```bash
$ documentation lint packages/**/index.js
```

PR #525, opened 2016-11-28
Update turf-simplify optional params

DenisCarriere:
This package uses `simplify-js`.

The reference for the added optional `@params` can be found here:

https://github.com/mourner/simplify-js/blob/master/simplify.js#L103-L113


DenisCarriere:
@mourner I'm pretty sure these optional params are correct based on your library, right?
```javascript
* @param {number} [tolerance=1] simplification tolerance
* @param {boolean} [highQuality=false] whether or not to spend more time to create
```

mourner:
@DenisCarriere yes.

Issue #286, opened 2015-10-15
overlaps 


jvail:
Could turf.overlaps (https://github.com/tcql/turf-overlaps) be added to the release or are there reasons I am not aware of that it has not been included so far? I have not used/tested turf-overlaps but a function a la intersects/overlaps (true/false) would be very useful.

WillieMaddox:
I second this. Specifically, having access to all the DE-9IM relationship tests is the only thing I see lacking in turf.  Everything else seems solid.

DenisCarriere:
👍 This would be a good idea.

stebogit:
Implemented with #856

PR #526, opened 2016-11-29
Added intersects from jsts

bozdoz:


DenisCarriere:
What's the purpose of this module `intersects`?

You can achieve the same `@return {boolean} true/false` by simply using `@turf/intersect` with `!!`.

```javascript
intersect = require('@turf/intersect')
console.log(!!intersect(poly1, poly2))
//=true/false
```

bozdoz:
The purpose is to avoid any costly aggregate function if possible.  Also, please note that this function exists in the jsts library.  I don't think it's outrageous to include it here.

DenisCarriere:
If you could open an [turf issue](https://github.com/Turfjs/turf/issues) and describe in detail what features you want implemented, that way we can have an open discussion on how to implement this.

Another similar issue was opened https://github.com/Turfjs/turf/issues/286 (turf.overlaps).

bozdoz:
Doesn't look like much of a discussion. Good luck!

DenisCarriere:
Unfortunately not... but it's better to open an Issue first, talk about it, then implement it.

PR #532, opened 2016-12-22
Upgrade geojson-area dependency

elentok:
The previous version (0.2.1):

* Couldn't be used with "use strict" (due to a global loop "i" variable)
  See https://git.io/v1hEy for more details.

* Showed a deprecation message (because it should be used as
  "@mapbox/geojson-area"


tmcw:
👍 Thanks David!

elentok:
No problem :)

When can this be released to npm?

PR #533, opened 2016-12-29
maintain linestring altitude data when simplifying a linestring with altitude data

udos:
extended functionality to maintain elevation in case the positions of a linestring contains elevation data as well (currently unsupported)

note: this is in line with specs which foresees the elevation as 3rd element of positions -> https://tools.ietf.org/html/rfc7946#section-3.1.1 (and http://geojson.org/geojson-spec.html#positions)

### DIFF
--- a/packages/turf-simplify/index.js
+++ b/packages/turf-simplify/index.js
@@ -147,17 +147,9 @@ function simpleFeature(geom, properties) {
 
 function simplifyLine(coordinates, tolerance, highQuality) {
     return simplify(coordinates.map(function (coord) {
-        if (coord.len === 2){
-            return {x: coord[0], y: coord[1]};
-        } else {
-            return {x: coord[0], y: coord[1], z: coord[2]};
-        }
+        return {x: coord[0], y: coord[1]};
     }), tolerance, highQuality).map(function (coords) {
-        if (coords.len === 2){
-            return [coords.x, coords.y];
-        } else {
-            return [coords.x, coords.y, coords.z];
-        }
+        return [coords.x, coords.y, coords.z];
     });
 }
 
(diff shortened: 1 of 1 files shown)


packages/turf-simplify/index.js
```diff
@@ -147,9 +147,17 @@ function simpleFeature(geom, properties) {
...
     return simplify(coordinates.map(function (coord) {
-        return {x: coord[0], y: coord[1]};
+        if (coord.len === 2){
+            return {x: coord[0], y: coord[1]};
+        } else {
+            return {x: coord[0], y: coord[1], z: coord[2]};
+        }
     }), tolerance, highQuality).map(function (coords) {
-        return [coords.x, coords.y];
+        if (coords.len === 2){
```
tmcw: would this work, given that `coords` is now an object, not an Array?

udos: yes. the code works. I changed it locally and run it as displayed here to be able to process linestrings containing elevation data.

tmcw: Sorry, this code doesn't work: if you run tests, it writes new files with `null` altitude values. The `length` check wouldn't work, because Objects don't have length.

DenisCarriere:
@udos I've got something working.

```javascript
function simplifyLine(coordinates, tolerance, highQuality) {
    return simplify(coordinates.map(function (coord) {
        return {x: coord[0], y: coord[1], z: coord[2]};
    }), tolerance, highQuality).map(function (coords) {
        return (coords.z) ? [coords.x, coords.y, coords.z] : [coords.x, coords.y];
    });
}
```

Also please include some test cases with Z values in `test/test.js`

tmcw:
@DenisCarriere's code is the right direction, but needs one small tweak:

```js
        return (coords.z !== undefined) ? [coords.x, coords.y, coords.z] : [coords.x, coords.y];
```

Avoids the case where a coordinate like `[-77, 20, 0]` becomes `[-77, 20]` - when `z` is cast to a boolean, so `0`, undefined, and null are all falsy. Making the check stricter avoids that case.

DenisCarriere:
Agreed! I always forget that `=== undefined` to catch null values or `0`.

PR #535, opened 2017-01-04
Fix typings for tin

garthk:
Otherwise `typescript@2.1.4` correctly complains:
```
ERROR in …/node_modules/@turf/tin/index.d.ts
(9,30): error TS2304: Cannot find name 'Points'.
```


DenisCarriere:
@garthk Thanks for catching this!

DenisCarriere:
Going to add some Typescript tests to try to catch more of these errors.

**turf-tin** [`test/types.ts`](https://github.com/Turfjs/turf/blob/master/packages/turf-tin/test/types.ts)

PR #536, opened 2017-01-04
Test all Typescript definitions

DenisCarriere:
Need some input on how to tackle this. 

**Option 1**
Add a single npm script in the root directory to test valid Typescript definitions via `test/types.ts` (current PR).

**Option 2**
Have tests running in each individual package? If that's the case, all Typescript dependencies needs to be included (@types/node, @types/geojson & typescript), this option sounds overkill.

@tmcw @morganherlocker

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "3.0.0",
   "description": "a node.js library for performing geospatial operations with geojson",
   "scripts": {
-    "test": "npm run lint && npm run types && lerna bootstrap && lerna run test",
+    "test": "npm run lint && lerna bootstrap && lerna run test && npm run types",
     "lint": "eslint packages",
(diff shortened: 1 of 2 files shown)


DenisCarriere:
Also can't figure out why MacOSX seems to crash when I run `npm run types`.

It works when I'm using bash/zsh/sh.
```
tsc **/test/types.ts --noImplicitAny
```

I've tested `npm run types` on Linux (Ubuntu) and it works... strange?

morganherlocker:
`Option 1` seems like a good plan to me.

DenisCarriere:
@morganherlocker Perfect, I also prefer Option 1.

I've already been able to spot & fix a few minor errors by adding this.

PR #537, opened 2017-01-06
[Fix] Fix an issue in lineSliceAlong when overshooting the line distance

aurelien-defossez:
If you overflow while parsing the line, the method returns an integer (the result of Array.push) instead of a LineString.


tmcw:
Thanks @aurelien-defossez !

PR #538, opened 2017-01-09
Current implementation judge some vertex on polygon as outside. This PR fix it.

kochizufan:
Current implementation judge some vertex on polygon as outside. This PR fix it.


kochizufan:
Error sample is included in PR as test.js

tmcw:
👍 Thanks!

PR #539, opened 2017-01-10
Give idw module a name for the doco that reflects it's module name

rowanwins:
Currently the idw module is called 'index' in the [turf doco](http://turfjs.org/docs/#index) which isn't ideal given that it is referenced by the name `idw` when calling the function.

PS I have some pull requests coming for the doco so Im just getting some changes made downstream here first


DenisCarriere:
@rowanwins Thanks!

Also it can't be searched either.

image

image

rowanwins:
Hi @DenisCarriere , now that I've tried up some of the upstream issues I plan on submitting some updates for the docs so stay tuned!

DenisCarriere:
Awesome stuff! Looking forward to it!

PR #540, opened 2017-01-10
Fix comparison operator throwing linting error in the tests

rowanwins:
ESlint was throwing an error because of a `==` equality operator so simply changed it `===`


DenisCarriere:
@rowanwins Thanks for the fix! Unfortunately there were some conflicts with a recent merge.

Fixed linting issues with this commit: https://github.com/Turfjs/turf/commit/89fbe792166e3f74897bb4da1104916688146780

PR #544, opened 2017-01-11
Turf.pointOnLine: add 'location' property

plepe:
I needed the distance along the line between start of the line and a closest point on the line (so I can play around with `turf.along`). Here's a patch that add's a `location` property to the result of `pointOnLine()`.

### DIFF
--- a/packages/turf-point-on-line/index.js
+++ b/packages/turf-point-on-line/index.js
@@ -10,7 +10,7 @@ var destination = require('@turf/destination');
  * @param {Feature<LineString>} line line to snap to
  * @param {Feature<Point>} pt point to snap from
  * @param {string} [units=kilometers] can be degrees, radians, miles, or kilometers
- * @return {Feature<Point>} closest point on the `line` to `point`. The properties object will contain two values: `index`: closest point was found on nth line part, `dist`: distance between pt and the closest point.
+ * @return {Feature<Point>} closest point on the `line` to `point`. The properties object will contain three values: `index`: closest point was found on nth line part, `dist`: distance between pt and the closest point, `location`: distance along the line between start and the closest point.
  * @example
  * var line = {
  *   "type": "Feature",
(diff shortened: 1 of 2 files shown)


packages/turf-point-on-line/index.js
```diff
@@ -86,20 +87,25 @@ module.exports = function (line, pt, units) {
         if (intersect) {
             intersectPt = point(intersect);
             intersectPt.properties.dist = distance(pt, intersectPt, units);
+            intersectPt.properties.location = length + distance(start, closestPt, units);
         }
 
         if (start.properties.dist < closestPt.properties.dist) {
             closestPt = start;
             closestPt.properties.index = i;
+            closestPt.properties.location = length + distance(start, closestPt, units);
```
DenisCarriere: I'm pretty sure you can drop the `distance()` function here since `closestPt === start` therefore always being a distance of zero.

```javascript
closestPt.properties.location = length;
```

DenisCarriere:
Overall it looks good to me 👍

Only downside is the extra `distance()` calculations will slightly slow down the performance for those who don't require the new `location` property.

Before merging, need some other opinions on this PR.

packages/turf-point-on-line/index.js
plepe: You are absolutely right. I added a commit. Or would you prefer to squash the change?

plepe:
What I wanted to achieve: To build a tangent to a line close to a point (which might  lie directly on the line, but not necessarily). Now I calculate the position along the line, create two points on the line near the closestPt (via turf.along). That's why I need the location. Maybe there's a different way which I overlooked?

It works fine with the submitted solution :-)

DenisCarriere:
@plepe Looks great! The changes are clear, you've included the tests cases & updated the docs 👍

plepe:
Thanks! \o/

PR #551, opened 2017-01-19
Update examples for testing in www

jvrousseau:
While implementing TurfJS/turf-www#93, I ran across some inconsistencies in the example section in the documentation. This blocked me from having a repeatable path for testing.

This PR cleans up those inconsistencies such that basic testing in turf-www passes.

Changes made to the examples consist of:

- examples for turf functions would need to be referenced as turf.{function}
- examples for turf meta functions would need to be referenced as turfMeta.{function}
- missed medial capitals
- incorrect order of parameters

This PR would allow me to go back and change the submodule in turf-www to point to the Turfjs organization's version of turf rather than my personal fork.

Please let me know if there are any changes that you would like to see, or if you see any issues with the PR.


tmcw:
👍 beautiful, reviewed and everything looks good.

Issue #550, opened 2017-01-19
Error in turf.concave when passed points are less than 3


harsh-a1:
Getting the following error when there are less than 3 point features int he array passed to the turf.concave(array,dist,unit) function -
```
Uncaught SyntaxError: Unexpected token u in JSON at position 0(…)
merge @ bundle.js:88087
concave @ bundle.js:88083
(anonymous function)....
```
Looks like bug or am i doing something wrong?

DenisCarriere:
@harsh-a1 Can you provide your input data or post it as a [Gist](https://gist.github.com).

harsh-a1:
Following reproduces it on my machine - 
```
   var pointsX = {
                type:"FeatureCollection",
                features : [ {
                    "type": "Feature",
                    "geometry": {
                        "type": "Point",
                        "coordinates": [13,79]
                    }
                }, {
                    "type": "Feature",
                    "geometry": {
                        "type": "Point",
                        "coordinates": [13,78]
                    }
                }]
            };

   var hullX = turf.concave(pointsX, 1000, 'kilometers');
```

dpmcmlxxvi:
The underlying issue seems to be the `merge` method used by `concave` which assumes at least 1 feature to merge. There should be a check on `tinPolys` having at least 1 feature.

You can also get the error for more than 2 points if the `maxEdge` parameter filters out too many triangles. Here is a [jsfiddle](https://jsfiddle.net/371r0m8u/4/) that can trigger the error both ways.

DenisCarriere:
Patch merged.

PR #552, opened 2017-01-22
Check turf-concave if not enough polygons to compute hull

dpmcmlxxvi:
Added a check and new tests to address #550. The new code still throws an exception like the old code but it now has a sensible message that there was a problem instead of the cryptic SyntaxError token thrown previously.


DenisCarriere:
@dpmcmlxxvi Looks great :+1:
Also thanks for the linting commit, I'm glad the automated testing is catching this in the PR's now.

dpmcmlxxvi:
No problem. Should have thought to run the full tests with linting before pushing the first time.

DenisCarriere:
No worries, that's why I :heart: GitHub, it tells you when you break stuff :nerd_face:

PR #556, opened 2017-01-23
Initial commit of turf-dissolve

rowanwins:

Gday @DenisCarriere 

First commit of turf-dissolve module.


packages/turf-dissolve/index.js
```diff
@@ -0,0 +1,161 @@
...
+var gju = require('geojson-utils');
+var getClosest = require('get-closest');
+
+/**
+ * Dissolves a FeatureCollection of polygons based on a property. Note that multipart features within the collection are not supported
+ *
+ * @name dissolve
+ * @param {FeatureCollection<Polygon>} featureCollection input feature collection to be dissolved
+ * @param {string} propertyName property name on which to dissolve features
+ * @return {FeatureCollection} a FeatureCollection containing the dissolved polygons
```
DenisCarriere: 1. JSDocs `@returns` instead of `@return`
2. Wouldn't it return `{FeatureCollection<Polygon>}`?

packages/turf-dissolve/index.js
```diff
@@ -0,0 +1,161 @@
...
+var Rbush = require('rbush');
+var gju = require('geojson-utils');
+var getClosest = require('get-closest');
+
+/**
+ * Dissolves a FeatureCollection of polygons based on a property. Note that multipart features within the collection are not supported
+ *
+ * @name dissolve
+ * @param {FeatureCollection<Polygon>} featureCollection input feature collection to be dissolved
+ * @param {string} propertyName property name on which to dissolve features
```
DenisCarriere: 1. Dissolving by property name should be optional.
2. Maybe we should introduce this option as a list of strings instead? That way this gives the ability to dissolve from multiple properties. `{string|Array<string>}` or only handle `{Array<string>}`

packages/turf-dissolve/index.js
```diff
@@ -0,0 +1,161 @@
...
+            minY: inputFeatureBbox[1],
+            maxX: inputFeatureBbox[2],
+            maxY: inputFeatureBbox[3],
+            origIndexPosition: polyIndex
+        };
+        treeItems.push(treeObj);
+    }
+    rtree.load(treeItems);
+
+    for (var i = 0; i < featureCollection.features.length; i++) {
```
DenisCarriere: This next comment is a personal preference, but you could make this `for loop` like the following:

```javascript
for (var polygon of featureCollection.features) {
  ...
```

Which eliminates of these lines:

```javascript
for (var i = 0; i < featureCollection.features.length; i++) {
  var polygon = featureCollection.features[i];
```
This does work in the browser, I've already tested this using Webpack & Browserify

packages/turf-dissolve/index.js
```diff
@@ -0,0 +1,161 @@
...
+            minX: polyBoundingBox[0],
+            minY: polyBoundingBox[1],
+            maxX: polyBoundingBox[2],
+            maxY: polyBoundingBox[3]
+        };
+        var potentialMatchingFeatures = rtree.search(searchObj);
+
+        var featureChanged = false;
+
+        for (var searchIndex = 0; searchIndex < potentialMatchingFeatures.length; searchIndex++) {
```
DenisCarriere: Same `for loop` comment from earlier

```javascript
for (var polygon of potentialMatchingFeatures) {
  ...
```

packages/turf-dissolve/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@turf/dissolve",
+  "version": "3.7.5",
+  "description": "Dissolves polygons based on an attribute",
+  "main": "index.js",
```
DenisCarriere: - Add `types: "index.d.ts"` once the`index.d.ts` is created.

Use this module as an example:
https://github.com/Turfjs/turf/tree/master/packages/turf-bearing

packages/turf/index.js
```diff
@@ -16,6 +16,7 @@ var turf = {
     within: require('@turf/within'),
     concave: require('@turf/concave'),
     difference: require('@turf/difference'),
+    dissolve: require('@turf/dissolve'),
```
DenisCarriere: Might want to remove this until all of the TravisCI tests pass.

We will re-add this afterwards once everything is good to go.

https://travis-ci.org/Turfjs/turf/builds/194420002#L780

packages/turf-dissolve/index.js
rowanwins: Happy to implement number 1, can't believe that didn't cross my mind!
Number 2 might be a bit trickier so for version 1 I'd prefer to stick to a single property.

packages/turf-dissolve/index.js
rowanwins: I need to hold onto the index variables unfortunately so can't really do this sorry.

dpmcmlxxvi: Another alternative is to use the array's `forEach` method. I find it to be expressive like the former but offers the option to pass in the index like the latter. Since you are not breaking out of the loop and only using `continue` shouldn't be an issue using `forEach`.

packages/turf-dissolve/index.js
DenisCarriere: That works for me, I just know that QGIS & ArcGIS can dissolve using multiple attributes, just something to keep in mind in the future.

packages/turf-dissolve/index.js
DenisCarriere: @rowanwins If you need to keep the index, you can use `in` instead of `of`.

```javascript
for (var i in featureCollection.features) {
  var polygon = featureCollection.features[i];
```
I dunno... personal taste I guess, I just hate seeing a `for loop` with `i++` it feels like the jQuery days.

packages/turf-dissolve/index.js
```diff
@@ -106,8 +106,10 @@ module.exports = function (featureCollection, propertyName) {
             }
             var matchFeature = featureCollection.features[matchFeaturePosition];
 
-            if (matchFeature.properties[propertyName] !== polygon.properties[propertyName]) {
-                continue;
+            if (propertyName !== null) {
```
DenisCarriere: Should be `undefined` instead of `null.
An empty param will result as `undefined`.

**Question**: Did this get picked up in your tests?

packages/turf-dissolve/index.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="geojson" />
+
+type Polygons = GeoJSON.FeatureCollection<GeoJSON.Polygon>;
+
+/**
+ * http://turfjs.org/docs.html#dissolve
+ */
+declare function dissolve(featureCollection: Polygons, propertyName: string): Polygons;
```
DenisCarriere: Include `?` to make it optional.

`propertyName?: string`

packages/turf-dissolve/test.js
```diff
@@ -1,13 +1,18 @@
 var turfDissolve = require('./index');
 var test = require('tape');
 var polys = require('./test/polys.json');
-var expectedOutput = require('./test/outputPolys.json');
+var expectedOutput = require('./test/outputPolysByProperty.json');
+var expectedOutput2 = require('./test/outputPolysWithNoProperty.json');
```
DenisCarriere: Instead of hardcoding your expected results, add a `fs.writeSync` function using an environment variable `REGEN=true` to output your expected results.

Here's an example of a test case using this:

https://github.com/Turfjs/turf/blob/master/packages/turf-buffer/test/test.js#L34-L38

```javascript
    buffered.features = buffered.features.concat(normalize(fixture).features);
    if (process.env.REGEN) {
      fs.writeFileSync(
          __dirname+'/fixtures/out/'+path.split('/')[path.split('/').length-1],
          JSON.stringify(buffered, null, 2));
    }
    var expected = JSON.parse(fs.readFileSync(
        __dirname+'/fixtures/out/'+path.split('/')[path.split('/').length-1]));
    t.deepEqual(expected, buffered);
  });
```

packages/turf/index.js
```diff
@@ -16,7 +16,7 @@ var turf = {
     within: require('@turf/within'),
     concave: require('@turf/concave'),
     difference: require('@turf/difference'),
-    dissolve: require('@turf/dissolve'),
+    // dissolve: require('@turf/dissolve'),
```
DenisCarriere: Damn.. still can't make the tests pass since it really needs that `@turf/dissolve` module to be present for `@turf/turf`, once the package is published once it will be ok.

https://travis-ci.org/Turfjs/turf/builds/194629215#L2380-L2383

You can uncomment this since the tests has passed all the way to the end.

DenisCarriere:
@rowanwins Looks good, just have a look at those recent comments I made and we should be good to merge and publish this module.

packages/turf-dissolve/index.js
DenisCarriere: @rowanwins Optional JSDocs are defined with square brackets. 

```javascript
 * @param {string} [propertyName] ...
```

packages/turf-dissolve/index.js
rowanwins: happy to change it to undefined. The existing tests don't pick up any errors even though one of the polys does not have the required property.

packages/turf-dissolve/test.js
rowanwins: I don't follow on this? What is the point? It seems much trickier than the current approach...

DenisCarriere:
FYI @rowanwins That last commit you sent 3 hours ago didn't have any updates.

DenisCarriere:
@rowanwins This module is Published! 👍 🎆 

Fixed some minor things: https://github.com/Turfjs/turf/commit/97c458e7f33c3ad60b255084a4eb00d93f7e00fa

PR #558, opened 2017-01-26
[line-slice-along] Fix a crash (accessing coord[i+1] when i is the last index) when overshooting a line in some cases.

aurelien-defossez:
When startDist = 0 and stopDist is greater than the line length, line-slice-along crashes because it tries to access the n+1 index of an array (OutOfBound).

I added a simple check to return the current slice if we've travelled until the end of the line.

### DIFF
--- a/packages/turf-line-slice-along/index.js
+++ b/packages/turf-line-slice-along/index.js
@@ -85,6 +85,10 @@ module.exports = function lineSliceAlong(line, startDist, stopDist, units) {
             slice.push(coords[i]);
         }
 
+        if (i === coords.length - 1) {
+            return lineString(slice);
+        }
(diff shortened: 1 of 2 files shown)


tmcw:
👍 Could you add a test to line-slice-along that would have crashed without this fix and no longer crashes, so that we can lock down the fix?

aurelien-defossez:
Yep, done!

DenisCarriere:
@aurelien-defossez Looks good, thanks for included the test cases.

```
# turf-line-slice -- line1 overshoot
ok 5 should be equal
ok 6 should be equal
ok 7 should be equivalent
ok 8 should be equivalent
```

PR #563, opened 2017-02-01
Round index up

bsudekum:
_Pulled from https://github.com/turf-junkyard/turf-point-on-line/pull/8_

The current behavior is biased towards the beginning of the line. In other words, if you’re midway between two vertices on the line, the returned index always corresponds to the earlier vertex, even if the point is closer to the later vertex. (The distance is consistent with this bias.) It’s as if the function “rounds down”. 

Here is the test case I added: 

>[Link](http://geojson.io/#id=gist:bsudekum/c2967fcaf2cbbef16d57401e40007ace&map=12/41.0328/-92.1629)

On master, the index will be `7`, on this PR the index would be `8` or the last (middle most) point. Visually, this makes sense; the point is closest to the endpoint of the line

We discovered this bug while porting this library to [swift](https://github.com/mapbox/MapboxNavigation.swift/blob/master/MapboxNavigation/Geometry.swift#L142-L177) and started to see some interesting edge cases when using this lib to draw maneuver arrows:

> _We use the index to know where to put the tip of the arrow_

/cc @DenisCarriere @tmcw @morganherlocker @1ec5

### DIFF
--- a/packages/turf-point-on-line/index.js
+++ b/packages/turf-point-on-line/index.js
@@ -99,7 +99,7 @@ module.exports = function (line, pt, units) {
         }
         if (stop.properties.dist < closestPt.properties.dist) {
             closestPt = stop;
-            closestPt.properties.index = i;
+            closestPt.properties.index = i + 1;
             closestPt.properties.location = length + sectionLength;
         }
         if (intersectPt && intersectPt.properties.dist < closestPt.properties.dist) {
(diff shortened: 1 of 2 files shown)


DenisCarriere:
Was scratching my head a bit trying to figure out why your test cases had `8` instead of `9`. It's because your geojson.io provided link wasn't the same as your test case 😕 (actual GeoJSON for test case [link](http://geojson.io/#id=gist:anonymous/c12062e10a5752b382b18665e09787be&map=12/41.0328/-92.1629)
Anyways, after I've mentally overcame that conundrum, everything made sense! 👍

Issue #569, opened 2017-02-08
Missing full license and copyright notice in sub modules


vlna:
In modules listed below files `LICENSE` with copyright notice are missing. Files `LICENSE` are missing in npm packages too. ISC license says `the above copyright notice and this permission notice appear in all copies`

Proper `LICENSE` files should be added or even better @tmcw and @morganherlocker should allow license change to MIT to keep whole turf under one license.

```
turf-area/package.json-  "author": "Tom MacWright",
turf-area/package.json:  "license": "ISC",
--
turf-bezier/package.json-  "author": "Morgan Herlocker",
turf-bezier/package.json:  "license": "ISC",
--
turf-concave/package.json-  "author": "Morgan Herlocker",
turf-concave/package.json:  "license": "ISC",
--
turf-convex/package.json-  "author": "Morgan Herlocker",
turf-convex/package.json:  "license": "ISC",
--
turf-intersect/package.json-  "author": "Morgan Herlocker",
turf-intersect/package.json:  "license": "ISC",
--
turf-invariant/package.json-  "author": "Tom MacWright",
turf-invariant/package.json:  "license": "ISC",
--
turf-meta/package.json-  "author": "Tom MacWright",
turf-meta/package.json:  "license": "ISC",
--
turf-random/package.json-  "author": "Tom MacWright",
turf-random/package.json:  "license": "ISC",
--
turf-simplify/package.json-  "author": "Morgan Herlocker",
turf-simplify/package.json:  "license": "ISC",
--
turf-union/package.json-  "author": "Morgan Herlocker",
turf-union/package.json:  "license": "ISC",
```

morganherlocker:
@vlna Thanks for flagging this. My intent was/remains to license turf and all related packages under the MIT license. I'll look into automating this consistently across the board.

morganherlocker:
@vlna, @tmcw implemented an automated test for fixing submodule license inconsistencies, which has been merged ( 🙇 🙇 🙇 ). All modules are now properly licensed under MIT (granted under my copyright) and authors are listed as "Turf Authors".

PR #570, opened 2017-02-09
license fix in package.json files #569

vlna:
Code not changed.

ISC just replaced with MIT in `package.json` files


tmcw:
👍  Thanks, merging and then I'll automate.

PR #578, opened 2017-02-23
initial commit of polygon split

rowanwins:
This is an inital commit of a polygon split module based on the polyk library
The main drawbacks are that it introduces polyk as a dependency (~5kb minified) and the line has to be a straight line. But it's better than nothing.

### DIFF
--- a/packages/turf-split/LICENSE
+++ b/packages/turf-split/LICENSE
@@ -1,0 +1,22 @@
+The MIT License (MIT)
+
+Copyright (c) 2017 Morgan Herlocker
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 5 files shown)


DenisCarriere:
@rowanwins Great stuff!

I'll have a look at it more tomorrow.

First off, it looks like the pre-install tests are failing, looks like there's an issue in the `package.json`.

Looks like there's an issue with `directory`.

```json
 "directories": {
    "test": "test"
  },
```
```
NpmUtilities.installInDir     ("./packages/turf-split", ["benchmark@^1.0.0","tape@^3.5.0","lodash.chunk@^4.2.0","polyk@^0.22.2"])
npm ERR! cb() never called!
@turf/simplify Errored while running BootstrapCommand.execute
npm ERR! cb() never called!
Waiting for 3 child processes to exit. CTRL-C to exit immediately.
```

rowanwins:
Hi @DenisCarriere 

There appears to be error with the polyk package on npm so I've logged a issue with them so stay tuned, once that is fixed all should be resolved. 

That `directories` tag in the package.json can also be deleted I think, I just cut and paste the package.json from somewhere else so it's just a hangover I think

Cheers
R

DenisCarriere:
@rowanwins Doesn't look like PolyK will work long term:

Fallbacks:
- Linestring is limited to two segments
- Dropped decimal precision points
- Output & Input is a very strange array [x1 y1, x2, y2, ...]

However, maybe we can look into the code and reshape it to make it work in a more GeoJSON friendly way, right now it's a giant hack to make it work, it doesn't look clean at all.

DenisCarriere:
I've added tests for PolyK (I had to peak inside your `index.js` "hack" to figure out how to make it work).

https://github.com/MartyWallace/PolyK/pull/10

rowanwins:
Gday @DenisCarriere 

Yeah it was a bit of a limited workaround, I suspect there is probably a fair bit of effort to build it ourselves but maybe we can get there, the thing that concerns me the most is supporting a line with more than 2 vertices and how much tweaking we'd need to do to the algorithm

DenisCarriere:
I still think including `PolyK` is a good idea for now, however in the future we can look into another option or even try to refactor the `Slice` component. It sounds painful to do at first, so lets start off by including it as a dependency and then look into those other options later.

My latest commits should of resolved that npm install issue, they just need to publish a new release now.

DenisCarriere:
Started a new branch `slice` https://github.com/Turfjs/turf/tree/slice

I've included some of your code and fixed the package issue in `PolyK`.

PR #579, opened 2017-02-23
Initial commit of @turf/unkink-polygon

rowanwins:
A new module to unkink a polygon based on [simplepolygon](https://github.com/mclaeysb/simplepolygon) (which in turn relies on turf :) )

One difference I have made from simplepolygon is that it returns some information about the unkinking process in the properties, I've removed these and simply assigned the input polygon properties instead. Happy to be guided by others opinions on this.


DenisCarriere:
Could the input include or test for the following input GeoJSON types:

- `FeatureCollection<Polygon>`
- `Feature<Polygon>`
- `FeatureCollection<MultiPolygon>`
- `Feature<MultiPolygon>`

Either returns an error with an inappropriate type or handles the input by using [`geojson-normalize`](https://github.com/mapbox/geojson-normalize) & [`geojson-flatten`](https://github.com/tmcw/geojson-flatten).

DenisCarriere:
Updated Benchmark tests, the results look pretty good!

```
hourglass.geojson x 16,597 ops/sec ±0.86% (88 runs sampled)
hourglassFeatureCollection.geojson x 8,074 ops/sec ±1.26% (89 runs sampled)
hourglassFeatureCollectionMultiPolygon.geojson x 4,866 ops/sec ±1.15% (86 runs sampled)
hourglassMultiPolygon.geojson x 6,925 ops/sec ±1.25% (87 runs sampled)
polygon.geojson x 34,603 ops/sec ±1.72% (82 runs sampled)
```

rowanwins:
Thanks @DenisCarriere

Issue #493, opened 2016-10-07
mask


andrewharvey:
Feature request for a mask function.

Given a geojson create the inverse/mask. Currently you need to manually create a polygon stretching the globe and do turf.difference for each polygon to get the mask.

tmcw:
For simple, non-intersecting polygons, this can be a lot faster and easier than using turf.difference: take the WGS84 bounding rectangle and make it the outer ring of a polygon, and you get the correct result:

![invert](https://cloud.githubusercontent.com/assets/32314/19193207/351dcf6a-8c77-11e6-8f51-c73da7d72a1a.gif)

andrewharvey:
@tmcw Thanks for that recommendation.

PR #589, opened 2017-03-02
New module for turf mask

rowanwins:
Initial commit of a mask module as per request in #493 .


DenisCarriere:
How will this module handle this?

https://gist.github.com/2884b2d9a1dba0fdcf4e14735ec58343

Sounds like you need to have `@turf/polygon-slice` before we start attempting this.

image

rowanwins:
@DenisCarriere hmmm I'd argue that that is not really what a mask is intended for...

DenisCarriere:
But what happens when that input is chosen.. does it break?

packages/turf-mask/index.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Takes a {@link Polygon|polygon} and an optional mask and returns the outside area,
+ * if you don't pass a mask polygon then the world is used.
+ *
+ * @name mask
+ * @param {Feature<Polygon>} polygon GeoJSON Polygon as input
```
DenisCarriere: Mask should also include as input: `FeatureCollection<Polygon>` & `Feature<MultiPolygon>`.

**Input two polygons**
https://gist.github.com/69de7993a2e8f485599f28b74a75b6a2
image

**Returns Mask with two holes**
https://gist.github.com/f44af2532298fd531ea13e27f0cf264c
image

packages/turf-mask/index.js
```diff
@@ -0,0 +1,31 @@
...
+ *      ]
+ *    }
+ * }
+ * var masked = turf.mask(polygon);
+ * //=masked
+ */
+module.exports = function (polygon, mask) {
+    var world = [[180, 90], [-180, 90], [-180, -90], [180, -90], [180, 90]];
+    if (mask !== undefined) {
+        polygon.geometry.coordinates.push(mask.geometry.coordinates);
```
DenisCarriere: I don't think this is robust enough. Also shouldn't the Polygon input show up as 2nd & 3rd polygons in the Mask (seems like this should be reversed). Consider making multiple holes in output Polygon.

packages/turf-mask/test.js
```diff
@@ -0,0 +1,56 @@
...
+        'coordinates': [
+          [[100, 0], [101, 0], [101, 1], [100, 1], [100, 0]],
+          [[100.2, 0.2], [100.8, 0.2], [100.8, 0.8], [100.2, 0.8], [100.2, 0.2]]
+        ]
+    }
+};
+
+test('turf collect module', function (t) {
+    var maskedWithNoMask = mask(poly);
+    t.deepEqual(maskedWithNoMask, maskedOutputWorld, 'Masks to the world');
```
DenisCarriere: Try to make your tests dynamic, hardcoding Polygons in the test cases is very difficult to troubleshoot and you can't easily expand the test cases by simply adding another folder with a couple geojsons.

I've converted a few modules in TurfJS so far with this concept.

You can also check out the latest one [`turf-line-intersect`](https://github.com/Turfjs/turf/blob/master/packages/turf-line-intersect/test.js).

Everything that's an INPUT should be placed in `test/in` and the `test/out` should be autogenerated by adding the environment variable **REGEN=true**.

DenisCarriere:
@rowanwins This should be the proposed JSDocs for mask.

```
/**
 * Takes a {@link Polygon|polygon} and an optional mask and returns the outside area,
 * if you don't pass a mask polygon then the world is used.
 *
 * @name mask
 * @param {FeatureCollection|Feature<Polygon|MultiPolygon>} polygon GeoJSON Polygon (used to cut holes in the mask).
 * @param {Feature<Polygon>} [mask] GeoJSON Polygon (defaults to world if undefined)
 * @return {Feature<Polygon>} Masked Polygon
```

packages/turf-mask/package.json
```diff
@@ -0,0 +1,36 @@
...
+  ],
+  "author": "Turf Authors",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Turfjs/turf/issues"
+  },
+  "homepage": "https://github.com/Turfjs/turf",
+  "dependencies": {},
+  "devDependencies": {
+    "benchmark": "^1.0.0",
```
DenisCarriere: Use benchmark `2.x`, also I've got an easy example that you could almost copy paste and only have to replace a few things.

https://github.com/Turfjs/turf/blob/master/packages/turf-line-intersect/bench.js

DenisCarriere:
Did a few more commits, it "works" but there's lots of work to make every test case pass.

PR #599, opened 2017-03-06
New module @turf/line-chunk

rowanwins:
Based on the work of @tcql in https://github.com/Turfjs/turf-line-chunk

I also added a parameter to return points instead of line segments so Im not sure if the line-chunk name is exactly the best but I dont have anything better right now...

Doesn't support MultiLineStrings currently, adding support for them would require some additional funky logic which I ran out of puff for.


packages/turf-line-chunk/index.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    var coordinates = line.geometry.coordinates.slice();
+    var result = [];
+
+
+    while (coordinates.length > 1) {
+        var endpt = along(linestring(coordinates), segment_length, units);
+        var tempcoords = [];
+        for (var i = 0; i < coordinates.length; i++) {
```
DenisCarriere: you should look into `coordReduce()` or `coordEach()` in `@turf/meta`

packages/turf-line-chunk/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+            if (i === coordinates.length - 1) {
+                coordinates = [];
+                result.push(tempcoords);
+            }
+        }
+    }
+
+    return fc(result.map(function (coords) {
+        if (!asPoints) {
```
DenisCarriere: Drop the asPoints.

Also try not to have any complex multi-line operations in the `return` it should simply have
```javascript
return featuresCollection(features);
```

packages/turf-line-chunk/index.js
```diff
@@ -0,0 +1,94 @@
+var along = require('@turf/along');
+var lineDistance = require('@turf/line-distance');
+var helpers = require('@turf/helpers');
+var flatten = require('@turf/flatten');
+var linestring = helpers.lineString;
+var point = helpers.point;
+var fc = helpers.featureCollection;
```
DenisCarriere: Call the helpers functions the way they are `fc` = `featureCollection`, `linestring` = `lineString`

packages/turf-line-chunk/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Divides a {@link LineString} into chunks of a specified length.
+ * If the line is shorter than the segment length then the orginal line is returned.
+ * @name lineChunk
+ * @param {Feature<LineString>} line the line to split
```
DenisCarriere: This module should support `{FeatureCollection|Feature<LineString|MultiLineString>}` as input

DenisCarriere:
When you make a new module, create a new branch in TurfJS instead of on your own personal repo.

packages/turf-line-chunk/index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+type FeatureCollectionPoints = GeoJSON.FeatureCollection<LineStringFeature>;
+type FeatureCollection = FeatureCollectionLines | FeatureCollectionPoints;
+
+type LineStringFeatures = LineString | LineStringFeature;
+
+interface LineChunk {
+    /**
+     * http://turfjs.org/docs/#lineChunk
+     */
+    (featureIn: LineStringFeatures,  segment_length: number, unit: Units, asPoints?: boolean): FeatureCollection;
```
DenisCarriere: Doesn't need to be an `interface` it can be a simple `export function`.

packages/turf-line-chunk/index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+type LineStringFeature = GeoJSON.Feature<GeoJSON.LineString>;
+type LineString = GeoJSON.LineString;
+type Point = GeoJSON.Point;
+
+type FeatureCollectionLines = GeoJSON.FeatureCollection<LineStringFeature>;
+type FeatureCollectionPoints = GeoJSON.FeatureCollection<LineStringFeature>;
+type FeatureCollection = FeatureCollectionLines | FeatureCollectionPoints;
+
+type LineStringFeatures = LineString | LineStringFeature;
```
DenisCarriere: Don't mix types together, the outputs can ONLY be one type at a time.

DenisCarriere:
You should be using `@turf/line-slice-along` module, that would eliminate A LOT of complexity from your code, the only you would need to do is iterate over the chunked lines by the distance length.

packages/turf-line-chunk/index.js
```diff
@@ -0,0 +1,94 @@
...
+ * @param {Number} segment_length how long to make each segment
+ * @param {String} units can be degrees, radians, miles, or kilometers
+ * @param {Boolean} [asPoints=false] return points instead of linestring segrments
+ * @return {FeatureCollection<LineString>|FeatureCollection<Point>} collection of line segments or points
```
rowanwins: You can a feature collection of points back if you pass in the optional 4th param as true. 

Getting line segments back is good, but there are times when you just want the points at which the segments break. Another option might be to create another module which converted a linestring to a set of points (more jsdocs & ts definitions I'll butcher :) )

DenisCarriere:
Post the Benchmark results if you have them working.

packages/turf-line-chunk/.npmignore
```diff
@@ -0,0 +1,2 @@
+test
+coverage
```
DenisCarriere: You can remove `.npmignore` since this is covered by `files:[*]` in the `package.json`

rowanwins:
Thanks for the comments @DenisCarriere , the code was pretty much just cut and paste from the original repo so yep it can probably be tidied and make use of some of the helper functions that are now available. And I take your point re using other methods for returning the points so I'll remove that bit (in my use case I was thinking of points hence why I implemented it in the module).

DenisCarriere:
👍 Points are very easy to create once you have your linestring, it's better to stick with an immutable output type. You'll notice many of the TurfJS modules only have a single output type for that reason (unless it's using JSTS which sometimes has a wild card of output types).

Once you've cleaned up some of the comments I mentioned, I'll help out with the multiLineString & featureCollection support. You should try to familiar yourself with the [`@turf/meta`](https://github.com/Turfjs/turf/tree/master/packages/turf-meta) operations, they are very handy and it also prevents any memory leaks on large GeoJSON datasets. Most of our test cases is for "small" GeoJSON datasets, the real problems occur on larger datasets.

rowanwins:
Gday @DenisCarriere 

I've put in a fair old overhaul. One note is that the current version relies on your new version of turf.flatten which isn't actually released to npm yet, yours seems to be better deal with returning feature collections.

It now caters for multilines, feature collections, lines and features.

DenisCarriere:
Benchmark results are pretty decent.

```
FeatureCollection.geojson x 32,454 ops/sec ±4.62% (88 runs sampled)
LineString.geojson x 74,091 ops/sec ±1.03% (96 runs sampled)
MultiLineString.geojson x 49,438 ops/sec ±1.21% (94 runs sampled)
```

DenisCarriere:
Add `reverse` param:
```javascript
/**
* @param {boolean}[reverse=false] reverses coordinates to start the first chunked segment at the end
```

PR #606, opened 2017-03-07
Changes to invariant based on comments

DenisCarriere:
Ok without losing too much performance on `getCoord()` I added a number validation function at the end and checks for `null` & `false`.

~Added a param for the validation, that way you can turn it on or off. (default=false)~

**Benchmark before**

```
getCoord.Point x 18,555,451 ops/sec ±1.18% (86 runs sampled)
```

**Benchmark with number Validation**

```
getCoord.Point x 45,853,256 ops/sec ±0.63% (88 runs sampled)
getCoord.LineString x 33,147,142 ops/sec ±1.29% (91 runs sampled)
getCoord.Polygon x 27,853,174 ops/sec ±0.78% (92 runs sampled)
```

**Benchmark without number Validation**

```
getCoord.Point x 51,871,316 ops/sec ±1.49% (89 runs sampled)
getCoord.LineString x 49,007,217 ops/sec ±2.44% (86 runs sampled)
getCoord.Polygon x 52,104,843 ops/sec ±0.74% (88 runs sampled)
```

CC: @tmcw

### DIFF
--- a/packages/turf-invariant/README.md
+++ b/packages/turf-invariant/README.md
@@ -2,6 +2,16 @@
 
 # getCoord
 
+Unwrap a coordinate from a Point Feature, Geometry or a single coordinate.
+
+**Parameters**
+
+-   `obj` **([Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)&lt;any> | [Geometry](http://geojson.org/geojson-spec.html#geometry) \| [Feature](http://geojson.org/geojson-spec.html#feature-objects)&lt;[Point](http://geojson.org/geojson-spec.html#point)>)** any value
+
+Returns **[Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array)&lt;[number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number)>** coordinates
+
+# getCoords
+
 Unwrap coordinates from a Feature, Geometry Object or an Array of numbers
 
 **Parameters**
--- a/packages/turf-invariant/bench.js
+++ b/packages/turf-invariant/bench.js
@@ -5,22 +5,15 @@ const invariant = require('./');
 const point = helpers.point([-75, 40]);
 const linestring = helpers.lineString([[-75, 40], [-70, 50]]);
 const polygon = helpers.polygon([[[-75, 40], [-80, 50], [-70, 50], [-75, 40]]]);
-
-const fixtures = [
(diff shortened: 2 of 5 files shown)


tmcw:
If we make validation optional, this change will have to be a major version bump for Turf. I'm kind of unconvinced here - like something that's hitting millions of ops / sec is taking millionths of seconds - like 21 nanoseconds. Shaving several nanoseconds in exchange for opening the door to uncaught errors doesn't seem like a worthwhile tradeoff.

DenisCarriere:
I totally agree with your last comment, I've already removed the `validation` option.

We shouldn't add it as an option even for a major release.

DenisCarriere:
I don't have any more commits to fire off, @tmcw I'll let you merge or let me know if you have any more comments and I can change them.

So far... nothing has changed the `@turf/invariant` behavior (still backwards compatible) except for the additional geometry support for `getCoord()`.

tmcw:
I feel like this is going to break things: getCoord is singular, and the Point-geometry checking was intentional. The new code will let `getCoord(Polygon) -> Array<Array<number>>`, which will break getCoord's contract.

In current master:

```js
> var distance = require('./packages/turf-distance/');
undefined
> distance({ type: 'LineString', coordinates: [[0,0],[1,1]] }, [3,4]);
NaN
```

Before the geometry check was introduced:

```js
> var distance = require('./packages/turf-distance/');
undefined
> distance({ type: 'LineString', coordinates: [[0,0],[1,1]] }, [3,4]);
Error: A coordinate, feature, or point geometry is required
    at getCoord (/Users/tmcw/src/turf/packages/turf-invariant/index.js:24:11)
    at module.exports (/Users/tmcw/src/turf/packages/turf-distance/index.js:49:24)
```

DenisCarriere:
**Proposed Solution**

Have two methods:

- `getCoord()` - Remains identical as before since it's meant to retrieve a single Coordinate (Point). Will throw an error if coordinates Array is not `Array<number>`.
- `getCoords()` - **New method** - retrieves coordinates which would normally be (Point/Line/Polygon) Also checks if array contains a number.

Thoughts?

DenisCarriere:
@tmcw Last commit https://github.com/Turfjs/turf/pull/606/commits/bce258e5103d7d1a99dd7e5bf8d1ad18d9e6e91e should address all the concerns you mentioned above.

- Implement getCoords() which retrieves coordinates
- Make sure that getCoord() doesn't break any backwards compatibility
  - Test for Null
  - Test for False
  - Test for LineString
  - Test for Arrays that must be an array of two numbers
- Add getCoords() to typescript definition
- Update README

Issue #597, opened 2017-03-05
bearing is initial bearing


espinielli:
The implementation of `bearing` is about the *initial* bearing as from cited [URL](http://www.movable-type.co.uk/scripts/latlong.html) in the sources.
IMHO this should be explicitly mentioned in the API. 
Maybe even implement *final* bearing as per above [URL](http://www.movable-type.co.uk/scripts/latlong.html)

> For final bearing, simply take the initial bearing from the end point to the start point and reverse it (using θ = (θ+180) % 360).

rowanwins:
Thanks for the report @espinielli , I'll get the docs updated to make that clearer. 

@DenisCarriere - what do you think about adding an optional third argument which flags to return the final bearing...

DenisCarriere:
I think adding an additional param would work.

This picture was helpful figuring out the difference between intial bearing (29 Deg) & final bearing (150 Deg).
![image](https://cloud.githubusercontent.com/assets/550895/23592786/edea6a90-01d3-11e7-9caf-08cae4fac336.png)

**Proposed JSDocs**

```javascript
/**
 * Takes two {@link Point|points} and finds the geographic bearing between them.
 *
 * @name bearing
 * @param {Feature<Point>} start starting Point
 * @param {Feature<Point>} end ending Point
 * @param {boolean} [final=false] calculates the final bearing if true
 * @returns {number} bearing in decimal degrees
```

espinielli:
I love the additional picture to explain it.
Also the additional, *optional* param is a good idea.

I forgot to mention that what is returned is angle east of North (see the [footnote link](http://mathforum.org/library/drmath/view/55417.html) in [URL](http://www.movable-type.co.uk/scripts/latlong.html) mentioned above).

![image](https://cloud.githubusercontent.com/assets/891692/23605112/af3954a4-025b-11e7-9f34-028b89572268.png)

DenisCarriere:
Fixed, will be released in the next Turf minor release.

You can do some cool stuff with final bearing! 👍 

@rowanwins This opens up a new potential module, `@turf/line-arc` which creates a "fake" line which represents a line based on a the great circle. It would be almost be the same as `@turf/line-chunk` except it has initial bearings & final bearings in the mix.

**Start**=Red
**End**=Blue
![image](https://cloud.githubusercontent.com/assets/550895/23670554/13f98316-0336-11e7-9250-e8a9db762df3.png)

rowanwins:
[...]

PR #607, opened 2017-03-08
New module @turf/great-circle

DenisCarriere:
# New Module! @turf/great-circle 🎉 

Calculate great circles routes as LineString.

**Contributors**

- @springmeyer (using [arc.js](https://github.com/springmeyer/arc.js) as dependency)
- @stepankuzmin (inspired by [`turf-arc`](https://github.com/stepankuzmin/turf-arc))
- @DenisCarriere

image

**JSDocs**

```javascript
/**
 * Calculate great circles routes as {@LineString}
 *
 * @name greatCircle
 * @param {Feature<Point>} start source point feature
 * @param {Feature<Point>} end destination point feature
 * @param {*} [properties={}] line feature properties
 * @param {number} [npoints=100] number of points
 * @param {number} [offset=10] offset controls the likelyhood that lines will
 * be split which cross the dateline. The higher the number the more likely.
 * @returns {Feature<LineString>} arc line feature
 * @example
 * var start = {
 *   "type": "Feature",
 *   "properties": {},
 *   "geometry": {
 *     "type": "Point",
 *     "coordinates": [-122, 48]
 *   }
 * }
 * var end = {
 *   "type": "Feature",
 *   "properties": {},
 *   "geometry": {
 *     "type": "Point",
 *     "coordinates": [-77, 39]
 *   }
 * }
 * var line = turf.greatCircle(start, end, {'name': 'Seattle to DC'});
 * //=line
 */
```

**Benchmark**

```
arc x 21,271 ops/sec ±5.14% (82 runs sampled)
```

Please fill in this template.

CC: @Turfjs/ownership #597


morganherlocker:
This is maybe a bit on the mushy philosophical side... but is this actually calculating a great circle as it describes?

https://en.wikipedia.org/wiki/Great_circle

I'm +1 on this module, but want to make sure we have precise description of what it's really doing, and that this is useful beyond being a rendering/projection hack.

DenisCarriere:
@morganherlocker good point, should we name the module `@turf/great-circle` instead? That would define the module a bit more from only the name.

I'm 👍 changing it to `great-circle`.

morganherlocker:
I think my question still is: Does this module (or arc.js) actually "`Calculate great circles routes as {@LineString}`"? Does a great circle actually make sense with our coordinate system outside of rendering usecases? Typically we have rejected modules that are purely projection or rendering related with no geospatial analysis value. I'm not saying we should reject this module, but I would like us to think through how this module relates to the scope of turf, which has always been laser focused on geo stats and analysis.

DenisCarriere:
I can't speak too much on the analysis part, however browsing quickly over the source code it references GDAL many times and I'm sure that @springmeyer made sure the calculations were correct.

I also agree that we shouldn't be including projection related modules, however this seems to be a unique case. You might of already seen Google Maps use this type of line representation in their measure tool.

image

Which looks exactly like the one that was created with `@turf/great-circle`.

This module gives the tools needed to create this "look in feel" distance calculator using GeoJSON and easily place that line inside Leaflet or MapboxGL.

DenisCarriere:
Also since the operations are really fast 20K ops/sec, someone would be able to draw a great circle line dynamically without having to let go the mouse which would result in a REALLY slick geo tool.

mourner:
I don't think this is projection-related. It's literally calculating the shortest path between two points on Earth.

DenisCarriere:
Will be published under TurfJS `3.13`

Issue #173, opened 2014-12-30
indexing


morganherlocker:
So far I have rejected anything related to indexing. I think of it as a problem for the data layer, rather than the processing layer. That said, I think it should be considered.
## pros
- we could use the indexes internally to speed up many algorithms ([turf-inside](https://github.com/Turfjs/turf-inside) immediately comes to mind).
- it is useful functionality, and many users will be unable to figure out how to use index modules in conjunction with turf effectively on their own
- several indexes (s2 cells, tile quadkeys, and geohashes) are data independent, so they scale across clusters (r-tree, for example, does not)
## cons
- I am not aware of an existing convention for indexes in GeoJSON, besides the id field, which is flimsy for this
- multiple types of indexes on the same dataset might clash
- some common indexing schemes do not scale well in clusters (r-tree)
- the fact that users have a hard time understanding indexes might just mean this allows them to do dumb things faster :smile: 
- scalable indexes tend to rely on "universal axioms" about planet earth; turf should work work gracefully without these contextual assumptions (turf is not currently perfect on this point: units of measurement, WGS84 encoding, assumptions about precision with 32bit floats)
## candidates for submodules
- [tile-cover](https://github.com/mapbox/tile-cover): fast and scales across clusters
- [rbush](https://github.com/mourner/rbush): fast but each box needs the index in memory (this might not matter, depending on some architecture decisions we make)
- [geohash](https://github.com/sunng87/node-geohash): fast and scalable, but does not work on lines or polygons (although maybe it would work fine with geohashed bbox sw & ne points)
- [tilebelt](https://www.npmjs.com/package/tilebelt): minimum bbox quadkey would be fast and scalable (same issues as geohash)

---

I am 100% on the fence with this. Input greatly appreciated from anyone who understands how indexing works at a low level **and** people who just need to use them). This has huge architecture implications, so we really need both sides.

tmcw:
see also: https://github.com/mikolalysenko/box-intersect-benchmark

jfirebaugh:
If there are big unknowns or tradeoffs in the choice of index datastructure, would it be possible to define algorithms like turf-inside in terms of an index _protocol_ rather than a specific concrete implementation? Turf would then provide implementations of the protocol for rbush, tile-cover, tilebelt, etc, and users would supply their choice when instantiating a particular algorithm.

stepankuzmin:
[...]

Issue #168, opened 2014-12-22
feature request: geohash


drewda:
@morganherlocker not sure if this would fit into your plans, but I'd find geohash encoding and decoding a useful addition to the turf collection. It's something I often use for indexing and for producing semi-readable output.

c.f. https://github.com/sunng87/node-geohash

rowanwins:
Im going to close this due to inactivity and given that we decided to not keep the turf/rbush module in favour of releasing it as [geojson-rbush](https://github.com/DenisCarriere/geojson-rbush). 

As mentioned by @morganherlocker the idea of indexing is beyond the scope of Turf (although we very much recognise it's usefulness and where relevant we'll use it within our modules).

PR #609, opened 2017-03-10
New module @turf/rbush

DenisCarriere:
# New Module 🎉 ! `@turf/rbush`

Creates an RBush Index Tree from a Feature Collection or GeometryCollection.

**TurfJS modules using RBush**

- [ ] [@turf/collect](https://github.com/Turfjs/turf/tree/master/packages/turf-collect)
- [ ] [@turf/dissolve](https://github.com/Turfjs/turf/tree/master/packages/turf-dissolve)
- [ ] [@turf/mask](https://github.com/Turfjs/turf/tree/master/packages/turf-mask) (will depend on `@turf/rbush`)
- [ ] [@turf/line-intersect](https://github.com/Turfjs/turf/tree/master/packages/turf-line-intersect) (will depend on `@turf/rbush`)
- [ ] [@turf/kinks](https://github.com/Turfjs/turf/tree/master/packages/turf-kinks) (will depend on `@turf/rbush`)

**Dependencies**

- [RBush](https://github.com/mourner/rbush)
- [@turf/bbox](https://github.com/Turfjs/turf/tree/master/packages/turf-bbox)
- [@turf/meta](https://github.com/Turfjs/turf/tree/master/packages/turf-meta)

**Contributors**

- @mourner (Author of [RBush](https://github.com/mourner/rbush))
- @jvrousseau (Inspired by [turf-index](https://github.com/jvrousseau/turf-index))
- @DenisCarriere

**JSDocs**

```javascript
/**
 * Creates a RBush Index Tree from a Feature Collection or GeometryCollection.
 *
 * @name rbush
 * @param {FeatureCollection|Feature<any>} features Features to be added to the RBush Tree index.
 * @returns {RBush} RBush Tree
```

**Example**

```javascript
var collection = {
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [[[-78, 41], [-67, 41], [-67, 48], [-78, 48], [-78, 41]]]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [[[-93, 32], [-83, 32], [-83, 39], [-93, 39], [-93, 32]]]
      }
    }
  ]
}
var tree = turf.index(collection);
//=tree

var search = tree.search(collection.features[0])
//=search => FeatureCollection
```

**Input - FeatureCollection with 3 polygons**

image

**Output (FeatureCollection with BBox)**

~an array of Objects with the added `index` which correlates to the FeatureCollection order~

```json
{
	"type": "FeatureCollection",
	"features": [
		{
			"type": "Feature",
			"properties": {},
			"geometry": {
				"type": "Polygon",
				"coordinates": [...]
			},
			"id": 0,
			"bbox": [
				-78,
				41,
				-67,
				48
			]
		},
		{
			"type": "Feature",
			"properties": {},
			"geometry": {
				"type": "Polygon",
				"coordinates": [...]
			},
			"id": 1,
			"bbox": [
				-93,
				32,
				-83,
				39
			]
		}
	]
}

```

**Benchmark**

```
linestrings x 774,236 ops/sec ±0.82% (87 runs sampled)
points x 897,735 ops/sec ±1.17% (86 runs sampled)
polygons x 713,494 ops/sec ±1.39% (83 runs sampled)
```

Please fill in this template.

CC: @Turfjs/ownership 
Issues: #173 #168


rowanwins:
Gday @DenisCarriere 

Nice one. So what was your thoughts on how this would get used? 
- Would it be used internally within modules eg turf/collect would call on it as a dependency
- Would users be responsible for creating their own index and then passing it around where required, eg a user chooses to pass the index into turf/collect

DenisCarriere:
@rowanwins This module would primarily be used internally to create high performance core TurfJS modules. I've already got [`@turf/mask` using a very similar syntax](https://github.com/Turfjs/turf/blob/master/packages/turf-mask/index.js#L185-L208).

Essentially `@turf/index` is a quick and easy way to load an entire FeatureCollection and returns an RBush Tree. Once the RBush index is created, it's up to your application/module to handle the rest using the RBush documentation.

mourner:
Nice! Is it beneficial to have this specialized output format? If it's a Turf module, I'd expect it to return the features array directly.

DenisCarriere:
@mourner I'm sure you've ran into this problem, if we load the RBush Tree with the **ALL** the GeoJSON features, wouldn't that be incredibly slow to load the RBush Tree and also create potential memory leak problems?

@rowanwins Are you proposing that the RBush contains the `feature`? The output would look something like this:

```json
[
	{
		"minX": -78,
		"minY": 41,
		"maxX": -67,
		"maxY": 48,
		"index": 0,
		"feature": {
			"type": "Feature",
			"properties": {},
			"geometry": {
				"type": "Polygon",
				"coordinates": [
					[
						[
							-78,
							41
						],
						[
							-67,
							41
						],
						[
							-67,
							48
						],
						[
							-78,
							48
						],
						[
							-78,
							41
						]
					]
				]
			}
		}
	},
	{
		"minX": -93,
		"minY": 32,
		"maxX": -83,
		"maxY": 39,
		"index": 1,
		"feature": {
			"type": "Feature",
			"properties": {},
			"geometry": {
				"type": "Polygon",
				"coordinates": [
					[
						[
							-93,
							32
						],
						[
							-83,
							32
						],
						[
							-83,
							39
						],
						[
							-93,
							39
						],
						[
							-93,
							32
						]
					]
				]
			}
		}
	},
	{
		"minX": -97.646484375,
		"minY": 43.197167282501276,
		"maxX": -88.681640625,
		"maxY": 48.28319289548349,
		"index": 2,
		"feature": {
			"type": "Feature",
			"properties": {},
			"geometry": {
				"type": "Polygon",
				"coordinates": [
					[
						[
							-97.646484375,
							43.197167282501276
						],
						[
							-88.681640625,
							43.197167282501276
						],
						[
							-88.681640625,
							48.28319289548349
						],
						[
							-97.646484375,
							48.28319289548349
						],
						[
							-97.646484375,
							43.197167282501276
						]
					]
				]
			}
		}
	}
]
```

DenisCarriere:
@mourner Would it be possible to "hack/tweak" RBush (within `@turf/index`) to insert a `GeoJSON Feature` instead of a BBox/minX & minY Object?

Something like this:

```javascript
var tree = turf.index(featureCollection);
var search = tree.search(polygon);
//=search Array<Feature> (empty array if no results found) or even output as a FeatureCollection
```

mourner:
@DenisCarriere there should be no performance difference between including feature references and number indices, since the first are just references — the objects themselves are not copied in memory. I didn't realize you're returning the RBush object itself — I assumed there's some kind of wrapper.

As an option, you extend RBush and override key methods so that they would map the input features in methods like `insert`/`load`, and map the output in methods like `search`. This could get a bit complicated though, so maybe the current API is fine. It's just that so far one of the API requirements for public Turf modules has been for it to be geographic, not implementation-specific. There are some exceptions though, and maybe it's another case where it's justified. If @morganherlocker is 👍 on the API, I'm 👍 merging.

morganherlocker:
😬 I'm a little 👎 on this for the following reasons:

- It creates a stateful object (tree), unlike any other turf module, which has been a very conscious choice. Every other module follows the strict rule of `data in` => `data out`.
- rbush already has an *excellent*, easy to use API (I'm all for using rbush directly internally)
- `index` is a very "assertive" namespace and I think overstates the general nature of an rtree API. Users may feel that this is always how their data should be indexed, which is not true. I use rbush in conjunction with turf and very rarely follow this exact pattern (ie: I use rbush all the time for indexing each segment of a linestring, but would not ever want to create a new geojson object for each segment due to RAM and CPU overhead). Other times, an rtree is not the best choice at all.

In cases like this, I almost always think it's best to leave this choice up to the user. The choice of an index is closely tied to the type of data being processed, as well as the environment where the processing is occurring. For example, if turf is being run in node against a dataset > 10GB, I would recommend using an rtree or similar index powered by a disk-based database like leveldb. In browser environments this would not be an option. In cases with < ~500MB, an in memory rtree like the one proposed here might be a perfect option.... unless you end up with overhead from keeping more full geojson objects in memory than you need. I think there are simply too many factors to build a generic geospatial index API that will work for everyone (unlike "calculate a bbox for a feature" or "measure the distance of a linestring" -- tightly scoped problems).

( @DenisCarriere 🙇 this is my least favorite type of comment to type. Great work regardless of my opinion on scope )

DenisCarriere:
@mourner 👍 Thanks for the advice, my last commit https://github.com/Turfjs/turf/pull/609/commits/f36aaab4479ac5f218cd5c0bbd59adef658626f6 should have a pure GeoJSON implementation.

- `load(Feature[])`
- `insert(Feature)`
- `search(Feature)` => FeatureCollection
- `all()` => FeatureCollection

DenisCarriere:
@morganherlocker No worries, it was fun to write! I learnt a lot from trying to implement a GeoJSON RBush implementation.

Have a look at the "new" GeoJSON commits I just passed, maybe you would change your mind.

If not.. no worries! 👍 

Here's a quick example of how it would work:

```javascript
import * as random from '@turf/random'
import * as index from '../'

// Fixtures
const points = random('points', 1000);
const polygons = random('polygons', 1000);

// Build Tree
const tree = index(points)

// Find All
const all = tree.all()
//= all => FeatureCollection
all.features.map(feature => {
    feature.properties
    feature.geometry
})

// Search
const search = tree.search(points.features[0])
//=search => FeatureCollection

// Insert one
tree.insert(points.features[0])
```

DenisCarriere:
Last commit https://github.com/Turfjs/turf/pull/609/commits/0bdf469f81bc853fe2fbbefd7913dd5996f9dce0 would make `@turf/index` a pure GeoJSON implementation of RBush.

GeoJSON as input => GeoJSON as output.

@morganherlocker we can discuss more about this over coffee :)

We could even call `@turf/rbush` if we don't want to call it `index`.

It just seems like there's MANY TurfJS modules that NEED some kind of spatial index to easily speed up the performance and to implement a basic RBush setup requires 15-20 LOC with 3+ dependencies, this pattern repeats itself over and over... this is the reason why I created this module.

We need basic building blocks for spatial indexing, lots has changed from the first commit to the last, I'm still open to suggestions (Thanks for all the quick feedback! 👍 ).

DenisCarriere:
Just a quick comment on:

>(ie: I use rbush all the time for indexing each segment of a linestring, but would not ever want to create a new geojson object for each segment due to RAM and CPU overhead). Other times, an rtree is not the best choice at all.

I've had to do the same thing for `@turf/mask` & `@turf/line-intersect`, internally I built my own custom "FeatureCollection" with the 2 coordinate line segments and then passed it to RBush. However... I've had to also include the `featureEach()`, `bbox()` and define the input as `minX` & `maxY` parameters, all of this syntax was becoming very repetitive.

There should be a few modules that are built for the internal workings of TurfJS such as `@turf/helpers` `@turf/invariant` & `@turf/meta`. If we ever want to remove `jsts` from our TurfJS stack we need to implement some kind of spatial index module (at least for some of the core modules).

DenisCarriere:
Quick code example of how `@turf/rbush` is currently working:

```javascript
const random = require('@turf/random')
const rbush = require('@turf/rbush')

// Fixtures
const points = random('points', 2);
const point = points.features[0];
const polygons = random('polygons', 2);
const polygon = polygons.features[0];

// Initialize GeoJSON RBush Tree
const tree = rbush()

// Load Tree with a FeatureCollection
tree.load(points);
tree.load(polygons);

// Insert by Feature
tree.insert(point)
tree.insert(polygon)

// Find All (returns FeatureCollection)
const all = tree.all()

// Search by Feature (returns FeatureCollection)
const search = tree.search(polygon)

// Collides by Feature (returns FeatureCollection)
const collides = tree.collides(polygon)

// Remove by Feature
tree.remove(point)
tree.remove(polygon)
```

DenisCarriere:
Thanks for the all the feedback @mourner & @morganherlocker.

After talking with Morgan, he advised me this type of module would better suited outside of Turf based on his previous comments mentioned above. 👍 

Which means...

## New library! [`geojson-rbush`](https://github.com/DenisCarriere/geojson-rbush) 🎉 

@mourner I took your advice and I'm building on top of RBush using `prototype.<method>.call` which you can see from the [`index.js`](https://github.com/DenisCarriere/geojson-rbush/blob/master/index.js).

Closing this PR

PR #617, opened 2017-03-15
Ignore the Calculate Final Bearing function so it doesn't appear in docs

rowanwins:
Currently the calculatefinalbearing function is appearing in the [docs](http://turfjs.org/docs/#calculatefinalbearing), it needs to be ignored by jsdocs


DenisCarriere:
👍 Yep, to hide those functions from appearing in the Turf docs, we've been using `@private` instead of `@ignore`.

rowanwins:
ah cool, thanks for the tip

PR #618, opened 2017-03-15
Initial commit of some new example maps

rowanwins:
Gday @DenisCarriere 

I've started populating some of the module jsdocs with the `addToMap` tag, just thought I'd get a quick review before proceeding with the rest.

A few things I've noticed
- In some of the docs there are things for styling the output features (eg turf-center), would we like to retain this?
- In others there is stuff which combines the input and output into a featureCollection (eg turf-bezier), I'd suggest we can tidy this out and make the docs clearer.

Thoughts?

### DIFF
--- a/packages/turf-bezier/index.js
+++ b/packages/turf-bezier/index.js
@@ -13,6 +13,7 @@ var Spline = require('./spline.js');
  * @param {number} [resolution=10000] time in milliseconds between points
  * @param {number} [sharpness=0.85] a measure of how curvy the path should be between splines
  * @returns {Feature<LineString>} curved line
+ * @addToMap line, curved
  * @example
  * var line = {
  *   "type": "Feature",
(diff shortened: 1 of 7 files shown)


packages/turf-bezier/index.js
```diff
@@ -13,6 +13,7 @@ var Spline = require('./spline.js');
  * @param {number} [resolution=10000] time in milliseconds between points
  * @param {number} [sharpness=0.85] a measure of how curvy the path should be between splines
  * @returns {Feature<LineString>} curved line
+ * @addToMap line, curved
```
DenisCarriere: Maybe only showing the final curved line would be useful here. Having both would most likely clutter the example map and be confusing (it's hard to display a before & after with a single map).

DenisCarriere:
@rowanwins I'm 👍 making the `@example` tidier, eliminating any extra lines that are not related to the module itself (like creating a featureCollection is not necessary).

A lot of these map examples will be Trial and error... so this looks good! 👍

packages/turf-bezier/index.js
rowanwins: This was where I was thinking the layer control box might be handy. On some maps it's going to make sense to show both the input and output (I think bezier is a good example where it's helpful to show both). It should be very simple to introduce

Issue #613, opened 2017-03-12
Centered turf-point-grid


DenisCarriere:
Would be handy to make the points from `@turf/point-grid` centered (divided equally from the rows & columns).

**Solution**: Add an extra param `@param {boolean} [centered=false]` with the new changes.

**Before (`centered=false`)**

https://gist.github.com/anonymous/1d5a64be9ef8c8eb0b8757d56526a99f

![image](https://cloud.githubusercontent.com/assets/550895/23836562/a91f716c-0750-11e7-9d8c-23c96046b072.png)

**Proposed (`centered=true`)**

![image](https://cloud.githubusercontent.com/assets/550895/23836509/0a49fe0e-0750-11e7-9f7e-e0f3f24f2b3d.png)

Please provide the following when reporting an issue:

- [x] GeoJSON data as a [gist file](https://gist.github.com/) or [geojson.io](http://geojson.io/) (filename extension must be `.geojson`).

DenisCarriere:
Implemented by PR https://github.com/Turfjs/turf/pull/650

Issue #390, opened 2016-06-01
redo turf-isolines


morganherlocker:
The existing implementation is finicky and now relies on [outdated internal dependencies](https://github.com/Turfjs/turf/pull/383).

rowanwins:
Just noting here that @StefanoGitHub has a [pull request open](https://github.com/Turfjs/turf-isobands/pull/7) on this in the old turf-isobands repo.

stebogit:
@DenisCarriere do I understand correctly that the `isolines` module has currently issues/bug as well? Don't get exactly which ones though.
Would you like me to take a look into this and maybe trying to replace `conrec` with `marchingsquare.isolines`?

stebogit:
@DenisCarriere Looking at the module's test it seems to me that the output is incorrect... am I wrong?

Other consideration: what's actually the difference between an isoline and an isoband? 
The more I think about this it seems to me the isoline is just a particular case of isoband, where its two limits (upper and lower) coincide. 
Couldn't we just use the same algorithm for both `isobands` and `isolines`? It would be probably mainly a matter of output format, `MultiPolygons` vs. `LinearRing` or `LineString`.
In [this example](http://emptypipes.org/2015/07/22/contour-comparison/) for instance it's also clear that the `conrec` algorithm/library does not "group" isolines of the same level, considering each line indipendent. I don't think this is a desired result for whom needs isolines.

DenisCarriere:
Well ideally we should only have 1 module, the only difference is the Polygon & LineString outputs.

Once we get `isobands` up and running we could drop isolines (turf 4.x) in favor of isobands.

morganherlocker:
> Well ideally we should only have 1 module, the only difference is the Polygon & LineString outputs.

Note that we used to have isobands and isolines, but isobands was deprecated due to limitations in the conrec implementation that did not apply to LineString output. Having both might be useful, and turning one format into the other can be surprisingly tricky for people looking for a quick result.

PR #619, opened 2017-03-15
New @turf-isobands module with MarchingSquares

stebogit:
- [X] Run `npm test` at the sub modules where changes have occured.
- [ ] Run `npm run lint` to ensure code style at the turf module level. --> __not sure how/if it works__


stebogit:
@DenisCarriere 
I tried to run eslint but I got so many errors (like 18000)... Is it possible that they are not only caused by my code? I noticed that it's asked to use `AirBnB` style guide, which has recently deprecated ES5, while `turf` uses ES5 only. Isn't that an issue?

I don't know how travis works, could you please give me some hint on how to set/fix it?

Thanks

DenisCarriere:
@stebogit 👍 Thanks attempting this, the linting is done at the TurfJS root level.

Once you are in the root folder:

```
$ npm install
$ npm run lint
```

So far the `isobands` turf linting related errors are:

You can also see this in the build logs https://travis-ci.org/Turfjs/turf/builds/211360094

```
/home/travis/build/Turfjs/turf/packages/turf-isobands/index.js
   12:1   error  Expected JSDoc for 'pointGrid' but found 'grid'                                                valid-jsdoc
   38:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
   40:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
   54:37  error  Missing semicolon                                                                              semi
   56:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
   68:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
   71:57  error  'pointsByLat' is defined but never used                                                        no-unused-vars
   73:22  error  Expected to return a value in this function                                                    array-callback-return
   73:46  error  'pointArr' is defined but never used                                                           no-unused-vars
   78:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
   91:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
   92:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  102:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  112:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  117:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  118:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  121:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  126:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  132:42  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
  140:13  error  Strings must use singlequote                                                                   quotes
  143:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  145:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  146:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  158:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  164:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  166:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  169:1   error  Trailing spaces not allowed                                                                    no-trailing-spaces
  196:46  error  Expected '===' and instead saw '=='   
```

packages/turf-isobands/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: node_js
+node_js:
+  - "6"
+  - "5"
+  - "4"
+before_install:
+  - npm install -g npm@~3.8.9
+
```
DenisCarriere: This file is not needed since TurfJS has `.travis.yml` at the root level

packages/turf-isobands/.npmignore
```diff
@@ -0,0 +1,2 @@
+test
+coverage
```
DenisCarriere: Not needed since this is covered in the `package.json` using `"files": ["index.js", "index.d.ts"]`

packages/turf-isobands/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2016 Stefano Borghi
```
DenisCarriere: We've been using TurfJS as a generic name for all licenses (also 2017)

packages/turf-isobands/bench.js
```diff
@@ -0,0 +1,18 @@
...
+  .add('turf-isobands',function () {
+    isobands(points, 'elevation', 15, [5, 45, 55, 65, 85,  95, 105, 120, 180], false);
+  })
+  .on('cycle', function (event) {
+    console.log(String(event.target));
+  })
+  .on('complete', function () {
+    
+  })
+  .run();
```
DenisCarriere: Feel free to write in ES6 👍 `bench.js` & `test.js` is your only chance to write arrow functions :)

packages/turf-isobands/index.d.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference types="geojson" />
+
+type Points = GeoJSON.FeatureCollection<GeoJSON.Point>;
+type LineStrings = GeoJSON.FeatureCollection<GeoJSON.LineString>;
+
+/**
+ * http://turfjs.org/docs/#isolines
+ */
+declare function isobands(points: Points, z: string, resolution: number, breaks: Array<number>): LineStrings;
+declare namespace isolines { }
```
DenisCarriere: namespace must be the same name as the function name.

This breaks the typescript definition.

packages/turf-isobands/index.js
```diff
@@ -0,0 +1,269 @@
...
+
+var turfHelpers = require('@turf/helpers');
+var turfFeaturecollection = turfHelpers.featureCollection;
+var turfPolygon = turfHelpers.polygon;
+var turfMultiPolygon = turfHelpers.multiPolygon;
+var turfExplode = require('@turf/explode');
+var turfInside = require('@turf/inside');
+var turfArea = require('@turf/area');
+
+/*******************************************************************
```
DenisCarriere: Follow the syntax of all other JSDocs Turf modules, simply using `/**` is enough.

packages/turf-isobands/index.js
```diff
@@ -0,0 +1,269 @@
...
+ * Takes a grid ({@link FeatureCollection}) of {@link Point} features with z-values and an array of
+ * value breaks and generates filled contour isobands.
+ *
+ *
+ * @module turf/isobands
+ * @category interpolation
```
DenisCarriere: `@module` & `@category` is optional, this would be defined in the `documentation.yml` in `turf-www`.

https://github.com/Turfjs/turf-www/blob/gh-pages/documentation.yml

However `@name` is required = "isobands"

packages/turf-isobands/index.js
```diff
@@ -0,0 +1,269 @@
...
+ * @param {FeatureCollection} grid of points, a FeatureCollection of {@link Point} features
+ * @param {string} z the property name in `points` from which z-values will be pulled
+ * @param {Array<number>} breaks where to draw contours
+ * @returns {FeatureCollection} a FeatureCollection of {@link MultiPolygon} features representing isobands
```
DenisCarriere: Define `FeatureCollection<MultiPolygon>` if that's the type that is returned

packages/turf-isobands/index.js
```diff
@@ -0,0 +1,269 @@
...
+ * @example
+ * // create random points with random
```
DenisCarriere: Explicitly define your examples since the JSDocs `@example` is actually being tested and will throw an error in the `turf-www` build.

packages/turf-isobands/index.js
```diff
@@ -0,0 +1,269 @@
...
+ * var cellWidth = 5;
+ * var units = 'miles';
+ * var pointGrid = turf.pointGrid(extent, cellWidth, units);
+ * for (var i = 0; i < pointGrid.features.length; i++) {
+ *     pointGrid.features[i].properties.elevation = Math.random() * 10;
+ * }
+ * var breaks = [0, 5, 8.5];
+ * var isobands = turf.isobands(pointGrid, 'z', breaks);
+ * //=isobands
+ ******************************************************************/
```
DenisCarriere: Same as above `**/`

stebogit:
thanks!
I see also `marchingsquare.js` has a bunch of styling issues, is it acceptable being that an external (minified) library?

packages/turf-isobands/index.js
```diff
@@ -0,0 +1,269 @@
...
+module.exports = function (pointGrid, z, breaks) {
+    
+    var points = pointGrid.features;
+    
+    /*####################################
+     divide points in pointGrid by latitude, creating a 2-dimensional data grid
+     ####################################*/
+    var pointsByLatitude = {};
+    for (var j = 0; j < points.length; j++) {
```
DenisCarriere: Try using one of the `@turf/meta` functions, `coordEach()` or `featureEach()`

packages/turf-isobands/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "turf-isobands",
+  "version": "2.0.0",
+  "description": "turf isobands module",
+  "main": "index.js",
```
DenisCarriere: Include `"types":"index.d.ts"` (next to main instead of at the bottom) & `"files": ["index.js", "index.d.ts"]`

packages/turf-isobands/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "scripts": {
+    "test": "tape test.js"
```
DenisCarriere: add benchmark to scripts

packages/turf-isobands/test.js
```diff
@@ -0,0 +1,17 @@
+var test = require('tape');
+var fs = require('fs');
+var isobands = require('./');
+
+test('isobands', function(t){
+  var points = JSON.parse(fs.readFileSync(__dirname + '/geojson/Points.geojson'));
```
DenisCarriere: Try to make your tests have dynamic data loading instead of hard coded fixtures in your tests.
That way someone can expand easily the test cases by simply adding a new `.geojson` to the `test/in`.

`@turf/union` has a good example of that:
https://github.com/Turfjs/turf/blob/master/packages/turf-union/test.js

DenisCarriere:
@stebogit As for the MarchingSquare.js file, I would not include it as a minified file, does it exists on npm?

stebogit:
@DenisCarriere 
no, `MarchingSquare.js` does not have an npm module (yet @RaumZeit ?). Is there a way to exclude it when linting?
Also, the name of the MS function `Isobands` is uppercase, which eslint doesn't like...

packages/turf-isobands/index.js
stebogit: Could you please elaborate a bit on this? First time I see this feature 😄 
What is missing? I just copied the JSDoc from the original version of the module

packages/turf-isobands/.npmignore
stebogit: What exactly is not needed? the `.npmignore` file?

packages/turf-isobands/package.json
stebogit: is `"bench": "node bench.js"` ok?

packages/turf-isobands/.npmignore
DenisCarriere: The entire `.npmignore` file is not required, I've changed every module in Turf so all of them have the same syntax & file structure.

Use any turf module as your starting point to propose a new module (except the `index.js`).
https://github.com/Turfjs/turf/tree/master/packages/turf-along

stebogit: 👍

DenisCarriere:
Quick eslint trick is you can include this around your code to ignore "valid" lines that cause errors.

```javascript
/*eslint-disable*/
var isobands = MarchingSquaresJS.IsoBands(gridData, lowerBand, upperBand - lowerBand);
/*eslint-enable*/
```

DenisCarriere:
@stebogit Glancing at the repo, I'm thinking we should drop (make optional) the `@param {string} z` in favor of the `geometry.coordinates` (x, y, z).

The GeoJSON specs only allows (longitude, latitude, elevation/altitude) to be a valid geometry.
http://geojson.org/geojson-spec.html#positions

**Proposed valid input**

```javascript
var points = {
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          7,
          42,
          300
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          10,
          45,
          230
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Point",
        "coordinates": [
          5,
          45,
          215
        ]
      }
    }
  ]
}
var results = isobands(points, [200, 250, 300])
```

packages/turf-isobands/test.js
```diff
@@ -2,30 +2,29 @@ const test = require('tape');
...
+    fixtures.forEach(({name, geojson, filename}) => {
+        const results = isobands(geojson, 'elevation', [0, 3, 5, 7, 10]);
 
-        t.ok(isobanded.features, 'should take a set of points with z values and output a set of filled contour MultiPolygons');
-        t.equal(isobanded.features[0].geometry.type, 'MultiPolygon');
-
-        t.deepEqual(isobanded, load.sync(directories.out + 'isobands' + idx), input.name);
+        if (process.env.REGEN) write.sync(directories.out + filename, results);
+        t.equal(results.features[0].geometry.type, 'MultiPolygon', name + ' geometry=MultiPolygon');
+        t.deepEqual(results, load.sync(directories.out + filename), name);
```
stebogit: this way the code looks definitely nicer, but the content of the output file, `points1.geojson`, is not a collection of `points`, but actually of `multipolygons` or `isobands`.
don't you think it might be a little confusing?

stebogit:
I get your point.
However the input are points, which might have multiple properties you want to chart (I'm not charting elevation for example, but estimated count of bugs 😄), so specifying the name might be useful.
If made optional the default might certainly be `elevation` or `altitude`

packages/turf-isobands/test.js
DenisCarriere: The `test/in` & `test/out` should be the same filename. That way it's clear what goes in and what goes out.

You can even call `points.geojson` `italy.geojson` and then the output would be `test/out/italy.geojson`

DenisCarriere: This becomes more evident when you start adding more and more files in the `test/in`, for example `@turf/union` has many files.

https://github.com/Turfjs/turf/tree/master/packages/turf-union/test/in

stebogit: then I'd call them `input1.geojson`, `input2.geojson` and so on 😄 
not a biggie though

DenisCarriere:
@stebogit 👍 Valid point, I think having it optional would be a good idea, or at least have that param at the end of the function.

```javascript
function (pointGrid, breaks, z)
```

We could have `z` as default `z="elevation"` or even better read the `z` coordinate from `[x,y,z]`.

We can add that logic at the beginning as part of the user input validation by reading the first point entry and figuring out if his `FeatureCollection<Point>` is valid or not (if he is using `properties` or `z` coordinates).

stebogit:
@DenisCarriere great! 🎉 
`z` at the end of the function makes it easy and effective.
Never thought about having an input validation, good point! 👍

DenisCarriere:
@stebogit I will have a stab at it over the weekend, also there's a lot of "complicated" logic inside the `index.js` that can be replaced by [`@turf/meta`](https://github.com/Turfjs/turf/tree/master/packages/turf-meta) methods (`coordEach()`, `featureEach()`, etc..).

stebogit:
@DenisCarriere Thanks, I'm not familiar with the `meta` functions and I haven't worked on my contour project so it might take a while for me to correctly make the changes. 
I believe you are more into `turf`/`GeoJSON` standards and details for input validation.
Let me know if/how I can help though

DenisCarriere:
No worries, I have this habit of refactoring code until the most minimal amounts of lines are needed.

At least we have a "functional" prototype of `isobands`, the next steps would simply be making sure the code is running as fast as it possibly can and without any memory leaks.

DenisCarriere:
@stebogit another idea for the `breaks` param. We could have both of the following as valid inputs:
- `Array<number>` (explicitly define each break)
- `number` (defined each break by a fix interval)

**Scenario**: User wants to create contour lines based on 25 meter intervals (elevation units must also be in meters or the desired unit which is the same as the breaks).

stebogit:
@DenisCarriere splendid idea!

RaumZeit:
MarchingSquaresJS is now available through npm as [marchingsquares](https://www.npmjs.com/package/marchingsquares) module. Note, however, that due to npm naming conventions I've reverted the public API function names to lower camel case since version 1.2.0 of MarchingSquaresJS, i.e. the `IsoBands` function is now named `isoBands`.

DenisCarriere:
@RaumZeit Awesome! Just tested the new npm module and it works just fine 👍

DenisCarriere:
After thinking about this, the `breaks` as `Array<number>|number` we can save that for later (the user could easily define that), no need to complicate the params at the moment.

DenisCarriere:
@stebogit big issue I just noticed, this doesn't work with [random points](https://github.com/stebogit/turf/blob/bac2c7556de3183a94a5fa0a20f18f923e9d8fb5/packages/turf-isobands/test/in/points-random.geojson)?

I'm pretty sure it has something to do when generating the `GridData`.

stebogit:
@DenisCarriere the [MarchingSquare algorithm](https://en.wikipedia.org/wiki/Marching_squares) requires, by definition, a 2D gridded data as input.
So we might integrate an input handling to transform random points into a grid into the isobands module, or we might create a helper function/module (maybe it's already there?).
I'd go for the second one, if not already available.

DenisCarriere:
@stebogit I don't think something like that already exists in TurfJS, however this type of grid seems very specific to Isobands.
Something else to think about is those newly created points, are we interpolating the elevation data between the points within the closest proxity (at first, lets just get the 2D grid, next would be to add the interpolation).

DenisCarriere:
I was looking at `@turf/isolines` and it seems it has to do the same 2D grid, it's simply using `@turf/point-grid` based on resolution. This looks like it would be easiest way to implement support for random points.

https://github.com/Turfjs/turf/blob/master/packages/turf-isolines/index.js#L42-L55

stebogit:
@DenisCarriere sorry The newly created points would be the gridded ones created from the random ones?
If that's the case, I believe we should not interpolate any value, but the "randomToGrid" function should able to generate a 2D grid with the right resolution so that all the random points well be actually points of the grid (no original points are left "outside" the resulting grid). 
That means the grid might get huge in some cases...

stebogit:
Well sure, if you need to generate a correct input.
But then I don't get the benefit of using random points. Just for example/testing purpose?

I for one was thinking about allowing any type of set of point as input of `isobands`; if actually of any usefulness.
I was trying to think about a use case where you would have a random set of 3D points that you might want to "traspose" into 2D, to justify a "randomToGrid" module.
If you where to represent a 3D function into a 2D plain you would have been doing (is that correct english?) what I am doing with my data, so you would probably already have all the necessary points on a grid (i.e. the space or their projection on the plain). So as you said it seems to be a very specific format for `isobands` and `isolines`.

DenisCarriere:
The benefit of testing for random points is because that's what you'll ALWAYS get from users, you can't expect users to perfectly give you a point grid.

stebogit:
While I totally agree on a general base, here we are basically dealing with a quite specific problem/need: representing a 3D surface on a 2D one. To be able to describe any surface you need a grid of coordinates, so the user in this case should already have the grid and so the "right" input.
What would be a different application where you have only a bunch of points and you want isolines on horizontal plain(s) out of them? I can't think of one, but I have not a bunch of experience in maps or data representation.
I see it as trying to calculate the area of a non closed line: either you close the line before passing it to `turf-area`, for instance, or `turf-area` closes it for you. In my understanding `turf-area` will return `null` if the geometry is not closed, i.e. it does not close the geometry for you (correctly, in my opinion).
So maybe a `pointsToGrid` module might be useful? 😄

DenisCarriere:
👍 it would be handy having some kind of `pointsToGrid` module. It would look something like this: https://github.com/Turfjs/turf/blob/master/packages/turf-isolines/index.js#L38-L77

stebogit:
Ok I'll work on that if it's ok with you.
What do you think about this approach?
```
[...]
    var pointsByLatitude = createPointsByLatitude(points);
    if (!isMatrix(pointsByLatitude)) {
        pointsByLatitude = createPointGrid(pointsByLatitude);
    }
    var gridData = createGridData(pointsByLatitude, property);
[...]
```
where `isMatrix` and `createPointGrid` are new methods

DenisCarriere:
That works, feel free to continue committing on this branch, I'll do the same. Just make sure when you start do a `git pull` to not have any conflicts.

RaumZeit:
@stebogit @DenisCarriere Sorry for me stalking your progress here ;) But I can not resist telling you that when I originally implemented the Marching Squares algorithm I also stumbled across the problem of creating a regular grid out of randomly distributed points. This lead to my  [Data2GridJS](https://github.com/RaumZeit/Data2Grid.js) module...
For that purpose, I may refer to a quite old publication [A Comparison of Interpolation Methods for Sparse Data: Application to Wind and Concentration Fields](http://journals.ametsoc.org/doi/abs/10.1175/1520-0450(1979)018%3C0761%3AACOIMF%3E2.0.CO%3B2) where I took some ideas from. Maybe this can be of interest for you as well? The algorithms in this paper are quite simple to implement and the results, depending on parametrization, are not that bad.

DenisCarriere:
@RaumZeit Thanks for the references, could you shine some light on the exact input & outputs of your `Data2Grid` library, could you provide a Quickstart example of how to use it.

RaumZeit:
Input to my [Data2Grid module](https://github.com/RaumZeit/Data2Grid.js) is an array of `x,y,z triples`, an `options object` that specifies the mode of transformation (interpolation method, grid dimensions, and other parameters), and a `resultsCallback function`. This callback then receives the gridded data once interpolation is finished.
I just added a [small example](https://github.com/RaumZeit/Data2Grid.js/tree/master/example) to my repo. However, the module is only for browser usage, and I think it actually doesn't support multiple grid conversions at the same time yet. I have to check that again, it's been a while that I've been playing arround with this module

stebogit:
@RaumZeit that's great! Thanks for stalking us then 😄 
It seems to me the two outputs are quite different in resolution. What's the difference between the two, maybe the Barnes method is much faster?
Looking quickly at `Data2Grid.js`, does it depend on `d3.js` (e.g. line 119 and 122)?

RaumZeit:
@stebogit Actually, it shouldn't. But you are right, there is a remnant function still included in `Data2Grid.js`. The `applyNeighborMethod` function needs to be thrown out. In my webapp where I use `Data2GridJS` I only apply the barnes or objective method. As far as I remember, Barnes is slower compared to the objective method. Another difference is that the Barnes method produces values for each grid point, while the objective method only smooths the values of an actual data point into the neighboring grid points. The radius of smoothing for both methods can be controlled by the `xr` and `yr` properties of the `options` object. In my example, default smoothing values are used which usually adapt the radius to the density of the input data. Also, please note, that my implementation was made for the browser. It contains several lines of code that one want to get rid of in a `Node.js` context, I guess. Anyway, I just wanted to suggest these two algorithms in case you want to start a new turf module for converting scattered 2D data into a regular scalar field.

stebogit:
@RaumZeit I played a little with your code to get an idea of how it works and I quickly created an additional [example](https://github.com/stebogit/Data2Grid.js/blob/98d02d52473c0e185b303adf8d73a39601d765f3/example/main.js), isolating the Objective method.
The result does not make sense to me, what am I doing wrong?

__p.s.__ not sure if this is the right place to discuss this, let me know if there is a better way

RaumZeit:
... just made some comments on your [latest commit](https://github.com/stebogit/Data2Grid.js/commit/98d02d52473c0e185b303adf8d73a39601d765f3)... we can continue discussing there

stebogit:
@DenisCarriere I added a couple of methods to handle random points using the snippet of code you pointed out from `isolines`. Please note the random input gets inevitably distorted when transformed into a `pointGrid`. I played also a little with Ronny's ` Data2Grid` module and also there a scattered input gets distorted, but I found more difficult to find a good way to tune the parameters for that algorithm to adapt to _any_ generic sparse input.
Let me know if you find any alternative solution to this issue.

Due to the distortion introduced by the conversion into grid, and the fact the I don't see any application requiring converting random points into contours, I still believe this module should only accept a `pointGrid`.
I would probably instead create a method/module to parse a bidimensional array (matrix) to `pointGrid`, useful for all the modules that require this kind of input (`isolines` for instance) and to allow users to easily map a 3D function in 2D.

stebogit:
@DenisCarriere I also noticed an error of the module: if you compare input and output (e.g. the `points-elevation.geojson` test) you should see the latter is "smaller", i.e. seems to exclude the top row and the rightmost column of points in the grid. This happens with both grid and random input.
I am not sure if it actually ignores those points or instead it erroneously rescales the contours, I could not find any error.
What do you think?
screen shot 2017-03-27 at 2 21 36 am
screen shot 2017-03-27 at 2 21 50 am

DenisCarriere:
I remember having this same issue with `@turf/square-grid` https://github.com/Turfjs/turf/pull/612, it would create the grid outside the giving extent.

We could include `completelyWithin` in the param of `@turf/point-grid` and that should solve the points being outside the `bbox`.

Also having the point grid centered might help the `isobands` https://github.com/Turfjs/turf/issues/613.

DenisCarriere:
@stebogit This was also happening before any refactoring https://github.com/Turfjs/turf/tree/c96d13ed4193ca2537adb519b28d555b8b7c1eb7/packages/turf-isobands/test/out (same output).

It's hard to pin-point exactly what's going on with this much source code, I would suggest making the `input` & `output` as GeoJSON that way you can save/preview the results each step of the way.

Keep making smaller and smaller functions that you can test individually is the best way to find out where the problem is.

stebogit:
@DenisCarriere I agree, the error might be due to #612 and probably #613 , in fact data is offset by one column and one row when transformed into `pointGrid`.
Besides, I recently noticed the same issue is also conditioning the current `isolines` output (see input/output comparison, #390 ).

DenisCarriere:
This is definitely going to be a tricky module to implement (I still have hope!! 👍 @stebogit Keep it up)

DenisCarriere:
@stebogit Let me know if you are happy to merge this module, looks good to me 👍 .

Always room for improvements/fixes in future releases.

Also, feel free to add your `matrix-to-grid` & `grid-to-matrix` module once it's stable.

stebogit:
@DenisCarriere personally I would remove the random input interpolation (the `createPointGrid ` function is not actually working correctly), just requiring the input to be a `pointGrid`. I'll work on fixing the output offset and I want to create few more tests to make sure the output is actually right (`matrix-to-grid` will be useful).

> Also, feel free to add your matrix-to-grid & grid-to-matrix module once it's stable.

Do you mean in this module? My idea was to use them in [`@turf/interpolate`](https://github.com/Turfjs/turf/compare/turf-interpolate) (still needs work..), which I thought would allow users to create/clean the input for `isobands` and `isolines`, leaving the input handling outside them. Or, we might then include `@turf/interpolate` once completed.

DenisCarriere:
Oh I thought that [`createGridData`](https://github.com/stebogit/turf/blob/2933158fe5b03bb535eb6afd145c4725722857ea/packages/turf-isobands/index.js#L255-L293) was `grid-to-matrix`, however it doesn't look like it's exactly the same thing.

👍 Yep, you can remove the random fixture, we can always add it later on in another minor release.

Looking forward to publishing this module! 🚀 

@stebogit Have a last glance at this module and merge this PR whenever you feel comfortable with it.

stebogit:
@DenisCarriere I created a few more tests for `isobands` and I can tell there are few things that are not quite right, the output is not reliable. `MarchingSquare` even seems to return some unexpected outputs sometimes, but I'm not sure yet.
I'll keep working on this, but unfortunately I believe it's not gonna be quick... sorry 😞

DenisCarriere:
@stebogit No rush... (take all the time in the world 👍 ).

> even seems to return some unexpected outputs sometimes

I agree, `MarchingSquare` is a bit foreign to me, you're the one with the most experience using this library.

Keep it up!

stebogit:
Fixed output bug, next will add more tests with graphically easy-to-read grid/matrix input values, then will evaluate performances.

stebogit:
@DenisCarriere please take a look at the module, and feel free to add more tests.
Now that I think about it, for some reason I removed the use of `z-coordinate` 😅 , we probably want to restore it.
I think we could consider this done as a first working version.
Lots of room for improvements though, in particular in performance and output validation for more strict maps like Google.

DenisCarriere:
Works for me! 👍 This will be released in the next minor release v4.3 (aiming to get that done tonight)

PR #622, opened 2017-03-16
Added category for translations of turfjs in other languages

shashanktomar:
Turfjs is used widely for geospatial calculations. We at [hammerhead](http://www.hammerhead.io/) translated it to golang, we use this golang implementation in three different codebases: server-side, android and ios.

Please fill in this template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,3 +89,6 @@ var point2 = {
   properties: {}
 };
 ```
+### Other languages
+
+Golang: <a href='https://github.com/hammerheadnav/turfgo'>TurfGo</a>


DenisCarriere:
Thanks for the efforts, however `turfgo` isn't a mirror of what is currently implemented in TurfJS.

In it's current state, I don't think we should be including this. 👎

shashanktomar:
Hi @DenisCarriere, thanks for considering it. All our existing implementations are exact replication of Turfjs features, including the test cases and the remaining one are under Feature tag in issues. Can you please elaborate a bit more on not being the exact mirror of TurfJs, it would help us in taking it forward.

DenisCarriere:
You're missing many of the core modules `@turf/meta`,`@turf/helpers`, `@turf/invariant` and so on. I would also recommend adding more documentation in your README.md
Keep up the hard work, however there seems to be a lot more work needed.

Another thing I noticed, you're [`Point` type](https://godoc.org/github.com/hammerheadnav/turfgo#Point) is using LatLng when the GeoJSON spec is suppose to be LngLat (xy), this would create massive confusion especially since TurfJS is entirely built on the GeoJSON spec.

Sorry, can't merge this at the moment.

shashanktomar:
Thanks a lot for the pointers @DenisCarriere we will keep contributing to the code.

Issue #649, opened 2017-04-07
ArcByCenterPoint


Adraesh:
Hi!

I did not find any method provided by the Turf lib which handle my case.

In fact, I have:

- startAngle (degrees)
- endAngle (degrees)
- radius (kilometers)
- centerPoint (long, lat) Center of the whole circle.

Basically I want to generate and ArcByCenterPoint with those data.

Is it possible ?

Thanks.

Adraesh:
Hi!

I just tested the turf-sector method and obviously there is a misunderstanding. Well the mode itself is working perfectly but this is not what I asked.

<img width="611" alt="screen shot 2017-04-09 at 6 07 22 pm" src="https://cloud.githubusercontent.com/assets/3873988/24838411/035c5504-1d50-11e7-8cb3-2bf5e9e6c18e.png">

In fact I got a circle sector but what I want is ONLY the line Arc not the entire sector :)

Thanks!

DenisCarriere:
I don't think there's any misunderstanding, what @stebogit implemented lines up perfectly with https://github.com/Turfjs/turf/issues/155.

This new module request would be called `@turf/line-arc` which would have this as a result (red line):

![image](https://cloud.githubusercontent.com/assets/550895/24838702/44ffa716-1d1b-11e7-8322-69b2402dcbbd.png)

@Adraesh Feel free to send a PR to TurfJS if you want specific modules implemented.

Adraesh:
Thanks Denis, but what I wanted to say is that my post can't be referenced to the turf-sector module newly implemented :)

stebogit:
@DenisCarriere @Adraesh I can easily implement a `turf-arc-line` module simply extracting a currently private function inside the `sector` module.

Adraesh:
Hey stebogit!

I believe you are talking about this method: 

function getArcLine(center, radius, angle1, angle2, steps, units)

I would be really appreciated if you can do that.

Thank you very much.

DenisCarriere:
👍 @stebogit Sounds like a plan, afterwards you can add `@turf/line-arc` as a dependency to `@turf/sector`.

DenisCarriere:
Thanks to @stebogit [`@turf/line-arc`](https://www.npmjs.com/package/@turf/line-arc) is now available on npm.

stebogit:
Tanks also @DenisCarriere! 👍 
I'll add `@turf/line-arc` as a dependency to `@turf/sector` then.

[...]

Issue #155, opened 2014-12-05
turf-sector


morganherlocker:
Need to make these for rounded miter joins

![](http://kineme.net/files/composition/gtoledo3/Pacman%20CoreImage.png)

``` js
sector(centerPoint, bearing1, bearing2, radius, units)
```

rowanwins:
Now available in [@turf/sector](http://turfjs.org/docs/#sector)

DenisCarriere:
Nom nom! Pacman 👍 This module put a huge smile on my face when we implemented this!

![image](https://cloud.githubusercontent.com/assets/550895/25118668/40433938-23e5-11e7-80af-f644e917afee.png)

Cheers to @morganherlocker @rowanwins @stebogit !

PR #653, opened 2017-04-09
New module @turf/sector

stebogit:
@DenisCarriere I implemented the `turf-sector` module (related to #155 and #649).
However I did not create the `README` file cause I thought it was created automatically from the documentation in the `index.js`, but it didn't. Should I create it manually?
Not sure if the `index.d.ts` and `bench.js` are correct, as I'm not familiar with them.

Is that ok if I created the branch in this repo or should I work on my forked one?


Adraesh:
Thank for the fast implementation.

Is it already ready in the turf.min.js ?

Thanks.

stebogit:
@Adraesh I just happened to be implementing it at the right moment apparently 😄 
It is not yet an official module, just created the PR. But you might want to try it as npm module and see if you find any issue with it.
BTW I don't really know how the process works to have it in the min version.

DenisCarriere:
> However I did not create the README file cause I thought it was created automatically from the documentation in the index.js, but it didn't. Should I create it manually?

It is created by the `index.js`, however when creating a new modules you would need to initially run [`/scripts/generate-readme`](https://github.com/Turfjs/turf/blob/master/scripts/generate-readmes).

> Not sure if the index.d.ts and bench.js are correct, as I'm not familiar with them.

- I can help review the Typescript definition.
- Benchmark is useful to determine the efficiency of the module, you can almost copy-paste any recent module for the benchmark template (ex: [`turf-bbox-clip`](https://github.com/Turfjs/turf/blob/master/packages/turf-bbox-clip/bench.js)

> Is that ok if I created the branch in this repo or should I work on my forked one?

Perfectly fine what you did, for every new PR generate a new forked branches from Turf's `master` branch.

CC: @stebogit

DenisCarriere:
@stebogit Look over your source code! 👍 I'm going to send over a few commits to add dynamic testing from GeoJSON point features in `test/in` & `test/out`.

stebogit:
@DenisCarriere so this is the way to go for any (new) module now, right?
> dynamic testing from GeoJSON point features in `test/in` & `test/out`.

I started this module by cloning the `circle` module which doesn't have that

DenisCarriere:
@stebogit ideally every module should have dynamic testing and you shouldn't have to touch the source code of the `test.js`, simply add `geojson` features to the `test/in` is the easiest way that others can contribute to improve the test cases. Also it's impossible to visually see the results when they are not being generated in `test/out`.

Just added dynamic tests via https://github.com/Turfjs/turf/pull/653/commits/57b07178cecab920d96835ee279934c83add6f32.

Noticed an issue with:
- `bearing1`= -50
- `bearing2`= -15

image

And an error with:
- `bearing1`= -15
- `bearing2`= -50

```
Error: Each LinearRing of a Polygon must have 4 or more Positions.
    at module.exports.polygon (/Users/mac/Github/turf-sector/packages/turf-sector/node_m
odules/@turf/helpers/index.js:84:19)
```

DenisCarriere:
This might be handy to have https://github.com/Turfjs/turf/pull/653/commits/65c3c365c43668e9a6a5ebc21fb921e61fc17cac
- extract required params from GeoJSON properties (only applied to required params and must be undefined)
- more verbose validation (to know exactly which param required)

stebogit:
@DenisCarriere I don't understand exactly how to run `./scripts/generate-readmes`, would you mind adding some more details about this on `CONTRIBUTING.md`, I think it might be beneficial for other less expert developers, like me, as well.

DenisCarriere:
@stebogit Are you running on MacOS or Linux? You can simply just execute it via the terminal/command line.

```
$ ./scripts/generate-readmes
```
In windows you might need to add `node` in front of the script

```
$ node ./scripts/generate/readmes
```

DenisCarriere:
@stebogit You might be interested to know that there's already a `getCoords()` function in `@turf/invariant`, it's an easy way to get `coords` from Geometry or Feature or an Array.

https://github.com/Turfjs/turf/blob/master/packages/turf-invariant/index.js#L22-L50

DenisCarriere:
Added test case with error:

`sector-error.geojson` seems to be having some issues, it should be the inverse.
- bearing1 = -15 (345)
- bearing2 = -50 (310)

packages/turf-sector/bench.js
```diff
@@ -15,12 +15,14 @@ const fixtures = fs.readdirSync(directory).map(filename => {
...
- * sector4 x 37,871 ops/sec ±3.13% (86 runs sampled)
+ * sector-full-360 x 29,179 ops/sec ±2.42% (76 runs sampled)
+ * sector-greater-360 x 296,103 ops/sec ±10.92% (65 runs sampled)
+ * sector1 x 189,709 ops/sec ±3.46% (75 runs sampled)
+ * sector2 x 72,365 ops/sec ±2.11% (77 runs sampled)
+ * sector3 x 38,093 ops/sec ±2.97% (77 runs sampled)
+ * sector4 x 210,468 ops/sec ±2.58% (78 runs sampled)
+ * sector5 x 26,438 ops/sec ±9.98% (70 runs sampled)
+ * sector6 x 29,032 ops/sec ±2.36% (70 runs sampled)
  */
```
stebogit: @DenisCarriere I thought to run `bench.js` and it turned out it worked 😆  (I guess)
I don't understand though if the result is good, I mean how do I evaluate these numbers, what do they mean?

stebogit:
@DenisCarriere since we have now more than one module having angles/bearing as input, I think the `convertAngleTo360` function (used in both `turf-sector` and `turf-line-arc`) could be moved into `turf-helpers`, what do you think?
Maybe you can find a better name.

packages/turf-sector/bench.js
DenisCarriere: Well the results will vary based on each machine, however it's a good performance gauge when you are trying to tweak a specific function.

Let's say you want to know if `map()` is faster than `featureEach()`, you can run the benchmark and then change the function and run the benchmark again to see which one has better performance.

Also when refactoring existing code, you might want to know if your "new" code is actually improving the performance or decreasing it. My `@turf/buffer` attempt is a good example of refactoring gone bad 👎  https://github.com/Turfjs/turf/pull/656 (fail me)

DenisCarriere:
@stebogit 👍 for `convertAngleTo360` that would fit the theme of `@turf/helpers`, there's already a few degrees methods https://github.com/Turfjs/turf/blob/master/packages/turf-helpers/index.js#L277-L320.

I'm sure there's a few other TurfJS modules that have this type of function internally built in, you could poke around and see if you could include it in other modules.

packages/turf-sector/bench.js
stebogit: so higher `ops/sec` means better performance, right?
then you might want to re-run `bench.js` here, as I'm sure my old macbook air here at home is slower than your machine...

stebogit:
@DenisCarriere may I add this task as a new issue and assign it to myself then?
Are you ok with the name of the function? I'm not 100% satisfied, but I can't find really a better one...

packages/turf-sector/bench.js
DenisCarriere: Yes higher `ops/sec` is better.

The Benchmark results are approximate, it gives you a good idea where the module stands.

DenisCarriere:
@stebogit Yep, make a new issue or open up a new PR and briefly explain it.

Usually I let a PR sit for a few days and review it again with a fresh mindset, you might find a better name for `convertAngleTo360` in a couple days.

stebogit:
@DenisCarriere what about `bearingToAngle`? 
After some research I find out that the 'bearing' is the angle between (generally) the north and a generic line/direction (that's why bearing is -180 and +180 degrees!), while an angle is generally express between 0 and 360 degrees (here again we might choose north as reference).
This is probably obvious to most people, but I would add some details in the `turf-bearing` description just to be more clear.

DenisCarriere:
@stebogit 👍 I like it a lot! A lot better than the previous method name.

I think you've got a winner there!

stebogit:
@DenisCarriere I believe this module might be ready for merging, after your blessing 😃

DenisCarriere:
@stebogit ✋ High Five! Looks great!

DenisCarriere:
## GeoJSON Pac-Man! 🚀 
https://github.com/Turfjs/turf/blob/master/packages/turf-sector/test/out/pacman.geojson
image

stebogit:
@DenisCarriere Great, thanks! 🎉 
How come that the branch turf-sector is still unmerged? Did you just copied the module over to master? Just curious
screen shot 2017-04-11 at 10 18 03 pm

DenisCarriere:
Strange, might be an issue on the GitHub Network graph. I simply merged it via this PR, nothing special (maybe check again tomorrow, might be a delay).

DenisCarriere:
@stebogit Reference the network graph. Maybe it's because when I merge I always select the **Squash and merge** vs. using **Create a merge commit** (I personally like having a single commit for each individual PRs vs. having 10 new commits for one fix/new module).

DenisCarriere:
Yep that was it, I just did a merge using **Create a merge commit** with `bearingToAngle` and the network graph merged.

image

PR #655, opened 2017-04-10
Arc line

stebogit:
@DenisCarriere I implemented module `@turf/line-arc` (although the branch is named arc-line, sorry my bad... 😅 ). Please check it out.
I fixed the two errors you found in `turf-sector`; I'm gonna use this module as dependency of `turf-sector` so that the errors are fixed there as well.

**N.B.:** the output of both `sector` and `line-arc` will be a `Polygon`/circle for equals bearings (a closed sector is actually a circle), however `line-arc` has `LineString` as output; should we include both types in `line-arc` output? 

Again I couldn't run the `generate-readmes` script, it's like the script is invisible, nothing happens when I run it...
Do I need to install anything in particular (I'm on MAC)?
I run `./scripts/generate-readmes` from the root folder and I have run `npm install`.
Also running `npm test` at the root doesn't produce the readme files.
Thanks


DenisCarriere:
@stebogit Thanks for opening this, however can you close this PR and make the new branch `arc-line` using the `master` branch and not to mix commits from `turf-sector`. (I made a fresh branch for you [`turf-line-arc`](https://github.com/Turfjs/turf/tree/turf-line-arc))

As for the README, it should of worked... might be a local config on your system.

stebogit:
@DenisCarriere I'm so sorry 😅 , I didn't know the thread of comments would remain the same even in a new PR.
Should I create a new PR for the new `turf-line-arc` branch?

Issue #659, opened 2017-04-10
@turf/circle translate GeoJSON properties


DenisCarriere:
Translate input GeoJSON properties of `center` to result circle Polygon.

PR #662, opened 2017-04-12
Translate input properties to output, updated test

stebogit:
@DenisCarriere I saw #659 and I thought it was easy-peasy since I worked on `line-arc`; check this out.


DenisCarriere:
👍 Looks great! Nice touch on adding those input validation, there's very minimal performance loss by including them so it's always nice to have whenever possible.

PR #663, opened 2017-04-12
Added `bearingToAngle` method to turf-helpers

stebogit:


packages/turf-helpers/index.js
```diff
@@ -331,5 +349,6 @@ module.exports = {
     multiPolygon: multiPolygon,
     radiansToDistance: radiansToDistance,
     distanceToRadians: distanceToRadians,
-    distanceToDegrees: distanceToDegrees
+    distanceToDegrees: distanceToDegrees,
+    bearingToAngle: bearingToAngle
```
DenisCarriere: 👍 Just a side note: This syntax "looks ugly" since now you can use **Shorthand property names in ES2015**. https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Object_initializer
However this is needed because TurfJS aims to be ES5 compatible.

DenisCarriere:
👍 Very handy little function, I know I've had to use this before in the past and it always takes at least 5-10 minutes to figure out the correct modulus operation to apply.

I added the function the main [`packages/turf/index.js`](https://github.com/Turfjs/turf/blob/master/packages/turf/index.js) & [`packages/turf/index.d.ts`](https://github.com/Turfjs/turf/blob/master/packages/turf/index.d.ts), [Lerna](https://github.com/lerna/lerna) bundles the dependencies as symbolic links, so don't hesitate to update the core repo whenever you need to add new features or new modules.

PR #664, opened 2017-04-12
First pass at adding redone maps

rowanwins:
Gday @DenisCarriere 

Starting the process of re-adding the maps. I've tested these modules locally and they are working well.

The only one which presents a bit funny is `bbox-clip` which needs a bit of love on the styling to get it to present well but im having a think about that.

Just thought I'd share these in case another other red flags jump about before i do the others...


packages/turf-area/index.js
```diff
@@ -29,8 +28,9 @@ var geomReduce = require('@turf/meta').geomReduce;
  *   }
  * }
  * var area = turf.area(polygon);
- * //=area => square meters
- * //=polygon
+ * //addToMap
+ * polygon.properties.area = area
+ * var addToMap = [polygon]
```
DenisCarriere: 🏅 Nicely done! Adding the results in the `properties` 👍

packages/turf-bbox/index.js
```diff
@@ -18,10 +17,8 @@ var each = require('@turf/meta').coordEach;
  * var bbox = turf.bbox(features);
  *
  * var bboxPolygon = turf.bboxPolygon(bbox);
- *
- * //=bbox
- *
- * //=bboxPolygon
```
DenisCarriere: Agreed, we can remove these comments if `//addToMap` is present.

packages/turf-bearing/index.js
```diff
@@ -11,33 +11,32 @@ var getCoord = require('@turf/invariant').getCoord;
...
+ *   "properties": {},
  *   "geometry": {
  *     "type": "Point",
  *     "coordinates": [-75.343, 39.984]
  *   }
  * };
  * var point2 = {
  *   "type": "Feature",
- *   "properties": {
- *     "marker-color": '#0f0'
```
DenisCarriere: 👍 Removing the any extra properties.

DenisCarriere:
@rowanwins 👍 Looks great!

DenisCarriere:
@rowanwins Delete the `updated-docs` branch if you are done with these changes or feel free to commit directly to the master branch for any `addToMap` related changes.

Issue #669, opened 2017-04-16
turf-flip incorrect if z-elevation is present


DenisCarriere:
If coordinates contain elevation, coordinates are flipped entirely (elevation is placed in first position).

```javascript
var serbia = {
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Point",
    "coordinates": [20.566406, 43.421008, 3000]
  }
};
turf.flip(serbia)
//=geometry.coordinates => [3000, 43.421008, 20.566406]
```

DenisCarriere:
Fixed via PR https://github.com/Turfjs/turf/pull/678

PR #678, opened 2017-04-18
Only flip [x,y] turf-flip #669

DenisCarriere:
Reference issue: https://github.com/Turfjs/turf/issues/669

- [x] Improved Typescript definition
- [x] Only reverse [x, y] coordinates (even if elevation coordinate is present)
- [x] Update tests to dynamic geojson
- [x] Confirmed input not being mutated.


DenisCarriere:
Merging, however it would be nice to drop the [`JSON.parse()` & `JSON.stringify()`](https://github.com/Turfjs/turf/blob/978a457cc0c95802f9929bfcb7d44918dc391648/packages/turf-flip/index.js#L28) methods in a major release `Turf v5.x` (i'll submit a formal issue for that change).

Issue #681, opened 2017-04-20
@turf/destination error with bearing=90


stebogit:
I noticed the returned point is not exactly at 90 degrees from the input, i.e. the latitude changes.
I'll create an additional test in `@turf/destination` to highlight the error.

DenisCarriere:
👍 Thanks for reporting this @stebogit, however I believe this might be an expected result from the Haversine Formula referenced in http://www.movable-type.co.uk/scripts/latlong.html.

It would be still good to provide some test fixtures regardless of the known or not known issue.

stebogit:
@DenisCarriere what do you think? I guess the behaviour might be unexpected to an unaware user (like me), but in fact correct.

rowanwins:
Gday @stebogit 

Perhaps we just add a short note to the doco for the module that the visual output might look slightly misaligned due to the curvature of the earth, im sure someone can come up with some slightly more accurate words than those.

DenisCarriere:
I'm still baffled by this issue, regardless of the Haversine Formula taking into account the curvature of the earth, one would assume by using `bearing=90` your destination point result should be at the same latitude (degrees or meters).

Let's keep this issue open until we find a solution (or give up the issue and accept the curvature of the earth distortion effect).

stebogit:
@DenisCarriere after some research I found out the reason why this module returns an unexpected output is that it follows a [great-circle distance](https://en.wikipedia.org/wiki/Great-circle_distance) to take in account the earth curvature, while somebody working on a 2D map would actually expect a movement along a [Rhumb line](https://en.wikipedia.org/wiki/Rhumb_line).

[`geodesy`](https://github.com/chrisveness/geodesy) is a great library that implements different [methods along those lines](http://www.movable-type.co.uk/scripts/latlong.html#rhumblines).

I'd be happy to implement like `@turf/rhumb-destination`, `@turf/rhumb-distance` and `@turf/rhumb-bearing` modules using that library, or maybe taking cues from it, as I believe it might result in a not lightweight dependency

stebogit:
A module like `@turf/rhumb-destination` would probably also "solve" #684, as it would properly move a point of the exact/expected distance.

DenisCarriere:
👍 @stebogit Sounds like a plan.

stebogit:
[...]

PR #682, opened 2017-04-20
add tests to @turf/destination

stebogit:
Ref. #681


DenisCarriere:
👍 Awesome, it's a lot easier to see the issue/known behavior from the `test/out` now.

image

PR #685, opened 2017-04-22
Linestrings to mulipolygon - Add test fixtures

DenisCarriere:
The goal of `linestring-to-polygon` is meant to be predictable without having to guess the output.

Here are the current mutations:

- (Multi)LineString => Polygon
- FeatureCollection of (Multi)LineString => MultiPolygon

With that information, it's up to the user to provide the correct input, it's the same scenario with `@turf/helpers` `polygon()`.

## multi-linestring-incomplete

> MultiLineString => Polygon

image

## linestrings-to-multipolygons

> FeatureCollection of (Multi)LineString => MultiPolygon

- **Blue**: MultiLineString
- **Red**: MultiLineString
- **Orange**: LineString
- **Green**: LineString

image

CC: @stebogit


stebogit:
@DenisCarriere the output of the `multi-linestring-incomplete.geojson` example should be a `Polygon`, not `MultiPolygon`.

Also, `@turf/helpers` `polygon()` does have an input validation so that the output is always an actual `Polygon`; here if the input is just a collection of `LineString` / `LinearRings` there is no check if the lines/rings are part of the same polygon or multiple polygons, which might result in just an incorrect output.

stebogit:
Sorry closed by mistake...

DenisCarriere:
> the output of the multi-linestring-incomplete.geojson example should be a Polygon, not MultiPolygon.

Agreed, the `test/in` fixture was a FeatureCollection, I changed it to a Feature and it returns a [Polygon](https://github.com/Turfjs/turf/pull/685/commits/ac194db85e43c307aa80a8938106bae51fc219f4#diff-c27ada88eab7779ea34b76a9f2a8b31fL8) now.

> there is no check if the lines/rings are part of the same polygon or multiple polygons

I'm using both `polygon` & [`multiPolygon` from `@turf/helpers`](https://github.com/Turfjs/turf/blob/master/packages/turf-helpers/index.js#L209-L231) for validation. I doubt we should be adding checks to see if the LineString is apart of the same (Multi)Polygon.

> which might result in just an incorrect output.

If the output is incorrect, so is the input. If we follow the logic of `Feature` => `Polygon` & `FeatureCollection` => `MultiPolygon`, then it would be up to the input to be correct to receive a valid output.

@stebogit : Any suggestions for validation?

stebogit:
@DenisCarriere it seems to me there is a fundamental flaw here, i.e. we assume that a `feature` (of single or multiple `lineString`) defines always a single `polygon`, while a `featureCollection` (again of single or multiple `lineString`, or `linearRings` if closed or being closed) defines a `multipolygon` (correct me if I'm wrong):

> Here are the current mutations:
>    - (Multi)LineString => Polygon
>    - FeatureCollection of (Multi)LineString => MultiPolygon
>
> With that information, it's up to the user to provide the correct input, it's the same scenario with @turf/helpers polygon().

I believe it is not that simple.
What defines if a bunch of `linearRings` (`multiLineString` feature), creates a single `polygon` (`feature`) or different `polygons` (as `featureCollection` of `polygons`, or a single `multipolygon` `feature`, it's just a different way to represent the same thing) is really if exists a single larger ring that includes all other rings, part of the same `polygon`, or there are more "larger" rings that identify multiple `polygons`.

I added the `multi-linestring-nested.geojson` test fixture (you can change the name if you want) that represents this instance and highlights [the issue](https://jsfiddle.net/jStefano/yd4cgy6b/):
screen shot 2017-04-23 at 4 39 54 pm

So, in my opinion the "is included" evaluation is necessary not only as validation, but to determine the right input.

I'll work on a routine to separate nested rings, i.e. single `polygons`, (I have to anyway for `isobands`) so you can take a look.

DenisCarriere:
@stebogit Can you push a commit or provide a gist link of both the LineStrings & Polygon of `multi-linestring-nested.geojson` you are referencing to.

DenisCarriere:
**Note:** Google Maps doesn't play very nice with GeoJSON Polygon (outer + inner + inner + inner).

The ordering is correct, don't know why it's not rendering correctly.

https://jsfiddle.net/yd4cgy6b/8/
https://gist.github.com/399ba4a6cbfd70d7fe92bcc145984893

**Google maps**
image

**geojson.io**
image

stebogit:
@DenisCarriere Sorry I thought I did commit.
Apparently GMap is really really strict, but reading the specs again it make sense:

> For Polygons with more than one of these rings, the first MUST be
      the exterior ring, and any others MUST be interior rings.  The
      exterior ring bounds the surface, and **the interior rings (if
      present) bound holes within the surface**.

In this example the outer doughnut is consider a `Polygon`, as the inner one, as the third `LinearRing` (from the outside) does not bound the same surface, but it is an exterior ring of another surface.
Together they would form a `MultiPolygon`.
Look at this: https://jsfiddle.net/jStefano/yd4cgy6b/11/

Man  that's tricky!

stebogit:
@DenisCarriere Man I've just realized I've overridden your commit! Fixed it.

DenisCarriere:
> Man that's tricky!

Oh my lord! My head is going to explode 🤕 💥 

> Look at this: https://jsfiddle.net/jStefano/yd4cgy6b/11/

That make sense, however there's no way we can implement such thing and still have a minimalistic code base (each Turf library shouldn't be more than 200 LOC **with exceptions).

DenisCarriere:
Added your [Outer & Inner doughnut Google Maps example](https://jsfiddle.net/jStefano/yd4cgy6b/11/) to the test cases (https://github.com/Turfjs/turf/pull/685/commits/4b00189a5814e7164eb1520e83d0d327333b5787 & https://github.com/Turfjs/turf/pull/686/commits/66e375a1a587210376ef066548ed7ec18e35d5ac).

Everything seems to work well 👍

DenisCarriere:
We've beaten to death this PR, I'm 👍 good for merging this, @stebogit any more concerns?

Also have a glance at https://github.com/Turfjs/turf/pull/686, I'm good for merging that one too.

stebogit:
@DenisCarriere if all tests pass then 👍  to me.
Sorry I can't work on this as much as I would like, so I'm slow. But I'll keep working on `isobands` as promised and I will try to use all this new good stuff, so if I find any issues I'll let you know (as I did so far, that's why I'm bombarding you with all these questions/issues! 😄 ).

Issue #692, opened 2017-04-30
Feature request: iterate over flattened features


dpmcmlxxvi:
Turf has a flatten method that creates a feature collection. It would be helpful to have a `flattenEach` method that traverses them with a callback without allocating them and consuming memory. I can submit a PR if there is interest in this feature.

Note, this would likely introduce `helpers` as a dependency. The `helpers` module in turn has no dependencies. So, this shouldn't be a problem unless it is desired to keep all these low level modules free of any dependencies.

DenisCarriere:
👍 Agreed having callback functionality, however `@turf/helpers` shouldn't be used to extend other modules and should be kept with zero dependencies.

One option that won't break any backwards compatibility is:
- Add optional `callback` param `flatten(geojson, callback)`, if callback is present then method should return void. If callback is not present `flatten(geojson)` then method should return FeatureCollection/Feature.

DenisCarriere:
👍 Agreed having callback functionality, however `@turf/helpers` shouldn't be used to extend other modules and should be kept with zero dependencies.

One option that won't break any backwards compatibility is:
- Add optional `callback` param `flatten(geojson, callback)`, if callback is present then method should return void. If callback is not present `flatten(geojson)` then method should return FeatureCollection/Feature.

@morganherlocker What's your opinion on including an optional `callback` param? Might be worth doing this to many other modules (I can already think of a few).

dpmcmlxxvi:
Just to be clear, I meant adding `@turf/helpers` as a dependency to `@turf/meta` since `meta` is where all the other `forEach` like methods are located. I didn't mean adding dependencies to `helpers`. It seemed like that would be fine since `@turf/helpers` is used as a dependency for many other modules. But if it is desired to keep `meta` free of dependencies then adding it to `flatten` is another option.

DenisCarriere:
Adding the current implementation of `@turf/flatten` to `@turf/meta` would defeat the purpose of iterating over the features since the current state of `flatten` outputs a `FeatureCollection`. If this isn't changed in the `@turf/flatten` module then the memory issue would still exist.

I'm 👍 adding a callback param directly in `@turf/flatten`, however this would require a significant overhaul of the module.

dpmcmlxxvi:
[...]

PR #707, opened 2017-05-02
Initial commit of boolean DE-9IM module

rowanwins:
Gday @DenisCarriere 

As discussed attached is my current approach on a boolean style module for DE-9IM. I've been using the [doco on this page](http://edndoc.esri.com/arcsde/9.0/general_topics/understand_spatial_relations.htm) as my guide so far.

## DE-9IM modules
- [x] Equal
- [x] Touch (although the linestring ones need fixing)
- [x] Contains
- [x] Within (based on the opposite of `contains`)
- [x] Disjoint (looks fairly straight forward)
- [ ] Intersects 
- [x] Overlap (looks straight forward)
- [x] Cross (looks straight forward)
- [ ] Covers

## Extra Testing
- [ ] donut polygons
- [ ] lines `[[0,0], [2,0]]` and `[[0,0], [1,0], [2,0]]` are equal but don't have identical coordinates.

There are a few helper style functions within the `helpers.js` file.

Anyway happy for any feedback. I'm sure some of the algorithms could be improved, they are pretty brute force at the moment.
Cheers
Rowan


DenisCarriere:
😄 Oh my! Looks great so far (quickly browsed the repo), I'll dive into further when I have hours and hours to dedicate to this 👍

dpmcmlxxvi:
Looks good. Just a thought, the motivation behind bringing up the `flattenEach` issue in #692 was that I was implementing some of the DE-9IM functions in another project and found that it gets much simpler if you flatten the features first. That way you really only have to deal with 3 geometries (i.e., point, line, and polygon) and a 3x3 decision matrix. Otherwise, you have to deal with all 6 and a 6x6 decision matrix.

DenisCarriere:
@dpmcmlxxvi Only having to deal with 3x3 makes a lot sense!

rowanwins:
Thanks for the comment/idea @dpmcmlxxvi , I wasn't sure how much overhead flattening the features would add in terms of performance so it's probably worth setting up some benchmark tests, it would be nice to cater for fewer scenarios :)

rowanwins:
Cross now mostly done...

DenisCarriere:
👍 Keep them coming! My next commit will extract each of these operations into their own modules, it's already starting to be confusing to test them.

DenisCarriere:
@rowanwins you might want to look into switch cases instead of a ton of `if` statements.

For example (using @dpmcmlxxvi's flatten approach):

```javascript
var multiFeature1 = flatten(feature1);
var multiFeature2 = flatten(feature2);

geomEach(multiFeature1, function (geom1) {
    geomEach(multiFeature2, function (geom2) {
        switch (geom1.type) {
        case 'Point':
            switch (geom1.type) {
            case 'Point':
            case 'LineString':
            case 'Polygon':
            }
        case 'LineString':
            switch (geom1.type) {
            case 'Point':
            case 'LineString':
            case 'Polygon':
            }
        case 'Polygon':
            switch (geom1.type) {
            case 'Point':
            case 'LineString':
            case 'Polygon':
            }
        }
    });
});
```

Depending on which boolean operations, you might want to use `geomReduce` instead of `geomEach`.

DenisCarriere:
@rowanwins Just noticed `checkIfCoordArraysMatch` & `checkIfDeepCoordArraysMatch` can be replaced using [`deep-equal`](https://www.npmjs.com/package/deep-equal).

```javascript
var deepEqual = require('deep-equal');

deepEqual(MultiPoint.geometry.coordinates[i], Point.geometry.coordinates)
```

packages/turf-boolean-contains/index.js
```diff
@@ -0,0 +1,208 @@
...
+            output = false;
+            break;
+        }
+    }
+    return output;
+}
+
+// TO DO - Work out how to check if line is in line (can potentially use line overlap module)
+// Also need to make sure lines are exactly the same, eg the second must be smaller than the first
+function isLineOnLine(LineString1, LineString2) {
```
DenisCarriere:
```javascript
 t.equal(contains(line1, line3), false, 'A line fails that lies partially outside the other line');
```
[Test fail](https://github.com/Turfjs/turf/pull/707/files#diff-083c7d1fa4cb2170f16604d55e772b48R42) because line should be **completely** on line (returns true because of partial match).

Might be worth adding a `isLineCompletelyOnLine(line1, line2)` method.

rowanwins: Gday @DenisCarriere 

As a general comment I feel like the line booleans are the trickiest to get right at this stage, acknowledge that there is some work needed there

rowanwins:
Just on the tests @DenisCarriere I dont think they are really too bad, if you want to test a individual boolean simply `cd` into the relevant directory such as `test/contains` and run `node test`, 

I think it would actually be a fair bit more overhead to split them all out.

packages/turf-boolean-contains/index.js
DenisCarriere: Never said it was easy 👍 Keep it up!

DenisCarriere:
> Just on the tests @DenisCarriere I dont think they are really too bad

I disagree, it's going to be a nightmare when there's 9 different test cases all in one repo.

Especially when we start adding fixtures to each repo (larger GeoJSON).

Also, we might end up including a test fixture folder for true/false.

- `test/true/<filename>.geojson`
- `test/false/<filename>.geojson`

That way it's easier for someone to submit a random GeoJSON that has the wrong expected output (true/false).

DenisCarriere:
Also for modules that depend on other modules, we would simply just add it as a dependency like `@turf/boolean-within` would depend on `@turf/boolean-contains`.

packages/turf-boolean-contains/index.js
```diff
@@ -0,0 +1,208 @@
...
+                foundAMatch++;
+                anyMatch = true;
+                break;
+            }
+        }
+        if (!anyMatch) {
+            return false;
+        }
+    }
+    return foundAMatch > 0 && foundAMatch < MultiPoint1.coordinates.length;
```
dpmcmlxxvi: The second condition here excludes the two geometries being equal. Not sure if that is actually in the definition of `contains`. According to [DE-9IM](https://en.wikipedia.org/wiki/DE-9IM) the definition is:

 - No point of `MultiPoint2` must be in the exterior of `MultiPoint1`
 - At least 1 point in the interior of `MultiPoint2` must be in the interior of `MultiPoint1`

Since the points themselves are their own interior, the two being equal would seem to meet that criteria. Could be wrong, the DE-9IM definitions are sometimes tricky to parse.

packages/turf-boolean-contains/index.js
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// http://stackoverflow.com/a/11908158/1979085
+function isPointOnLine(LineString, Point) {
```
dpmcmlxxvi: This seems to be allowing the point to land on the line end points. That would place the point on the line boundary and therefore no point lies on the interior of the line (see above definition of contains). I think the end points should be excluded.

packages/turf-boolean-contains/index.js
```diff
@@ -0,0 +1,208 @@
...
+    for (var i = 0; i < LineString.coordinates.length - 1; i++) {
+        if (isPointOnLineSegment(LineString.coordinates[i], LineString.coordinates[i + 1], Point.coordinates)) {
+            output = true;
+            break;
+        }
+    }
+    return output;
+}
+
+function isMultiPointOnLine(LineString, MultiPoint) {
```
dpmcmlxxvi: Same issue as above. No check that any point is in the interior of line. As you probably have noticed already the edge cases are the real pain when implementing these booleans functions.

packages/turf-boolean-contains/index.js
```diff
@@ -0,0 +1,208 @@
...
+        }
+    }
+    return output;
+}
+
+function isPointInPoly(Polygon, Point) {
+    return inside(Point, Polygon);
+}
+
+function isMultiPointInPoly(Polygon, MultiPoint) {
```
dpmcmlxxvi: Same interior point issue as above.

packages/turf-boolean-contains/index.js
```diff
@@ -0,0 +1,208 @@
...
+    if (!overlappingLine) return false;
+    overlappingLine = getGeom(overlappingLine);
+    if (overlappingLine.coordinates[0] === LineString1.coordinates[0] &&
+        overlappingLine.coordinates[overlappingLine.coordinates.length - 1] === LineString1.coordinates[LineString1.coordinates.length - 1]) {
+        output = false;
+    }
+    return output;
+}
+
+function isLineInPoly(Polygon, Linestring) {
```
dpmcmlxxvi: I think this approach will only work for convex polygons. Concave or polygons with holes can fail this. Think of a donut with a line that only crosses the hole.

I think it would be safer to first check if the line intersects the polygon boundaries (including holes). If it doesn't then check that at least one point is in polygon interior.

packages/turf-boolean-contains/index.js
```diff
@@ -0,0 +1,208 @@
...
+            output = false;
+            break;
+        }
+    }
+    return output;
+}
+
+// See http://stackoverflow.com/a/4833823/1979085
+// TO DO - Need to handle if the polys are the same, eg there must be some outer coordinates different
+function isPolyInPoly(Polygon1, Polygon2) {
```
dpmcmlxxvi: I think this has the same issue as `isLineInPoly` above.

packages/turf-boolean/equal.js
```diff
@@ -0,0 +1,18 @@
...
+* @name contains
+* @param {feature1} feature1
+* @param {feature2} feature2
+* @returns {Boolean}
+* @example
+* var along = turf.contains(line, point);
+*/
+
+module.exports = function (feature1, feature2) {
+    return checkIfDeepCoordArraysMatch(feature1.geometry.coordinates, feature2.geometry.coordinates) && feature1.geometry.type === feature2.geometry.type;
```
dpmcmlxxvi: Might want to add a warning that this is only testing that the coordinates are identical. It is not testing that the geometries are equal. For example, the lines `[[0,0], [2,0]]` and `[[0,0], [1,0], [2,0]]` are equal but don't have identical coordinates.

packages/turf-boolean/index.js
```diff
@@ -0,0 +1,21 @@
...
+*
+* @name boolean
+* @param {feature1} feature1
+* @param {feature2} feature2
+* @returns {Boolean}
+* @example
+* var along = turf.contains(line, point);
+*/
+
+module.exports = {
```
dpmcmlxxvi: @DenisCarriere Just wondering what the policy is on namespacing functionality. I thought I recalled reading in some other thread that Turf enforced flat structure and namespacing functions was discouraged. May be mis-remembering.

packages/turf-boolean/touch.js
```diff
@@ -0,0 +1,98 @@
...
+            onStartOrEnd++;
+            continue;
+        }
+        for (var lineI = 0; lineI < LineString.geometry.coordinates.length - 1; lineI++) {
+            if (isPointOnLineSegment(LineString.geometry.coordinates[lineI], LineString.geometry.coordinates[lineI + 1], MultiPoint.geometry.coordinates[i])) {
+                pointIsSomewhereOnLine++;
+            }
+        }
+    }
+    return onStartOrEnd === 1 && pointIsSomewhereOnLine === 0;
```
dpmcmlxxvi: Just a nit-pick. You can probably return inside the loop if the second criteria is broken. That way you don't iterate through all the points if you don't need to. Also, I think the first criteria should read `onStartOrEnd > 0`. It could be both ends that are touched.

packages/turf-boolean/touch.js
```diff
@@ -0,0 +1,98 @@
...
+    for (i = 0; i < Polygon.geometry.coordinates[0].length - 1; i++) {
+        if (isPointOnLineSegment(Polygon.geometry.coordinates[0][i], Polygon.geometry.coordinates[0][i + 1], Point.geometry.coordinates)) {
+            output = true;
+            break;
+        }
+    }
+    return output;
+}
+
+function isMultiPointOnPolygon(MultiPoint, Polygon) {
```
dpmcmlxxvi: I think you also need a check that none of the points are in the interior of the polygon.

dpmcmlxxvi:
@rowanwins I made a bunch of comments on the code. Hope they help. It's always easier to review than implement! Keep it up. I'd really like to see these included in turf.

packages/turf-boolean/equal.js
DenisCarriere: This one sounds like it would be real tricky to write. Would be worth including as a test case or add it to the JSDocs description.

packages/turf-boolean/index.js
DenisCarriere: Agree, that's why I suggested to have all the `boolean` modules within their own unique repos. I've only converted `@turf/boolean-contains` so far from this PR.

@rowanwins Lets stick with 1 PR per `boolean` module & all within their own repos at first, `@turf/boolean` can group them all up afterwards.

DenisCarriere:
@dpmcmlxxvi Thanks for your valuable feedback! 👍

rowanwins:
Thanks for the feedback @dpmcmlxxvi 

Im still working out the interpretation of how the DE-9IM relationships work so yep happy to take feedback on board.  I think I'll try and run a similar bunch of tests against a third party implementation (eg in QGIS) just to check I understand the interpretations.

packages/turf-boolean-contains/index.js
DenisCarriere: 👍 Polygons with holes will be hard to implement, @rowanwins I think at first lets just get it working with only outer rings and afterwards we work on the inner ring validation (donuts).

DenisCarriere:
@rowanwins Helpful reference for the `overlaps` module. https://github.com/tcql/turf-overlaps

dpmcmlxxvi:
@rowanwins @DenisCarriere Just a quick note on `turf-overlaps`, it checks for overlaps by testing for intersections. However, you can have polygons with edges that intersect without actually overlapping.

rowanwins:
Just as a heads up on this @DenisCarriere I've created a `turf-boolean-helpers` module where I'm storing lots of the common code, 

For example `isPointOnLineSegment` gets used in pretty much every module. As a further FYI I've now adjusted it to include a flag for isPointOnInteriorOnly as per @dpmcmlxxvi suggestion.

Any qualms with this approach? I'm not sure how many functions will end up in here but probably half a dozen I'd guess.

DenisCarriere:
👍 I like the idea of having `@turf/boolean-helpers`, it makes sense to have a collection of methods that are used in many modules, and if one of those `boolean-helpers` methods because too elaborate then we can convert it to it's own repo.

DenisCarriere:
@rowanwins I would refrain from adding any `eslint-disable` since most of these warning message are valuable.

**Example:**
```javascript
return !isPointOnLine(geom2, geom1) ? true : false; //eslint-disable-line no-unneeded-ternary
```
**Refactored to:**
```javascript
return !isPointOnLine(geom2, geom1) || false;
```

rowanwins:
ah that's how you get rid of that stupid error, thanks @DenisCarriere , learn something new everyday!

packages/turf-boolean-cross/test.js
```diff
@@ -17,7 +17,8 @@ test('turf-boolean-cross', t => {
 
     t.equal(cross(line1, line2), true, 'True if lines cross');
     t.equal(cross(line1, line3), false, 'False if lines cross');
-    t.equal(cross(line1, line4), false, 'False if lines only touch');
+    // **== ERROR ==**
+    // t.equal(cross(line1, line4), false, 'False if lines only touch');
```
dpmcmlxxvi: Just wondering if the test library provides a skip method to gracefully skip a test so that it still shows up listed but just not executed. That way it doesn't get forgotten but won't cause a failure. I know mocha lets you do that by just throwing in `.skip(` before the test. According to the tape docs it looks like they do.

DenisCarriere: That's most likely a better idea, there's still tons of tests still missing, lots of work left to do

DenisCarriere:
@rowanwins have you tried this library? [geojson-equality](https://github.com/geosquare/geojson-equality) which was used in [`turf-equal`](https://github.com/Turfjs/turf-equal/blob/master/index.js)

DenisCarriere:
@rowanwins Added a few extra params to `equal`.

```javascript
/**
 * @name equal
 * @param {Geometry|Feature<any>} feature1 GeoJSON Feature or Geometry
 * @param {Geometry|Feature<any>} feature2 GeoJSON Feature or Geometry
 * @param {Boolean} [direction=false] direction of LineString or Polygon (orientation) is ignored if false
 * @param {number} [precision] coordinate decimal precision
```

**Tests**
```javascript
test('turf-boolean-equal -- reduce coordinate precision', t => {
    const pt1 = point([30.2, 10]);
    const pt2 = point([30.22233, 10]);

    t.true(equal(pt1, pt2, false, 1));
    t.false(equal(pt1, pt2, false, 3));
    t.end();
});


test('turf-boolean-equal -- apply direction (orientation) rule', t => {
    const line1 = lineString([[30, 10], [10, 30], [40, 40]]);
    const line2 = lineString([[40, 40], [10, 30], [30, 10]]);

    t.true(equal(line1, line2, true));
    t.false(equal(line1, line2, false));
    t.end();
});
```

DenisCarriere:
@rowanwins Don't mind if we close this PR, I'm pretty sure you have this stored locally and we can leave the branch active until we port over everything.

PR #712, opened 2017-05-04
Add flattenEach/flattenReduce to @turf/meta.

dpmcmlxxvi:
This addresses issue #692. There are a few design choices made that deserve scrutiny:

 - `@turf/helpers` is added as a dependency to `@turf/meta`. Needed to create features as geometries get flattened.
 - The `geomEach` callback function was changed to add an additional third argument that passes the properties of the source feature from which the geometry came. Wasn't sure if adding this to the argument list was a good idea since a geometry doesn't have a property per the GeoJSON spec but since `geomEach` is essentially a flattening function the source will usually be a feature. It's backward compatible as the last argument and existing users can just ignore it.
 - The `flattenEach` and `flattenReduce` callback function returns an `index` argument (increases for each geometry) and an additional `subIndex` argument that increases for each sub-geometry if the geometry being flattened was a multi-geometry (MultiPoint, MultiLineString, MultiPolygon). For example, for a MultiPoint with, say, 3 points each point will have the same `index` but their `subIndex` will be 0, 1, and 2. I'm open to making this a single growing index but I thought this would be helpful if the caller wanted to index back into some property for multi-geometries.

Finally, does Turf have a unit test coverage functionality somewhere? I didn't see any and it would be good to know if the tests are hitting all the right code.

Please fill in this template.

### DIFF
--- a/packages/turf-meta/README.md
+++ b/packages/turf-meta/README.md
@@ -300,7 +300,7 @@ Iterate over each geometry in any GeoJSON object, similar to Array.forEach()
 **Parameters**
 
 -   `layer` **[Object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object)** any GeoJSON object
--   `callback` **[Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function)** a method that takes (currentGeometry, currentIndex)
+-   `callback` **[Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function)** a method that takes (currentGeometry, currentIndex, currentProperties)
 
 **Examples**
 
@@ -326,9 +326,10 @@ var features = {
     }
   ]
 };
-turf.geomEach(features, function (currentGeometry, currentIndex) {
+turf.geomEach(features, function (currentGeometry, currentIndex, currentProperties) {
   //=currentGeometry
   //=currentIndex
+  //=currentProperties
 });
 ```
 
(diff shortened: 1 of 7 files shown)


DenisCarriere:
> @turf/helpers is added as a dependency 

Don't think this would be necessary, we can include a basic `feature()` method internally to `@turf/meta`. No dependencies for this `@turf/meta` would be ideal.

> The geomEach callback function was changed to add an additional third argument that passes the properties of the source feature

👍 That works for me

> returns an index argument (increases for each geometry) and an additional subIndex argument that increases for each sub-geometry 

I like it! 👍 A growing index could get messy if one needs to find a single part of a MultiFeature.

Few things I did in the recent commits:
- Converted tests to ES6
- Dropped `@turf/helpers` by including a private `feature()` method (I don't think the helpers validation is necessary in this scenario).
- Added `flattenEach` & `flattenReduce` to typescript definition
- Updated/Refactored entire typescript definition

dpmcmlxxvi:
OK, will keep in mind newer conventions like ES6 when making any other edits.

DenisCarriere:
@dpmcmlxxvi To clarify, ES6 would only apply for the testing (since it's being executed with NodeJS), the `index.js` syntax would still remain as ES5.

DenisCarriere:
@dpmcmlxxvi BTW great job on the PR 🥇 👍 

> Finally, does Turf have a unit test coverage functionality somewhere? I didn't see any and it would be good to know if the tests are hitting all the right code.

100% agreed, we should have coverage functionality in Turf. I believe I've got a pretty good minimalistic solution for this, i'll send over a new PR for this since we might want to apply this solution to every module.

Issue #328, opened 2016-03-06
tape -> tap, set minimum coverage


tmcw:

tmcw:
Going to stay with tape for 3.x, only switch to tap once we figure some other details out.

DenisCarriere:
@tmcw Looking to hear your feedback, `tap` seems like a good solution for Turf.
Can I ask what those details to figure out were?

https://github.com/Turfjs/turf/pull/713

DenisCarriere:
Just found out that `tap` seamlessly integrates with `coveralls.io` without any additional configurations or changes in the `package.json`.

http://www.node-tap.org/cli/

> If a COVERALLS_REPO_TOKEN environment variable is set, then coverage is captured by default and sent to the coveralls.io service.

**package.json** - no change (other than `node test.js` => `tap test.js`)
```json
  "scripts": {
    "test": "tap test.js",
    "bench": "node bench.js"
  },
```

**test.js** - only require changes (NodeJS >=6)

```javascript
const {test} = require('tap');
```

DenisCarriere:
Woot! It works! level of effort **Zero**!

![image](https://cloud.githubusercontent.com/assets/550895/26460657/538ef9b6-4148-11e7-8948-dc3c677ae271.png)

DenisCarriere:
Added tap as main testing suite, using `tap` in each repo is optional, both `tape` or `tap` works.

PR #713, opened 2017-05-04
Add coverage functionality

DenisCarriere:
Adding coverage functionality to Turf would allow for better tests.

Ref Article: [Testing `tape` vs `tap`](https://remysharp.com/2016/02/08/testing-tape-vs-tap)

## Proposal

- Replace `tape` for `tap` (no change in test code)
- Minimal effort to implement (1 LOC change)
- Easy to switch back to `tape`
- Generates easy to understand web report with coverage
- **Fallbacks**:  source code mapping is off since it gets compiled with [`nyc`](https://github.com/istanbuljs/nyc) (might need some extra configuration to get it working).

```bash
$ npm t
```

image

Also, for more in depth coverage reports you can use:

```bash
$ tap test.js --cov --coverage-report=lcov
```

image

image

Ref: https://github.com/Turfjs/turf/pull/712 https://github.com/Turfjs/turf/issues/328
Looking to hear your feedback: @morganherlocker @dpmcmlxxvi @tmcw

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+.nyc_output
+coverage
 lib-cov
 *.seed
 *.log
--- a/packages/turf-meta/package.json
+++ b/packages/turf-meta/package.json
@@ -9,7 +9,7 @@
     "index.d.ts"
   ],
   "scripts": {
-    "test": "node test.js",
+    "test": "tap test.js --cov",
     "bench": "node bench.js"
   },
   "keywords": [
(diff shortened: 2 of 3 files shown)


DenisCarriere:
Just realized that we can simply just use `nyc` in front of our existing `tape` tests and we receive coverage results (would only need to be used internally).

```bash
$ nyc node test.js
```

image

Going to close this PR for now, at the moment I don't think making the switch to another testing suite is worth the effort for a few reasons:
- issue with source code mapping (difficult to troubleshoot errors)
- extra overhead from source code being compiled (nyc needs to compile)
- unknown - how to implement coverage results using Lerna (monorepo).

dpmcmlxxvi:
Just tried `nyc` and the whole project and works like a charm. It has a helpful `--exclude` argument to exclude the test files and check coverage on just the production code. FYI, currently at 90%.

DenisCarriere:
Awesome, that's good news that it's at 90%.

Issue #635, opened 2017-03-28
Update distance measurement factors


dhivehi:
Turf line distance calculation is not accurate according to google earth. Is there any way to use custom radius in calculation?

DenisCarriere:
Please provide more context for you issue.

`@turf/distance` is using the [Haversine formula](http://en.wikipedia.org/wiki/Haversine_formula) which takes into consideration the curvature of the earth.

Closing this issue.

dhivehi:
its not the formula issue, but the problem is the earth radius being used is not the radius used in google earth.

dhivehi:
Official google earth's radius is 6378137meters

DenisCarriere:
👍 Agreed

You could send a PR with a fix for https://github.com/Turfjs/turf/blob/master/packages/turf-helpers/index.js#L264-L276

dhivehi:
I have send a PR, but it failed for some reason. I donno why. I'm new to github PR. This was my 1st PR in github.

dpmcmlxxvi:
My guess would be that, since the helper functions are used everywhere, changing their numerical behavior has affected unit tests throughout the entire library. One of the errors on travis-ci is an expected output not matching a fixture. Changing these constants will probably require going through and addressing all downstream unit tests that depend on the helper functions that rely on these constants.

dhivehi:
So could you suggest a fix?

dpmcmlxxvi:
I suggest you look at the travis-ci report and inspect at each unit test error returned in your PR. The error tells you what was the expected value and what was the actual value. Track down the cause of the difference and you'll likely see the some distance calculation is the root cause. Then update those test fixtures.

DenisCarriere:
@dpmcmlxxvi Exactly.

@dhivehi I pushed a commit to your PR https://github.com/Turfjs/turf/pull/636/commits/b7e826aedc4c32c8883832ba44cbd51e5c5699b0 which should of updated the the libraries that depend on `@turf/helpers` & `@turf/distance`.

The only things that really changed was the `factors` from your commits https://github.com/Turfjs/turf/pull/636/commits/a65c2370f08284ad260cbccd98a5ffb014090a92.

```javascript
var factors = {
    miles: 3963.190592,
    nauticalmiles: 3443.918467,
    degrees: 57.2957795,
    radians: 1,
    inches: 251107755.9055118,
    yards: 6975215.442,
    meters: 6378137,
    metres: 6378137,
    kilometers: 6378.137,
[...]

PR #715, opened 2017-05-06
angle conversion helper functions, modified precision of conversion

stebogit:
- Adds `radians2degrees` and  `degrees2radians` to `turf-helpers`;
- Modified precision of degrees conversion reducing Earth radius to it's [average value](https://en.wikipedia.org/wiki/Great-circle_distance#Radius_for_spherical_Earth)


DenisCarriere:
## To be removed 👎 
- **Modifying Earth Radius**, we've previously discussed this issue here https://github.com/Turfjs/turf/issues/635#issuecomment-292011500 (also such change would require a major release).
- Remove `round()` to methods, unnecessary processing.

## Worth including 👍 
- `radians2degrees` method (without `round`)
- `degrees2radians` method (without `round`)
- `round` can be included in `@turf/helpers`. Looks to be a useful method and has already been used in [`@turf/truncate`](https://github.com/Turfjs/turf/blob/master/packages/turf-truncate/index.js#L67) - however the `multiplier/factor` was calculated prior to doing any iterations.

DenisCarriere:
@stebogit Oops, I thought you did more modifications to the `factor`. Just realized that you've only changed it from **6373km** to **6371km** which lines up with what @mourner was discussing in the previous thread.

👍 This change, I'm not too sure if this should be done as a minor release or major release since this will affect a lot of modules.

stebogit:
@DenisCarriere Why don't we add an optional parameter `earthRadius` that would default to **6371km**, which I agree with @mourner is more correct, or any other value, and the user could decide what value to use.
This way `Turf` could even land on Mars! 😃

DenisCarriere:
@mourner's comment does reflect your change, he was mentioning we should use the **mean radius** which is **6371km**, currently TurfJS is using **6373km** (dates back [over a year ago](https://github.com/Turfjs/turf/blame/c81fc99576a37c38af91527462fd812b6c2bc1d2/packages/turf-helpers/index.js#L265-L275))

DenisCarriere:
@stebogit Going to implement `radians2degrees`, `round` & `degrees2radians`, however going to leave the `earthRadiusIn/factor` changes for another PR (since a lot of changes need to happen in dependent modules).

DenisCarriere:
Merging, [build passes on `master`](https://travis-ci.org/Turfjs/turf/builds/230115126?utm_source=github_status&utm_medium=notification) branch (newly updated `@turf/buffer`)

mourner:
> Why don't we add an optional parameter earthRadius [...], and the user could decide what value to use. This way Turf could even land on Mars! 😃

I don't think there's a need. Distance calculations depend on radius linearly, meaning that for any calculated value using Earth radius you can do `distance * mars_radius / earth_radius` to adjust the distance value to Mars.

DenisCarriere:
@mourner Can you confirm/approve the `earth radius` change (**6373km** => **6371km**) 

> Should only be implemented in next major release (v5.0)

Ref: https://github.com/Turfjs/turf/issues/635#issuecomment-292011500 https://github.com/Turfjs/turf/issues/635#issuecomment-292013737
CC: @morganherlocker

mourner:
@DenisCarriere I don't know where 6373 comes from, but the mean radius is 6371: https://en.wikipedia.org/wiki/Earth_radius

DenisCarriere:
@mourner Agreed, however **6373km** has been in TurfJS since the beginning, I didn't catch this when I reverted the last earth radius commit since I only I reverted it back to its original state.

https://github.com/Turfjs/turf/blame/c81fc99576a37c38af91527462fd812b6c2bc1d2/packages/turf-helpers/index.js#L273

Issue #88, opened 2014-02-14
removing jsts dependency


tmcw:
I see some TODO-like comments in the source https://github.com/morganherlocker/turf/blob/4a4772433b3deb6ae5f8dc9d24afbdd9c6555911/lib/erase.js#L4 but it seems like a good idea to veer away from too much jsts lineage since it's a rather huge project with its own lingo and internal formats.

---
- Polygon clipping modules require a new clipping algorithm like [greiner-hormann](https://github.com/tcql/greiner-hormann). Greiner-Hormann is approaching release readiness, but still has some outstanding clipping issues.
  - [ ] turf-intersect
  - [ ] turf-difference
  - [ ] turf-union
- Buffering requires a solution to buffering arbitrary polygons. @morganherlocker has made progress buffering other geometries, but polygons are still not ready.
  - [ ] turf-buffer

Evaluate this for an implementation guide:

https://github.com/akavel/polyclip-go/blob/master/clipper.go

tmcw:
Adding to the 3.0.0 milestone: @morganherlocker can you top-edit this so that it describes the current overall state of this task - which turf modules still need de-JSTS'ing and what algorithms are required to do it, and what state are those tasks in?

korczis:
Any news?

tcql:
@tmcw edited OP to list which modules need de-JSTSing, and a bit about what the state of those solutions are.

@korczis It hasn't moved too much lately. We'll be working on scheduling time to knock out the remaining algorithmic challenges soon

gajus:
We are only using `turf-buffer` to describe a circle around a point:

```js
import turfPoint from 'turf-point';
import turfBuffer from 'turf-buffer';
import turfInside from 'turf-inside';

type TypePoint = [number, number];

/**
 * Check if `point` is within the circle with center `center` and radius `radius`
 */
export default (point: TypePoint, center: TypePoint, radius: number) : boolean => {
    let centerPoint,
        circleAroundCenter,
        insidePoint;

    centerPoint = turfPoint(center);
    insidePoint = turfPoint(point);

    circleAroundCenter = turfBuffer(centerPoint, radius, 'meters');

    return turfInside(insidePoint, circleAroundCenter);
};

```

Is there a away to do it without `turfBuffer` and get rid off `jsts`?

tmcw:
turf-buffer depends on jsts, so, unfortunately, no

gajus:
[...]

PR #718, opened 2017-05-07
Improved/fixed @turf/buffer using equidistant buffers

DenisCarriere:
## Issue

Buffers were being distorted the further away it was from [0, 0]. Distances from north to south were different than west to east when measuring the output buffer. The expected output of a buffer should have its geometry equidistance in all directions.

## Summary
- Project coordinates using TransverseMercator centered/rotated on GeoJSON before applying `jsts` buffer calculation, unproject coordinates (to degrees) for final output.
- Added a few more input validation
- Dropped `@turf/dissolve` since `jsts` was faster with multiPoint
- Added Geometry Object tests
- Added prevent input mutation tests
- Used @w8r's [`moscow-rings`](https://github.com/w8r/moscow-rings) as inspiration.

Ref: https://github.com/Turfjs/turf/issues/660 https://github.com/Turfjs/turf/issues/88
CC: @flightsurvey @bjornharrtell @rowanwins 

image

image

## Equidistance Polygon Buffer

image

image


rowanwins:
Looks like a big improvement @DenisCarriere , nice one!

PR #724, opened 2017-05-08
Use @turf/meta's flattenEach in @turf/flatten

DenisCarriere:
Use `@turf/meta`'s newly implemented method `flattenEach`.

120+ LOC of was removed from `@turf/flatten` and replaced with only this:

```javascript
var results = [];
flattenEach(geojson, function (feature) {
    results.push(feature);
});
return featureCollection(results);
```

👍 Thanks @dpmcmlxxvi for the `flattenEach` & `flattenReduce` methods, works great!


dpmcmlxxvi:
Looks good. Glad to see the tests carried over without edits.

DenisCarriere:
@dpmcmlxxvi 👍 I'm also glad!

PR #726, opened 2017-05-10
Support GeometryCollection in @turf/line-chunk.

dpmcmlxxvi:
Addresses #710 for `line-chunk` to add GeometryCollection support. Updates include:
 - Replaces `featureEach`/`flatten` with `flattenEach`.
 - Adds GeometryCollection test fixtures.
---
Please fill in this template.

- [X] Use a meaningful title for the pull request. Include the name of the package modified.
- [X] Have read [How To Contribute](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#how-to-contribute).
- [X] Run `npm test` at the sub modules where changes have occurred.
- [X] Run `npm run lint` to ensure code style at the turf module level.

PR #727, opened 2017-05-10
line-segment: Simplify implementation and add GeometryCollection test fixture

dpmcmlxxvi:
Originally planned on adding Geometry support per #726 for `line-segment` until I realized that @DenisCarriere beat me to the punch. So, instead
- Simplified the implementation using `flattenEach`
- Added a GeometryCollection test fixture.

Question regarding current implementation. Is it intended for the method to silently ignore input geometries that are not lines or polygons?

I can see how that would be helpful for a user with a large feature collection that only wants the lines and polygons processed without having to create a new layer. But since Turf does a lot of input validation elsewhere, I wasn't sure if this was the intended design.

---

Please fill in this template.


DenisCarriere:
> Question regarding current implementation. Is it intended for the method to silently ignore input geometries that are not lines or polygons?

It was intended that Point & MultiPoints are simply ignored since they don't have lines, but if one would have a FeatureCollection of random stuff (Points, Lines, Polygon) and wanted to retrieve line segments from it then it would break if one point would exist in the dataset. Currently it's not harmful to have the `Points` included in the FeatureCollection vs. having the method throw an error would seem more annoying at times instead of helpful.

If there's a scenario that's worth throwing an error for Points in `line-segment` let me know and we can discuss this further.

packages/turf-line-segment/index.js
```diff
@@ -1,18 +1,13 @@
...
-var lineString = helpers.lineString;
-var featureCollection = helpers.featureCollection;
+var lineString = require('@turf/helpers').lineString;
 
 /**
  * Creates a {@link FeatureCollection} of 2-vertex {@link LineString} segments from a {@link LineString|(Multi)LineString} or {@link Polygon|(Multi)Polygon}.
  *
  * @name lineSegment
- * @param {Geometry|FeatureCollection|Feature<LineString|MultiLineString|MultiPolygon|Polygon>} geojson GeoJSON Polygon or LineString
+ * @param {GeometryCollection|Geometry|FeatureCollection|Feature<LineString|MultiLineString|MultiPolygon|Polygon>} geojson GeoJSON Polygon or LineString
```
DenisCarriere: These JSDocs are becoming so long!... 😮 
I wonder if we could shorthand the Multi Geometries
```javascript
/**
 * @param {Geometry|FeatureCollection|Feature<(Multi)LineString|(Multi)Polygon>}
```

packages/turf-line-segment/index.js
```diff
@@ -32,56 +27,35 @@ module.exports = function (geojson) {
...
-        break;
-    case 'GeometryCollection':
-        geomEach(geojson, function (geometry) {
-            lineSegment(geometry, results);
-        });
-        break;
-    default:
-        lineSegment(geojson, results);
-    }
+    flattenEach(geojson, function (feature) {
```
DenisCarriere: So Clean!! ❤️

packages/turf-line-segment/index.js
```diff
@@ -32,56 +27,35 @@ module.exports = function (geojson) {
...
-    case 'MultiPoint':
-    case 'MultiLineString':
-    case 'MultiPolygon':
-        geojson = feature(geojson);
+        coords = getCoords(geometry);
+        break;
+    case 'LineString':
+        coords = [getCoords(geometry)];
     }
-    geomEach(flatten(geojson), function (geometry) {
```
DenisCarriere: Love it! 🏅

Issue #714, opened 2017-05-06
New module proposal - Rhumb line


DenisCarriere:
# New module proposal - Rhumb Line

Variations of Rhumb Line implementations.

- `@turf/rhumb-bearing`
- `@turf/rhumb-destination`
- `@turf/rhumb-distance`

![image](https://cloud.githubusercontent.com/assets/550895/25773667/9479b450-324f-11e7-8947-ec289f5a27d3.png)

## `@turf/rhumb-destination`

Similar to [`@turf/destination`](https://github.com/Turfjs/turf/blob/master/packages/turf-destination/index.js), however instead of using the Haversine formula it uses the Rhumb Line path. Can be implemented using [`rhumbDestinationPoint` from `geodesy`](https://github.com/chrisveness/geodesy).

## `@turf/rhumb-bearing`

Similar to [`@turf/bearing`](https://github.com/Turfjs/turf/blob/master/packages/turf-bearing/index.js), however instead of using the Haversine formula it uses the Rhumb Line path. Can be implemented using [`rhumbBearingTo ` from `geodesy`](https://github.com/chrisveness/geodesy).

## `@turf/rhumb-distance`

Similar to [`@turf/distance`](https://github.com/Turfjs/turf/blob/master/packages/turf-distance/index.js), however instead of using the Haversine formula it uses the Rhumb Line path. Can be implemented using [`rhumbDistanceTo ` from `geodesy`](https://github.com/chrisveness/geodesy).

![image](https://cloud.githubusercontent.com/assets/550895/25774118/8a5bf318-3257-11e7-9e77-4cb35efd677d.png)

CC: @stebogit 
Ref: https://github.com/Turfjs/turf/issues/681 https://github.com/Turfjs/turf/issues/684

stebogit:
@DenisCarriere I'm gonna work on this. 
Should I create a branch for each module, or just one is ok?

DenisCarriere:
They seem pretty easy to implement, one branch should be good enough.

stebogit:
Implemented with PR #728

Issue #684, opened 2017-04-22
`degreesToDistance` different for Latitude and Longitude


stebogit:
Looking at `@turf/helpers` `distanceToRadians` and `distanceToDegrees` methods I noticed that if you move a point E-W or N-S of the same amount of degrees and calculate the distance between the original point and the moved point, you actually get two different distances.
Is that for some reason correct?
If so, why? 
Then, how could you create for example a square from a `point' that actually looks like a square? Using `@turf/bearing`, due to #681, at 90˚ the end points are not at the same latitude, while varying the coordinates you don't get a square:

![screen shot 2017-04-22 at 12 14 46 am](https://cloud.githubusercontent.com/assets/12717225/25302428/a08e966a-26f2-11e7-9117-4e8c828bb98a.png)

I added also some [test on `@turf/helpers`](https://github.com/Turfjs/turf/commit/709eff00c1de771c39b04932ab5c55f93d60a6ee) to highlight the issue.

DenisCarriere:
## Web Mercator

>  if you move a point E-W or N-S of the same amount of degrees and calculate the distance between the original point and the moved point, you actually get two different distances.

The Leaflet map you have in the background is using [Web Mercator projection](https://en.wikipedia.org/wiki/Web_Mercator) which is an equidistance projection (distances are equal only in meters NOT degrees).

Because of the earth's curvature, calculating the distance of 1 degree North does NOT equal 1 degree West. The map you are showing is **accurate** to represent those points the way they are. 

If you decided to visualize the same thing using a Geodetic map projection ([WGS84](https://en.wikipedia.org/wiki/World_Geodetic_System)) then those two points will "look" like they have an equal distance.

## Square

> how could you create for example a square from a point' that actually looks like a square?

I don't think using `@turf/bearing` would be the appropriate tool to create a square. You could convert your [LatLng coordinates to Meters](https://github.com/DenisCarriere/global-mercator#lnglattometers), apply your distance calculations to get your 4 corners and then convert your 4 coordinates back [Meters to LatLng](https://github.com/DenisCarriere/global-mercator#meterstolnglat).

DenisCarriere:
Going to close this issue, however feel free to continue the discussion if you have more questions/comments.

stebogit:
@DenisCarriere so the output of `@turf/point-grid`, which creates new points by coordinates (i.e. degrees right?), if represented on a Geodetic map would not result to be a square?
[...]

PR #728, opened 2017-05-10
Rhumb line modules

stebogit:
Added
`@turf/rhumb-bearing`
`@turf/rhumb-destination`
`@turf/rhumb-distance`

Solves #714 and #681 
Ref: #684


stebogit:
@DenisCarriere I don't understand why the tests fail, on my computer everything pass... 😞

DenisCarriere:
This is the error, I'll have a look at it.

```
Errored while running npm script 'test' in '@turf/along'
Error: Command failed: npm run test 
/home/travis/build/Turfjs/turf/packages/turf-rhumb-destination/index.js:43
    if (distance === undefined || distance === null || distance < 0) throw new Error('distance is required');
```

packages/turf-rhumb-destination/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@turf/destination",
```
DenisCarriere: @stebogit Here's your problem 🤦‍♂️  You've named it `@turf/destination` and when `lerna boostrap` creates the symbolic links it overwrites the "real" `@turf/destination` with the rhumb module. I'll fix this and the tests should pass.

packages/turf-rhumb-destination/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "version": "4.3.0",
```
DenisCarriere: For new modules, define them as `4.0.0` (base number of major release) because whenever I publish with Lerna it will automatically bump up the appropriate version (using the same version number as the next release will cause a version conflict with Lerna).

stebogit: got it 👍

packages/turf-rhumb-destination/package.json
stebogit: thanks!

packages/turf-rhumb-bearing/index.js
```diff
@@ -0,0 +1,60 @@
+// https://en.wikipedia.org/wiki/Rhumb_line
+// http://www.movable-type.co.uk/scripts/latlong.html#rhumblines
+var getCoords = require('@turf/invariant').getCoords;
```
DenisCarriere: Since you're only dealing with points you can use [`getCoord`](https://github.com/Turfjs/turf/tree/rhumb/packages/turf-invariant#getcoord) which has some extra point validation built-in.

packages/turf-rhumb-bearing/index.js
```diff
@@ -0,0 +1,60 @@
...
+var GeodesyLatLon = require('geodesy').LatLonSpherical;
+
+/**
+ * Takes two {@link Point|points} and finds the bearing angle between them along a Rhumb line
+ * i.e. the angle measured in degrees start the north line (0 degrees)
+ *
+ * @name rhumbBearing
+ * @param {Feature<Point>} start starting Point
+ * @param {Feature<Point>} end ending Point
```
DenisCarriere: Might be worth adding `Geometry Object` & `Array<number>` as valid input (input type supported when using `getCoord()`).

```javascript
* @param {Geometry|Feature<Point>|Array<number>} start starting Point
```

packages/turf-rhumb-destination/index.js
```diff
@@ -0,0 +1,75 @@
...
+var point = helpers.point;
+var radiansToDistance = helpers.radiansToDistance;
+var degrees2radians = helpers.degrees2radians;
+
+/**
+ * Returns the destination {@link Point} having travelled the given distance along a Rhumb line from the
+ * origin Point with the (constant) given bearing.
+ *
+ * @name rhumbDestination
+ * @param {Point|Array<number>} origin starting point
```
DenisCarriere: Same JSDocs as above `Geometry|Feature<Point>|Array<number>`

packages/turf-rhumb-destination/index.js
```diff
@@ -0,0 +1,75 @@
...
+ * var point = {
+ *   "type": "Feature",
+ *   "properties": {},
+ *   "geometry": {
+ *     "type": "Point",
+ *     "coordinates": [-75.343, 39.984]
+ *   }
+ * };
+ * var distance = 50;
+ * var bearing = 90;// function (from, distance, bearing, units) {
```
DenisCarriere: `// function (from, distance, bearing, units) {` ?

packages/turf-rhumb-destination/index.js
```diff
@@ -0,0 +1,75 @@
...
+ */
+module.exports = function (origin, distance, bearing, units) {
+    // validation
+    if (!origin) throw new Error('origin is required');
+    if (distance === undefined || distance === null || distance < 0) throw new Error('distance is required');
+    if (bearing === undefined || bearing === null) throw new Error('bearing is required');
+
+    units = units || 'kilometers';
+
+    var distanceInMeters;
```
DenisCarriere: Might be worth having `distanceInMeters` as a method in `@turf/helpers`

packages/turf-rhumb-bearing/index.js
stebogit: I haven't understood the difference between the two yet... Maybe some additional explanation on the doc would help?
I'll use getCoord anyway 👍

packages/turf-rhumb-destination/index.js
stebogit: 👍 I was thinking about it; maybe `distanceInKm` (just * 1000 to have meters) as that is `Turf`'s default in many cases?

packages/turf-rhumb-destination/index.js
```diff
@@ -0,0 +1,76 @@
...
+ * destination.properties['marker-color'] = '#f00';
+ * point.properties['marker-color'] = '#0f0';
+ * var addToMap = [point, destination]
+ */
+module.exports = function (origin, distance, bearing, units) {
+    // validation
+    if (!origin) throw new Error('origin is required');
+    if (distance === undefined || distance === null) throw new Error('distance is required');
+    if (bearing === undefined || bearing === null) throw new Error('bearing is required');
+    if (!(distance >= 0)) throw new Error('distance must be greater than 0');
```
DenisCarriere: Changed this if statement from `(distance < 0)`. If your input is text, then the result would be `false` since text cannot be less than 0.

```bash
> -4 < 0
true
> 5 < 0
false
> 'miles' < 0
false
```

packages/turf-rhumb-destination/index.js
```diff
@@ -0,0 +1,76 @@
...
+
+    units = units || 'kilometers';
+
+    var distanceInMeters;
```
DenisCarriere: This entire switch statement can be replaced with only:

```javascript
var distanceInMeters = radiansToDistance(distanceToRadians(distance, units), 'meters');
```

packages/turf-rhumb-bearing/test.js
```diff
@@ -0,0 +1,64 @@
...
+        const finalBearing = rhumbBearing(start, end, true);
+
+        const result = {
+            "initialBearing": initialBearing,
+            "finalBearing": finalBearing
+        };
+        if (process.env.REGEN) write.sync(directories.out + name +'.json', result);
+        t.deepEqual(load.sync(directories.out + name + '.json'), result, name);
+
+        // TODO adopt the following graphical output once rhumbDestination is published
```
DenisCarriere: 👍 Sounds like a plan!

packages/turf-rhumb-distance/test.js
```diff
@@ -0,0 +1,46 @@
...
+            greatCircleDistance: round(distance(pt1, pt2, 'kilometers'), 6),
+            radians: round(rhumbDistance(pt1, pt2, 'radians'), 6),
+            degrees: round(rhumbDistance(pt1, pt2, 'degrees'), 6)
+        }
+
+        if (process.env.REGEN) write.sync(directories.out + name + '.json', distances);
+        t.deepEqual(distances, load.sync(directories.out + name + '.json'), name);
+
+        // Removed since (point1 = true & point2 = false)
+        // t.ok(distances.kilometers < distances.greatCircleDistance, name + ' distance comparison');
```
DenisCarriere: @stebogit Can you have a look at this failing test case, I don't know if the GreatCircle distance is always greater than Rhumb Line Distance (I would assume that your test statement is correct, however the output says otherwise)

DenisCarriere:
@stebogit Ok I'm done with commits 🤓 Have a glance at it again and give me a 👍 if you're good to merge this, looks good to me!

packages/turf-rhumb-bearing/index.js
DenisCarriere: Yes more documentation is always better 👍 
The Gist is `getCoord` if you want to retrieve a Point Array (with point validation), `getCoords` for the entire coordinates as an Array (limited validation).

packages/turf-rhumb-destination/index.js
DenisCarriere: Yes many libraries use meters as the unit of measurement.

I solved this using:

```javascript
var distanceInMeters = radiansToDistance(distanceToRadians(distance, units), 'meters');
```
Might be worth having a method that wraps this behavior. Doesn't have to be to Meters it can be to any distance unit type.

```javascript
convertDistance(1000, 'miles', 'meters')
```

I dunno about the method name, but something like that.

packages/turf-rhumb-distance/test.js
stebogit: The correct test is reverse, as the GreatCircle line is the shortest path:
`distances.kilometers < distances.greatCircleDistance`
However it still fails now because of the approximation error due to the incorrect earth radius. I ran the corrected test with 6371km radius and it passes for both cases.

packages/turf-rhumb-bearing/index.js
stebogit: Why then not renaming them `getCoords` and `getCoordsArray` (I guess this would be suitable for `LineString`, `Polygon`, etc. right?); I believe that would make it more clear and intuitive.

packages/turf-rhumb-destination/index.js
stebogit: very simple and elegant! 👍 👍

stebogit:
Looks good to me 😃

Issue #368, opened 2016-05-04
Implement parallel_offset


danvk:
Turf has a [buffer method](http://turfjs.org/static/docs/module-turf_buffer.html). It would be great if it had a [parallel_offset method](http://toblerity.org/shapely/manual.html#object.parallel_offset) as well, ala shapely.

cc @nrathi

danvk:
This might be as simple as calling [`setSingleSided`][1] on a `BufferParameters` object.

[1]: http://bjornharrtell.github.io/jsts/1.1.1/apidocs/org/locationtech/jts/operation/buffer/BufferParameters.html#setSingleSided-boolean-

clhenrick:
Related recent question on gis.stackexchange: http://gis.stackexchange.com/questions/85358/how-to-add-an-offset-to-a-geojson-layer-in-leaflet/

rowanwins:
I needed this so I've made v1 - just finalising a few tests etc but should have a pull request soon!

DenisCarriere:
👍 Sweet! That's awesome

DenisCarriere:
Implemented in PR https://github.com/Turfjs/turf/issues/729.

Will be available in the next minor release `v4.4.0`

PR #729, opened 2017-05-10
New module @turf/line-offset

rowanwins:
# New Module `@turf/line-offset`

Adds a new lineOffset module as per [this issue](https://github.com/Turfjs/turf/issues/368). Basically takes an input line and returns a new line offset by the distance.

## Geometry Support

- [x] LineString => LineString
- [x] MultiLineString => MultiLineString

## Geometry **NOT** supported
- Point => LineString (alternative `@turf/circle`?)
- MultiPoint => MultiLineString (alternative `@turf/circle`?)
- Polygon => LineString (alternative `@turf/buffer` / `@turf/transform-scale`)
- MultiPolygon => MultiLineString (alternative `@turf/buffer` / `@turf/transform-scale`)
- FeatureCollection/GeometryCollection (might contain mixed geometries and cause errors)

## To-Do
- [x] Start and end points of the resulting line shall not be connected.
- [ ] Equidistance offset (similar to issue https://github.com/Turfjs/turf/issues/660 / https://github.com/Turfjs/turf/pull/718)

## JSDocs

```javascript
/**
 * Takes a {@link LineString|line} and returns a {@link LineString|line} at offset by the specified distance.
 *
 * @name lineOffset
 * @param {Geometry|Feature<LineString>} line input line
 * @param {number} offset distance to offset the line (can be of negative value)
 * @param {string} [units=kilometers] can be degrees, radians, miles, kilometers, inches, yards, meters
 * @returns {Feature<LineString>} Line offset from the input line
 * @example
 * var line = {
 *   "type": "Feature",
 *   "properties": {},
 *   "geometry": {
 *     "type": "LineString",
 *     "coordinates": [[-83, 30], [-84, 36], [-78, 41]]
 *   }
 * };
 *
 * var offsetLine = turf.lineOffset(line, 2, 'miles');
 *
 * //addToMap
 * var addToMap = [offsetLine, line]
 */
```

## Examples
image


DenisCarriere:
@rowanwins Can you provide more information in your Initial Description, that way when we reference this PR we have all the information available for the module at the top.

Use this as a rough example: https://github.com/Turfjs/turf/pull/700

> And for some reason the benchmark isn't printing results?

You need to upgrade the benchmark version (v1.0.0 has issues with ES6 syntax).

```bash
$ yarn add --dev tape benchmark
```

packages/turf-line-offset/bench.js
```diff
@@ -0,0 +1,64 @@
...
+      ],
+      [
+        -77.01982498168944,
+        38.89236892551996
+      ]
+    ]
+  }
+};
+
+var route = JSON.parse(fs.readFileSync(__dirname + '/test/fixtures/route.geojson'));
```
DenisCarriere: Try to make the new `benchmark.js` files as minimalistic as possible, use a few of the latest modules as examples:
https://github.com/Turfjs/turf/blob/0cfbda88456a89c7cd988ac17afd6056a42f59a8/packages/turf-rhumb-distance/bench.js

packages/turf-line-offset/index.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="geojson" />
+
+type LineString = GeoJSON.Feature<GeoJSON.LineString>;
+
+/**
+ * http://turfjs.org/docs/#lineOffset
+ */
+declare function lineOffset(line: LineString | GeoJSON.LineString, distance: number, units?: string): LineString;
```
DenisCarriere: Recently I've been adding the `Units` types pulled from `@turf/helpers`.
https://github.com/Turfjs/turf/blob/0cfbda88456a89c7cd988ac17afd6056a42f59a8/packages/turf-rhumb-distance/index.d.ts

packages/turf-line-offset/index.js
```diff
@@ -0,0 +1,70 @@
...
+var intersection = require('intersection');
+var helpers = require('@turf/helpers');
+var lineString = helpers.lineString;
+var distanceToDegrees = helpers.distanceToDegrees;
+
+/**
+ * Takes a {@link LineString|line} and returns a {@link LineString|line} at offset by the specified distance.
+ *
+ * @name lineOffset
+ * @param {Feature<LineString>|Geometry<LineString>} line input line
```
DenisCarriere: You can group the JSDoc types together
```javascript
/**
 * @param {Geometry|Feature<LineString>} line input line
```

packages/turf-line-offset/bench.js
rowanwins: Doh didn't push the right benchmark file!

packages/turf-line-offset/index.js
```diff
@@ -0,0 +1,70 @@
...
+            var seg1 = {start: {x: segment[0][0], y: segment[0][1]}, end: {x: segment[1][0], y: segment[1][1]}};
+            var seg2 = {start: {x: seg2Coords[0][0], y: seg2Coords[0][1]}, end: {x: seg2Coords[1][0], y: seg2Coords[1][1]}};
+            var int = intersection.intersect(seg1, seg2);
+            segment[1][0] = int.x;
+            segment[1][1] = int.y;
+            seg2Coords[0][0] = int.x;
+            seg2Coords[0][1] = int.y;
+        }
+    });
+    var finalCoords = segments.map(function (segment) {
```
DenisCarriere: Might be worth using `forEach` or `for` loop since it will be significantly faster than using `map` (~30-40% faster - Use benchmark to get an exact %)

packages/turf-line-offset/index.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+module.exports = function (line, offset, units) {
+    var segments = [];
+    var offsetDegrees = distanceToDegrees(offset, units);
+    var coords = getCoords(line);
+
+    coordEach(line, function (currentCoords, currentIndex) {
+        if (currentIndex !== coords.length - 1) {
+            var outCoords = processSegment(currentCoords, coords[currentIndex + 1], offsetDegrees);
+            segments.push(outCoords);
```
DenisCarriere: Iterating over all coords, then iterating over all segments, then iterating over segments again.

Looks like this could be done by only iterating your coordinates once.

You'll need to convert this code block into it's own method and use a `callback` to prevent the double iterations.

> I can tackle this if you don't feel comfortable using callbacks

packages/turf-line-offset/index.js
```diff
@@ -0,0 +1,70 @@
...
+            segment[1][1] = int.y;
+            seg2Coords[0][0] = int.x;
+            seg2Coords[0][1] = int.y;
+        }
+    });
+    var finalCoords = segments.map(function (segment) {
+        return segment[0];
+    });
+    finalCoords.push(segments[segments.length - 1][1]);
+    return lineString(finalCoords);
```
DenisCarriere: Should we also translate the properties from the input LineString?

packages/turf-line-offset/index.js
```diff
@@ -0,0 +1,70 @@
...
+};
+
+// Inspiration taken from http://stackoverflow.com/questions/2825412/draw-a-parallel-line
```
DenisCarriere: 👍 Nice reference

packages/turf-line-offset/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "url": "git://github.com/Turfjs/turf.git"
+  },
+  "keywords": [
+    "line",
+    "linestring",
+    "turf",
+    "offset"
+  ],
+  "author": "Turf Authors",
+  "license": "MIT",
```
DenisCarriere: Add yourself as `contributors`.

packages/turf-line-offset/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "bugs": {
+    "url": "https://github.com/Turfjs/turf/issues"
+  },
+  "homepage": "https://github.com/Turfjs/turf",
+  "devDependencies": {
+    "benchmark": "^1.0.0",
```
DenisCarriere: Upgrade `benchmark` & `tape`

packages/turf-line-offset/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "tape": "^3.5.0",
+    "@turf/helpers": "^4.1.0"
+  },
+  "dependencies": {
+    "@turf/helpers": "^4.1.0",
+    "@turf/invariant": "^4.2.0",
+    "@turf/meta": "^4.2.0",
+    "intersection": "0.0.1"
```
DenisCarriere: I'd drop `intersection` in favor of using [`@turf/line-intersect`](https://github.com/Turfjs/turf/tree/master/packages/turf-line-intersect).

packages/turf-line-offset/index.js
```diff
@@ -0,0 +1,70 @@
...
+        if (currentIndex !== coords.length - 1) {
+            var outCoords = processSegment(currentCoords, coords[currentIndex + 1], offsetDegrees);
+            segments.push(outCoords);
+        }
+    });
+    segments.forEach(function (segment, index) {
+        if (index !== segments.length - 1) {
+            var seg2Coords = segments[index + 1];
+            var seg1 = {start: {x: segment[0][0], y: segment[0][1]}, end: {x: segment[1][0], y: segment[1][1]}};
+            var seg2 = {start: {x: seg2Coords[0][0], y: seg2Coords[0][1]}, end: {x: seg2Coords[1][0], y: seg2Coords[1][1]}};
```
DenisCarriere: Pretty sure this can be handle by `@turf/line-intersect`.
We could include `Array<Array<number>>` support in `line-intersect` (geometry.coordinates)

packages/turf-line-offset/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "author": "Turf Authors",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/Turfjs/turf/issues"
+  },
+  "homepage": "https://github.com/Turfjs/turf",
+  "devDependencies": {
+    "benchmark": "^1.0.0",
+    "tape": "^3.5.0",
+    "@turf/helpers": "^4.1.0"
```
DenisCarriere: Missing `write-json-file` & `load-json-file` devDependencies

packages/turf-line-offset/index.js
rowanwins: Yeah it is a bit messy at the moment, I'll probably wont get a chance to look at it for another few nights so if you've got time and want to tackle it go nuts :)

packages/turf-line-offset/package.json
rowanwins: `intersection` works out where lines would intersect based on their direction, even if they don't actually intersect, eg it works irrespective of length. See [this interactive example](http://www.mathopenref.com/coordintersection.html).

packages/turf-line-offset/index.js
rowanwins: See comment above about how the `intersect` module differs from `line-intersect`

packages/turf-line-offset/index.js
rowanwins: could be faster, although the benchmark is performing 600,000 operations per second so we're talking about shaving milli-milli-milli-mill seconds :) Can probably be tackled in the overall refactoring of those loops

packages/turf-line-offset/bench.js
DenisCarriere: :) nice

packages/turf-line-offset/index.js
DenisCarriere: I'll have a look at the refactoring, sometimes it's not worth it, but sometimes it's an easy change for a pretty significant performance increase. Not a deal breaker whatsoever

packages/turf-line-offset/index.js
DenisCarriere: Will do :) I'll have a look at it more tomorrow, it's a good starting point so far!
+1

packages/turf-line-offset/package.json
DenisCarriere: Oooohh, yes that's completely different, I wonder if it's worth including that logic directly in Turf. I'm not a big fan of external dependencies, especially if it can be refactored in less than 30-50 LOC

packages/turf-line-offset/index.js
DenisCarriere: +1 yep makes sense

DenisCarriere:
To make the tests pass you might need to wrap the results with `@turf/truncate` since the error is coming from coordinates being slightly off of 0.000001.

packages/turf-line-offset/package.json
rowanwins: yeah it probably could be included, I had a bit of a look at the source code and it's not huge but it was enough of a hassle that I thought I'd pull in the external first just to test the concept first.

rowanwins:
Gday @DenisCarriere 

Have just updated the whole looping approach and its now much tidier I think. 

Only thing is the benchmark results look like they are slower than before but in reality I think it's now twice as fast, either a) my laptop is running on battery and goes slower, or b) I changed a benchmark test parameter. Anyway I've run the same benchmark test on both variations on the code this evening and the latest commit is twice as fast as the old.

Cheers

DenisCarriere:
Woot! Passing tests, one of the reasons why I decided to build `@turf/truncate`, it was a total pain having 0.000001 offset causing the `deepEqual` to fail.

rowanwins:
Updated to support lines with only 1 segment and lines that are completely straight (horizontal and vertical).

rowanwins:
PS Turns out that having my battery plugged in does effect the performance of my benchmark tests!
```
linestring-feature with changer === 1M ops/sec
linestring-feature with battery === 300K ops/sec
````

DenisCarriere:
👍 Yea the benchmark results are very dependent on CPU performance, but it's a good gage of performance. Also closing 10+ Google Chrome browser windows help 🤓

DenisCarriere:
@rowanwins Have a glance at the module now, I've included the `intersection` library to `index.js`.

DenisCarriere:
Great job 👍 @rowanwins, this module looks awesome! This will be really cool to do a demo of `@turf/line-offset` with an interactive web map.

image

rowanwins:
Nice work @DenisCarriere , thanks for bringing that dependency in. I reckon this will be a nifty little module!

DenisCarriere:
For sure, there's going to be a few of those methods that can be re-used.

DenisCarriere:
@rowanwins Just an update on this PR, once we get a few other geometries supported this will be ready to merge.

rowanwins:
Gday @DenisCarriere 

It did just cross my mind that this could conceivable be used as a replacement for a buffer module, particularly for polys... Not sure how performance would go but worth a quick trial.

DenisCarriere:
Dropping Polygon or MultiPolygon support since this algorithm is not designed for this type of offset. Below are examples of attempts using `line-offset` on Polygons:

**Red**: Input
**Blue**: Results

> 1000 miles - Inner rings become larger than outer rings

image

> -500 miles - Doesn't make any sense

image

> Start & End vertices are disconnected (might be worth looking into for LineString)

image

CC: @rowanwins

DenisCarriere:
> be used as a replacement for a buffer module, particularly for polys

I thought so too! Minus the crazy outputs! Lets stick with LineStrings on this module.

I really like it so far, but not with Polygons.

DenisCarriere:
@rowanwins Added MultiLineString support, I'm ready to merge this if ever you are 👍 with it.

DenisCarriere:
@rowanwins Another issue that can be addressed at a later time is the Equidistance offset, same issue we had with `@turf/buffer` https://github.com/Turfjs/turf/pull/718.

## Issue in Northern Latitudes

50km offset north/south directions, 12km offset west/east direction

https://github.com/Turfjs/turf/blob/bc34c926a11a424992b9e08643603fa81d54329f/packages/turf-line-offset/test/out/northern-line.geojson
image

rowanwins:
merge away. I might have a tinker with the polygon issue because im sure the code is there, just might need a bit of refactoring

stebogit:
@rowanwins @DenisCarriere I believe the output with (Multi)Polygons is actually the correct/expected result for this module, as it creates a parallel (and elongated/shortened) line alongside the original one. A Polygon just happens to be a closed line, however the module I think should not connect the start and end points of the resulting line.
If that was the actual need, then it would probably be more appropriate a scale transformation (#747) of the Polygon.
Ergo, I'd say is correct applying this module only to (opened) lines.

DenisCarriere:
> the module I think should not connect the start and end points of the resulting line.

👍 Agreed

> is correct applying this module only to (opened) lines.

Agreed, only supporting LineString would be best for this module.

stebogit:
@rowanwins @DenisCarriere I believe [these formulas]((https://en.wikipedia.org/wiki/Scale_(map)#Mercator_projection)) might help to correct the distortion here and in other modules.

DenisCarriere:
>  these formulas might help to correct the distortion here 

That's what I did for `@turf/buffer`.
https://github.com/Turfjs/turf/blob/master/packages/turf-buffer/index.js#L117-L163

rowanwins:
Hi @DenisCarriere & @stebogit 

One thought on the whole distortion issue, my understanding is that  projections have been designed differently, some projections are designed for preserving shape, while others are designed to preserve area etc etc

With the lines @DenisCarriere references above we're making that judgement for ppl as to what the best projection is (one that preserves shape). D3 actually [supports a bunch ](https://github.com/d3/d3-geo/blob/master/README.md#projections) which are useful for different purposes.

Rather than us forcing an choice on people perhaps we should be getting them to pass in their data with  the projection they deem appropriate?

Thoughts?

stebogit:
I personally don't have experience with projections different from [(Web)Mercator](https://en.wikipedia.org/wiki/Web_Mercator) (i.e. Google Maps). 
I'm wondering also how many users use `Turf` in a non-Mercator map, being Pseudo-Mercator the 
> "de facto standard for Web mapping applications"

and `Turf.js` a JavaScript (i.e. web) library.

Just my thought.
I'd be genuinely curious to know about interesting applications in other environments with different projections.

DenisCarriere:
> Rather than us forcing an choice on people perhaps we should be getting them to pass in their data with the projection they deem appropriate?

👎 TurfJS shouldn't be providing projection support.

@w8r explained it well in https://github.com/Turfjs/turf/issues/660#issuecomment-298678494.

>  The most important point is that you have to do the geometrical calculations in the R2 space where they make sense, so abstract equidistant does the trick.

DenisCarriere:
@rowanwins Using that equidistance projection will solve the offset being different. This isn't a projection "visual effect", the offset is legitimately different (12km & 50km).
image

DenisCarriere:
Going to merge this since it's 95% complete, only tiny issue is situations up north (can be in a later release or a new PR).

rowanwins:
Gday @DenisCarriere 

Just checking back in on the projection discussion. I wasn't proposing for turf to support projections, there is already [proj4js](http://proj4js.org/) to do that. I just wanted to check that by forcing a projection during a process like `buffer`, we weren't setting ourselves up for trouble. It's been way too long since I thought about datums and projections so apologies for the misuse of terminology etc :)

DenisCarriere:
Agree we should do the same as the buffer, however we should make the project/reproject process as a module or external dependency.

PR #744, opened 2017-05-15
Extended `turf-inside` input to any feature

stebogit:
@DenisCarriere I found useful extending the capability of this module

### DIFF
--- a/packages/turf-inside/index.js
+++ b/packages/turf-inside/index.js
@@ -1,6 +1,8 @@
 var invariant = require('@turf/invariant');
 var getCoord = invariant.getCoord;
 var getCoords = invariant.getCoords;
+var getGeomType = invariant.getGeomType;
+var explode = require('@turf/explode');
 
 // http://en.wikipedia.org/wiki/Even%E2%80%93odd_rule
 // modified from: https://github.com/substack/point-in-polygon/blob/master/index.js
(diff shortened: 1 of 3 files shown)


dpmcmlxxvi:
While I like the idea of extending this capability to all geometries, the implementation here is not actually checking that the input geometry is inside. It is only checking that its vertices are inside, which is not the same thing. For example, a doughnut like polygon with a line crossing just the inner polygon would return true even though the line is not inside. I think any user would expect the geometry to be inside and not just the vertices if it returns true.

Besides, @rowanwins is currently, working on the DE-9IM implementations of this in #707.

stebogit:
@dpmcmlxxvi you are totally right!
I implemented this cause `@turf/isobands` currently uses the same logic to test if `LinearRings`/`Polygons` belong to a `Polygon`. However the example you mentioned might result in an error. Good catch!
I look forward to include the new module(s) in `isobands` 😃 

I guess we can close this PR then?

DenisCarriere:
Lets leave `@turf/inside` as is and focus our efforts towards building robust `true/false` modules using the `boolean` themed modules.

**@turf/inside** would most likely fall under [**@turf/boolean-contains**](https://github.com/Turfjs/turf/tree/f2b95318e372ba6c6f98cc5965ccd2ca8975b635/packages/turf-boolean-contains).

DenisCarriere:
@stebogit Feel free to include your test fixtures in any of the relevant `boolean` modules.

Issue #722, opened 2017-05-08
pointOnLine.location does not return correct distance


Andi-Lo:
I currently wonder if the @turf/pointOnLine package does return a wrong location value. 
Here is a short fiddle of the problem:

http://jsfiddle.net/andi_lo/95q4Lgqc/

I would like to set a point (which is a train station in my project) onto the polyline (representing the routes shape) via `turf.pointOnLine` and use the returning location property as of how far the train traveled. However if I use turf.along with the distance provided by `turf.pointOnLine(...).properties.location` I will get a point that does not match my expectations. I would expect to get a point that lays onto the station but I get a point that is positioned falsely.

![turf_pointonline](https://cloud.githubusercontent.com/assets/7545568/25802370/5ac2c874-33f2-11e7-89b6-f7c0e6e9512b.jpg)

The used geojson is inside the fiddle. Feel free to ask for clarification if needed.

DenisCarriere:
@Andi-Lo Thanks for reporting this issue. I've managed to get the result closer to the station point by applying `reverse()` to the LineString's coordinates, however it's still a bit off, one would assume that doing `pointOnLine`+ `along` would result in nearly the same location.

We should try to replicate this behavior in the TurfJS tests.

## Example - `reverse()`

```javascript
lineString.features[0].geometry.coordinates.reverse();
```
![image](https://cloud.githubusercontent.com/assets/550895/25817242/be7611e4-33f4-11e7-89b1-10025d8b26fe.png)

DenisCarriere:
Related To-Do on `@turf/along`: https://github.com/Turfjs/turf/issues/691

Andi-Lo:
Thanks for reply,

until it is fixed I will use a workaround (measuring all sub-segments with turfs lineDistance module)

DenisCarriere:
@Andi-Lo Feel free to send a PR if you do find a solution for the `@turf/point-on-line` issue you mentioned.

dpmcmlxxvi:
@Andi-Lo Can you double check that this is resolved with PR #750 so the issue can be closed?

Andi-Lo:
@dpmcmlxxvi not sure how to validate.

I would need to checkout the turf master branch and get myself a bundled turf.min.js file and include that in my project to see if results are not as expected or is there a faster way?

dpmcmlxxvi:
[...]

Issue #691, opened 2017-04-25
along returns a point that is not found by pointOnLine


grundid:
When I generate a point using the along function the pointOnLine function does not see this point as on line. It also returns a NaN value for location.
Example:

```javascript
const line1 = turf.lineString([[7, 50], [8, 50], [9, 50]]);
const pointAlong = turf.along(line1, 10);
const pointOnLine = turf.pointOnLine(line1, pointAlong);
```
The resulting pointOnLine value looks like this:

```javascript
{ type: 'Feature',
  properties: { dist: 0.057664328657654904, location: NaN, index: 0 },
  geometry: { type: 'Point', coordinates: [ 7.139803396792796, 50 ] } }
```
It is not documented that location can become NaN. 

I would assume that the pointOnLine should match the pointAlong. This is the case if I pass for example the point [8, 50]. The result is:

```javascript
{ type: 'Feature',
  properties: { dist: 0, index: 1, location: 71.49662609671813 },
  geometry: { type: 'Point', coordinates: [ 8, 50 ] } }
```

DenisCarriere:
**Nice to do:** Can you use **```javascript** to style your Markdown codeblocks (my eyes are bad at differentiating comments vs. code blocks).

I also noticed location = `NaN` while updating the test cases today.

> location: distance along the line between start and the closest point.

I would assume if the start point & closest point are the same, then the "distance along the line" would be `0` instead of `NaN`.

dpmcmlxxvi:
@grundid  Can you double check that this is resolved with PR #750 so the issue can be closed?

grundid:
I'm sorry, but I don't have a test setup for this repo. I'll try after the release and reopen if this is still an issue.

DenisCarriere:
Added tests for this issue, `location` does not return `NaN` anymore.

PR #750, opened 2017-05-20
Fix location of intersect point for @turf/point-on-line.

dpmcmlxxvi:
This addresses #691 and #722. A few issues to point out:
 - The `location` property of the `intersectPt` was being computed using the `closestPt` which was initialized to `Infinity` which yielded spurious results. I believe it should be computed using `intersectPt`.
 - I think the current implementation was passing its tests because the expected fixture values were using the values from the algorithm itself. For example, the test `turf-point-on-line - points on top of line` uses a line with length `0.30576` and places `10` equally spaced points on it. So the first point should have been at `0.030576` but the test was expecting it at `0.021111`. Hence the fixtures have been updated accordingly.

I checked the results and #691 and #722 appear to have been fixed. Could use a double check though.

---
Please fill in this template.


DenisCarriere:
Thanks, this change makes sense, no more `NaN`

Issue #747, opened 2017-05-17
New module proposal: Transform


stebogit:
Inspired by PR #729 and the necessity to use a rotated `gridPoint` as `turf-isobands` input, I'd like to implement the following basic transformation functions, if of interest in `Turf`:

# `@turf/transform-translate` 
- [x] Moves a feature of `distance` kilometers (or `units`) on a specific `direction` angle
- [x] 3rd (z) coordinates support
```javascript
/**
@param {GeoJSON}` **object** to be translated
@param {Number}` **distance** length of the motion
@param {Number}` **direction** angle/bearing of the motion, in degrees from North (positive clockwise)
@param {String}` **units** (optional, default Km) unit for distance
@returns {Feature<any>}` the translated Feature
```
![screen shot 2017-05-17 at 12 11 25 am](https://cloud.githubusercontent.com/assets/12717225/26143351/749942d4-3a99-11e7-8148-3d234c6ee87b.png)

# `@turf/transform-scale`
- [x] Enlarges or shrinks a feature by a scale `factor`.
- [x] 3rd (z) coordinates support
```javascript
/**
@param {GeoJSON}` **object** to be scaled
@param {Number}` **factor** scale factor (positive number)
@param {Feature|Geometry<Point>|Array<number>}` **origin** point (of the Feature) around which the scaling will occur, defaults to centroid
@returns {Feature<any>}` the scaled Feature
```

![screen shot 2017-05-17 at 12 14 20 am](https://cloud.githubusercontent.com/assets/12717225/26143359/785a2c6c-3a99-11e7-814c-58764d72d5fb.png)

# `@turf/transform-rotate`
- [x] Rotates a feature of `angle` degrees around a fixed `point` (default to its center/center of mass/centroid)
- [x] 3rd (z) coordinates support (rotation only by z axis, i.e. on the x,y plane)
```javascript
/**
@param {GeoJSON}` **object** to be rotated
@param {Number}` **angle** extent of the rotation, in degrees from North (positive clockwise)
@param {Feature|Geometry<Point>|Array<number>}` **pivot**, point around which the rotation will occur, defaults to `centroid`
@returns {Feature<any>}` the rotated Feature
```

![screen shot 2017-05-17 at 12 18 13 am](https://cloud.githubusercontent.com/assets/12717225/26143363/7bcdfd88-3a99-11e7-9a3d-cd59ca5bfd46.png)
 
## Geometry Support
- `Geometry|Feature<Point>` => `Feature<Point>`
[...]

PR #752, opened 2017-05-22
Turf Transform

stebogit:
Implemented first module `@turf/transform-translate`
**Note**: I used `@turf/rhumb-destination` to create the translation vector, which defines the motion; needs to be checked if the length of the resulting translation matches the provided `distance`.

Please feel free to comment, suggest, correct, add tests.

(Ref: #747)


DenisCarriere:
👍 awesome stuff @stebogit I'm going to find the time this week to have a look at this in more details.

Keep it up!

stebogit:
@dpmcmlxxvi @morganherlocker @DenisCarriere I implemented the first module `transform-translate` using a matrix multiplication as I thought it would have been easy to use the same pattern with all the other transformations.
Although it works fine (setting the distance issue aside for now), I realized that for example in `turf-rotation` it would mean first translate the feature to the origin (i.e. `point([0,0])`), then perform the rotation, then translate it back to the `centroid`. This I believe might result slow(er) and introduce approximation errors.
The alternative would be using `turf-rhumb-destination` on each point for the translation, and I guess `turf-arc-line` on each point for the rotation.
However, having matrix operations as a base "format" for the transformation I do believe might be helpful pattern to create other transformations (flip, mirror, projections, etc.). It also helps to handle the third coordinate.

Thoughts? Suggestions?

stebogit:
Also, I worked based on the assumption that the transformations were meant to be on the plain, forgetting about, or better ignoring on purpose, the earth's curvature. This would make the transformations valid only in relatively "small" (compared to earth radius) areas.

DenisCarriere:
Using the `rhumb` modules is perfectly fine! I even prefer this type of transformation since it plays "nicer" with Web Applications using Web Mercator projection (Google maps/Bing/Mapbox/ESRI).

dpmcmlxxvi:
I agree that the `rhumb` approach might be more appropriate. It would be nice to have a common implementation (like transformation matrices) that could be re-used by multiple transforms but I think most users would expect a transformation in a geospatial library to perform geospatial friendly transformations and take the earth's curvature into account.

stebogit:
👍 I'll rewrite it then

stebogit:
I implemented `turf-rotate` on the `x,y` plane; for now I did not include a rotation by the other axes. 
Do we want to add that capability?

packages/turf-transform-rotate/index.d.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference types="geojson" />
+import {Point} from '@turf/helpers'
+
+export type GeometryObject = GeoJSON.GeometryObject;
+export type Feature<Geom extends GeometryObject> = GeoJSON.Feature<Geom>;
+
+/**
+ * http://turfjs.org/docs/#transform-rotate
+ */
+declare function rotate(geojson: Feature<any> | GeometryObject,
```
DenisCarriere: 🥇 These Typescript definitions are not obvious, the goal with this would be to have the Input be equal to the Ouptut.

Which would equal to:
- `Geometry<Point>` => `Feature<Point>`
- `Feature<Polygon>` => `Feature<Polygon>`

Right now the current state mutates the geometry type to `any`:
- `Geometry<Point>` => `Feature<any>`
- `Feature<Polygon>` => `Feature<any>`

**Example from translate**

```typescript
export type GeometryObject = GeoJSON.GeometryObject;
export type Feature<Geom extends GeometryObject> = GeoJSON.Feature<Geom>;

/**
 * http://turfjs.org/docs/#transform-translate
 */
declare function translate<Geom extends GeometryObject>(
    geojson: Feature<Geom> | Geom,
    distance: number,
    direction: number,
    units?: string): Feature<Geom>;
```
> **Note:** I'll push a commit to update those definitions.

packages/turf-transform-rotate/bench.js
```diff
@@ -0,0 +1,37 @@
...
+ * multiPoint x 66,089 ops/sec ±2.21% (76 runs sampled)
+ * multiPolygon x 2,478 ops/sec ±10.86% (56 runs sampled)
+ * no-rotation x 4,989 ops/sec ±17.30% (40 runs sampled)
+ * point x 49,158 ops/sec ±25.83% (42 runs sampled)
+ * polygon-with-hole x 3,125 ops/sec ±16.74% (44 runs sampled)
+ * polygon x 41,617 ops/sec ±14.82% (51 runs sampled)
+ */
+const suite = new Benchmark.Suite('turf-transform-rotate');
+for (const {name, geojson} of fixtures) {
+    let {angle, pivot} = geojson.properties || {};
```
DenisCarriere: `let` is used if the variable will be reassigned, in this scenario `const` should be used since it doesn't change after it's been declared.
This code will still work, but some aggressive linters have started to catch this as an linting error.

packages/turf-transform-translate/index.js
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * Moves any geojson Feature or Geometry of a specified distance on the provided direction angle.
+ *
+ * @name translate
+ * @param {Geometry|Feature<any>} geojson object to be translated
+ * @param {number} distance length of the motion; negative values determine motion in opposite direction
+ * @param {number} direction of the motion; angle from North between -180 and 180 decimal degrees, positive clockwise
+ * @param {string} [units=kilometers] in which `distance` will be express; miles, kilometers, degrees, or radians
+ *     (optional, default `kilometers`)
```
DenisCarriere: `(optional, default "kilometers")` is not required since it's already defined as optional using the square brackets.

```javascript
/**
 * @param {string} [units=kilometers]
```

packages/turf-transform-translate/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@turf/distance": "^4.3.0",
+    "benchmark": "^2.1.4",
+    "load-json-file": "^2.0.0",
+    "tape": "^4.6.3",
+    "write-json-file": "^2.0.0"
+  },
+  "dependencies": {
+    "@turf/helpers": "^4.3.0",
+    "@turf/invariant": "^4.3.0",
+    "@turf/meta": "^4.3.0",
```
DenisCarriere: Since `@turf/meta` does not exist in the `index.js` there's a new test in [`package/turf/test.js`](https://github.com/Turfjs/turf/blob/master/packages/turf/test.js#L109-L121) that will catch this as an error.

packages/turf-transform-translate/test.js
```diff
@@ -0,0 +1,58 @@
+const test = require('tape');
+const path = require('path');
+const fs = require('fs');
+const load = require('load-json-file');
+const write = require('write-json-file');
+const helpers = require('@turf/helpers');
```
DenisCarriere: Since the tests are now written in ES6 (NodeJS 6+) you can use the following when importing multiple methods from a module:

```javascript
const {point, featureCollection} = require('@turf/helpers');
```

packages/turf-transform-translate/test.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+});
+
+test('translate', t => {
+    for (const {filename, name, geojson}  of fixtures) {
+        let {distance, direction, units, zTranslation} = geojson.properties || {};
+
+        const translated = translate(geojson, distance, direction, units, zTranslation);
+
+        // style result
```
DenisCarriere: Wrap these styled results as their own dedicated functions (ex: `colorize(geojson)`).

That way the tests can be more readable and the `colorize` method can be reused if need be.

```javascript
function colorize(geojson) {
    // style result
    if (geojson.geometry.type === 'Point' || geojson.geometry.type === 'MultiPoint') {
        geojson.properties['marker-color'] = '#F00';
        geojson.properties['marker-symbol'] = 'star';
    } else {
        geojson.properties['stroke'] = '#F00';
        geojson.properties['stroke-width'] = 4;
    }
    return geojson;
}
```

packages/turf-transform-rotate/README.md
```diff
@@ -0,0 +1,51 @@
...
+(optional, default `centroid`)
+
+**Examples**
+
+```javascript
+var poly = turf.polygon([[0,29],[3.5,29],[2.5,32],[0,29]]);
+var rotatedPoly = turf.rotate(poly, 85);
+
+//addToMap
+var addToMap = [rotatedPoly];
```
DenisCarriere: Might want to include "before" & "after", simply seeing 1 polygon won't show you much (might want to change the colour of the fill).

packages/turf-transform-rotate/index.d.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference types="geojson" />
+import {Point} from '@turf/helpers'
+
+export type GeometryObject = GeoJSON.GeometryObject;
+export type Feature<Geom extends GeometryObject> = GeoJSON.Feature<Geom>;
```
DenisCarriere: Since this modules are single export methods, you can only have 1 dedicated export, can't have multiple exports.
```
index.d.ts(15,1): error TS2309: An export assignment cannot be used in a module with
 other exported elements.
```

packages/turf-transform-rotate/index.js
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @name rotate
+ * @param {Geometry|Feature<any>} geojson object to be rotated
+ * @param {number} angle of rotation (along the vertical axis), from North in decimal degrees, negative clockwise
+ * @param {Geometry|Feature<Point>|Array<number>} [pivot=`centroid`] point around which the rotation will be performed
+ * @param {number} [xRotation=0] angle of rotation, respect to the horizontal (x,y) plain, about the x axis;
+ * @param {number} [yRotation=0] angle of rotation, respect to the horizontal (x,y) plain, about the y axis;
+ * @returns {Feature<any>} the rotated GeoJSON feature
+ * @example
+ * var poly = turf.polygon([[0,29],[3.5,29],[2.5,32],[0,29]]);
```
DenisCarriere: invalid polygon 🤦‍♂️ , needs an extra `[]`

DenisCarriere: We should add these tests directly in the main Turf tests [`package/turf/test.js`](https://github.com/Turfjs/turf/blob/master/packages/turf/test.js).

A test that actually tests if all the JSDocs are being tested, it's currently doing this in `turf-www`.

@rowanwins

packages/turf-transform-rotate/test.js
```diff
@@ -0,0 +1,66 @@
...
+
+        // style result
+        if (rotated.geometry.type === 'Point' || rotated.geometry.type === 'MultiPoint') {
+            rotated.properties['marker-color'] = '#F00';
+            rotated.properties['marker-symbol'] = 'star';
+        } else {
+            rotated.properties["stroke"] = "#F00";
+            rotated.properties["stroke-width"] = 4;
+        }
+        if (pivot) {
```
DenisCarriere: This type of logic makes the tests confusing.

Simplified this as a single function.

```javascript
function makePivot(pivot, geojson) {
    if (!pivot) return centroid(geojson);
    return point(getCoord(pivot), {'marker-symbol': 'circle'});
}
```

packages/turf-transform-rotate/test/in/polygon.geojson
```diff
@@ -0,0 +1,40 @@
+{
+  "type": "Feature",
+  "properties": {
+    "angle": 200,
+    "pivot": {
```
DenisCarriere: Not too sure if `pivot` should be nested as a GeoJSON point, could simply be an Array `[2.48291015625, 27.819644755099446]`

DenisCarriere: Noticed you applied this change to other fixtures (looks like this was the only one) 👍

packages/turf-transform-rotate/index.js
```diff
@@ -0,0 +1,114 @@
...
+ * var rotatedPoly = turf.rotate(poly, 100, 35);
+ *
+ * //addToMap
+ * var addToMap = [rotatedPoly];
+ */
+module.exports = function (geojson, angle, pivot, xRotation, yRotation) {
+    // Input validation
+    if (!geojson) throw new Error('geojson is required');
+    if (angle === undefined || angle === null || isNaN(angle)) throw new Error('angle is required');
+    if (pivot && !Array.isArray(pivot) && pivot.geometry.type !== 'Point' && pivot.type !== 'Point') throw new Error('invalid pivot');
```
DenisCarriere: This validation breaks if the inputs are `Geometry|Feature<Point>|Array<number>`.

`getCoord` already does point validation and will cause an error if the input is not a point.

Replaced with:

```javascript
if (pivot) pivot = point(getCoord(pivot));
else pivot = centroid(geojson);
```

packages/turf-transform-rotate/index.js
```diff
@@ -0,0 +1,114 @@
...
+        return polygon(rotatedCoords, properties);
+    case 'MultiPolygon':
+        rotatedCoords = getCoords(geojson).map(function (polyCoords) {
+            return polyCoords.map(function (ringCoords) {
+                return rotate(ringCoords, angle, pivot);
+            });
+        });
+        return multiPolygon(rotatedCoords, properties);
+    default:
+        throw new Error('geometry ' + type + ' is not supported');
```
DenisCarriere: Impossible to hit this `default` switch statement.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/switch
> If no default clause is found, the program continues execution at the statement following the end of switch.

packages/turf-transform-rotate/test.js
```diff
@@ -0,0 +1,66 @@
...
+
+    t.end();
+});
+
+test('rotate -- throws', t => {
+    const line = lineString([[10, 10], [12, 15]]);
+
+    t.throws(() => rotate(null, 100), 'missing geojson');
+    t.throws(() => rotate(line, null), 'missing angle');
+    t.throws(() => rotate(line, 56, 'notApoint'), 'invalid pivot');
```
DenisCarriere: Sometimes it's best to explicitly define which error you are expecting.

Ex: `invalid point`

```javascript
t.throws(() => rotate(line, 56, 'notApoint'), /coordinates must only contain numbers/, 'invalid pivot');
```

packages/turf-transform-rotate/test.js
```diff
@@ -0,0 +1,66 @@
...
+
+    t.end();
```
DenisCarriere: Added two more tests **Geometry Support** & **Input mutation**, I've been adding these to each tests since they are pretty standard and easy to drop in without too much effort.

```javascript
test('rotate -- mutated input', t => {
    const line = lineString([[10, 10], [12, 15]]);
    const lineBefore = JSON.parse(JSON.stringify(line));

    rotate(line, 100);
    t.deepEqual(line, lineBefore, 'input should not be mutated');
    t.end();
});

test('rotate -- geometry support', t => {
    const line = lineString([[10, 10], [12, 15]]);
    const pt = point([10, 10]);

    t.assert(rotate(line.geometry, 100), 'geometry line support');
    t.assert(rotate(line.geometry, 100, pt.geometry), 'geometry pt support');
    t.assert(rotate(line.geometry, 100, pt.geometry.coordinates), 'pt coordinate support');
    t.end();
});
```

packages/turf-transform-rotate/index.js
```diff
@@ -0,0 +1,114 @@
...
+ * @name rotate
+ * @param {Geometry|Feature<any>} geojson object to be rotated
+ * @param {number} angle of rotation (along the vertical axis), from North in decimal degrees, negative clockwise
+ * @param {Geometry|Feature<Point>|Array<number>} [pivot=`centroid`] point around which the rotation will be performed
+ * @param {number} [xRotation=0] angle of rotation, respect to the horizontal (x,y) plain, about the x axis;
+ * @param {number} [yRotation=0] angle of rotation, respect to the horizontal (x,y) plain, about the y axis;
+ * @returns {Feature<any>} the rotated GeoJSON feature
+ * @example
+ * var poly = turf.polygon([[0,29],[3.5,29],[2.5,32],[0,29]]);
+ * var rotatedPoly = turf.rotate(poly, 100, 35);
```
DenisCarriere: `pivot` param is also invalid 🤦‍♂️ 
```javascript
var poly = turf.polygon([[[0,29],[3.5,29],[2.5,32],[0,29]]]);
var rotatedPoly = turf.rotate(poly, 100, [15, 15]);
```

packages/turf-transform-rotate/index.js
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * Rotates any geojson Feature or Geometry of a specified angle, around its `centroid` or a given `pivot` point;
+ * all rotations follow the right-hand rule: https://en.wikipedia.org/wiki/Right-hand_rule
+ *
+ * @name rotate
+ * @param {Geometry|Feature<any>} geojson object to be rotated
+ * @param {number} angle of rotation (along the vertical axis), from North in decimal degrees, negative clockwise
+ * @param {Geometry|Feature<Point>|Array<number>} [pivot=`centroid`] point around which the rotation will be performed
+ * @param {number} [xRotation=0] angle of rotation, respect to the horizontal (x,y) plain, about the x axis;
```
DenisCarriere: @stebogit `xRotation` & `yRotation` isn't being used, should we drop this params or you intend on adding that in your next commits?

These options seem a bit too complex, simply having `angle` & `pivot` is plenty enough, no need to over complicate this module.

stebogit: Those are the (only?) way to support the `z` coordinate and rotations by the `x` and/or `y` axes; we'd need matrix operations for that though. No problem for me to add this feature, maybe after we have a first working version of the three `transform-` modules.
Having support of the third coordinate in some (e.g. `turf-transform-translate`) but not all the modules would be inconsistent/unexpected, wouldn't it?

In any way I still need to implement z-coordinate support (at least let it pass through the rotation "as is").

stebogit:
Thanks 😃

DenisCarriere:
> I agree that the rhumb approach might be more appropriate. It would be nice to have a common implementation 

> I'll rewrite it then

No need for rewrite, we can just name these modules as `@turf/transform-rhumb-rotate` & `@turf/transform-rhumb-translate`.

We might also think about making `distance`/`bearing`/`destination` modular by grouping all the `rhumb` methods into one single module `@turf/rhumb` and also create another group for `@turf/haversine`. That way it's easier to build a **transform** module by just hot swapping haversine/rhumb. 

CC: @stebogit @dpmcmlxxvi

DenisCarriere:
Woot! First try!! 👍 

image

DenisCarriere:
Since this is the only implementation of `transform-rotate` & `transform-translate`. I'm fine keeping the name as is (for now) until we add another transform module which supports `haversine`.

packages/turf-transform-rotate/index.js
DenisCarriere: We can keep the params and add `(not implemented)` in the description, I'm good with that.

stebogit:
>No need for rewrite, we can just name these modules as `@turf/transform-rhumb-rotate` & `@turf/transform-rhumb-translate`.

I believe the translated feature is not actually at the expected distance.
I'm pretty sure replacing the matrix multiplication with `turf-rhum-destination` for each point will solve the issue and actually move the feature along a rhumb line.
I'll give it a try later tonight.

>We might also think about making distance/bearing/destination modular by grouping all the rhumb methods into one single module @turf/rhumb

👍 I like that! 
Why not doing that also for the `transform` functions in a single module?
We might even have a `transform/rhumb` and `transform/haversine` grouping.

DenisCarriere:
Each module should still remain separated since this complicates the testing and the entire package structure. However we can definitely make a "group" module which imports all the sub modules and maybe has the distance/bearing matrixes.

The goal would be to create an entirely new `haversine` transform set of modules but without re-writing the entire logic, simply load a new distance matrix.

We can finish this first, and then later on we can think about all that grouping & nested modules.

DenisCarriere:
@stebogit 👍 I'm good with a merge, whenever you're ready with it click the **Squash and Merge** button.

packages/turf-transform-rotate/index.js
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * Rotates any geojson Feature or Geometry of a specified angle, around its `centroid` or a given `pivot` point;
+ * all rotations follow the right-hand rule: https://en.wikipedia.org/wiki/Right-hand_rule
+ *
+ * @name rotate
+ * @param {Geometry|Feature<any>} geojson object to be rotated
+ * @param {number} angle of rotation (along the vertical axis), from North in decimal degrees, negative clockwise
+ * @param {Geometry|Feature<Point>|Array<number>} [pivot=`centroid`] point around which the rotation will be performed
+ * @param {number} [xRotation=0] angle of rotation, respect to the horizontal (x,y) plain, about the x axis (not implemented);
```
dpmcmlxxvi: @stebogit Just some random thoughts. It states here that x and y rotations are not yet implemented. This will require some careful thought on how to combine with the given rhumb approach for a few reasons:
 - I don't know of a natural geospatial way to rotate along those axes.
 - The rhumb approach is not really a z rotation but a quasi-z rotation that rotates the points in x, y, and z along the earth surface.
 - Most importantly, the order of rotations in 3D are not interchangeable. So, just providing a scalar value for `xRotation` and `yRotation` would not be enough to define a unique rotation.

packages/turf-transform-rotate/index.js
```diff
@@ -0,0 +1,115 @@
...
+
+    // copy properties to avoid reference issues
+    var properties = {};
+    Object.keys(geojson.properties).forEach(function (key) {
+        properties[key] = geojson.properties[key];
+    });
+
+    var rotatedCoords;
+
+    var type = geojson.geometry.type;
```
dpmcmlxxvi: @stebogit @DenisCarriere Since the rotation is being applied on each coordinate separately, couldn't you use `coordEach` and transform each coordinate. The `coordEach` method seems to be passing around references so you could mutate them. That way you don't have to deal with all these switch cases. You just clone, run `coordEach`, and apply `rotate`. I think that would make the implementation much simpler and you don't have to worry about feature types.

packages/turf-transform-rotate/index.js
stebogit: @dpmcmlxxvi great point, I didn't think about the order of rotation. 👍 
I was concerned as well about a geospatial way to rotate a feature. Maybe we can just decide to drop this function all together.

@DenisCarriere ?

packages/turf-transform-rotate/index.js
stebogit: @dpmcmlxxvi I tried that, but `coordEach` truly passes the `callback` each single coordinate of the `coordinates` array of the feature, so the problem there is that for multi-features or composite features you loose the capability to recreate the single rotated (sub)feature (i.e. `LineString` in `MultiLineString`, or `LinearRing` of `Polygon`).
That's why the `switch` case.

packages/turf-transform-rotate/index.js
dpmcmlxxvi: Giving it some more thought, I suppose rotating x and y could be the equivalent of rotating along just a latitude rhumb line or just a longitude rhumb line but the order of rotation would still be an issue

packages/turf-transform-rotate/index.js
dpmcmlxxvi: Not sure what you mean. Unless I'm misunderstanding something it seems like the single coordinates are exactly what you want. For example, I get the following for a `MultiLineString`. I bumped each coordinate and got an output `MultiLineString` with each coordinate bumped. Isn't that basically what you want to do, but with a `rotate` transformation?

## Code
```javascript
const multiline = turf.multiLineString([ [[1,1], [2,2]], [[3,3], [4,4]]]);
turf.coordEach(multiline, function (coord, index) {
    coord[0] = coord[0] + 1;
    coord[1] = coord[1] + 1;
    console.log(index, coord);
});
console.log(JSON.stringify(multiline, null, 4));
```

## Output

```shell
0 [ 2, 2 ]
1 [ 3, 3 ]
2 [ 4, 4 ]
3 [ 5, 5 ]
{
    "type": "Feature",
    "properties": {},
    "geometry": {
        "type": "MultiLineString",
        "coordinates": [
            [
                [
                    2,
                    2
                ],
                [
                    3,
                    3
                ]
            ],
            [
                [
                    4,
                    4
                ],
                [
                    5,
                    5
                ]
            ]
        ]
    }
}
```

packages/turf-transform-rotate/index.js
DenisCarriere: I think rotating using a Z axis is a bit too intense and wouldn't be worth investing all that energy towards building that support that isn't needed (I personally won't be using it, the angle & pivot is plenty enough for me).

packages/turf-transform-rotate/index.js
DenisCarriere: I'm on @dpmcmlxxvi side for this one, using `coordEach` would simplify things, it's a bit of a hack way of thinking things. Once you get it running it's pretty clean, I've used this approach in [`@turf/truncate`](https://github.com/Turfjs/turf/blob/master/packages/turf-truncate/index.js#L46-L49).

stebogit: Man I'm so so dumb! Of course it would work that way, I was simply doing it all wrong.
Thank you very much @dpmcmlxxvi , I'll make it right right away!

dpmcmlxxvi: No problem. Those meta `Each` methods come in real handy.

DenisCarriere: @stebogit 👍 I personally add a `mutate=false` param to be able to properly test the benchmark results without being throttled by the [`JSON.parse(JSON.stringify(geojson))`](https://github.com/Turfjs/turf/blob/master/packages/turf-truncate/index.js#L41) combo which can be a pretty significant performance loss if one doesn't care having your input data mutated or not.

stebogit: I got it working, but I encounterd two issues I don't understand:
- why this does not work?
```javascript
turf.coordEach(multiline, function (coord, index) {
    var newCoord = ...rotating truff...;
    coord = newCoord;
});
```
only works assigning the two coordinates separately as @dpmcmlxxvi did.

- this approach actually transforms the `geojson` object, while I believe the expected and correct stateless result would be instead returning a translated (or rotated) *copy* of the original object. 
Ergo, what is the best way to copy/clone the original `geojson` object before performing the transformation?

stebogit: Found [clone](https://github.com/pvorb/clone), and it works well.
However, `benchmark` shows a significant decrease in performances... 😞

packages/turf-transform-rotate/test.js
```diff
@@ -45,8 +45,6 @@ test('rotate -- throws', t => {
     t.throws(() => rotate(featureCollection([line]), 100), /FeatureCollection is not supported/, 'featureCollection');
     t.throws(() => rotate(geometryCollection([line.geometry]), 100), /GeometryCollection is not supported/, 'geometryCollection');
```
stebogit: @DenisCarriere this currently fails because `geojson.type == 'Feature'` so it won't trigger the `Error`; was it working before?

stebogit:
Man now the code looks so nice! Love it! 😃 

@DenisCarriere in regard to [this your comment](https://github.com/Turfjs/turf/pull/752#issuecomment-303784180) I'm not sure I got what you mean.
Currently `transform-translate` uses a matrix for the transformation, which I think is neither a `rhumb` nor a `haversine` motion/transformation, am I right?
I was thinking to rewrite it with `turf-rhumb-destination` and `coordEach` applying there the same clean logic just used for `transform-rotate`.
What do you think?

stebogit:
Added a test case for the z coord. 
I noticed that the extra coordinate value is actually correctly handled by the transformation, however [`turf-truncate` in `test.js`](https://github.com/Turfjs/turf/blob/922e6e18547fa49d3cb5689ce9d6955d67c81d0b/packages/turf-transform-rotate/test.js#L30) removes it, so the test fails.

Is that an expected result of `truncate`?

packages/turf-transform-rotate/index.js
DenisCarriere: 👍 Yep! You have to include that `parse` + `stringify` combo to prevent input mutation. `Clone` is significantly slower so I wouldn't use that vs. the method suggested.

DenisCarriere:
@stebogit Turf truncate removes the `z` coordinate by default, however you can add coordinate 3 to [include the z coordinate](https://github.com/Turfjs/turf/blob/master/packages/turf-truncate/index.js#L9), maybe we should change the default value to 3 coordinates instead of 2 since GeoJSON allows 6 decimals & 3 coordinates.

DenisCarriere:
@stebogit Are you planning on refactoring `@turf/transform-translate` to `coordEach()`?

I can help refactor it after you've done your first pass at it.

packages/turf-transform-rotate/index.d.ts
```diff
@@ -11,7 +11,8 @@ declare function rotate<Geom extends Geoms>(
     angle: number,
     pivot?: Point,
     xRotation?: number,
-    yRotation?: number): Geom;
+    yRotation?: number,
```
stebogit: Are we keeping `xRotation` and `yRotation`?

stebogit:
@DenisCarriere [yes I am](https://github.com/Turfjs/turf/pull/752#issuecomment-303943164) 😃 , but I won't be able to do it until tonight (PST).
Feel free to go on if you want and have time to do it. I'll still have `turf-transform-scale` to play with tonight 😄

DenisCarriere:
👍 I'll let you tackle that one, lots of stuff is going on in that repo and it's going to go a lot faster if you do it since you built it, i'd get stomped if I get an error at the end. However `@turf/transform-rotate` is looking pretty slick, it's a good example to use as a starting point.

DenisCarriere:
These coveralls notifications are cool, but also a bit annoying.

Going to remove this for now, but I'm looking forward to have coverage feedback in the PRs.

packages/turf-transform-translate/index.js
```diff
@@ -0,0 +1,171 @@
...
+    if (distance < 0) {
+        distance = -distance;
+        direction = -direction;
+    }
+    var properties = geojson.properties || {};
+
+    // Clone geojson to avoid side effects
+    if (mutate === false || mutate === undefined) geojson = JSON.parse(JSON.stringify(geojson));
+
+    var motionVector = getCoords(rhumbDestination([0, 0], distance, direction, units));
```
dpmcmlxxvi: @stebogit  Not sure if this has been discussed but this `motionVector` based translation approach is not doing what you may be expecting it to do. The rhumb line should be calculated from the given point not from the origin and then translated. Won't give the same answer. You should be able to apply the same logic from the scale approach and just apply the rhumb transformation in the `coordEach` loop.

dpmcmlxxvi: OK, just saw the comment where you were planning on doing this.

packages/turf-transform-translate/test.js
```diff
@@ -56,18 +56,13 @@ test('rotate -- mutated input', t => {
...
-    // t.assert(translate(featureCollection([line]), 100, 50), 'featureCollection support');
-    // t.assert(translate(line.geometry, 100, 50), 'geometry line support');
-    // t.assert(translate(line.geometry, 100, 50), 'geometry pt support');
+    const line = lineString([[10, 10], [12, 15]]);
+    t.assert(translate(geometryCollection([line.geometry]), 100, 50), 'geometryCollection support');
+    t.assert(translate(geometryCollection([line.geometry]).geometry, 100, 50), 'geometryCollection support');
+    t.assert(translate(featureCollection([line]), 100, 50), 'featureCollection support');
+    t.assert(translate(line.geometry, 100, 50), 'geometry line support');
+    t.assert(translate(line.geometry, 100, 50), 'geometry pt support');
     // t.assert(translate(line.geometry.coordinates, 100, 50), 'pt coordinate support');
```
stebogit: @DenisCarriere I believe this test is actually wrong; we are passing here an array of coordinates, which is not a GeoJSON object

DenisCarriere: That's true, this should throw an error 👍

packages/turf-transform-translate/index.js
```diff
@@ -35,137 +30,26 @@ module.exports = function (geojson, distance, direction, units, zTranslation, mu
...
-        ];
-        var resultVector = multiplyMatrices(translationMatrix, coordsVector);
-        resultVector.pop(); // remove last element
-        if (!zTranslation) {
-            resultVector.pop(); // remove z value
-        }
-        var translatedCoords = [].concat.apply([], resultVector); // flatten translated vector
-        return translatedCoords;
+    // Translate each coordinate
+    coordEach(geojson, function (pointCoords) {
```
DenisCarriere: Wow, what a big change! 👍 Looks too good to be true!

stebogit:
@DenisCarriere how come that the branch has been merged but it is shown as still separate in the graph?
screen shot 2017-05-26 at 8 24 54 am

Can I still work on the `transform` branch to implement `transform-scale`?
and, if yes, once I'm done should I create a new PR on this same branch?

DenisCarriere:
@stebogit We've noticed that in the past, I believe the reason is because when I merged I chose "**Squash and merge**". I don't know if that's an intended feature/bug in GitHub.

It's best to start a new branch (like you've done).

Issue #745, opened 2017-05-17
Handling negative buffer with NaN coordinates


DenisCarriere:
Based on https://github.com/Turfjs/turf/pull/736#issuecomment-301937747, I wanted to add this **morphological closing** behavior to the test cases, however I noticed when entering a `radius` which is greater than the polygon, the geometry's coordinates will result in `NaN` values.

**Question:** Is this expected output correct? Or should we return `undefined` or an empty `FeatureCollection`?

### Example

https://jsfiddle.net/pvLx12n1/
```javascript
const poly = turf.polygon([[[11, 0], [22, 4], [31, 0], [31, 11], [21, 15], [11, 11], [11, 0]]]);
turf.buffer(poly, -500, 'miles').geometry.coordinates;
//=[ [ NaN, NaN ] ]
```

### Output Geometry

> Seems like this output would be an invalid GeoJSON Polygon

```json
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Polygon",
    "coordinates": [[null, null]]
  }
}
```

CC: @rafaelalmeida

rafaelalmeida:
I would be in favor of returning `undefined`. See the map of the polygon you made:

http://geojson.io/#id=gist:anonymous/7d16198bde157fea2b8783f2fd29dfcb&map=4/12.77/16.00

The vertical line is about 800 miles long (check by clicking it and going to the "Info" tab). With a -500 miles buffer, the middle part of the polygon will shrink 500 miles from each side, meaning they would "overlap" - shrunk bottom would go above shrunk top, basically "folding the polygon inside out".

Theoretically, we could arbitrarily _define_ the negative buffer operation such that the Polygon would "go around" if the negative buffer is large enough. However, I think it would not be a good idea, because:

1. Not sure if the underlying JSTS algorithm would support that, or even the practicality of making such an algorithm from scratch;
2. The results could be very confusing - given a large enough magnitude, your negative buffer would yield a polygon larger than the one you started with

That being said, since it's not clear how this would behave, and we did not _define_ the behavior, `undefined` would be the semantically correct way to represent this (unless there's other JS-wide or project-wide conventions around the semantics of `undefined` which would override this).

DenisCarriere:
👍 Thanks for the feedback @rafaelalmeida, I agree we return `undefined` if GeoJSON has no geometry. If the input is a FeatureCollection and none of the outputs have geometries then an empty FeatureCollection should be returned.

[...]

PR #753, opened 2017-05-23
Update docs in case of negative buffer @turf/buffer

DenisCarriere:
Ref: https://github.com/Turfjs/turf/issues/745
CC: @rafaelalmeida


rafaelalmeida:
👍

Issue #628, opened 2017-03-24
Modules to support z coordinate from geometry


DenisCarriere:
### Proposing

Enable modules that require having a `z` coordinate be able to retrieve this altitude data from the 3rd coordinate of the GeoJSON geometry `[x,y,z]`. The `z` param would still be available to retrieve the altitude from properties, however this would now be optional.

### Modules

List of modules that should be able to manipulate `z` coordinate directly from geometry.

- [x] [@turf/isobands](https://github.com/Turfjs/turf/pull/619)
- [x] [@turf/planepoint](https://github.com/Turfjs/turf/tree/master/packages/turf-planepoint)
- [ ] [@turf/tin](https://github.com/Turfjs/turf/tree/master/packages/turf-tin)
- [x] [@turf/isolines](https://github.com/Turfjs/turf/tree/master/packages/turf-isolines)
- [ ] [@turf/idw](https://github.com/Turfjs/turf/tree/master/packages/turf-idw)
- [ ] Missed modules??

Any thoughts?

CC: @marie15 @morganherlocker @tmcw @stebogit @rowanwins

marie15:
You didn"t miss any modules, nice. But my problem is I don't have z coordinate, so how can I get this coordinate?

DenisCarriere:
@marie15 Z coordinates is defined by elevation/altitude, that would be retrieved from your own personal datasets, TurfJS won't be able to help you calculate datasets without this information.

marie15:
As I said before I don't have this information, how can I get this
coordinate from latitude and longitude?
Thank you

2017-03-28 17:51 GMT+00:00 Denis <notifications@github.com>:

morganherlocker:
@marie15 You will need to source this information yourself and implement your own method of querying the value at a coordinate. This is outside the scope of Turf, but I would do some web searching and check out datasets like those provided by [USGS](https://nationalmap.gov/elevation.html). This data is normally delivered in a bulk raster format like a [DEM](https://en.wikipedia.org/wiki/USGS_DEM), which is not the type of data that Turf is designed to handle.

marie15:
I understood, thank you for clarification, I will do research about what
you suggested.

2017-04-05 0:48 GMT+00:00 Morgan Herlocker <notifications@github.com>:

DenisCarriere:
Ref: https://github.com/Turfjs/turf/issues/628 https://github.com/Turfjs/turf/issues/2

rowanwins:
[...]

PR #754, opened 2017-05-24
Support z-coordinate as input for @turf/pointplane

dpmcmlxxvi:
Addresses #628 for `@turf/pointplane`. PR has the following edits:
 - Uses z-coordinate if triangle properties `a`, `b`, or `c` are undefined. So, should be backwards compatible.
 - Replaces original test with simpler more meaningful test values. Besides, original test was only testing for a truthy response. Using simpler coordinates seemed easier to verify.
 - Updated `test.js` and `bench.js` syntax to ES6.

@DenisCarriere Let me know if this is what you had in mind.

---
Please fill in this template.


DenisCarriere:
@dpmcmlxxvi 👍 Yep that's what I had in mind to support Z coordinates.

My last commit added Geometry support using `@turf/invariant` to normalize the input.

dpmcmlxxvi:
@DenisCarriere  Thanks for the review/edits. On the inputs, it now accepts an array for the input point. I thought the general philosophy of Turf was to have only GeoJSON as input and output for geographical features.

DenisCarriere:
@dpmcmlxxvi Agreed, however the `Point` is the only exception since many times your point data will consist of `[x, y]` and not be in a Geometry/Feature (it's a pain having to convert your point data to Geometry or Feature if the only valid input is a point).

### Rules to allow `Array<number>`
If the input is ONLY a point and you must use [`getCoord` method](https://github.com/Turfjs/turf/blob/master/packages/turf-invariant/index.js#L8-L21) from `@turf/invariant` which performs the point validation.

PR #756, opened 2017-05-24
Validation for unused dependencies not required by index.js

DenisCarriere:
An error will now occur if you have extra `dependencies` that are not being required in `index.js`.

## Error example

```
not ok 2 turf-concave @turf/meta is not required in index.js
  ---
    operator: fail
    at: Test.t (/Users/mac/Github/turf-tests/packages/turf/test.js:124:54)
  ...

1..2
```

### DIFF
--- a/packages/turf-concave/package.json
+++ b/packages/turf-concave/package.json
@@ -35,7 +35,6 @@
   },
   "dependencies": {
     "@turf/distance": "^4.3.0",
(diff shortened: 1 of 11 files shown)


rowanwins:
ooo nice little test @DenisCarriere - I'm guilty of that one!

DenisCarriere:
👍 I'm guilty of this too! That's why I added this

PR #759, opened 2017-05-27
New module @turf/transform-scale

stebogit:
Ref: #747

@DenisCarriere @dpmcmlxxvi 
The output is currently not correct: it is scaled successfully, but for some reason it does not reflect the `factor` value.
I implemented [my proposal for the `origin`](https://github.com/Turfjs/turf/issues/747#issuecomment-304428783); I look forward to knowing your opinion.
Also, the scaling along the z-coordinate is not along a Rhumb line, but a simple multiplication by 'factor'. Is there a way, if it make sense at all, to perform such a scaling on `z`?

For the moment I'd keep it simple and consider only uniform scalings, i.e. scaling factor equal in all directions.


DenisCarriere:
Keeping it simple is always the winner in my book! 👍 Looks good so far

DenisCarriere:
Strange that the coverage doesn't compute correctly, it might be looking at the first tests and then upload right away.

image

packages/turf-transform-scale/index.js
```diff
@@ -0,0 +1,71 @@
...
+
+
+/**
+ * Moves any geojson Feature or Geometry of a specified distance along a Rhumb Line
+ * on the provided direction angle.
+ *
+ * @name scale
+ * @param {GeoJSON} geojson object to be scaled
+ * @param {number} factor of scaling, positive values greater than 0
+ * @param {boolean} [fromCenter=true] point from which the scaling will occur; it will default to the most
```
DenisCarriere: 👍 Love it!

We might need to re-word the JSDocs, sounds like the "default" is saying `False` but in fact it's `True`.

> point from which the scaling will occur; **`it will default`** to the most south-west point of a `bbox` containing the geojson, if true the scaling will origin from the `centroid` of the geojson

packages/turf-transform-scale/index.js
```diff
@@ -0,0 +1,71 @@
...
+ * //addToMap
+ * scaledPoly.properties = {stroke: '#F00', 'stroke-width': 4};
+ * var addToMap = [poly, scaledPoly];
+ */
+module.exports = function (geojson, factor, fromCenter, mutate) {
+    // Input validation
+    if (!geojson) throw new Error('geojson is required');
+
+    if (geojson.type === 'Point')
+        return (mutate === false || mutate === undefined) ? geojson : JSON.parse(JSON.stringify(geojson));
```
DenisCarriere: ?? Why the multiple `mutate === undefined` This should be on it's own and only defined once.

packages/turf-transform-scale/index.js
```diff
@@ -0,0 +1,71 @@
...
+    // Shortcut no-scaling
+    if (factor === 1)
+        return (mutate === false || mutate === undefined) ? geojson : JSON.parse(JSON.stringify(geojson));
+
+    // Define origin
+    var origin;
+    if (fromCenter === true || fromCenter === undefined) {
+        origin = centroid(geojson);
+    } else {
+        var box = bbox(geojson); // [ minX, minY, maxX, maxY ]
```
DenisCarriere: Calculating `BBox` can be slightly computational heavy, so if a `bbox` is already provided in the geojson (which is defined in the spec) we can use that vs. re-calculating it.

http://geojson.org/geojson-spec.html#bounding-boxes
```javascript
var box = (geojson.bbox) ? geojson.bbox : bbox(geojson);
```

packages/turf-transform-scale/index.js
```diff
@@ -0,0 +1,71 @@
...
+        origin = point([box[0], box[1]]);
+    }
+
+    // Clone geojson to avoid side effects
+    if (mutate === false || mutate === undefined) geojson = JSON.parse(JSON.stringify(geojson));
+
+    // Scale each coordinate
+    coordEach(geojson, function (pointCoords) {
```
DenisCarriere: Going to be using `coord` as variable name since the singular term implies it's a point:
- `coord === [x, y]`
- `coords === [[x1, y1], [x2, y2]]`

DenisCarriere:
@stebogit Got an idea for defining the point of origin.

What if allow a param called `fromCorner` with 5 options:

```javascript
var west = bbox[0];
var south = bbox[1];
var east = bbox[2];
var north = bbox[3];
```

- `ne`/`northeast`/ `topright` => `[east, north]`
- `nw`/`northwest`/ `topleft` => `[west, north]`
- `se`/`southeast`/ `bottomright` => `[east, south]`
- `sw`/`southwest`/ `bottomleft` => `[west, south]`
- `undefined` => center

This would allow an application to scale any GeoJSON in any 5 direction with relative ease.

stebogit:
@DenisCarriere super! :+1:

DenisCarriere:
@stebogit Cool! I'll add that, shouldn't be too long.

packages/turf-transform-scale/README.md
```diff
@@ -0,0 +1,52 @@
+# @turf/transform-scale
+
+# scale
+
+Moves any geojson Feature or Geometry of a specified distance along a Rhumb Line
```
stebogit: @DenisCarriere probably you can find a better description for the module, explaining the Rhumb "motion" of the pointa; this however is the `translate` description

packages/turf-transform-scale/index.js
stebogit: good to know!

packages/turf-transform-scale/index.js
```diff
@@ -0,0 +1,69 @@
...
+ * var addToMap = [poly, scaledPoly];
+ */
+module.exports = function (geojson, factor, fromCenter, mutate) {
+    // Input validation
+    if (!geojson) throw new Error('geojson is required');
+    if (typeof factor !== 'number' && factor === 0) throw new Error('invalid factor');
+    if (!factor) throw new Error('factor is required');
+
+    // Default params
+    var isMutate = mutate === false || mutate === undefined;
```
dpmcmlxxvi: Is this logic correct? If `mutate` is false then `isMutate` is true? Also, if `mutate` is not defined then `isMutate` is true. The logic below says if `isMutate` is true then you clone the data. I think either the logic or the variable names are backwards. You pick.

packages/turf-transform-scale/index.js
```diff
@@ -33,38 +33,36 @@ var getCoords = invariant.getCoords;
...
+    var isCenter = fromCenter === true || fromCenter === undefined;
+    var origin;
 
     // Shortcut no-scaling
-    if (factor === 1)
-        return (mutate === false || mutate === undefined) ? geojson : JSON.parse(JSON.stringify(geojson));
+    if (factor === 1) return geojson;
```
stebogit: @DenisCarriere what about:
`    if (factor === 1 || geojson.geometry.type === 'Point' || geojson.type === 'Point') return geojson;`
Scaling a `Point` is pointless 😄

DenisCarriere: Single Point true, MultiPoint would make sense

dpmcmlxxvi:
@stebogit Scaling z by a multiplication factor is fine. Rhumb lines are just along the earth's surface.

packages/turf-transform-scale/index.js
stebogit: this would skip only `Point`, not `MultiPoint`. Wouldn't it?

packages/turf-transform-scale/index.js
stebogit: I'd say just:
`if (mutate === true) geojson = JSON.parse(JSON.stringify(geojson));`
dropping `isMutate`

dpmcmlxxvi:
@stebogit You might consider allowing the `fromCorner` to be either the five strings mentioned by @DenisCarriere or a `Point`. Plus the name `fromCorner` might be a little misleading since, for example, the center is not a corner. Maybe just `origin`.

packages/turf-transform-scale/index.js
DenisCarriere: Yep MultiPoint won't be skipped. I'll add that in my next commit, wait out

packages/turf-transform-scale/index.js
```diff
@@ -0,0 +1,69 @@
...
+var getCoords = invariant.getCoords;
+
+
+/**
+ * Moves any geojson Feature or Geometry of a specified distance along a Rhumb Line
+ * on the provided direction angle.
+ *
+ * @name scale
+ * @param {GeoJSON} geojson object to be scaled
+ * @param {number} factor of scaling, positive values greater than 0
```
DenisCarriere: @stebogit Couldn't it also be negative scaling?

stebogit: @DenisCarriere how do you scale negative?
`0 - 1` => shrink
`1 - Infinity` => enlarge
`< 0` ?

packages/turf-transform-scale/index.js
dpmcmlxxvi: So if `mutate` is true then you clone the data? Seems backwards.

stebogit:
@dpmcmlxxvi 👍 
> Maybe just origin.

Why not `originPosition` then? Clear, unmistakable.

packages/turf-transform-scale/index.js
stebogit: Arrrgh! You're right! Dumb me...
Then
`if (!mutate === true) geojson = JSON.parse(JSON.stringify(geojson));`
Still no need for `isMutate`
Right?

stebogit:
> allowing the `fromCorner` to be [...] a Point

@dpmcmlxxvi  @DenisCarriere could the origin be anywhere?
i.e. do we want to allow this:
scale picture

I personally find it weird, not obvious, but if you think otherwise let's `origin` be a `Point`!

packages/turf-transform-scale/index.js
dpmcmlxxvi: That should work. Though there should be enough tests to ensure that it works and the data is not cloning when it should be and vice versa.

However, I usually prefer logical conditions to be explicit and not implicit which require the reader to sit there and figure out what is going on and have to remember javascript falsey rules. Just my preference. But if it works it works.

dpmcmlxxvi:
I think lots of scaling operations would be performed from origin points that are not related to the features. For example, fixing sensor mis-calibrations, mis-registrations, etc.

packages/turf-transform-scale/index.js
```diff
@@ -30,26 +30,19 @@ var getCoords = invariant.getCoords;
...
-    // Define origin
-    if (isCenter) origin = centroid(geojson);
-    else {
-        var box = (geojson.bbox) ? geojson.bbox : bbox(geojson); // [ west, south, east, north ]
-        origin = point([box[0], box[1]]);
-    }
+    if (factor === 1 || isPoint) return geojson;
 
     // Clone geojson to avoid side effects
     if (isMutate) geojson = JSON.parse(JSON.stringify(geojson));
```
stebogit: @DenisCarriere I'd put this before any `return`, since the `mutate` behaviour should be respected even if there is no actual `scaling`. That's why my original [multiple `mutate === undefined`](https://github.com/Turfjs/turf/pull/759#discussion_r118824374)

packages/turf-transform-scale/index.js
```diff
@@ -67,3 +60,45 @@ module.exports = function (geojson, factor, fromCenter, mutate) {
...
+
+/**
+ * Define Origin
+ *
+ * @private
+ * @param {GeoJSON} geojson GeoJSON
+ * @param {string} [fromCorner] sw/se/nw/ne/center
+ * @returns {Feature<Point>} Point origin
+ */
+function defineOrigin(geojson, fromCorner) {
```
stebogit: Great job! 👍

DenisCarriere:
> I personally find it weird, not obvious (it looks more like a scale + translate), but if you think otherwise let's origin be a user-defined Point!
We can add `Geometry|Feature<Point>|Array<number>` as a valid input.

DenisCarriere:
Added `origin` param, check it out now.

DenisCarriere:
Good to merge on my side.

@stebogit @dpmcmlxxvi anything we missing?

stebogit:
@DenisCarriere let me check the output of the module, I'm not sure it's actually correct.

DenisCarriere:
👍 Perfect! Merge away afterwards, don't worry about the coverage fail (all of these are at 100%)

stebogit:
@DenisCarriere @dpmcmlxxvi I fixed a couple of errors in the calculation and added some more tests. Now I believe the output is correct.
However I believe `centroid` is unexpected as default central `origin`, I'd rather use `center`. What do you think?

Example `MultiPolygon`:
screen shot 2017-05-28 at 12 45 32 am
screen shot 2017-05-28 at 12 46 47 am

DenisCarriere:
@stebogit @dpmcmlxxvi Shouldn't it be best for Multi(Feature) to have the origin point centered on each individual shape when doing the scale operations.

We could use `FlattenEach` for that (shouldn't be too much effort, very similar to `coordEach` syntax).

DenisCarriere:
## Handling FeatureCollection

@stebogit FeatureCollection should be handled as an exception, if `origin` is `undefined` or `string` we should apply the scale operation on each feature individually. If `origin` is provided as a **Point** then the entire FeatureCollection will be scaled using that Point.

### Proposed (Multiple centroids per feature)
image

### Before (Single centroid origin)
image

DenisCarriere:
> However I believe centroid is unexpected as default central origin, I'd rather use center. What do you think?
@stebogit Yep agreed, `center` might be a better "default" value.

dpmcmlxxvi:
@stebogit It looks good to me. @DenisCarriere As for scaling each feature separately in a FeatureCollection, it's an interesting approach but I'm wondering what you expect the use case to be? What geospatial application would someone need to scale each feature separately from, say, their own centroid? I would think scaling each separately would be the exception and by default treat the collection as the others.

DenisCarriere:
Actually `centroid` looks like it would be the better option vs. `center`.

## Centroid
image


## Center

image

DenisCarriere:
> As for scaling each feature separately in a FeatureCollection, it's an interesting approach but I'm wondering what you expect the use case to be? 

@dpmcmlxxvi  If you provide a FeatureCollection using the bare minimum parameters `scale(geojson, 2)`, what would you expect?

I personally would expect this:
image

DenisCarriere:
@dpmcmlxxvi  @stebogit Good news!! No issues up in the Northern Latitudes (~75-80 lat)

image

Bad (expected) news, some issues with a Fiji MultiPolygon.

> We can ignore this for now, but another day we look for a solution.

image
image

stebogit:
To me the difference between `FeatureCollection` and `Multi(Feature)` is the relationship between the `features`: the first case we have a collection of independent features (`FeatureCollection` is like a container for the features), while the second is actually a single feature described by multiple similar features. Therefore I'd expect a transformation on a collection as independently performed.

DenisCarriere:
@stebogit Agreed, FeatureCollection is simply a container of Features which `scale` should be applied to each individual feature. MultiFeature should be treated as an entire object when using Scale (not to use `flatten` transformation).

DenisCarriere:
The FeatureCollection handling logic is pretty straightforward.
```javascript
// Scale each Feature separately
if (geojson.type === 'FeatureCollection') {
    featureEach(geojson, function (feature, index) {
        geojson.features[index] = scale(feature, factor, origin);
    });
    return geojson;
}
// Scale Feature/Geometry
return scale(geojson, factor, origin);
```

dpmcmlxxvi:
@DenisCarriere I guess when I think of a scaling operation I normally expect there to be one fixed origin from which to scale the entire dataset. However, I'm fine with the current implementation because if the user wants a scaling from a single origin they can just provide a `Point` as the `origin`, correct?

stebogit:
@DenisCarriere @dpmcmlxxvi I guess we'd need this then (note `originIsPoint` is just a logical evaluation):
```javascript
// Scale each Feature separately
if (geojson.type === 'FeatureCollection' && !originIsPoint) {
    featureEach(geojson, function (feature, index) {
        geojson.features[index] = scale(feature, factor, origin);
    });
    return geojson;
}
// Scale Feature/Geometry
return scale(geojson, factor, origin);
```
However this behaviour would need to be clearly documented, as I don't think is obvious

dpmcmlxxvi:
@stebogit Actually, I think the current implementation handles that case as-is, it is just scaling the coordinates of each feature separately, as opposed to the coordinates of the entire collection, but either way it should be the same effect since the `origin` is the same.

DenisCarriere:
> However this behaviour would need to be clearly documented, as I don't think is obvious

Agreed, would need to clarify the FeatureCollection behavior.

stebogit:
@DenisCarriere the 180th Meridian problem is fascinating, I didn't know about it! 
Clearly we need to work on `turf-rhumb-destination` (which of course presents the same issue) and once fixed that we should have fixed also all the transformations, all affected.
I'll create an issue and work on it right away.

DenisCarriere:
👍 Yes the 180th Meridian problem is a fun one! Don't sweat it too much since it's a common problem for A LOT of GIS applications to handle. Especially once you start handing a Geometry which spans on both sides of the 180th Meridian (50km buffer around Fiji is a good example of that).

stebogit:
@DenisCarriere @dpmcmlxxvi I guess once we merge #770 and re-test everything we're good to merge this. Anything else missing?
Once done here I'll test the other `transform` modules against the 180th meridian issue.

DenisCarriere:
Should be good to go! We can update the fixtures once both PRs are in the master branch.

Issue #702, opened 2017-05-01
Getting error while finding intersection between a multipolygon and a polygon


amishas157:
I am getting following error while finding intersection between a multipolygon and a polygon:

```
/Users/AmishaSingla/Documents/node_modules/turf-intersect/node_modules/jsts/dist/jsts.min.js:14
this.resultGeom=this.computeGeometry(this.resultPointList,this.resultLineList,this.resultPolyList,t)},labelIncompleteNode:function(t,e){var n=this.ptLocator.locate(t.getCoordinate(),this.arg[e].getGeometry());t.getLabel().setLocation(e,n)},copyPoints:function(t){for(var e=this.arg[t].getNodeIterator();e.hasNext();){var n=e.next(),i=this.graph.addNode(n.getCoordinate());i.setLabel(t,n.getLabel().getLocation(t))}},findResultAreaEdges:function(t){for(var e=this.graph.getEdgeEnds().iterator();e.hasNext();){var n=e.next(),i=n.getLabel();i.isArea()&&!n.isInteriorAreaEdge()&&ii.isResultOfOp(i.getLocation(0,cn.RIGHT),i.getLocation(1,cn.RIGHT),t)&&n.setInResult(!0)}},computeLabelsFromDepths:function(){for(var t=this.edgeList.iterator();t.hasNext();){var e=t.next(),n=e.getLabel(),i=e.getDepth();if(!i.isNull()){i.normalize();for(var r=0;2>r;r++)n.isNull(r)||!n.isArea()||i.isNull(r)||(0===i.getDelta(r)?n.toLine(r):(f.isTrue(
```

The two polygons are : 
  - https://gist.github.com/amishas157/b7604c9021c0b49ec8c3dd237d4b0a80
  - https://gist.github.com/amishas157/21f43cbbdad585839ee0b155fb5defd0

Source Code:
https://jsfiddle.net/ob0b6w0g/1/

Looks like something is wrong with Multipolygon.

rowanwins:
Hi @amishas157 

Yes this is a known issue, unfortunately `JSTS` (on which the @turf/intersect module relies) does not handle MultiPolygons very well. We're currently working on a replacement which is hopefully only a few weeks away.

Cheers
Rowan

DenisCarriere:
👍 @amishas157 Thanks reporting this, will be sure to include these test fixtures in the next `@turf/intersect`.

amishas157:
[...]

Issue #580, opened 2017-02-26
New module @turf/slice


DenisCarriere:
New module **`@turf/slice`**, currently on [**`slice`** branch](https://github.com/Turfjs/turf/tree/slice)

Contributions and comments are welcomed before we publish to TurfJS.

**Example**
```javascript
var sliced = turf.slice(polygon, linestring);
```

**JSDocs**
```javascript
/**
 * Takes a {@link Polygon} and cuts it with a {@link Linestring}. Note the linestring must be a straight line (eg made of only two points).
 * Properties from the input polygon will be retained on output polygons. Internally uses [polyK](http://polyk.ivank.net/) to perform split.
 *
 * @name slice
 * @param {Feature<Polygon>} poly - single Polygon Feature
 * @param {Feature<LineString>} line - single Polyline Feature
 * @return {FeatureCollection<Polygon>} A FeatureCollection of polygons
 */
```

**Input**

![image](https://cloud.githubusercontent.com/assets/550895/23344200/12550b58-fc46-11e6-9658-9ab6294ca397.png)

**Output**

![image](https://cloud.githubusercontent.com/assets/550895/23344205/224fc58e-fc46-11e6-915d-c93cefb04012.png)

CC: @Turfjs/ownership

mourner:
Not a fan of bringing yet another computational geometry library as a dependency...

DenisCarriere:
@mourner Thanks for your input, I totally agree with your comment. Recently I've sent over many commits to the library (https://github.com/MartyWallace/PolyK/commits/master) before including it as a dependency.

Mourner you are the expert in Javascript computational geo libraries (big fan).

I'm willing to "Turfify" this library to Zero dependency (only Turf dependencies).

@mourner Can you review it after it's done (~1-2 weeks).

mourner:
@DenisCarriere oh, don't take my comment as the final word — it's just my first thought. Worth discussing. Additional questions that come to mind are:

- How much size does PolyK add to the bundle?
- How stable/robust/fast is the algorithm?
- How much of an overlap there is between JSTS (already a dependency) and PolyK?
- Is it really necessary to have this as a core package, or would it be easier to just publish as a separate module and free to use PolyK/any other deps needed?

@morganherlocker @tcql thoughts?

morganherlocker:
Same questions as @mourner for me, and one more. I definitely think this could be useful as a core module.

[...]

PR #767, opened 2017-05-28
Turf polygonize

NickCis:
# New Module `@turf/polygonize`

Polygonizes a set of Geometrys which contain linework that represents the edges of a planar graph. It's basically an implementation of [GEOS's Polygonizer](https://github.com/echoz/xlibspatialite/blob/master/geos/include/geos/operation/polygonize/Polygonizer.h).

I ended up porting this to _javascript_ because I wanted the same functionality as [Shapely's polygonize function](http://toblerity.org/shapely/manual.html#shapely.ops.polygonize). This python framework uses GEOS'S C api as backend.

## JSDoc

```javascript
/** Implementation of GEOSPolygonize function (`geos::operation::polygonize::Polygonizer`).
 *
 * Polygonizes a set of lines that represents edges in a planar graph. Edges must be correctly
 * noded, i.e., they must only meet at their endpoints. LineStrings must only have two coordinate
 * points.
 *
 * The implementation correctly handles:
 *
 * - Dangles: edges which have one or both ends which are not incident on another edge endpoint.
 * - Cut Edges (bridges): edges that are connected at both ends but which do not form part
 *     of a polygon.
 *
 * @param {FeatureCollection<LineString>} geoJson - Lines in order to polygonize
 * @returns {FeatureCollection<Polygon>} - Polygons created
 */
```

## Example

This example is the test found in [`test/in/complex.geojson`](https://github.com/NickCis/turf/blob/turf-polygonize/packages/turf-polygonize/test/in/complex.geojson)

```javascript
const polygonize = require('@turf/polygonize'),
    fs = require('fs'),
    input = JSON.parse(fs.readFileSync('./test/in/complex.geojson'));

console.log(JSON.stringify(polygonize(input)));
```

An example of [input's LineString](https://github.com/NickCis/turf/blob/turf-polygonize/packages/turf-polygonize/test/in/complex.geojson):

captura de pantalla de 2017-05-27 23-39-11

Turned into polygons:

captura de pantalla de 2017-05-27 23-42-17


DenisCarriere:
@NickCis 👍 This is a great new module! I'll send over a review

DenisCarriere:
👎 All source code should be written in ES5 (with the exceptions of tests could be written in ES6).

All TurfJS modules should be very minimal with only a single `index.js` & `test.js` file.

You should consider publishing `polygonize` as it's own module on npm and TurfJS could simply include it as a dependency.

- [ ] Convert code to ES5
- [ ] Only include a single `index.js`
- [ ] Only include single `test.js`
- [ ] Convert `/src` to external dependency (ex: `polygonize`)
- [ ] Add `@private` to other utils functions

packages/turf-polygonize/index.js
```diff
@@ -0,0 +1,18 @@
...
+ * Polygonizes a set of lines that represents edges in a planar graph. Edges must be correctly
+ * noded, i.e., they must only meet at their endpoints. LineStrings must only have two coordinate
+ * points.
+ *
+ * The implementation correctly handles:
+ *
+ * - Dangles: edges which have one or both ends which are not incident on another edge endpoint.
+ * - Cut Edges (bridges): edges that are connected at both ends but which do not form part
+ *     of a polygon.
+ *
```
DenisCarriere: Add `@name polygonize`

packages/turf-polygonize/test.js
```diff
@@ -0,0 +1,51 @@
...
+const directories = {
+  in: path.join(__dirname, 'test', 'in') + path.sep,
+  out: path.join(__dirname, 'test', 'out') + path.sep,
+};
+
+function getFullPath(filename, type='in') {
+  return path.join(directories[type], filename);
+}
+
+function readJsonFile(filename, type='in') {
```
DenisCarriere: You can use `load-json-file` which handles reading GeoJSON/JSON, most of the recent modules are using this.

NickCis:
1. I've published [polygonize](https://github.com/NickCis/polygonize) as a separate module.
2. The convertion to ES5 was done with babel, the transpiled code is the one published on [npm](https://www.npmjs.com/package/polygonize).
3. The PR includes only one `index.js` and one `test.js` (There is a folder called `test` with all the fixtures)

Travis is now complaining on a few things:
* `bench.js`: Is it OK to copy the `bench.js` structure of another module and to run the same test fixtures that `test.js` is running?
* `yarn.lock`: I've never used yarn, i would appreciate help with this one
* `index.d.ts`:  Is a transcript definition file or something of the sort?,  i've never written transcript, i really don't now what this file must include.

packages/turf-polygonize/test.js
```diff
@@ -0,0 +1,51 @@
...
+  polygonize = require('./'),
+  fs = require('fs'),
+  path = require('path');
+
+const directories = {
+  in: path.join(__dirname, 'test', 'in') + path.sep,
+  out: path.join(__dirname, 'test', 'out') + path.sep,
+};
+
+function getFullPath(filename, type='in') {
```
DenisCarriere: Don't know the purpose of this? Since the files are being saved to `test/out` you can simply use:

```javascript
directories.out + filename
```

packages/turf-polygonize/test.js
```diff
@@ -0,0 +1,51 @@
...
+    return JSON.parse(fs.readFileSync(getFullPath(filename, type)));
+  } catch(e) {
+    if (e.code !== 'ENOENT')
+      throw e;
+    return undefined;
+  }
+}
+
+function writeJsonFile(filename, data, type='out') {
+  fs.writeFileSync(getFullPath(filename, type), JSON.stringify(data));
```
DenisCarriere: Use `write-json-file` for this behavior

packages/turf-polygonize/test.js
```diff
@@ -0,0 +1,51 @@
...
+}
+
+fs.readdirSync(directories.in)
+  .filter(filename => !filename.startsWith('.'))
```
DenisCarriere: Which files are being filtered?

packages/turf-polygonize/test.js
```diff
@@ -0,0 +1,51 @@
...
+  .map(filename => {
+    return {
+      filename,
+      name: path.parse(filename).name,
+      input: readJsonFile(filename, 'in'),
+      output: readJsonFile(filename, 'out'),
+    }
+  })
+  .forEach(({filename, name, input, output}) => {
```
DenisCarriere: I would recommend your testing template with one of the recent modules.

https://github.com/Turfjs/turf/blob/master/packages/turf-transform-translate/test.js
https://github.com/Turfjs/turf/blob/master/packages/turf-transform-rotate/test.js

packages/turf-polygonize/test.js
```diff
@@ -0,0 +1,51 @@
...
+    test(`turf-polygonize :: ${name}`, t => {
+      const result = polygonize(input);
+      if (output) {
+        t.deepEqual(result, output);
+      } else {
+        t.skip(`${name} not found :: writing`);
+        writeJsonFile(filename, result);
```
DenisCarriere: Writing GeoJSON should only be saved if `process.env.REGEN` is true (or exists). Shouldn't be saved each time.

packages/turf-polygonize/test.js
```diff
@@ -0,0 +1,51 @@
...
+      }
+      t.end();
+    });
+  });
```
DenisCarriere: Add the following tests:
- [ ] Geometry Support
- [ ] Input mutation

Examples here: https://github.com/Turfjs/turf/blob/master/packages/turf-transform-rotate/test.js

packages/turf-polygonize/test.js
NickCis: I work with vim, vim generates swap files in the file's folder (`.*.swp` files). When editing tests you often forget to close the input file in order to eliminate the swap file.

packages/turf-polygonize/test.js
NickCis: The geoJsons are only written when no output test file exists.

I didn't know that turf has the convention of using that environmental variable. I'll change the test.

DenisCarriere:
- [x] Babel compile 👍 That's perfect
- [ ] `bench.js`: Every new module should copy one of the latest published modules, this would be the best example of benchmark tests. [`@turf/transform-rotate` `bench.js`](https://github.com/Turfjs/turf/blob/master/packages/turf-transform-rotate/bench.js)
- [ ]  `yarn.lock`: you can install it globally and run it (might need to modify .gitignore since `yarn.lock` was there in your branch).
```
$ npm install -g yarn
$ yarn
```
- [ ] `index.d.ts`:  I can help with that. Or attempt to create it using this [`@turf/transform-rotate index.d.ts`](https://github.com/Turfjs/turf/blob/master/packages/turf-transform-rotate/index.d.ts) as an example.

packages/turf-polygonize/test.js
NickCis: I don't understand what you are asking with geometry support. (I'm really new to geojson), should i test using a GeometryCollection as input, instead of always using FeatureCollection?.

(Is there a difference between Geometry(Collection) / Feature(Collection)?, Or is it just for naming / cohesion purposes ?)

DenisCarriere: A GeometryCollection is a geometry type of a Feature where a FeatureCollection is a list of Features.

The Geometry tests just make sure the module doesn't break when you input a `Geometry<LineString>` or a `Feature<LineString>`.

There's also MultiLineString that you could add as support.

packages/turf-polygonize/test.js
DenisCarriere: 😕 woot for vim

DenisCarriere:
> LineStrings must only have two coordinate points.

Should be removed, you can use [`@turf/line-segment`](https://github.com/Turfjs/turf/tree/master/packages/turf-line-segment) to easily solve this.

DenisCarriere:
This module should also support the following Geometry Types:

- [ ] `Geometry<LineString>`
- [ ] `Geometry<MultiLineString>`
- [ ] `Feature<LineString>`
- [ ] `Feature<MultiLineString>`
- [x] `FeatureCollection<LineString>`
- [ ] `FeatureCollection<MultiLineString>`

NickCis:
> Should be removed, you can use @turf/line-segment to easily solve this.

I've used [`@turf/meta :: coordReduce`](https://github.com/NickCis/polygonize/commit/944c737fed2a9b17705676d994b9543a5bf2d49a#diff-5c80696e0289692939b32d8ffea8ddf3R47) in order to iterate over the coordinates and segment the _LineString_.

I'm working on the _MultiLineString_ support. Sorry for the delay.

DenisCarriere:
@NickCis I've got MultiLineString support by adding `@turf/line-segment` at the top level. Converts all the Geometry into 2 vertex LineStrings.

```javascript
module.exports = function (lines) {
    return polygonize(lineSegment(lines));
};
```

NickCis:
I'm working some issues in the `polygonize` module, i'll update the `yarn.lock` file when finished

DenisCarriere:
Awesome stuff! So far everything looks good to me.

NickCis:
Finished update of `polygonize` package :)

packages/turf-polygonize/index.js
```diff
@@ -19,6 +18,4 @@ var lineSegment = require('@turf/line-segment');
  * @returns {FeatureCollection<Polygon>} Polygons created
  * @throws {Error} if geoJson is invalid.
  */
-module.exports = function (geojson) {
-    return polygonize(lineSegment(geojson));
-};
+module.exports = polygonize;
```
DenisCarriere: Going to keep the basic function syntax in case anyone wants to expand the TurfJS method they can easily do so.

```javascript
module.exports = function (geojson) {
  return polygonize(geojson);
}
```

packages/turf-polygonize/index.js
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * The implementation correctly handles:
+ *
+ * - Dangles: edges which have one or both ends which are not incident on another edge endpoint.
+ * - Cut Edges (bridges): edges that are connected at both ends but which do not form part of a polygon.
+ *
+ * @name polygonize
+ * @param {FeatureCollection|Geometry|Feature<LineString|MultiLineString>} geojson Lines in order to polygonize
+ * @returns {FeatureCollection<Polygon>} Polygons created
+ * @throws {Error} if geoJson is invalid.
```
DenisCarriere: If you are going to include this `@throws {Error}` in the JSDocs at least include the tests in TurfJS to make those errors fail.

Noticed you're making Polygon fail, any reason why this would fail? I can already envision this error being very annoying, example if you have a collection of LineStrings and you `polygonize` them you then get a FeatureCollection of Polygons, if you keep adding more LineStrings to that FeatureCollection and re-run `polygonize` then it would fail since the FeatureCollection will have mixed contents.

It seems like throwing an Error against Polygons would make less practical.

DenisCarriere:
Noticed you're making Polygon fail, any reason why this should fail? I can already see this error being very annoying, example if you have a collection of LineStrings and you polygonize them you then get a FeatureCollection of Polygons, if you keep adding more LineStrings to that FeatureCollection and re-run polygonize then it would fail since the FeatureCollection will have mixed contents.

It seems like throwing an Error against Polygons would make this module less practical.

NickCis:
@DenisCarriere I thought it would be more coherent to throw an error than just ignore another input that isn't a LineString ([Collection]Feature/Geometry<[Multi]LineString>). The exception is a good way to tell the user "Hey, are you sure that you want to send me all these things that i do not understand?".

But, as you are a maintainer of the library, i think that you have a better understanding of the target audience. If you explain me what will the expected behavior of the user, i'll happily make the necessary changes.

DenisCarriere:
@NickCis As for Polygon Features, you can simply just return itself and not need to polygonize Polygons.

```javascript
if (type === 'Polygon') return feature
```
The only thing that really need to be strict is the output, as long as the output is always the same.

As for Inputs the more geometry types the better, you could even attempt to properly support `GeometryCollection` (I don't know if that works with your library yet).

DenisCarriere:
Would be nice to `polygonize` with mixed contents (polygons & lines).

**Red:** LineString
**Blue:** Polygon
image

DenisCarriere:
Come to think of it, polygons with holes might get problematic... 😕 Might be best to leave out Polygons from the supported input Geometry types.

@NickCis Keep throwing that Error 👍

NickCis:
The library correctly supports `GeometryCollections` (there is also a [test for this](https://github.com/NickCis/polygonize/blob/master/test/in/geometry-collection-two-polygons.geojson), i think that i've forgotten to add it in turf).

As regards `(Multi)Polygons` as input, is a valid request, in a way i'll have to suppose they are a especial case of a `LineString` which starts and ends in the same point, i could transform them using `@turf/polygon-to-linestring`. I cannot store it and just returned, because, if there are is a line which creates a polygon inside the polygon, the starting polygon should not be returned.

And example is the following:

[captura de pantalla de 2017-06-01 12-25-20](https://gist.github.com/anonymous/df0862c236177dbe00a7ae4de48010a8)
**Red**: MultiLineString
**Blue**: Polygon

[captura de pantalla de 2017-06-01 12-29-09](https://gist.github.com/anonymous/530c60679f7130aa39ae75891ab06993)
The valid result is the 4 inner polygons, and not the one that contains them.

The problem with this approach is that the algorithm still assumes that lines are correctly nodded (i.e.: they only met at their endpoints). So, if the polygon doesn't have a vertex where the line starts, it won't work correctly. Perhaps, it could be useful to add a non default option to find all intersections and divide lines and polygons. I think that it will have a huge impact on performance. Does turf have a module that given `polygon`s and `LineString`s, it finds all intersections and add them as new vertex? (if not, well, i think that i'll be starting to write a new module, haha).

NickCis:
@DenisCarriere I didn't read you answer before posting, sorry!

DenisCarriere:
@NickCis This looks awesome, maybe that can be an entirely other module, I had attempted this before but it was getting too complex for the basic implementation I had.

I never finished it but this would be ideal for `@turf/polygon-slice`.

More info here:
https://github.com/Turfjs/turf/issues/580

NickCis:
Ok!, I'll look the issue and the branch! We'll follow the topic there!. Thanks!

maphew:
I'm not getting expected results from polygonize. Where is the appropriate place to ask questions about it? I'm a raw newbie to Turf and barely familiar with Node.

DenisCarriere:
@maphew this is the right place to ask questions, or you can submit an issue with a sample code. Use this issue https://github.com/Turfjs/turf/issues/702 as an example.

Issue #755, opened 2017-05-24
Add properties param to @turf/centroid


DenisCarriere:
Allow properties from GeoJSON input or properties param to be translated down to point.

End state would be to do the following:

```javascript
centroid(geojson, {'marker-symbol': 'circle'})
```

Other related modules to apply properties param:
- [x] [`@turf/centroid`](https://github.com/Turfjs/turf/tree/master/packages/turf-centroid)
- [x] [`@turf/center`](https://github.com/Turfjs/turf/tree/master/packages/turf-center)
- [x] [`@turf/center-of-mass`](https://github.com/Turfjs/turf/tree/master/packages/turf-center-of-mass)

stebogit:
@DenisCarriere I noticed this issue when working on `transform-scale`! 👍 
I'll submit a PR with the fix.

DenisCarriere:
Fixed in https://github.com/Turfjs/turf/pull/769

PR #769, opened 2017-05-29
Adds properties parameter to @turf/centroid and @turf/center

stebogit:
@DenisCarriere in order to add the parameter to `center-of-mass` we need to merge this to `master` first, as `center-of-mass` depends on `centroid`.

Ref: #755


DenisCarriere:
Interestingly you can actually update all 3 at once and `center-of-mass` will use the latest version of `centroid` since `lerna` uses symbolic links to all the dependencies of each modules.

DenisCarriere:
@stebogit Updated significantly the tests:

- Add properties support to center-of-mass
- Update Readmes
- Update typescript definitions
- Update tests using glob
- Update test fixtures
- Update benchmark results
- Added switch statement CenterOfMass
- Added Typescript tests (`types.ts`)

PR #771, opened 2017-05-29
Fixes 180th meridian crossing for @turf/rhumb-destination

### LINKED ISSUE #770: @turf/rhumb-destination issue at 180th meridian
## [The 180th meridian problem](https://macwright.org/2016/09/26/the-180th-meridian.html)

When `rhumb-destination` crosses the 180th meridian from east to west, for example from an `origin` with negative longitude (i.e. -179.5) to a destination `point` 100 kilometers west of it, it returns a positive longitude for the `destination`, which is generally interpreted by the Map (i.e. Bing/Google) as a point at the other side of the map.

#### Example:
```javascript
var destination = getCoord(rhumbDestination(point([-179.5, 16.5]), 100, -90));
```
Current output: `//= [ 179.56205350889832, -16.5]`
screen shot 2017-05-29 at 1 44 28 am

## Solution
A quick solution that works for most Web Mercator Maps is to return a longitude value greater than 180 degrees.
Correct/expected output:`//= [-180.4379464911, -16.5]`
screen shot 2017-05-29 at 1 52 47 am

stebogit:
Fixes #770, returning if necessary longitude greater than 180 degrees.

I'd see this as a bug fix rather than a breaking compatibility with previous version.

Modules depending on `@turf/rhumb-destination`:
- [ ] [`@turf/transform-translate`](https://github.com/Turfjs/turf/tree/master/packages/turf-transform-translate)
- [ ] [`@turf/transform-rotate`](https://github.com/Turfjs/turf/tree/master/packages/turf-transform-rotate)
- [ ] [`@turf/transform-scale`](https://github.com/Turfjs/turf/tree/master/packages/turf-transform-scale) (still under-construction)

@DenisCarriere we might want to mention this behaviour to the README?

Also, I was wondering if we needed now to check if the input has longitude greater than 180 degrees, so that it's normalized between -180..180 before the computation.

Finally, do you think we should add the same correction to `@turf/destination` and other modules?

cc: @dpmcmlxxvi

### DIFF
--- a/packages/turf-rhumb-destination/index.js
+++ b/packages/turf-rhumb-destination/index.js
@@ -51,5 +51,8 @@ module.exports = function (origin, distance, bearing, units) {
     var pt = new GeodesyLatLon(coords[1], coords[0]);
     var destination = pt.rhumbDestinationPoint(distanceInMeters, bearing);
 
+    // compensate the crossing of the 180th meridian (https://macwright.org/2016/09/26/the-180th-meridian.html)
+    // solution from https://github.com/mapbox/mapbox-gl-js/issues/3250#issuecomment-294887678
+    destination.lon += (destination.lon - coords[0] > 180) ? -360 : (coords[0] - destination.lon > 180) ? 360 : 0;
     return point([destination.lon, destination.lat]);
 };
--- a/packages/turf-rhumb-destination/test.js
+++ b/packages/turf-rhumb-destination/test.js
@@ -28,10 +28,10 @@ test('turf-rhumb-destination', t => {
         dist = (dist !== undefined) ? dist : 100;
         units = units || 'kilometers';
 
-        const resultPoint = rhumbDestination(inputPoint, dist, bearing, units);
-        const line = truncate(lineString([getCoords(inputPoint), getCoords(resultPoint)], {"stroke": "#F00", "stroke-width": 4}));
+        const destinationPoint = rhumbDestination(inputPoint, dist, bearing, units);
+        const line = truncate(lineString([getCoords(inputPoint), getCoords(destinationPoint)], {"stroke": "#F00", "stroke-width": 4}));
         inputPoint.properties['marker-color'] = '#F00';
-        const result = featureCollection([line, inputPoint, resultPoint]);
+        const result = featureCollection([line, inputPoint, destinationPoint]);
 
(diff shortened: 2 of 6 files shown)


packages/turf-rhumb-destination/index.js
```diff
@@ -51,5 +51,8 @@ module.exports = function (origin, distance, bearing, units) {
     var pt = new GeodesyLatLon(coords[1], coords[0]);
     var destination = pt.rhumbDestinationPoint(distanceInMeters, bearing);
 
+    // compensate the crossing of the 180th meridian (https://macwright.org/2016/09/26/the-180th-meridian.html)
+    // solution from https://github.com/mapbox/mapbox-gl-js/issues/3250#issuecomment-294887678
+    destination.lon += (destination.lon - coords[0] > 180) ? -360 : (coords[0] - destination.lon > 180) ? 360 : 0;
```
DenisCarriere: Ok as a patch, but what if the input is `[-720, 65]`.

I wrote a full library for this type of fix, this "simple" but really annoying to implement.

https://github.com/DenisCarriere/bbox-dateline

DenisCarriere: Example Point

```json
{
	"type": "Feature",
	"properties": {
		"bearing": -90,
		"dist": 100
	},
	"geometry": {
		"type": "Point",
		"coordinates": [
			-539.5,
			-16.5
		]
	}
}
```

DenisCarriere: Oh nevermind, it seems to work just fine.

image

DenisCarriere:
> Also, I was wondering if we needed now to check if the input has longitude greater than 180 degrees, so that it's normalized between -180..180 before the computation.

No we should keep the input coordinates the same, there's merit to having coordinates with +180 longitude, it's really up to the GIS application to handle those coordinates, Leaflet & MapboxGL can handle it just fine.

> Finally, do you think we should add the same correction to @turf/destination and other modules?

Yep I think we should apply this to these modules as well, this isn't a breaking change since this only affects Fiji or calculations crossing the 180th Meridian.

CC: @morganherlocker @mourner Feel free to jump in if you think otherwise.

DenisCarriere:
This fix reflects @tmcw's excellent [The 180th Meridian](https://macwright.org/2016/09/26/the-180th-meridian.html) article/blog.

PR #781, opened 2017-06-07
New @turf/isolines based on MarchingSquares.js

### LINKED ISSUE #390: redo turf-isolines

stebogit:
Fixes #390


DenisCarriere:
?? Why all the changes to `turf-bbox`??

Looks like you merged an old branch to this PR, might be best to close this PR and only send the changes for `@turf/isolines`.

👎

stebogit:
@DenisCarriere Wow man, I didn't see those! 😳 
It seems actually that it was my last `yarn` update (all I did was running `yarn` in my terminal) which generated all those changes in `turf-bbox` and `turf-helpers`; everything before was ok.
Shouldn't have I done that maybe?

DenisCarriere:
Humm.. strange, I don't think `yarn` would of caused that.

Who knows!

I've been starting to use [GitHub Desktop](https://desktop.github.com/) which allows you to easily `Update from Default Branch`. That will make your "stale" branch current to the Master branch.

image

stebogit:
@DenisCarriere you can easily see that those changes occurred only in [the last commit](https://github.com/Turfjs/turf/pull/781/commits/cced2cf4fd6f1ad52d324fdbf5d54fb2224d6975); and I swear I did only run `yarn` cause I thought it would have updated the `yard.lock` file (which it did, among the other changes).
And you can see from the [chart](https://github.com/Turfjs/turf/network) I branched out from the HEAD of master.

However do you want me to close this PR and copy all the good changes on a new branch? I believe though we might just undo the last commit and all should be fine.

DenisCarriere:
Undo the commit works. 👍

DenisCarriere:
That's really strange... Ah well... no worries 👍 

image

stebogit:
So, for the next time, should I run `yarn` to update the `yarn.lock` file or it get's updated maybe at merge or when Travis runs?

DenisCarriere:
You can still run `yarn`, but just keep track of your `git commit` & `git push` files.

stebogit:
Yay!! Undo done 😃

packages/turf-isolines/index.d.ts
```diff
@@ -1,11 +1,8 @@
...
-
-type Points = GeoJSON.FeatureCollection<GeoJSON.Point>;
-type LineStrings = GeoJSON.FeatureCollection<GeoJSON.LineString>;
+import {Points, MultiLineStrings} from '@turf/helpers'
 
 /**
  * http://turfjs.org/docs/#isolines
  */
-declare function isolines(points: Points, z: string, resolution: number, breaks: Array<number>): LineStrings;
+declare function isolines(points: Points, breaks: Array<number>, property?: string): MultiLineStrings;
```
DenisCarriere: Looks like you're missing `options`

You would need to build an interface for that.

```typescript
interface Options {
  isolineProperties: any[],
  commonProperties: any
}
```
Then declare the option as a param.

packages/turf-isolines/index.js
```diff
@@ -1,28 +1,27 @@
...
- * @param {string} z the property name in `points` from which z-values will be pulled
- * @param {number} resolution resolution of the underlying grid
- * @param {Array<number>} breaks where to draw contours
- * @returns {FeatureCollection<LineString>} isolines
+ * @param {Array<number>} breaks where to draw isolines
+ * @param {string} [zProperty='elevation'] the property name in `points` from which z-values will be pulled
+ * @param {Object} [options={}] options on output
+ * @param {Array<Object>} [options.isolineProperties=[]] GeoJSON properties passed, in order, to the correspondent
+ * isoline (order defined by breaks)
+ * @param {Object} [options.commonProperties={}] GeoJSON properties passed to ALL isolines
```
DenisCarriere: I would most likely change the name of the param to `properties`.

packages/turf-isolines/index.js
```diff
@@ -1,28 +1,27 @@
...
  *
  * @name isolines
  * @param {FeatureCollection<Point>} points input points
- * @param {string} z the property name in `points` from which z-values will be pulled
- * @param {number} resolution resolution of the underlying grid
- * @param {Array<number>} breaks where to draw contours
- * @returns {FeatureCollection<LineString>} isolines
+ * @param {Array<number>} breaks where to draw isolines
+ * @param {string} [zProperty='elevation'] the property name in `points` from which z-values will be pulled
+ * @param {Object} [options={}] options on output
```
DenisCarriere: If you decide to use options, then all "optional" params should be included (`zProperty` in options)

packages/turf-isolines/index.js
```diff
@@ -1,28 +1,27 @@
...
+ * @param {Array<Object>} [options.isolineProperties=[]] GeoJSON properties passed, in order, to the correspondent
```
DenisCarriere: I don't quite understand this param...? We might want to improve the description a bit

packages/turf-isolines/index.js
DenisCarriere: Or just drop `options` all together and just add 2 extra more params without the `Object`, there's very rare amounts of TurfJS modules that use this type of syntax.

packages/turf-isolines/index.js
```diff
@@ -34,67 +33,107 @@ var Conrec = require('./conrec');
...
-        for (var j = 0; j < tinResult.features.length; j++) {
-            var triangle = tinResult.features[j];
-            if (inside(pt, triangle)) {
-                pt.properties = {};
-                pt.properties[z] = planepoint(pt, triangle);
-            }
-        }
+module.exports = function (points, breaks, zProperty, options) {
+    // Input validation
+    var isObject = function (input) {
```
DenisCarriere: No need to nest this function inside another function.

packages/turf-isolines/index.js
```diff
@@ -34,67 +33,107 @@ var Conrec = require('./conrec');
...
+    };
+    collectionOf(points, 'Point', 'Input must contain Points');
+    if (!breaks || !Array.isArray(breaks)) throw new Error('breaks is required');
+    if (options.commonProperties && !isObject(options.commonProperties)) {
+        throw new Error('commonProperties is not an Object');
+    }
+    if (options.isolineProperties && !Array.isArray(options.isolineProperties)) {
+        throw new Error('isolineProperties is not an Array');
+    }
+    if (zProperty && typeof zProperty !== 'string') {
```
DenisCarriere: No need to check if `zProperty` exists or not
```javascript
if (typeof zProperty !== 'string') {
```

stebogit:
Thanks @DenisCarriere, always good to improve the code 😃 
I'll clean up as soon as I have few minutes

packages/turf-isolines/index.js
stebogit: If you look at the `bigMatrix` test it is how you can pass for example different colors to the `isolines`. The first property of the `isolineProperty` will be assigned to the first isoline (defined by the first `breaks` value), the second `isolineProperty` to the second isoline and so on.
Don't know how to better describe it in a concise way though...

packages/turf-isolines/index.js
stebogit: >  (would break backwards compatibility)

Well actually compatibility is already an issue, since the `resolution` parameter is not used/useful anymore (it is defined by the `pointGrid` input itself).
Maybe we keep this module update for the next major release?

> there's very rare amounts of TurfJS modules that use this type of syntax

We'll need to update `turf-isoband` as well then, as it follow this same syntax?
This would be a breaking change as well.

stebogit:
@DenisCarriere I implemented your suggestions, please take a look to `index.d.ts` cause I might have done it wrong.
I [named the last optional parameter `properties`](https://github.com/Turfjs/turf/pull/781#discussion_r121051232), which defines `properties.perIsoline` and `properties.toAllIsolines`, hopefully now is more clear.
Let me know what you think.

packages/turf-isolines/index.d.ts
```diff
@@ -1,11 +1,13 @@
-/// <reference types="geojson" />
+import {Points, MultiLineStrings} from '@turf/helpers'
 
-type Points = GeoJSON.FeatureCollection<GeoJSON.Point>;
-type LineStrings = GeoJSON.FeatureCollection<GeoJSON.LineString>;
+interface Properties {
+    perIsoline: Array<Object>,
```
DenisCarriere: 👍 Looks good, only thing is since `perIsolines` & `toAllIsolines` are optional, you will need to add `:?` to make them optional.

```typescript
interface Properties {
    perIsoline?: Array<Object>,
    toAllIsolines?: Object
}
```

However, it might be best to not have those `properties` tags as an Object and just by their own param.

```typescript
declare function isolines(
    points: Points,
    breaks: number[],
    zProperty?:string,
    propertiesPerIsoline?: Object[],
    propertiesToAllIsolines?: Object): MultiLineStrings;
```

**javascript code**
```javascript
isolines(points, breaks, 'temperature', [{foo: 'bar'}], properties)
```

DenisCarriere: Also `propertiesPerIsoline`, seems a bit more complex to use vs. the `propertiesToAllIsolines`. We might want to reverse them.

```javascript
const properties = {foo: 'bar'}
isolines(points, breaks, 'temperature', properties)
isolines(points, breaks, 'temperature', properties, [{name: 'break1'}, {name: 'break2'])
```

DenisCarriere:
👍 @stebogit You've outdone yourself this time, this module looks great! Can't wait to try this out on some big datasets.

image

DenisCarriere:
**Updates**
- Divide properties={} to two params
- Simplify Catch error logic (index.js)
- Update Typescript definition
- Add types.ts to test Typescript definition
- Add single task to benchmark
- Update Yarn.lock
- Simplify bench & test logic
- Save fixtures out to GeoJSON
- perIsoline overlaps properties from toAllIsolines

DenisCarriere:
👍 Going to merge, feel free to send commits on the master branch for any more little tweaks.

Issue #33, opened 2013-11-03
cluster


morganherlocker:
Point clustering for rendering lots of points

morganherlocker:
Ok, an O(n**2-ish) option that might be a good standin because I already wrote it :stuck_out_tongue_winking_eye: 

```js
var featurecollection = require('turf-featurecollection');
var distance = require('turf-distance');

module.exports = function(pts, maxDistance, units) {
  var clusters = turf.featurecollection([]);
  for(var i = 0; i < pts.features.length; i++) {
    var exemplar = true;
    for(var k = 0; k < clusters.features.length; k++){
      if(turf.distance(pts.features[i], clusters.features[k], units) < maxDistance) {
       exemplar = false;
     }
    }
    if(exemplar) clusters.features.push(pts.features[i]);
  }

  return clusters;
}
```

A slower and more memory hungry version would accumulate the cluster points and compute the centroid at the end. I think the tradeoffs might be worthwhile, but I'm going to stew on this for a bit.

stepankuzmin:
Hi @morganherlocker! Is there are any progress with this?

rowanwins:
Hi @stepankuzmin it looks like this has largely been tackled with the [supercluster ](https://github.com/mapbox/supercluster) library from mapbox. Documentation is a little sparse but hopefully if you follow the tests you'll get a few ideas.

Cheers
Rowan

cyrilchapon:
Clustering on a map and clustering some points is a very different feature.

I personaly need K-mean clustering inside an API that serves geojson.

So, is there any progress with this ?

morganherlocker:
> Clustering on a map and clustering some points is a very different feature.
> I personaly need K-mean clustering inside an API that serves geojson.
> So, is there any progress with this ?

Agree. No progress to report, but I still think this would be a great feature.

donmccurdy:
Would this differ from what 'supercluster' provides? Note that it can be used serverside, in which case it takes geojson points as input, and outputs geojson points for the clusters, annotated with point counts.

cyrilchapon:
Very thanks on documentation.

Well the zoomlevel thing config is philosohically anoying, but I should be able to deal with it 😄

I'm trying hard, I'll tell you

cyrilchapon:
I'm just struggling at making it work.

It aggregates all my points to a single point.

morganherlocker:
> Would this differ from what 'supercluster' provides? 
> [...]
[...]

PR #787, opened 2017-06-13
New module @turf/clusters

stebogit:
First draft of `@turf/cluster`, Ref: #33 

```javascript
var clusters = require('@turf/clusters')
var points = // featurecollection of points
var numberOfClusters = 4
var clustered = clusters(points, numberOfClusters)

// clustered is an object with the following attributes:
// - `points`: the input (featurecollection of) points with each point given a `cluster` number property
// - `centroids`: featurecollection of centroid points, each with its `cluster` number as property
```
screen shot 2017-06-13 at 1 22 42 am

Uses [`clusters`](https://github.com/NathanEpstein/clusters) for simplicity, but there might be a faster/better implementation.
Not sure if the output is ok, I tried to follow @tmcw's [suggestion](https://github.com/Turfjs/turf/issues/33#issuecomment-264036583).

Open to suggestions and further testing.


rowanwins:
ooo this looks nifty @stebogit

DenisCarriere:
- [ ] Change the module name to `cluster` => so far all the TurfJS module names are singular.
- [ ] 50% agree with the output type, at the moment I can't really think of a better way to do it 😕

DenisCarriere:
- [ ] All `properties` are lost, when creating these clusters it would be useful to preserve the existing properties.

stebogit:
@DenisCarriere I replaced `clusters` with [`skmeans`](https://github.com/solzimer/skmeans) which is incredibly faster (like thousands times!) and allows to conserve the input and all its properties.

stebogit:
@DenisCarriere 
> so far all the TurfJS module names are singular.

The more I think about it the more I believe this should be called 'clusters', since it's returning (and really expected to return) multiple clusters. The majority of Turf modules are singular because they deal with or generate a singular "thing", but few of them do return multiple elements and they have plural names, like `isobands`, `isolines`, `kinks`, `polygon-tangents`, `helpers`(functions).

DenisCarriere:
👍 Ok that works, we can name it `clusters`.

DenisCarriere:
>  I replaced clusters dependency with skmeans which allows to compute incredibly faster (like thousands times!) and to conserve the input and all its properties.

That's great, yea it was really slow before... I'm glad it's a lot faster now

packages/turf-cluster/index.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+module.exports = function (points, numberOfClusters) {
+    // Input validation
+    collectionOf(points, 'Point', 'Input must contain Points');
+    // Default Params
+    var count = points.features.length;
+    numberOfClusters = numberOfClusters || Math.round(Math.sqrt(count / 2));
+
+    // collect points coordinates
+    var data = featureReduce(points, function (prevValue, currentFeature) {
```
DenisCarriere: Using `coordEach` makes it about ~30-80% faster.

```javascript
    // collect points coordinates
    var data = [];
    coordEach(points, function (coord) {
        data.push(coord);
    });
```

DenisCarriere:
- [ ] Number of points must be greater or equal to `numberOfClusters` (add as input validation)

packages/turf-cluster/index.js
stebogit: 👍  good to know! I still don't master these methods 😕

stebogit:
@DenisCarriere I added the validation on `numberOfClusters`

DenisCarriere:
@stebogit 👍 Looks good!

cyrilchapon:
Awesome

DenisCarriere:
We should be pushing a new minor release soon with all these new modules being merged in the past week.

cyrilchapon:
What about the complexity on adding a `maxDistance` factor, as described in #33 ?

stebogit:
@cyrilchapon I'm not an expert in clustering, even less in clustering algorithms, however it seems to me an algo requiring a `maxDistance` would be substantially different that the `k-means` one.
Even reading through #33 I don't quite get what kind of feature you're suggesting; could you please give an example of use and application for that clustering method?

cyrilchapon:
@stebogit yep.. I'm not an expert too in custering algos, but it seems indeed that it would be a very different algorithm.

Speaking of API, I would be suggesting something like

```
var clusters = require('@turf/clusters')
var points = // featurecollection of points

//kmeans strategy
var options = {
  numberOfClusters: 4
}
var 
var kmeansClustered = clusters(clusters.consts.algorithms.KMEANS, points, options)

//OR

//maxDistance strategy
var options = {
  maxDistance: 3,
  maxDistanceUnit: 'km'
}
var kmeansClustered = clusters(clusters.consts.algorithms.SOMENAME, points, options)
```

or maybe

```
var clusters = require('@turf/clusters')
var points = // featurecollection of points

//kmeans strategy
var numberOfClusters = 4
var kmeansClustered = clusters.kmeans(points, numberOfClusters)

//OR

//maxDistance strategy
var maxDistance = 3
var maxDistanceUnit = 'km'
var kmeansClustered = clusters.somename(points, maxDistance, maxDistanceUnit)
```

But I'm not able to implement such a thing

DenisCarriere:
Both strategies shouldn't be implemented in the same module, they should be separate modules with unique names.

A good to read blog post from @mourner https://www.mapbox.com/blog/supercluster/ about Clustering.

`Supercluster` isn't meant for TurfJS, however some of the source code is good to look at for inspiration on how to tackle the **distance** strategy.

Another good library built by @mourner is [`kdbush`](https://github.com/mourner/kdbush) which is used by `Supercluster` and this would most likely be needed in this new cluster library.

Issue #220, opened 2015-02-12
A grid that accepts a polygon


xrmx:
Am replicating the walkability estimation applied to a whole italian region which is irregularly shaped. So i ended up doing this:

```
            var grid = turf.grid(polypiemonte.bbox_piemonte, 1000);
            // remove grid points outside the region shape
            grid = turf.featurecollection(grid.features.filter(function(f) {
                return turf.inside(f, polypiemonte.poly_piemonte);
            }));
```

Was wondering if it would be possible to do this in one pass avoiding creating a second collection to save some memory / cpu time.

stebogit:
Implemented with PR #791

xrmx:
Thanks!

loaiabdalslam:
Hey , is That Avaliable RIght now for HexGrid ?

stebogit:
I don't think so

PR #791, opened 2017-06-15
Adds mask option to `@turf/point-grid`

stebogit:
Enhances `@turf/point-grid` so that it can interpret `bbox` as mask (if `(Multi)Polygon`)  to filter the generated points.
Ref: #220 

**Example:**
```js
var polypiemonte = // piedemont polygon;
var grid = turf.pointGrid(polypiemonte, 15, 'miles', true, true);
```
screen shot 2017-06-15 at 12 58 52 am


DenisCarriere:
👍 Looks great

PR #796, opened 2017-06-17
@turf/booleanCrosses

rowanwins:

Gday @DenisCarriere

Well I think this is hopefully almost done.

- There are 20+ tests (although realistically I think we could still do more!)
- I've run the tests against shapely and confirmed the same results
The benchmark results are looking not too bad although I think our line intersections algorithm could be improved
- I suggest we make the name `booleanCrosses`, as 'crosses' (rather than cross) is the name of the predicate everywhere else

````
 * Benchmark Results
 * LineDoesNotCrossButTouches x 71,945 ops/sec ±2.04% (71 runs sampled)
 * LineDoesNotCrossLine x 88,084 ops/sec ±2.56% (70 runs sampled)
 * LineDoesNotCrossPolygon x 86,024 ops/sec ±2.80% (71 runs sampled)
 * MultiPointNotCrossLine x 11,976,750 ops/sec ±1.61% (93 runs sampled)
 * MultiPointNotCrossLineEnd x 10,191,949 ops/sec ±3.51% (85 runs sampled)
 * LineCrossesLine x 68,764 ops/sec ±2.53% (72 runs sampled)
 * LineCrossesPolygon x 49,268 ops/sec ±2.70% (80 runs sampled)
 * LineCrossesPolygonPartial x 63,313 ops/sec ±2.87% (71 runs sampled)
````

### CODE REVIEW

File: packages/turf-boolean-crosses/index.js [hunk 1]
```diff
@@ -0,0 +1,153 @@
...
+    case 'LineString':
+        switch (type2) {
+        case 'LineString':
+            return doLineStringsCross(geom1, geom2);
+        case 'Polygon':
+            return doLineStringAndPolygonCross(geom1, geom2);
+        }
+        throw new Error('feature2 ' + type2 + ' geometry not supported');
+    default:
+        throw new Error('feature1 ' + type1 + ' geometry not supported');
```
DenisCarriere: `type1` cannot be Point/Polygon/MultiLineString? or you are still working on this?

DenisCarriere:
👍 that's great that we've got Shapely validation, do you have that code on GitHub anywhere?

packages/turf-boolean-crosses/index.js
```diff
@@ -0,0 +1,153 @@
...
+    var foundExtPoint = false;
+    var pointLength = multiPoint.coordinates.length;
+    var i = 0;
+    while (i < pointLength && !foundIntPoint && !foundExtPoint) {
+        for (var i2 = 0; i2 < lineString.coordinates.length - 1; i2++) {
+            var incEndVertices = true;
+            if (i2 === 0 || i2 === lineString.coordinates.length - 2) {
+                incEndVertices = false;
+            }
+            if (isPointOnLineSegment(lineString.coordinates[i2], lineString.coordinates[i2 + 1], multiPoint.coordinates[i], incEndVertices)) {
```
dpmcmlxxvi: Is this excluding both end points of both end segments? Shouldn't it just be the actual `LineString` end points?

packages/turf-boolean-crosses/index.js
```diff
@@ -0,0 +1,153 @@
...
+function doLineStringsCross(lineString1, lineString2) {
+    var doLinesIntersect = lineIntersect(lineString1, lineString2);
+    if (doLinesIntersect.features.length > 0) {
+        for (var i = 0; i < lineString1.coordinates.length - 1; i++) {
+            for (var i2 = 0; i2 < lineString2.coordinates.length - 1; i2++) {
+                var incEndVertices = true;
+                if (i2 === 0 || i2 === lineString2.coordinates.length - 2) {
+                    incEndVertices = false;
+                }
+                if (isPointOnLineSegment(lineString1.coordinates[i], lineString1.coordinates[i + 1], lineString2.coordinates[i2], incEndVertices)) {
```
dpmcmlxxvi: Why are you checking that the line segment 1 end points fall on line segment 2? `cross` only requires that they share some but not all interior points and that the intersections be points. They don't have to be the vertex points.

packages/turf-boolean-crosses/index.js
rowanwins: Crosses only supports a very limited set of types, see [this link](http://desktop.arcgis.com/en/arcmap/10.3/manage-data/using-sql-with-gdbs/relational-functions-for-st-geometry.htm#GUID-4686F4AF-C498-43D4-8F4A-5FBEA3352E61).  So as far as I'm concerned I think they're all covered off

packages/turf-boolean-crosses/index.js
rowanwins: ooo yeah that is a good pickup, a slight modification is needed on the `isPointOnLineSegment` function

packages/turf-boolean-crosses/index.js
rowanwins: again you are correct @dpmcmlxxvi this was an oversight late last night. I need to steal and adapt the `lineIntersects` algorithm

packages/turf-boolean-crosses/index.js
rowanwins: actually now I remember why I did this, I worked out that this didn't matter as the actual 2nd vertex get's picked up in the second segment.

rowanwins:
Gday @dpmcmlxxvi 

Thanks for the review. Perhaps the best thing to do is if you've got a concern is perhaps add some more test cases and we'll see how they come out.

dpmcmlxxvi:
@rowanwins Will do. It maybe a little bit before I get some times to work on it but will work on putting some tests together.

DenisCarriere:
@rowanwins @dpmcmlxxvi As for the types, shouldn't the opposite input types also be valid?

- [x] LineString => Polygon
- [ ] Polygon => LineString **(Currently throws an error)**
- [x] MultiPoint => Polygon
- [ ] Polygon => MultiPoint **(Currently throws an error)**
- [ ] Other combinations

> ST_Crosses returns 1 or t (TRUE) if the intersection results in a geometry that has a dimension that is one less than the maximum dimension of the two source geometries and the intersection set is interior to both source geometries. ST_Crosses returns 1 or t (TRUE) for only ST_MultiPoint/ST_Polygon, ST_MultiPoint/ST_LineString, ST_Linestring/ST_LineString, ST_LineString/ST_Polygon, and ST_LineString/ST_MultiPolygon comparisons.

**Question**: Shouldn't we return `False` instead of throwing an Error?

dpmcmlxxvi:
Agreed that the opposite input types are valid and should be allowed. Should be easy enough to implement. Use same functions just flip the `geom#` arguments.

On the return value. I can go either way. I do think there is some value to the way @rowanwins has it now. It might be helpful to make a distinction between having valid vs. invalid inputs. Returning `False` for DE-9IM functions that are not defined for certain inputs types might be misleading.

DenisCarriere:
I think if we include the reverse inputs we will have very little "invalid" input types.

Ex: Point => Point - should be invalid since two points cannot cross each other.

DenisCarriere:
@rowanwins FYI: Going to include your [shapely script](https://gist.github.com/rowanwins/e23208be1c5b27c1af2abdffab546444) into [`/scripts`](https://github.com/Turfjs/turf/tree/master/scripts)

rowanwins:
Gday @DenisCarriere 

Good thinking, it looks like the inverse can be covered in `crosses` ( I guess crosses is a bit different from of the other booleans in that regard), I'll include those in.

DenisCarriere:
@rowanwins Wow, this actually worked, I was able to include the shapely tests directly in the Tape tests.

```javascript
glob.sync(path.join(__dirname, 'test', 'true', '*.geojson')).forEach(filepath => {
    const {name} = path.parse(filepath);
    const geojson = load.sync(filepath);
    const [feature1, feature2] = geojson.features;
    if (process.env.SHAPELY) shapely('crosses', feature1, feature2).then(result => t.true(result, '[true] shapely - ' + name));
    t.true(crosses(feature1, feature2), '[true] ' + name);
});
```

Just make sure you define your environment variable as `SHAPELY=true` and you have Python + `shapely` installed.

DenisCarriere:
Merging, all the tests pass (including Shapely results).

We can always improve this library by adding more test fixtures, so far it looks good to me.

PR #803, opened 2017-06-19
Add boolean-shapely as external module

DenisCarriere:
## [`boolean-shapely`](https://github.com/DenisCarriere/boolean-shapely) as external testing module
Makes it easier to test/validate all DE-9IM boolean operations using [Shapely](https://github.com/Toblerity/Shapely).

```javascript
const shapely = require('boolean-shapely');

const line1 = lineString([[-2, 2], [1, 1]]);
const line2 = lineString([[1, 1], [1, 2], [1, 3], [1, 4]]);

shapely.crosses(line1, line2);
    .then(result => console.log(result));
//= true/false
```

FYI: @rowanwins


rowanwins:
Very neat indeed @DenisCarriere !

DenisCarriere:
@rowanwins Let me know if it works for you, so far it works for both Python 2 & 3 on my Mac and works in TravisCI.

PR #807, opened 2017-06-20
Add Multi Geometry support to `@turf/boolean-crosses`

DenisCarriere:
Support all geometries to `@turf/boolean-crosses`:

- [ ] Point
- [x] LineString
- [x] Polygon
- [ ] MultiLineString
- [ ] MultiPolygon
- [x] MultiPoint

Shapely does support MultiLineString & all other geometries.

## Failing Tests (new)
- [ ] `doLineStringsCross` not handling [Polygon => LineString](https://github.com/Turfjs/turf/blob/a21401534a49f61cbf6337471f74c1557ba05e1a/packages/turf-boolean-crosses/test/true/Polygon/LineString/Polygon-LineString-1.geojson)

Image below returns `False`
image


rowanwins:
Gday @DenisCarriere 

Yeah i was thinking at midnight last night that I needed to start including multi-geometries (not to mention polys with holes!). I think the flatten approach is good, looks like you've got it well structured.

DenisCarriere:
Multi-Geometries is definitely needed, I'm not too sure how to handle Polygon with holes since I think you only need to cross the outer ring and I'm pretty that's all you need to test for.

It's nice that we have the Shapely tests to reference to

DenisCarriere:
Another day for this PR...

PR #810, opened 2017-06-21
Renamed `cellSize` param in grid modules

### LINKED ISSUE #623: distance between two hexagons in hexgrid
I am trying to find distance between two hexagons in a hex grid and as far as my understand goes, it should be equal to the cell size we pass while creating the hexgrid.
`hexGrid = turf.hexGrid(bbox, 1,"kilometers")`

but when I try to find the distance between centroid of two cells using `distance` api, the value I get is less than the cell size. I am using the below code.
`turf.centroid(tile)`
`turf.distance(from, to,"kilometers")`

The cell size is 1kilometers and what get back as distance is 0.866 kilometers.

Please help!

stebogit:
- renamed `cellSize` parameter to `cellSide` in `@turf/point-grid`;
- renamed `cellSize` parameter to `cellDiameter` in `@turf/hex-grid`
- added distance between  hexagon centroids
- modified docs

Fixes #623

### DIFF
--- a/packages/turf-hex-grid/README.md
+++ b/packages/turf-hex-grid/README.md
@@ -2,25 +2,25 @@
 
 # hexGrid
 
-Takes a bounding box and a cell size in degrees and returns a [FeatureCollection](http://geojson.org/geojson-spec.html#feature-collection-objects) of flat-topped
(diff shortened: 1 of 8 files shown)


DenisCarriere:
👍

Issue #811, opened 2017-06-21
Proposal new module distance-clustering


stebogit:
Continuing the discussion started in #787, ref. also #33 

@cyrilchapon as @DenisCarriere said we should not merge the two clustering strategies, but rather create a new module.
My question about the use of the distance clustering was not on the API, but more about an example of application, to try to understand better how to design input and output.
Reading the [article](https://www.mapbox.com/blog/supercluster/) provided by @DenisCarriere, [`kdbush`](https://github.com/mourner/kdbush) seems to be a great tool to index the points, but then you need to query the generated tree to perform the clustering.
What I don't quite get yet is how you would exactly use `maxDistance` to group the points...
Or maybe for each point you want a `FeatureCollection` of points that are within that distance?

Once we defined that we might want to define two clustering modules:
- [x]  `@turf/clusters-kmeans` (currently `@turf/clusters`)
- [x]  ~~`@turf/clusters-distance`~~ `@turf/clusters-dbscan`

cyrilchapon:
@stebogit, my original use-case was [this one](https://stackoverflow.com/questions/40806348/aggregate-by-geonear). It was db-oriented question but the main purpose is now I'm using to preprocess some stuff for insert/updates on mongo. In that particular case, it's a common use-case tu aggregate some location points from IOT networks for example.
maxDistance (actualité this is lire like minDistance) factor would be the "maximal distance between 2 points before jumping to a new cluster". Or maybe the "minimal distance between 2 points to be considered as a cluster". But it shouldn't "jump" from point to point, it would be more like the mac radius of a circle containing all points of a cluster.

DenisCarriere:
Another library that might be useful to create the distance cluster is [`geokdbush`](https://github.com/mourner/geokdbush) which will allow you to find the closest points using [`geokdbush.around`](https://github.com/mourner/geokdbush#geokdbusharoundindex-longitude-latitude-maxresults-maxdistance-filterfn).

DenisCarriere:
Lets develop both for now, and lets see at the end if we should combine them together or not.

@stebogit I'll be working on `clusters-distance` today, I've got an idea how I can tackle it.

stebogit:
Implemented with #812

[...]

PR #812, opened 2017-06-21
Implement `@turf/clusters-distance` module

DenisCarriere:
## First draft of `@turf/clusters-distance` Ref. #811

> **Note:** this module might be merged into `clusters` and support both `distance` & `numberOfClusters` params (still up in the air).

Compared to the `kmeans` cluster this is roughly 5x-50x times faster (the `kdbush` index does help a lot -- The performance will most likely slow down as weighted clusters becomes implemented).

## To-Do

- [x] Translate existing GeoJSON properties
- [x] Weighted clusters based on distance (right now the first point from each new cluster is the "center")
- [x] Not too sure if we should be adding **centroid/points** in the outputs (instead of simply a `FeatureCollection<Point>`). Using the `cluster` property param is already enough to be able to create a centroid/center/centerOfMass.
- [x] Only supports **kilometers** as distance

## More To-Dos
- [x] change library name to `clusters-dbscan`
- [x] add a dbscan property
- [x] drop the centroid
- [x] Prevent input mutation

## Examples

**points1.geojson**

image

**points2.geojson**

image


stebogit:
@DenisCarriere how can I help?

DenisCarriere:
👍 Agreed about including the `centroid`.

As for help, when the clusters ID's are being associated, it [only matches the first found](https://github.com/Turfjs/turf/pull/812/files#diff-1320fee909185d17f5301bbfd39410d7R44) and then skips all the other ones. What would need to happen is to sort/group the clusters by closest distance (not the first match).

stebogit:
@DenisCarriere I don't know if this is exactly the issue you were referring to, but I feel there is something not right with the approach we are using for this clustering.
Take a look at the [gif image](https://cdn-images-1.medium.com/max/800/0*wBezECSY50VQZYOU.gif) that shows the greedy clustering algorithm (which we are basically using here) of [superclaster](https://blog.mapbox.com/clustering-millions-of-points-on-a-map-with-supercluster-272046ec5c97); I added a few points to show the issue I'm referring to:
screen shot 2017-07-05 at 7 58 26 am
As you can see all points are grouped (which is the goal of that clustering method), however the green points 2 and 3 technically belong to two different `maxDistance` circles/cluster, but they are associated with a (`maxDistance`) cluster that is first calculated.

The problem I think is that for the `@turf/clusters-distance` circles (i.e. clusters) should not overlap, but it's impossible to completely cover an area using circles...

DenisCarriere:
Yes this is the issue I was referencing, I actually used [`supercluster`](https://github.com/mapbox/supercluster) as inspiration for this module (removed all the WebGL & vector tile grouping).

What if we define the `cluster` param as `clusters: Array<number>` instead?

That way the points 2 & 3 would have this as a param `clusters: [2, 3]`

This approach might help solve the ambiguous clusters.

Points that only have 1 cluster would still be an Array except with only 1 number.

@stebogit thoughts?

DenisCarriere:
By the way... Thanks @mourner for another great library 👍  ([supercluster](https://github.com/mapbox/supercluster/graphs/contributors))

stebogit:
@DenisCarriere I think we should better define the expected result of the module.
It seems to me that we are kind of mixing the (current) `@turf/cluster` functionality/output (a [partitional clustering](https://en.wikipedia.org/wiki/Cluster_analysis#Centroid-based_clustering)) with the `supercluster` one (a [hierarchical clustering](https://en.wikipedia.org/wiki/Hierarchical_clustering)).

What if we assigned the `clusters` (array) parameter to each point, where the array contains all the index/reference to al the points that are within the `maxDistance`?
There wouldn't be an actual `centroid` for each cluster, as actually there wouldn't be any cluster (so this method might not be called `clusters-`)...

I have no idea if this would be of any utility though.

DenisCarriere:
Well how would one go about to create clusters based on a distance?

Right now we can create clusters based on `kmeans`, but if you wanted to cluster points based on a `maxDistance`, then this would be the module for it.

No clue what the perfect outcome should be, however I'm sure we can come up with a creative way to make such module.

packages/turf-clusters-distance/index.js
```diff
@@ -25,17 +25,12 @@ module.exports = function (points, maxDistance) {
     collectionOf(points, 'Point', 'Input must contain Points');
 
     // Create index
-    const load = points.features.map(function (point, index) {
```
DenisCarriere: 👍 Love it! I think my first approached used the `id` fields to do the matching, however I might of dropped that workflow mid way and forgot to remove the `.map()`.

@stebogit Good catch!

packages/turf-clusters-distance/bench.js
```diff
@@ -19,16 +19,16 @@ const fixtures = fs.readdirSync(directory).map(filename => {
...
- * many-points x 8,193 ops/sec ±4.55% (81 runs sampled)
- * points-with-properties x 663,783 ops/sec ±1.61% (83 runs sampled)
- * points1 x 250,386 ops/sec ±1.53% (88 runs sampled)
- * points2 x 166,679 ops/sec ±1.40% (88 runs sampled)
+ * fiji: 1.692ms
+ * many-points: 14.448ms
+ * points-with-properties: 0.164ms
+ * points1: 0.087ms
+ * points2: 0.694ms
+ * fiji x 1,320,371 ops/sec ±1.72% (80 runs sampled)
```
DenisCarriere: 🚗 💨 Zoom Zoom!

stebogit:
> Well how would one go about to create clusters based on a distance?
Right now we can create clusters based on `kmeans`, but if you wanted to cluster points based on a `maxDistance`, then this would be the module for it.

@DenisCarriere I've been thinking a lot about this clustering problem and I keep wondering: `maxDistance` from what? There's no reference from which to calculate the distance.

Should we add a (required) `centroids` input and assign a centroid/cluster to each point within `maxDistance` from that centroid?
This would not necessary assign a cluster to all the points in the set ([see _partial_ clustering](https://www-users.cs.umn.edu/~kumar/dmbook/ch8.pdf)).
However, would this be a useful/common clustering method, that 'popular' to deserve a dedicated module?

In my research on clustering methods I keep seeing these three as most popular (i.e. I guess mostly used/useful) methods:
- **K-means** (each point of the cluster is closest to its centroid that any other centroid, thus this is a clustering method based on distance)
- **Agglomerative Hierarchical Clustering** (similar to [`supercluster`](https://github.com/mapbox/supercluster/graphs/contributors) and the current implementation of this module)
- [**DBSCAN**](https://en.wikipedia.org/wiki/DBSCAN) (density based, clusters are defined as regions of higher density, i.e. number of points per unit area, than the remainder of the data set)

DenisCarriere:
Thanks @stebogit for that initial research.

For me, my initial intent with this distance cluster would most likely reflect [`DBSCAN`'s clustering algorithm](https://en.wikipedia.org/wiki/DBSCAN). This approach makes the most sense to me and would be "relatively" easy to implement using [`geokdbush`](https://github.com/mourner/geokdbush), I think I've already got an idea how to do it.


Using this approach would probably solve your question about (`maxDistance` from what?)
image

> In this diagram, minPts = 4. Point A and the other red points are core points, because the area surrounding these points in an ε radius contain at least 4 points (including the point itself). Because they are all reachable from one another, they form a single cluster. Points B and C are not core points, but are reachable from A (via other core points) and thus belong to the cluster as well. Point N is a noise point that is neither a core point nor directly-reachable.

@stebogit Are you on the same page if we use the DBSCAN approach for `@turf/clusters-distance`?

CC: @morganherlocker & @mourner Feel free to shim in anytime if you feel this is going in the wrong direction.

DenisCarriere:
@stebogit I couldn't help myself to get this done in one night.

Check out the new implementation, it's pretty solid and it uses some of the DBSCAN logic (minus not providing noise points - they are simply excluded).

## Many Points
> Clusters will keep growing as long as clusters are within maximum distance

image

## Points 2
> Outliers can be identified here by showing clusters of a single point (can be removed if `minPoints` param is defined greater than 1).

image

stebogit:
👍 👏  @DenisCarriere I'll take a look at this tomorrow.
Quick thought, we could call this module `@turf/clusters-density`, to differentiate it from `kmeans` which does cluster points based on *distance* from the `centroid`.
Or we might just stick with the algorithm name, so `@turf/clusters-dbscan`, which is probably more clear for everybody.

DenisCarriere:
👍 For `@turf/clusters-density` name, I don't think the average user would know what `dbscan` means, keeping the module names simple is best.

👍 renaming `@turf/clusters` to `@turf/clusters-kmeans`, that way the cluster process is better defined/scoped by using the name of the module.

DenisCarriere:
The reason why I wouldn't call it `@turf/clusters-dbscan` is because this is more of an "inspiration" of **dbscan** and not a complete implementation of it.

We can always improve on this at a later date...

stebogit:
@DenisCarriere I would still output the `points` plus the `centroids`, like in `@turf/clusters-kmeans`), cause I still believe it's faster and convenient to do it while handling all the points.

BTW out of curiosity is there any particular reason why you preferred to implement this module from scratch?
I found [this](https://github.com/uhho/density-clustering) implementation of DBSCAN which, if you don't mind, I'd like to test and see if it works (we can even pass our own distance function!); that would keep a lower complexity in the Turf module and base it on a standard and well known algorithm (like `kmeans`).
We might then compare outputs and performances.

DenisCarriere:
If we go to the `points/centroids` approach we should also include the `noise` points.

Reason built this module from scratch, mostly wanted to try out `Geokdbush`, it would be good to have a comparison, might even be easier or faster (hopefully) using the DBSCAN library.

There's a few things in this library that isn't optimized.

Definitely try out that DBSCAN implementation, if it's easier doing it using that, then that's even better!

DenisCarriere:
One thing to consider is `Geokdbush` has implemented taking Earth curvature and date line wrapping into account.

Whereas the DBSCAN is using a basic euclidean distance calculation, we would have to test both...

stebogit:
@DenisCarriere looking at the `many-points` output I realized the centroid here are really not useful, because of the irregular shape of the clusters, compared to the `k-means` centroids where they actually represents the "thing" that aggregate the points. I think we should actually drop it (😅 ...)

>Whereas the DBSCAN is using a basic euclidean distance calculation, we would have to test both...

I actually used `@turf/distance` as distance function, since [`density-clustering`](https://github.com/uhho/density-clustering) allows you to pass your function. That should take Earth curvature into account, right?
I tryed also with `@turf/rhumb-distance` and the output is exactly the same (all test pass), I guess because we are clustering points enough close to each other.

DenisCarriere:
@stebogit You win the 👑 for refactoring my code => `index.js` -200 removed lines and only +50 added, good job!

DenisCarriere:
+1 your last comments. Keeping the centroids would keep it consistent between clusters modules, it is pretty neat to have those centroids created for the user

DenisCarriere:
Oh that's nice that you can add your own distance calculations, didn't notice it, but would of soon realized it from the easy added units support

stebogit:
@DenisCarriere Actually there seems to be an issue with the `many-points` test: the output contains 505 total points, the input 500... 🤔 

These three points for example are actually a cluster (thus the `centroid`), but two of them are also in the noise array, so they are also marked as noise:

screen shot 2017-07-14 at 12 51 21 am

All the other tests though are ok.
I'm gonna take a look at the dbscan library, also because I bet it could be faster if implemented with `geokdbush`

DenisCarriere:
As for your last observation, the points are "properly" being tagged as cluster, except they are in both FeatureCollections (noise + points).

image

In my implementation, those noise points get dropped in the `points`.

DenisCarriere:
> the output contains 505 total points (`points` + `noise`), the input 500... 🤔 

Looks fine to me, might be because you are calculating the points differently, here's how I did it:

**test.js**
```js
console.log(geojson.features.length);
const clustered = clustersDistance(geojson, distance, units, minPoints);
console.log(clustered.points.features.length);
```
**output**
```
# clusters-distance
6
6
ok 1 fiji
500
500
ok 2 many-points
25
25
ok 3 noise
```

DenisCarriere:
Instead of pushing all points to the output, we can only push the ones with associated cluster properties.

Just added this:

```js
var newPoints = [];
var centroids = [];
var noise = [];
//...
        clusterIds.forEach(function (idx) {
            //...
            cluster.push(clusterPoint);
            newPoints.push(clusterPoint);
        });
//...
return {
    points: featureCollection(newPoints),
    centroids: featureCollection(centroids),
    noise: featureCollection(noise)
};
```

Count reflects the drop in points now.
```
# clusters-distance
6
6
ok 1 fiji
500
481
ok 2 many-points
```

DenisCarriere:
@stebogit Woot, got it to work, it sounds like it's a bug in the DBSCAN library, it associated noise points in both Noise & Cluster.. doesn't seem right to me.

Regardless, was pretty easy to fix:
```js
    // handle noise points, if any
    dbscan.noise.forEach(function (noiseId) {
        var noisePoint = points.features[noiseId];
        // Skip Noise if cluster is already associated
        // This might be a slight deviation of DBSCAN (or a bug in the library)
        if (noisePoint.properties.cluster) return;
        noise.push(noisePoint);
    });
```

image

DenisCarriere:
@stebogit Looks like those **Noise** + **ClusterID** points are the "edges" within a cluster.

Using the image below:
- Noise => N
- Edges => B & C (these points would be in both `points` & `edges`
- Core Points => A (all `points` which form a cluster)

image

DenisCarriere:
Actually, just thinking about this, instead of making 3 different FeatureCollections, couldn't we just simply add a `dbscan` property to each point which we can identify them by :

- `noise`
- `core`
- `edge`

stebogit:
> add a dbscan property

👍 

> drop the centroid

👍 

> Prevent input mutation

👍

DenisCarriere:
haha! 👍 

Also, I'm sure you agree with this as well, we should change the of this to `clusters-dbscan` (density is an easier terminology, however we are using DBSCAN for the library and outputting DBSCAN specific properties, might be best to use the same lingo for the module).

stebogit:
> change the name of this to `clusters-dbscan`

sure 👍 😄 
Shall we rename also `@turf/clusters-kmeans` now or wait until next major release?

stebogit:
@DenisCarriere why don't we return this instead:
```js
return {
    points: FeatureCollection<Points>, // with `clusterId` property
    edges: Array<Array<number>>, // edges ids
    centroids: Array<Array<number>>, // centroids ids
    noise: Array<Array<number>> // noise ids
};
```
This would speed up the calculation (no feature creation) and slim the output, which seems now a little bloated to me, while still allowing iteration through each cluster or point type.

packages/turf-clusters-distance/index.js
```diff
@@ -73,12 +78,13 @@ module.exports = function (points, maxDistance, units, minPoints) {
...
         // Skip Noise if cluster is already associated
         // This might be a slight deviation of DBSCAN (or a bug in the library)
-        if (noisePoint.properties.cluster) return;
+        if (noisePoint.properties.cluster) return edges.push(noisePoint);
         noise.push(noisePoint);
     });
 
     return {
         points: featureCollection(newPoints),
+        edeges: featureCollection(edges),
```
stebogit: typo:
```
- edeges: featureCollection(edges),
+ edges: featureCollection(edges),
```

DenisCarriere:
> Shall we rename also `@turf/clusters-kmeans` now or wait until next major release?

We just recently published that module... We change it to `kmeans` at the next major release, or include both (for now) and `clusters` will simply import the new module name `clusters-kmeans` and we can add a `console.warning()` message when using it.

DenisCarriere:
> This would speed up the calculation (no feature creation) and slim the output, which seems now a little bloated to me, while still allowing iteration through each cluster or point type.

This is already deviating A LOT from the main purpose of TurfJS, all of the inputs/outputs should be in GeoJSON.

We shouldn't even be pushing out an Array of FeatureCollection (if we don't have too).

DenisCarriere:
I like simply including the `dbscan` property, it's easy to understand the output.

Quick update on the Typescript definition

```typescript
interface Point extends GeoJSON.Feature<GeoJSON.Point> {
    properties: {
        dbscan?: 'core' | 'edge' | 'noise';
        [key: string]: any;
    }
}

interface Output {
    type: 'FeatureCollection'
    features: Point[];
}
```

DenisCarriere:
@stebogit Next commit should include a lot of changes, have a review:

image

DenisCarriere:
> @stebogit This would speed up the calculation (no feature creation) and slim the output, which seems now a little bloated to me, while still allowing iteration through each cluster or point type.

These types of modules are probably better to be abstracted out of TurfJS (only dealing with 2D points) and afterwards TurfJS would wrap it to simplify the GeoJSON integration.

@mourner's library are mainly all 2D (`Array<number>`) which can easily be used in all sorts of different environment, whereas TurfJS is mostly focused on having pure GeoJSON outputs.

packages/turf-clusters-distance/test.js
```diff
@@ -63,35 +63,41 @@ test('clusters -- prevent input mutation', t => {
...
-        centroid.properties['marker-symbol'] = 'star';
-        centroid.properties['marker-size'] = 'large';
-        points.push(centroid);
-    });
-    featureEach(clustered.noise, function (point) {
-        point.properties['marker-color'] = '#AEAEAE';
-        point.properties['marker-symbol'] = 'circle-stroked';
-        point.properties['marker-size'] = 'medium';
-        points.push(point);
+    featureEach(clustered, function (point) {
```
stebogit: I'd change this to:
```js
switch (point.properties.dbscan) {
    case 'core':
    case 'edge': {
        const coreColor = colours[point.properties.cluster];
        const edgeColor = chromatism.brightness(-20, colours[point.properties.cluster]).hex;
        point.properties['marker-color'] = (point.properties.dbscan === 'core') ? coreColor : edgeColor;
        point.properties['marker-size'] = 'small';
        points.push(point);
        break;
    }
    case 'noise': {
        point.properties['marker-color'] = '#AEAEAE';
        point.properties['marker-symbol'] = 'circle-stroked';
        point.properties['marker-size'] = 'medium';
        points.push(point);
    }
}
```
as `edges` are not really a major feature to highlight.
But it's just a finesse 😄

packages/turf-clusters-distance/index.js
```diff
@@ -17,7 +15,7 @@ var featureCollection = helpers.featureCollection;
  * @param {FeatureCollection<Point>} points to be clustered
  * @param {number} maxDistance Maximum Distance between any point of the cluster to generate the clusters (kilometers only)
  * @param {string} [units=kilometers] in which `maxDistance` is expressed, can be degrees, radians, miles, or kilometers
- * @param {number} [minPoints=1] Minimum number of points to generate a single cluster, points will be excluded if the
+ * @param {number} [minPoints=3] Minimum number of points to generate a single cluster, points will be excluded if the
```
stebogit: 👍

packages/turf-clusters-distance/index.js
```diff
@@ -17,7 +15,7 @@ var featureCollection = helpers.featureCollection;
...
  *     cluster does not meet the minimum amounts of points.
  * @returns {Object} an object containing a `points` FeatureCollection, the input points where each Point
  *     has given a `cluster` property with the cluster number it belongs, a `centroids` FeatureCollection of
```
stebogit: Should be updated with the new `dbscan` property

DenisCarriere: 👍 Yep! It's defined in the Typescript definition, but yes it needs to be documented

packages/turf-clusters-distance/index.js
```diff
@@ -45,47 +43,33 @@ module.exports = function (points, maxDistance, units, minPoints) {
...
+            clusterPoint.properties.cluster = clusterId;
+            clusterPoint.properties.dbscan = 'core';
         });
-        var centroid = centerOfMass(featureCollection(cluster), {cluster: clusterId});
-        centroids.push(centroid);
     });
+
     // handle noise points, if any
+    // Skip Noise if cluster is already associated
+    // This might be a slight deviation of DBSCAN (or a bug in the library)
```
DenisCarriere: 👍

packages/turf-clusters-distance/test.js
DenisCarriere: lol Yep! Looks good.

I really like this Switch statement, makes it really easy to control those clustered points.

packages/turf-clusters-distance/test.js
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * @private
+ * @param {FeatureCollection|Feature[]} geojson GeoJSON
+ * @param {string} property Name of property of GeoJSON Properties to bin each feature - This property will be translated to each centroid's properties
+ * @param {*} properties Properties to translate down to each centroid (2nd priority)
+ * @returns {FeatureCollection<Point>} centroids
+ * @example
+ * var centroids = centroidFromProperty(points, 'cluster');
+ */
+function centroidFromProperty(geojson, property, properties) {
```
DenisCarriere: @stebogit Added back your centroids points 😄 

image

stebogit: Mmmh... in density clustering the centroids are less useful/identifying/important than in k-means, I guess.  🤔 
But I might be wrong.

stebogit: Oh, I see now, this is only in `test.js`. Good 👍

stebogit:
Cool! 😃  🎊 🚀

DenisCarriere:
29 commits later.. 👍

packages/turf-clusters-distance/test.js
DenisCarriere: Yes! :) only for `test.js` This where this module would be used... we wouldn't apply this directly in the modules, but for display purposes.

Also this can be applied against Polygons or any Geometry Types, finding the "centroid" of stuff.

PR #813, opened 2017-06-22
Fixes for docs

rowanwins:
Various little fixes for the docs as I'm about to regenerate them all :)

Mostly just fixing camel-casing and including `transform` in the name


DenisCarriere:
lineStringToPolygon uses a capital "S".

https://github.com/Turfjs/turf/blob/20d9d3b65a049f253c8981ea7be560dd9a4d7dbc/packages/turf/index.js#L74

Other then that 👍

rowanwins:
Missed it by this much `''`

DenisCarriere:
👍

PR #827, opened 2017-07-03
Modified `@turf/clone` docs

stebogit:
@DenisCarriere I believe this is a more clear description for the module


DenisCarriere:
👍 Awesome, thanks @stebogit! Love the new description, a lot better 👍

Issue #829, opened 2017-07-06
Isoline from example returning MultiLineString with no coordinates


bakerac4:
Following the example found at http://turfjs.org/docs/#isolines after downloading the minified turf.js file and including it in my project ends up giving me 10 features that each have 1 geometry object inside, with type MultiLineString that has an empty array of coordinates. 

Here is a GIST of the GeoJSON I am getting back
https://gist.github.com/bakerac4/89656a65e7388b290d223facdaf5dc75

The code I am using seems to be a copy and paste from the docs.
```javascript
var points = turf.random('point', 100, {
  bbox: [0, 30, 20, 50]
});
for (var i = 0; i < points.features.length; i++) {
  points.features[i].properties.temperature = Math.random() * 10;
}
var breaks = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
var isolines = turf.isolines(points, breaks, 'temperature');
```

DenisCarriere:
Huh... interesting.

Also getting the same result:

```javascript
//= isoloines.features[0]
{ type: 'Feature',
  properties: { temperature: 1 },
  geometry: { type: 'MultiLineString', coordinates: [] } }
```

Thoughts? @stebogit

stebogit:
@bakerac4 `turf-isolines` (like [`turf-isobands`](http://turfjs.org/docs/#isobands)) expects a [`pointGrid`](http://turfjs.org/docs/#pointgrid) as input, which will be internally converter into a matrix (`Array<Array<number>>`) of `zProperty` values to be able to calculate the isolines. With just random points there is no matrix (that's why a [`@turf/interpolate`](https://github.com/Turfjs/turf/issues/24) module has been desirable for long time).
If you have though a matrix and you want to convert it into a `pointGrid` you could use [matrix-to-grid](https://github.com/stebogit/matrix-to-grid#readme).

I'm surprised (and sorry) the documentation have not been correctly updated when the module was re-implemented, I'll commit the appropriate correction.

stebogit:
Done. I did also fix a small bug in `isobands` and made sure also its example was working.
@DenisCarriere how can I update also the website?

bakerac4:
Thank you @stebogit for clearing that up. That leaves me a bit confused about something (and Im sure its just me misunderstanding something). Does `pointGrid` create a grid of points that are always evenly spaced out?  

[...]

Issue #24, opened 2013-10-27
New module @turf/interpolate


morganherlocker:
Takes a set of points and estimates values on a grid.

stebogit:
@DenisCarriere this seems a lot similar to the "validation" step we tried to add inside `isobands` (random points to grid).
In order to reduce the complexity of that module, which should only do one thing, what if I implement this one and we keep the input of `isobands` (and `isolines`) as `pointGrid`, since this module will be available?

DenisCarriere:
👍 @stebogit Sounds like a good idea.

Many of the "complex" modules usually have 3-4 modules that can be discovered/created prior to making the main module (ex: `isobands` in this case).

stebogit:
Just to make sure we are on the same page about what this module should do.
My understanding is, based for example on `isobands` and `isolines` input requirement, to take any set of points with an `elevation`/`z` value and transpose them to a `pointGrid`, interpolating said "vertical" value on the grid.
Is that correct @morganherlocker and @DenisCarriere ?

DenisCarriere:
👍 Sounds right to me.

stebogit:
Implemented with #832

PR #772, opened 2017-05-29
Support z-coordinate as input for @turf/turf-tin

dpmcmlxxvi:
Addresses #628 for `@turf/tin`. PR has the following edits:
 - Uses z-coordinate of input points only if input z-property is undefined. So, should be backwards compatible.
 - Updates test fixtures to include z-coordinate.
 - Updated `test.js` and `bench.js` syntax to ES6.

---
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Have read [How To Contribute](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#how-to-contribute).
- [x] Run `npm test` at the sub modules where changes have occurred.
- [x] Run `npm run lint` to ensure code style at the turf module level.

PR #832, opened 2017-07-08
New module `@turf/interpolate` (proposal)

stebogit:
Implements #24

## JSDocs

```js
/**
 * Takes a set of points and estimates their 'property' values on a grid using the Inverse [Distance Weighting (IDW) method.](https://en.wikipedia.org/wiki/Inverse_distance_weighting).
 *
 * @name interpolate
 * @param {FeatureCollection<Point>} points with known value
 * @param {number} cellSize the distance across each grid point
 * @param {string} [gridType='square'] defines the output format based on a Grid Type (options: 'square' | 'point' | 'hex' | 'triangle')
 * @param {string} [property='elevation'] the property name in `points` from which z-values will be pulled
 * @param {string} [units=kilometers] used in calculating cellSize, can be degrees, radians, miles, or kilometers
 * @param {number} [weight=1] exponent regulating the distance-decay weighting
 * @returns {FeatureCollection<Point>} grid of points with 'property'
 */
module.exports = function (points, cellSize, property, units, weight) {
```

## Example
```js
var points = turf.random('points', 15, {
    bbox: [-75.9221, 39.11, -75.05, 39.987]
});
turf.featureEach(points, function(point) {
    point.properties.elevation = Math.random() * 50;
});
var cellSize = 5;
var grid = turf.interpolate(points, cellSize);
```

screen shot 2017-07-08 at 12 40 20 am

- [x] add `zValue` as 3rd coordinate if `property` is not defined.
- [x] add tests
- [x] add all other grid types to library.
- [ ] might replace/integrate [`@turf/idw`](http://turfjs.org/docs/#idw) ?

@DenisCarriere following #829 I thought maybe it would make sense to implement this module with the IDW method. In fact I don't really know any other interpolation methods, but I'm open to suggestions.
However, this differs from `@turf/idw` only in outputting a `pointGrid` instead of a `squareGrid`. Does it make sense to have two modules here?
If not I would rather merge the two in a single `turf-interpolate` (I think is more clear to an average user like me, who ignored the meaning of `IDW` since this morning...) and add to the module an `outputType` parameter (default to one of the two).

@bakerac4 maybe you want to try this module out and see if it fits your needs.


stebogit:
@DenisCarriere I don't know why the test pass in my computer, but not in Travis... 😞

DenisCarriere:
I'll comment more later on this weekend, however for the Travis build fail, I'm guessing it's a decimal precision error, even using Node 6 vs. 7 you could have slight differences. Solution would be to use the `@turf/truncate` module on the result test/out.

Quick glance, this looks like an awesome module!

packages/turf-interpolate/test.js
```diff
@@ -0,0 +1,58 @@
...
+        geojson: load.sync(directories.in + filename)
+    };
+});
+
+test('turf-interpolate', t => {
+    for (const {filename, name, geojson}  of fixtures) {
+        const {property, cellSize, units, weight} = geojson.properties;
+
+        const grid = interpolate(geojson, cellSize, property, units, weight);
+        const result = colorize(grid, property);
```
DenisCarriere: Added `@turf/truncate` to results (reduce coordinates to 6 decimals).

Node 7 & 8 were passing, but Node 6 was failing (0.0000001 of a difference).

packages/turf-interpolate/index.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="geojson" />
+
+import {Points} from '@turf/helpers';
```
DenisCarriere: Since `@turf/helpers` is not required in the main `package.json`, it might be better to include the Full GeoJSON typescript definition instead of importing it (in case the dependencies wouldn't include `@turf/helpers`).

```typescript
type Point = GeoJSON.FeatureCollection<GeoJSON.Point>;
```

packages/turf-interpolate/index.d.ts
```diff
@@ -0,0 +1,10 @@
...
+
+/**
+ * http://turfjs.org/docs/#interpolate
+ */
+declare function interpolate(points: Points, cellSize: number, property?: number, units?: string, weight?:number): Points;
```
DenisCarriere: `property?: number` should be a `string`

DenisCarriere: Contradicts my previous comment, but `units?: string` could use `Units` instead of `string`.

```javascript
import {Units} from '@turf/helpers'
```

packages/turf-interpolate/index.js
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * //addToMap
+ * var addToMap = grid
+ */
+module.exports = function (points, cellSize, property, units, weight) {
+    // validation
+    if (!points) throw new Error('points is required');
+    collectionOf(points, 'Point', 'input must contain Points');
+    if (!cellSize) throw new Error('cellSize is required');
+    if (cellSize === undefined || cellSize === null) throw new Error('cellWidth is required');
```
DenisCarriere: I'm going to assume this is a duplicate of `cellSize`.

`CellSize` cannot be 0, correct?

DenisCarriere:
**Question**: Should we add a `round/precision` property? That way the `elevation` properties can be rounded to a specific precision (starting to have a lot of options).

CC: @stebogit

packages/turf-interpolate/test/types.ts
```diff
@@ -0,0 +1,16 @@
+import * as linearc from '../index'
```
DenisCarriere: 🤦‍♂️ Looks like this file came from the previous repo 🤣 

I've been putting `types.ts` at the root directory that way we don't forget it.

One day... all modules will have `types.ts` to test for the Typescript definitions.

So far this one looks like this:

```typescript
import * as interpolate from './';
import {point, featureCollection} from '@turf/helpers';

const cellSize = 1;
const property = 'pressure';
const weight = 0.5;
const units = 'miles';
const points = featureCollection([
  point([1, 2]),
  point([12, 13]),
  point([23, 22]),
]);

const grid = interpolate(points, cellSize, property, units, weight);
grid.features[0].properties.pressure;
```

DenisCarriere:
@stebogit Should we add the 3rd coordinate as `zValue` if `property` is not defined? Like we did for (https://github.com/Turfjs/turf/issues/628, https://github.com/Turfjs/turf/pull/772).

DenisCarriere:
> might replace/integrate @turf/idw 

Agreed, however the only difference with `@turf/idw` is the output is a `FeatureCollection<Polygon>`, if we can manage to have `interpolate` do both or simply integrate this library into `idw` but keep the output as polygons.

CC: @stebogit

stebogit:
> Should we add a round/precision property? 

@DenisCarriere I don't see the number of decimals as an issue, the user can simply round it, if necessary, to the desired precision when using the interpolated value(s)

DenisCarriere:
> I don't see the number of decimals as an issue
👍 That works, yea it's not really an issue.

DenisCarriere:
@stebogit Have a quick glance at it before we merge, I'm done with any edits.

DenisCarriere:
> If not I would rather merge the two in a single turf-interpolate (I think is more clear to an average user like me, who ignored the meaning of IDW since this morning...) and add to the module an outputType parameter (default to one of the two).

👍 Agreed, maybe once we bump up to a major version we could do that.

I also prefer the term `interpolate` vs. `IDW` (even though this term is commonly used in GIS).

DenisCarriere:
> If not I would rather merge the two in a single turf-interpolate (I think is more clear to an average user like me, who ignored the meaning of IDW since this morning...) and add to the module an outputType parameter (default to one of the two).

👍 Agreed, maybe once we bump up to a major version we could do that.

I also prefer the term `interpolate` vs. `IDW` (even though this term is commonly used in GIS).

CC: @morganherlocker Thoughts?

stebogit:
> Agreed, however the only difference with `@turf/idw` is the output is a `FeatureCollection<Polygon>`, if we can manage to have `interpolate` do both or simply integrate this library into `idw` but keep the output as polygons.

@DenisCarriere yes, having this module doing both is really easy (I'll implement that).
> I also prefer the term interpolate vs. IDW (even though this term is commonly used in GIS).

👍  we could add an `outputType = 'polygon'|'point'` parameter, defaulting to `'polygon'`.
This module should also be faster than the current implementation of `idw`.

When I implemented this I was wondering if @morganherlocker had actually a different idea/algorithm of interpolation in mind when suggesting this module

stebogit:
BTW anybody knows if there is a more appropriate default value for `weight`?

packages/turf-interpolate/index.d.ts
stebogit: ah! nice to know, I didn't noticed the types in `@turf/helpers` 👍

stebogit:
@DenisCarriere I added the `outputType` parameter.
I thought to define it as required, so the user has to consciously choose the output; also if we wanted to extend it to `hexGrids` or `triangleGrid` we could easily.

DenisCarriere:
👍 I like the option of defining the different `outputType`, we could call it `gridType` and the avaible options should be exactly the same singular name as the Grid module (`point` instead of `points`).

Defining a default param should be fine, we can use `square` as the default since it's the default for `idw` and if we replace that module with `interpolate` then it wouldn't be such a drastic change.

```js
/**
 * @param {string} [gridType='square'] defines the output format based on a Grid Type (options: 'square' | 'point')

DenisCarriere:
Last edits before we merge:
- [ ] Add all other grid types to library.

stebogit:
@DenisCarriere good to merge I believe. 
Nice and colorful module, particularly handy in combo with `isobands` and `isolines`! 😃

DenisCarriere:
Love it!! Super excited to release this module in the next minor release!

image

PR #836, opened 2017-07-12
Set `centered` param default to `True`

### LINKED ISSUE #833: @turf/point-grid - change centered param default to True
## [`@turf/point-grid`](https://github.com/Turfjs/turf/tree/master/packages/turf-point-grid) - change `centered` param default to True
Seems like the `centered` property should default to `True` instead of `False`/`undefined` since this option would be most useful all the time.

```javascript
poinGrid(box, cellSize, units)
//= centered param would equal True (even if defined as undefined)
```

## JSDocs

```javascript
/**
 * Creates a {@link Point} grid from a bounding box, {@link FeatureCollection} or {@link Feature}.
 *
 * @name pointGrid
 * @param {Array<number>|FeatureCollection|Feature<any>} bbox extent in [minX, minY, maxX, maxY] order
 * @param {number} cellSide the distance between points
 * @param {string} [units=kilometers] used in calculating cellSide, can be degrees, radians, miles, or kilometers
 * @param {boolean} [centered=true] adjust points position to center the grid into bbox
 * @param {boolean} [bboxIsMask=false] if true, and bbox is a Polygon or MultiPolygon, the grid Point will be created
 * only if inside the bbox Polygon(s)
 * @returns {FeatureCollection<Point>} grid of points
 */
module.exports = function (bbox, cellSide, units, centered, bboxIsMask) {
```

Ref: Issue https://github.com/Turfjs/turf/issues/613
CC: @stebogit

stebogit:
Fixes #833.

@DenisCarriere I added a note in the doc that notify/warn the user that `centered` will be dropped in the next major release, to try to discourage its use.
Is that ok?


packages/turf-point-grid/index.d.ts
```diff
@@ -3,6 +3,6 @@ import {BBox, Points, Units, Feature, Features} from '@turf/helpers';
 /**
  * http://turfjs.org/docs/#pointgrid
  */
-declare function pointGrid(bbox: BBox | Feature<any> | Features<any>, cellSide: number, units?: Units, centered?: boolean, bboxIsMask?: boolean): Points;
+declare function pointGrid(bbox: BBox | Feature<any> | Array<number>, cellSide: number, units?: Units, centered?: boolean, bboxIsMask?: boolean): Points;
```
DenisCarriere: @stebogit 1. Why the removal of `Features<any>` this is to support FeatureCollection, I guess we should use the full name of the Type (`FeatureCollection<any>`) instead of making the types plural.

2. BBox is already defined - `Array<number>` is not needed

`BBox` => `[number, number, number, number]`

DenisCarriere: I need to take some time and update all the Typescript definitions for each module, make them all consistent.

stebogit: @DenisCarriere it'd be nice if we had some kind of `types` file (`TYPES.md`?) maybe in the root as reference for all Turf modules. With that I could help updating all the modules and would be useful, when creating new models, for who like me doesn't know typescript

DenisCarriere: That would be a good idea, maybe I'll create that before updating all the definitions.

My approach to Typescript is not as strict as others like [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) which enforce strict linting (don't think we are there yet for TurfJS - might add a basic TS linting config).

PR #838, opened 2017-07-12
Added MultiLineString support for @turf/point-on-line

### LINKED ISSUE #835: MultiLineString support for point-on-line
It would be nice if the point-on-line module could accept either a LineString or a MultiLineString parameter.

stebogit:
Fixes #835 

Still the module returns a single `point`, the closest of all the `lines` of the `MultiLineString`.
Is it correct?

CC: @jbccollins @DenisCarriere


packages/turf-point-on-line/index.js
```diff
@@ -44,24 +47,22 @@ var lineString = helpers.lineString;
...
-    var coords;
-    if (line.type === 'Feature') {
-        coords = line.geometry.coordinates;
-    } else if (line.type === 'LineString') {
-        coords = line.coordinates;
-    } else {
-        throw new Error('input must be a LineString Feature or Geometry');
+module.exports = function (lines, pt, units) {
+    // validation
+    if (lines.type !== 'Feature' && lines.type !== 'LineString' && lines.type !== 'MultiLineString' && lines.geometry.type !== 'LineString' && lines.geometry.type !== 'MultiLineString') {
```
DenisCarriere: Condensed this validation step to this:

```js
var type = (lines.geometry) ? lines.geometry.type : lines.type;
if (type !== 'LineString' && type !== 'MultiLineString') {
    throw new Error('lines must be LineString or MultiLineString');
}
```

packages/turf-point-on-line/index.js
```diff
@@ -1,18 +1,21 @@
...
+var coordAll = meta.coordAll;
 var lineString = helpers.lineString;
+
 /**
- * Takes a {@link Point} and a {@link LineString} and calculates the closest Point on the LineString.
+ * Takes a {@link Point} and a {@link LineString} and calculates the closest Point on the (Multi)LineString.
  *
  * @name pointOnLine
- * @param {Feature<LineString>} line line to snap to
+ * @param {Feature<LineString>|Feature<MultiLineString>} lines lines to snap to
```
DenisCarriere: To support both Feature & Geometry, so far the JSDocs have been declared like this:

```js
/**
 * @param {Geometry|Feature<LineString|MultiLineString>} lines lines to snap to
 * @param {Geometry|Feature<Point>|number[]} pt point to snap from
```

packages/turf-point-on-line/index.js
stebogit: 🤦‍♂️ of course! Dumb me 😅

DenisCarriere:
Using happening when the point is between the MultiLineString.
image

packages/turf-point-on-line/index.js
DenisCarriere: No no... it's just simpler to read this way.

stebogit: exactly!
simpler => better!!  😃

DenisCarriere: Agreed! 😄

stebogit:
Fixed with `flattenEach`.

screen shot 2017-07-12 at 11 52 17 pm

Added also a more complex test:
screen shot 2017-07-12 at 11 58 07 pm

DenisCarriere:
@stebogit Awesome! Looks good to me

PR #840, opened 2017-07-14
Suggested DBSCAN implementation for `@turf/clusters-distance`

stebogit:
@DenisCarriere what do you think about this implementation?
It returns a single object with `centroids`, `noise` points and `clustered` points.
I added support for any distance `unit` and some more input validation.

It is significantly slower than yours, in particular because it generates all the `FeatureCollection`s for the output object. However I believe your machine is much more powerful than mine and might perform a little better.
Not sure if it should output an empty `FeatureCollection`, rather than `undefined` or `null`, if there isn't any `noise` points

**Example**: `maxDistance`  = 75 (km), `minPoints` = 3
screen shot 2017-07-13 at 11 00 48 pm

### DIFF
--- a/packages/turf-clusters-distance/bench.js
+++ b/packages/turf-clusters-distance/bench.js
@@ -18,17 +18,19 @@ const fixtures = fs.readdirSync(directory).map(filename => {
 /**
  * Benchmark Results
  *
- * // Clusters distance
- * fiji: 1.692ms
- * many-points: 14.448ms
- * points-with-properties: 0.164ms
- * points1: 0.087ms
- * points2: 0.694ms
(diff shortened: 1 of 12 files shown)


DenisCarriere:
Since the outputs are FeatureCollections they can simply have an empty Array `[]` of features, if you were expecting an `Feature` then that could result in `null` or `undefined` (which isn't the case for this module).

As for the Benchmark performance, I don't think I updated the Benchmark tests, so I'm pretty sure they are suppose to be that slow (`5,000 ops/sec` sounds about right).

Great job! 👍 The output looks really solid!

DenisCarriere:
This approach is way better, going to merge it and then we can work on the `clusters-distance` branch for any last chnages.

Issue #845, opened 2017-07-16
Cluster modules - New modules/features


DenisCarriere:
# Cluster Modules/Features

The past few cluster related modules have helped us define some of the "nice to have" features related to GeoJSON clustering. I propose we start a collection of `clusters-<name>` modules for specific clustering algorithms (mainly focusing on the GeoJSON implementation & leveraging an external dependency for the algorithm) and use `@turf/clusters` as a utility module for common clustering methods & operations:

- [x] `@turf/clusters-kmean` https://github.com/Turfjs/turf/pull/787 (Algorithm dependency [`skmeans`](https://github.com/solzimer/skmeans))
- [x] `@turf/clusters-dbscan`https://github.com/Turfjs/turf/pull/812   (Algorithm dependency [`density-clustering`](https://github.com/uhho/density-clustering))
- [x] `@turf/clusters` https://github.com/Turfjs/turf/issues/841 (only TurfJS dependencies)
  - [x] `getCluster`
  - [x] `clusterEach`
  - [x] `clusterReduce`

## To-Do
- [x] Rename `@turf/clusters` => `@turf/clusters-kmeans`
- [x] Add deprecated warning message to `@turf/clusters` in favor of using `@turf/clusters-kmeans` (~major release `v5.0` calling it the Cluster Release~ `@turf/clusters` was never actually published to `v4.5.0`, going to bump this change to `v4.6.0`)
- [x] ~Rename `@turf/clusters-dbscan` => `@turf/cluster-dbscan`~
- [ ] ~(**Optional**)~ Change output of `@turf/clusters-kmeans` to pure GeoJSON output `FeatureCollection<Point>` (each feature will still include `cluster` property) in favor of leveraging the `@turf/clusters` utility tools to generate derived centroids and/or concave polygons.

So far it's been @stebogit & @DenisCarriere driving the direction of these `clusters` modules, however I would love to hear some feedback from @Turfjs/ownership before we publish the next release of these cluster modules.

> EDIT: Changed `cluster` => `clusters` (plural module name to prevent any accidental conflicts with `cluster` property).

DenisCarriere:
@stebogit Ok finished adding all `@turf/clusters` methods (`getCluster`, `clusterEach`, `clusterReduce`). Looking forward to your feedback (no rush)

stebogit:
Implemented with PR #847

DenisCarriere:
[...]

PR #847, opened 2017-07-17
New @turf/clusters module (getCluster, clusterEach, clusterReduce)

DenisCarriere:
# New `@turf/clusters` module

Ref: https://github.com/Turfjs/turf/issues/845

## Methods

- [x] `getCluster`
- [x] `clusterEach`
- [x] `clusterReduce`

## To-Do

- [x] Improve input validations
- [x] Drop support of Array of Features.
- [x] Use `@turf/clusters-kmeans` for `@example` (improve documentation)
- [x]  Leverage `@turf/helpers` & `@turf/meta` for GeoJSON compatibility 

## JSDocs - `getCluster`

```javascript
/**
 * Get Cluster
 *
 * @param {FeatureCollection} geojson GeoJSON Features
 * @param {*} filter Filter used on GeoJSON properties to get Cluster
 * @returns {FeatureCollection} Single Cluster filtered by GeoJSON Properties
 * @example
 * var geojson = turf.featureCollection([
 *     turf.point([0, 0]),
 *     turf.point([2, 4]),
 *     turf.point([3, 6]),
 *     turf.point([5, 1]),
 *     turf.point([4, 2])
 * ]);
 *
 * // Create a cluster using K-Means (adds `cluster` to GeoJSON properties)
 * var clustered = turf.clustersKmeans(geojson);
 *
 * // Retrieve first cluster (0)
 * var cluster = turf.getCluster(clustered, {cluster: 0});
 */
function getCluster(geojson, filter) {
```

## JSDocs - `clusterEach`

```js
/**
 * Callback for clusterEach
 *
 * @callback clusterEachCallback
 * @param {FeatureCollection} cluster The current cluster being processed.
 * @param {*} clusterValue Value used to create cluster being processed.
 * @param {number} currentIndex The index of the current element being processed in the array.Starts at index 0
 * @returns {void}
 */

/**
 * clusterEach
 *
 * @param {FeatureCollection} geojson GeoJSON Features
 * @param {string|number} property GeoJSON property key/value used to create clusters
 * @param {Function} callback a method that takes (cluster, clusterValue, currentIndex)
 * @returns {void}
 * @example
 * var geojson = turf.featureCollection([
 *     turf.point([0, 0]),
 *     turf.point([2, 4]),
 *     turf.point([3, 6]),
 *     turf.point([5, 1]),
 *     turf.point([4, 2])
 * ]);
 *
 * // Create a cluster using K-Means (adds `cluster` to GeoJSON properties)
 * var clustered = turf.clustersKmeans(geojson);
 *
 * // Iterate over each cluster
 * clusterEach(clustered, 'cluster', function (cluster, clusterValue, currentIndex) {
 *     //= cluster
 *     //= clusterValue
 *     //= currentIndex
 * })
 */
function clusterEach(geojson, property, callback) {
```

## JSDocs - `clusterReduce`

```js
/**
 * Callback for clusterReduce
 *
 * The first time the callback function is called, the values provided as arguments depend
 * on whether the reduce method has an initialValue argument.
 *
 * If an initialValue is provided to the reduce method:
 *  - The previousValue argument is initialValue.
 *  - The currentValue argument is the value of the first element present in the array.
 *
 * If an initialValue is not provided:
 *  - The previousValue argument is the value of the first element present in the array.
 *  - The currentValue argument is the value of the second element present in the array.
 *
 * @callback clusterReduceCallback
 * @param {*} previousValue The accumulated value previously returned in the last invocation
 * of the callback, or initialValue, if supplied.
 * @param {FeatureCollection} cluster The current cluster being processed.
 * @param {*} clusterValue Value used to create cluster being processed.
 * @param {number} currentIndex The index of the current element being processed in the
 * array. Starts at index 0, if an initialValue is provided, and at index 1 otherwise.
 */

/**
 * Reduce clusters in GeoJSON Features, similar to Array.reduce()
 *
 * @name clusterReduce
 * @param {FeatureCollection} geojson GeoJSON Features
 * @param {string|number} property GeoJSON property key/value used to create clusters
 * @param {Function} callback a method that takes (previousValue, cluster, clusterValue, currentIndex)
 * @returns {*} The value that results from the reduction.
 * @example
 * var geojson = turf.featureCollection([
 *     turf.point([0, 0]),
 *     turf.point([2, 4]),
 *     turf.point([3, 6]),
 *     turf.point([5, 1]),
 *     turf.point([4, 2])
 * ]);
 *
 * // Create a cluster using K-Means (adds `cluster` to GeoJSON properties)
 * var clustered = turf.clustersKmeans(geojson);
 *
 * // Iterate over each cluster and perform a calculation
 * var initialValue = 0;
 * var total = turf.clusterReduce(clustered, 'cluster', function (previousValue, cluster, clusterValue, currentIndex) {
 *     //=previousValue
 *     //=cluster
 *     //=clusterValue
 *     //=currentIndex
 *     return previousValue++;
 * }, initialValue);
 */
function clusterReduce(geojson, property, callback, initialValue) {
```

## Benchmark Results
```
testing -- createBins x 2,954,562 ops/sec ±1.29% (89 runs sampled)
testing -- propertiesContainsFilter x 10,902,619 ops/sec ±2.11% (85 runs sampled)
testing -- filterProperties x 27,860,084 ops/sec ±1.12% (88 runs sampled)
testing -- applyFilter x 18,971,594 ops/sec ±7.14% (77 runs sampled)
getCluster -- string filter x 5,873,843 ops/sec ±1.30% (87 runs sampled)
getCluster -- object filter x 883,469 ops/sec ±1.28% (93 runs sampled)
getCluster -- aray filter x 3,602,147 ops/sec ±1.30% (89 runs sampled)
clusterEach x 1,133,904 ops/sec ±0.90% (89 runs sampled)
clusterReduce x 1,057,572 ops/sec ±1.13% (89 runs sampled)
```


packages/turf-clusters/index.js
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * // Create a cluster by an Array of Strings
+ * var cluster3 = turf.getCluster(geojson, ['cluster', 'foo']);
+ */
+function getCluster(geojson, filter) {
+    // Validation
+    if (!geojson) throw new Error('geojson is required');
+    if (filter === undefined || filter === null) throw new Error('filter is required');
+    if (geojson.type === 'FeatureCollection') geojson = geojson.features;
+    if (!Array.isArray(geojson)) throw new Error('invalid geojson');
```
tmcw: Seems like an appropriate place to use turf/meta to handle any sort of data. This currently wouldn't support valid GeoJSON constructs like bare feature objects, and would support invalid geojson constructs, like an array of features.

DenisCarriere:
> It looks like this module would be mostly used after turf-cluster-kmeans or another geometrical clustering algorithm

Agreed, this module will be used after `kmeans` & `dbscan` or any other clustering algorithms, which will allow us to focus on only adding a `cluster` property to each feature.

**Note:** `@turf/clusters-kmeans` will be changed to a single GeoJSON output (with extra cluster property), then we will generate centroids using the `clusterEach` method.

> because it essentially is a filter module? 

Agreed! 😄 Kinda wanted to call it `groupBy` or `filterBy`, the initial topic that started all this was about clustering.

> Seems like something that's mostly to be solved in the documentation

Agreed, once we have `dbscans` & `kmeans` using this module would be a lot easier to understand. Good idea, we should use `kmeans` directly in the `@example`.

@tmcw

packages/turf-clusters/index.js
DenisCarriere: `@turf/invariant` right? To test for a valid FeatureCollection.
> would support invalid geojson constructs, like an array of features.

Agreed.. 👍 That's my "cheeky" style, i'll drop it.

DenisCarriere:
@tmcw 👍 Thanks for the review

packages/turf-clusters/index.js
DenisCarriere: > Seems like an appropriate place to use turf/meta to handle any sort of data. 
Oh, you mean `featureEach` in `@turf/meta`?

tmcw: Oh, noticed that the docs say it's a FeatureCollection specifically - so, yep! turf/invariant would be useful to enforce that requirement. turf/meta would be useful if the docs promised that the input was any geojson object.

DenisCarriere: Originally I had it as `FeatureCollection|Feature[]` (`FeatureCollection|Array<Feature>`), now that we've dropped the Array of features (since it's not a valid GeoJSON), there isn't much complex validation to do which would require using `@turf/invariant`. Using [`geojsonType`](https://github.com/Turfjs/turf/blob/master/packages/turf-invariant/index.js#L74-L89) is already being handled [here](https://github.com/Turfjs/turf/pull/847/files#diff-b96c64619e5de1f4ad32bb28a082aa44R28):
```js
if (geojson.type !== 'FeatureCollection') throw new Error('geojson must be a FeatureCollection');
```

DenisCarriere: I did however add `featureEach` from `@turf/meta` to handle the iteration of the FeatureCollection.

packages/turf-clusters/README.md
```diff
@@ -0,0 +1,172 @@
+# @turf/clusters
+
+# getCluster
+
+Get Cluster
+
+**Parameters**
+
+-   `geojson` **[FeatureCollection](http://geojson.org/geojson-spec.html#feature-collection-objects)** GeoJSON Features
+-   `filter` **Any** Filter used on GeoJSON properties to get Cluster
```
stebogit: Now that I think about it, only in `k-means` the number of clusters is known before applying the module (which often is considered its drawbacks), so you can easily "query" for a specific cluster `id` for example; on the other side if you apply `dbscan` or other algorithms you don't know a priori how many clusters will be identified.
It would be useful, I guess, to have a method that returns the total number of clusters in the `FeatureCollection` (basically how many groups of points have the same property with different values); it could be this one when passed `null` or `'all'` as filter.

packages/turf-clusters/README.md
```diff
@@ -0,0 +1,172 @@
...
+var geojson = turf.featureCollection([
+    turf.point([0, 0]),
+    turf.point([2, 4]),
+    turf.point([3, 6]),
+    turf.point([5, 1]),
+    turf.point([4, 2])
+]);
+
+// Create a cluster using K-Means (adds `cluster` to GeoJSON properties)
+var clustered = turf.clustersKmeans(geojson);
```
stebogit: Although their primary use is actual clustered points, I mean as output of a clustering module, these functions could be useful to identify any group of points with a common property, even among points where not all have said property (like identify all points with a certain `marker-symbol`).
I think it could be beneficial to the user to see examples that are not necessarily related to clustered (in the above mentioned meaning) points, in order to suggest other uses and applications for this module's functions.

packages/turf-clusters/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "turf",
+    "geojson",
+    "cluster",
+    "clusters",
+    "clustering"
+  ],
+  "author": "Turf Authors",
+  "contributors": [
+    "Denis Carriere <@DenisCarriere>",
+    "Stefano Borghi <@stebogit>"
```
stebogit: I didn't really do anything in here, it's all your own work 😅

packages/turf-clusters/index.js
```diff
@@ -0,0 +1,268 @@
...
+        if (properties.hasOwnProperty(key)) newProperties[key] = properties[key];
+    }
+    return newProperties;
+}
+
+module.exports = {
+    getCluster: getCluster,
+    clusterEach: clusterEach,
+    clusterReduce: clusterReduce,
+    createBins: createBins, // Only exposed for testing purposes
```
stebogit: @DenisCarriere out of curiosity, how do you exclude these functions from the official release (if that's the plan)?

DenisCarriere: Going to comment them out, or... we leave them there and no document their usage.

At the moment all of these `utils` functions have tests cases for them.

packages/turf-clusters/package.json
DenisCarriere: Got preserved from the last module, I can drop you ;) lol

packages/turf-clusters/README.md
DenisCarriere: Agreed, feel free to change some of these examples or add a 2nd example.

packages/turf-clusters/README.md
DenisCarriere: This type of questions would be better answered using `clusterEach` or `clusterReduce`.

Calculating how many total clusters you could use `clusterReduce` and `previousValue++` which would give you the total, or if you want to figure all of the values which created a bin you could also use `clusterReduce`.

> I'll add some more examples to `clusterReduce` & `clusterEach`

packages/turf-clusters/README.md
DenisCarriere: ✅ Added

```js
var geojson = turf.featureCollection([
    turf.point([0, 0], {'marker-symbol': 'circle'}),
    turf.point([2, 4], {'marker-symbol': 'star'}),
    turf.point([3, 6], {'marker-symbol': 'star'}),
    turf.point([5, 1], {'marker-symbol': 'square'}),
    turf.point([4, 2], {'marker-symbol': 'circle'})
]);

// Create a cluster using K-Means (adds `cluster` to GeoJSON properties)
var clustered = turf.clustersKmeans(geojson);

// Retrieve first cluster (0)
var cluster = turf.getCluster(clustered, {cluster: 0});
//= cluster

// Retrieve cluster based on custom properties
turf.getCluster(clustered, {'marker-symbol': 'circle'}).length;
//= 2
turf.getCluster(clustered, {'marker-symbol': 'square'}).length;
//= 1
```

packages/turf-clusters/package.json
DenisCarriere: ✅ Dropped you 😄

packages/turf-clusters/index.js
DenisCarriere: Come to think about it, these are ONLY going to be exposed in the `@turf/clusters` module, however they will not be exposed to `@turf/turf` (which is the main repo that most users will be using).

The only people who will actually know that these methods are exposed will be the ones that dive into the source code or `require('@turf/clusters`)` and see which methods the module exposes.

Since there's test cases for each of these functions already I'm going to keep them exposed, I don't see any benefit "commented" them out for the sake of it.

morganherlocker:
Incredible work! 🎉

DenisCarriere:
Thanks @morganherlocker!

PR #851, opened 2017-07-20
Drop array of features from clusters-dbscan

DenisCarriere:
# Updates to `@turfl/clusters-dbscan`

- [x] Dropped array of Features
- [ ] Create centroids & concave polygons using `clusterEach` (for testing purposes only)

### DIFF
--- a/packages/turf-clusters-dbscan/index.d.ts
+++ b/packages/turf-clusters-dbscan/index.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="geojson" />
 
-import {Units, Points, Point} from '@turf/helpers';
+import {Units, Points} from '@turf/helpers';
 
 interface Output {
     type: 'FeatureCollection'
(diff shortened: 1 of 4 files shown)


DenisCarriere:
Made a fun interactive leaflet map with `@turf/clusters-dbscan`! 🎉 

https://github.com/DenisCarriere/turf-example-clusters-dbscan

image

@stebogit I can try to do one for `@turf/clusters-kmeans` too.

stebogit:
@DenisCarriere I'm astonished. That is super fun, and instructive! 😃

DenisCarriere:
@stebogit I showed a few friends this tonight and we just kept making clusters for about 20 minutes  😝

If I ever create another example/project with `dbscan` or any other clustering algorithms I'll make sure to share it here.

The entire source code to make that `dbscan` example is hosted here:
https://github.com/DenisCarriere/turf-example-clusters-dbscan

Pretty straightforward static website, no Webpack, no Babel, no Typescript, only vanilla Javascript + HTML.

PR #856, opened 2017-07-22
New module `@turf/boolean-overlaps`

stebogit:
Solves #286.
@DenisCarriere can you please take a look at the Typescript file?

Thanks @tcql for the great job! 👍 

**To-Do**:
- [x] add support for `Point` & `MultiPoint`
- [x] update `index.d.ts`
- [x] refactor using `@turf/line-intersects`, `segmentEach`, ` flattenEach` 
- [x] rename module to `@turf/boolean-overlap`
- [ ] add/test geometry support

```js
/**
 * Takes two features and returns true or false whether or not they overlap, i.e. whether any pair 
 * of edges on the two polygons intersect. If there are any edge intersections, the polygons 
 * overlap.
 *
 * @name booleanOverlaps
 * @param  {Geometry|Feature<LineString|MultiLineString|Polygon|MultiPolygon>} feature1 input
 * @param  {Geometry|Feature<LineString|MultiLineString|Polygon|MultiPolygon>} feature2 input
 * @returns {boolean} true/false
 * @example
 * var poly1 = turf.polygon([[[18.70,-34.19],[18.93,-34.19],[18.93,-34],[18.70,-34],[18.70,-34.19]]]);
 * var poly2 = turf.polygon([[[18.52,-34.36],[18.79,-34.36],[18.79,-34.10],[18.52,-34.10],[18.52,-34.36]]]);
 * var line = turf.lineString([[18.62,-34.39],[18.87,-34.21]]);
 *
 * turf.booleanOverlaps(poly1, poly2)
 * //=true
 * turf.booleanOverlaps(poly2, line)
 * //=false
 */
module.exports = function (feature1, feature2) {
```


DenisCarriere:
@stebogit Couldn't we also support `Point` & `MultiPoint`? Sounds like it would be an easy one to do and that way we could use this module for Every GeoJSON type.

DenisCarriere:
The definition looks good, however if we did support all Geometry Types then it would look something like this:

```typescript
/// <reference types="geojson" />

type Feature = GeoJSON.Feature<any> | GeoJSON.GeometryObject

/**
 * http://turfjs.org/docs/#boolean-overlaps
 */
declare function overlaps(feature1: Feature, feature2: Feature): boolean;
declare namespace overlaps { }
export = overlaps;
```

packages/turf-boolean-overlaps/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Returns if the two segments intersect
+ *
+ * @private
+ * @param {Array<Array<number>>} line1 array of a segment coordinates
+ * @param {Array<Array<number>>} line2 array of a segment coordinates
+ * @returns {boolean} segments intersect
+ */
+function doSegmentsIntersect(line1, line2) {
```
DenisCarriere: Couldn't this be replaced with `@turf/line-intersects`? The module goes really fast if you only provide 2 vertex line segments.
https://github.com/Turfjs/turf/blob/master/packages/turf-line-intersect/index.js#L85

packages/turf-boolean-overlaps/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Returns if the two lineStrings intersect
+ *
+ * @private
+ * @param {Array<Array<number|Array<number>>>} ring1 array of a LineString coordinates
+ * @param {Array<Array<number|Array<number>>>} ring2 array of a LineString coordinates
+ * @returns {boolean} lines intersect
+ */
+function doLinesIntersect(ring1, ring2) {
```
DenisCarriere: We should build a `segmentEach` & `segmentReduce` module in [`@turf/meta`](https://github.com/Turfjs/turf/tree/master/packages/turf-meta) for iterating over line segments. (most logic can be pulled from [`@turf/line-segment`](https://github.com/Turfjs/turf/tree/master/packages/turf-line-segment)).

packages/turf-boolean-overlaps/index.js
```diff
@@ -0,0 +1,105 @@
...
+    var coords1 = getCoordinatesArr(feature1),
+        coords2 = getCoordinatesArr(feature2);
+
+    // Since we don't care about the overlap amount, or it's geometry, but rather just whether overlap
+    // occurs, polygon overlap can most simply be expressed by testing whether any pair of edges on the two polygons
+    // intersect. If there are any edge intersections, the polygons overlap.
+
+    // This looks completely stupid ridiculous to have so many nested loops, but it supports
+    // multipolygons nicely. In the case of polygons or linestrings, the outer loops are only one iteration.
+    return coords1.some(function (rings1) {
```
DenisCarriere: We have [`flattenEach`](https://github.com/Turfjs/turf/tree/master/packages/turf-meta#flatteneach) which solves this "stupid ridiculous" pattern 😝  (pulling directly from comment above).

DenisCarriere:
Since this is one of the DE-9IM modules, we should use the same naming convention `overlap` instead of `overlaps`.

http://edndoc.esri.com/arcsde/9.0/general_topics/understand_spatial_relations.htm

@rowanwins had already started working on this in the past, might be worth including some of his test cases (PR https://github.com/Turfjs/turf/pull/707 &  [turf/packages/boolean-overlap](https://github.com/Turfjs/turf/tree/f40b3d13368e2f3873e5cd4e019a252594f73a65/packages/turf-boolean-overlap)).

stebogit:
> Couldn't we also support Point & MultiPoint? 

👍  great idea!
so, just to be all on the same page:
- **`true`** if a`Point` overlaps (i.e. equals geometries) another `Point`
- **`true`** if a`Point` overlaps one point of a `MultiPoint`
- **`true`** if _ANY_ `Point` of a `MultiPoint` overlaps one point of a `MultiPoint` feature
- **`false`** otherwise

correct?

I'll implement the changes you suggested using the available modules.
I'll also try to define the `segmentEach` & `segmentReduce` modules, I like the idea!

> Since this is one of the DE-9IM modules, we should use the same naming convention overlap instead of overlaps.

Sure thing

DenisCarriere:
@stebogit You are spot on https://github.com/Turfjs/turf/pull/856#issuecomment-317218072 👍

rowanwins:
Gday @stebogit 

For verifying scenarios it's also worth using @DenisCarriere handy `boolean-shapely` [module](https://github.com/DenisCarriere/boolean-shapely), you can see it in action [here](https://github.com/Turfjs/turf/blob/master/packages/turf-boolean-contains/test.js). Basically it allows us to check our definitions against a widely used python library, you'll just need to make sure you have python installed and the shapely lib.

DenisCarriere:
Thanks @rowanwins for sharing, that `boolean-shapely` module is pretty handy for testing purposes, `overlaps` is supported.

packages/turf-boolean-overlaps/index.js
DenisCarriere: Done, `segmentEach` & `segmentReduce` is now available in PR https://github.com/Turfjs/turf/pull/863.

You can use it right now (after PR merged) by using `lerna bootstrap` on an updated branch.

stebogit:
@DenisCarriere @rowanwins 
looking at the [spatial relations article](http://edndoc.esri.com/arcsde/9.0/general_topics/understand_spatial_relations.htm), it seems to me that the current implementation of `@turf/boolean-overlap` is actually kind of a mix of **overlap** and **cross**.
In fact it should only
> compares two geometries of the **same dimension** and returns t (TRUE) if their intersection set results in a geometry **different** from both but of the **same dimension**.

Shall I implement this behavior?
It would only apply then to `(Multi)Point`, `(Multi)LineString` and `(Multi)Polygon`; following `@turf/boolean-crosses` example I would throw an error on `Point` input(s).

rowanwins:
Gday @stebogit 

Yep to aim is to follow the implementation provided by those guidelines. If in doubt about how the rules play out check the scenarios against shapely.

Issue #113, opened 2014-05-08
Convex hull (buffering tin polys) is slow


atdrago:
I've been playing with the `convex` method and notice that it can get very slow (> 1s at times) because of the buffer that happens: 

``` javascript
mergePolys = t.merge(t.buffer(tinPolys, .05, 'miles'));
```

When I take out the buffer, `convex` is much faster (< 100ms) (presumably because of the amount of polygons that `tin` creates, which then must be buffered; untested). What do you think about modifying `convex` to remove the `buffer` (potentially breaking change), or creating a separate "quickConvex" method that doesn't buffer the polygons before merging? I'd be willing to put together whatever solution you feel is best, and then issue a pull request. 

Thank you!

PR #857, opened 2017-07-22
Fixed position of documentation.js '//addToMap' tag

stebogit:
Ref. [Turf-www #113](https://github.com/Turfjs/turf-www/issues/113#issuecomment-317166131)

Updated `@examples` doc for updating the website


packages/turf-bezier/index.js
```diff
@@ -33,9 +33,9 @@ var Spline = require('./spline.js');
  * };
  *
  * var curved = turf.bezier(line);
+ * curved.properties = { stroke: '#0f0' };
  *
  * //addToMap
- * curved.properties = { stroke: '#0f0' };
  * var addToMap = [line, curved]
  */
```
rowanwins: so rather than moving the `curved.properties` above the `//addToMap` tag, I generally move it to below the `var addToMap = [line, curved]`. That benefit of that is that it isn't rendered as part of the code sample (which might potentially confuse people)

So I'd make it look like
````
  * var curved = turf.bezier(line);
   *
   * //addToMap
   * var addToMap = [line, curved];
   * curved.properties = { stroke: '#0f0' };
   */
````

Comment (DenisCarriere) [inline packages/turf-bezier/index.js]: 👍 Agreed, so anything that's below`//addToMap` is not displayed in the website.

Comment (stebogit) [inline packages/turf-bezier/index.js]: will do! 👍

**Comment by DenisCarriere** [comment]
Just pushed the latest `@example` tests to this PR, we should be able to detect those JSDocs example errors a lot easier now.

**Comment by DenisCarriere** [comment]
Ok fixed most of the major `@example` errors are fixed.

I've excluded some of the new modules (6 skipped) from the tests since they haven't been added to `@turf/turf` yet.

```
$ tap test.example.js
test.example.js ................................... 100/106
  Skipped: 6
    clustersDbscan
    clustersKmeans
    getCluster
    clusterEach
    clusterReduce
    interpolate

total ............................................. 100/106

  100 passing (1s)
  6 pending
```

DenisCarriere:
@rowanwins 👍 Updated the `addToMap` JSDocs.

packages/turf-bezier/index.js
stebogit: @DenisCarriere @rowanwins ok the code after `addToMap = ...` is not rendered in the map on the website. So what's the purpose of adding properties in the example, which highlight the result on the map, if they are not actually used for the map?

DenisCarriere: There's two parts, the map which uses the addToMap as a featureCollection with all the properties. The second is the JSDocs example being rendered on the website which does not include the addToMap details.
Most of the extra properties added is to properly visualize the map example.

Hopefully that helps answer your question

PR #858, opened 2017-07-22
Initial commit of booleanPointOnLine

rowanwins:
Please fill in this template.

## booleanPointOnLine
Returns true if a point is on a line. Accepts a optional parameter to ignore the start and end vertices of the linestring.

This is a snippet that we've used across a number of the boolean modules now so figured it was worth separating out.


stebogit:
@rowanwins @DenisCarriere 
looking at this new module (:+1: btw), may I suggest (for the following major release) to rename `@turf/point-on-line` to something like `@turf/nearest-point-on-line`?
Otherwise it's a little confusing and the user _has_ to go to the docs or the code to understand the difference.
We might even "group" the `nearest` modules, allowing an easy addition of other modules alike, renaming also `@turf/nearest` to `@turf/nearest-point`

DenisCarriere:
> @turf/point-on-line to something like @turf/nearest-point-on-line?
Otherwise it's a little confusing and the user has to go to the docs or the code to understand the difference.
We might even "group" the nearest modules, allowing an easy addition of other modules alike, renaming also @turf/nearest to @turf/nearest-point

👍 I like this change for a next major release, it is a big confusing when they aren't grouped and the name is too generalized.

packages/turf-boolean-point-on-line/index.js
```diff
@@ -0,0 +1,65 @@
...
+ * @example
+ * var pt = turf.point([0, 0]);
+ * var line = turf.lineString([[-1, -1],[1, 1],[1.5, 2.2]]);
+ * var isPointOnLine = turf.booleanPointOnLine(pt, line);
+ * //=true
+ */
+module.exports = function (point, linestring, ignoreEndVertices) {
+    var pointCoords = getCoords(point);
+    var lineCoords = getCoords(linestring);
+    for (var i = 0; i < lineCoords.length - 1; i++) {
```
DenisCarriere: 👍 Code looks good.

We really need a `segmentEach` module (also mentioned here https://github.com/Turfjs/turf/pull/856#discussion_r128902548)

packages/turf-boolean-point-on-line/index.js
```diff
@@ -0,0 +1,65 @@
+var getCoords = require('@turf/invariant').getCoords;
+
+/**
+ * Returns true if a point is on a line. Accepts a optional parameter to ignore the start and end vertices of the linestring.
+ *
+ * @name booleanPointOnLine
+ * @param {Geometry|Feature<Point>} point GeoJSON Feature or Geometry
+ * @param {Geometry|Feature<LineString>} linestring GeoJSON Feature or Geometry
+ * @param {Boolean} [ignoreEndVertices=false] whether to ignore the start and end vertices.
+ * @returns {Boolean} true/false
```
DenisCarriere: I remember having this conversation with @tmcw about `Boolean` & `boolean`.

From what I can read on JSDocs, it's saying to use `boolean`
http://usejsdoc.org/tags-type.html

Also DocumentationJS linting flags `Boolean` as a warning (which we will be adding soon)

This change applies to many modules, so don't need to fix this right now.

packages/turf-boolean-point-on-line/index.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="geojson" />
+
+type Feature = GeoJSON.Feature<any> | GeoJSON.GeometryObject;
+
+/**
+ * http://turfjs.org/docs/#booleanPointOnLine
```
DenisCarriere: All lowercase for the URL.

Example http://turfjs.org/docs/#centerOfMass (doesn't work)

packages/turf-boolean-point-on-line/bench.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Benchmark Results
+ *
+ * PointOnLineEnd: 0.355ms
+ * PointOnLineMidpoint: 0.019ms
+ * PointOnLineStart: 0.021ms
+ * PointOnLineEnd x 13,957,263 ops/sec ±0.53% (91 runs sampled)
+ * PointOnLineMidpoint x 17,388,052 ops/sec ±0.46% (94 runs sampled)
+ * PointOnLineStart x 17,036,405 ops/sec ±1.34% (91 runs sampled)
```
DenisCarriere: 🏎  💨 Zoom Zoom!

packages/turf-boolean-point-on-line/index.js
rowanwins: The only thing I find tricky about the invariant type modules is a lack of flow control, but in a lot of cases they are super-handy!

DenisCarriere: What do you mean by lack of flow control?

packages/turf-boolean-point-on-line/index.d.ts
```diff
@@ -1,12 +1,11 @@
...
 type Point = GeoJSON.Feature<GeoJSON.Point> | GeoJSON.Point;
-type LineString = GeoJSON.LineString;
-type Line = GeoJSON.Feature<LineString> | LineString;
+type Line = GeoJSON.Feature<GeoJSON.LineString> | GeoJSON.LineString;
 
 /**
  * http://turfjs.org/docs/#booleanpointonline
  */
-declare function booleanPointOnLine(feature1: Point, feature2: Line, ignoreEndVertices: boolean): boolean;
+declare function booleanPointOnLine(point: Point, linestring: Line, ignoreEndVertices?: boolean): boolean;
```
DenisCarriere: @rowanwins Another thing with these definitions, I try to use the same parameter names as the `index.js`

And `ignoreEndVertices` is optional, you can add `?` to define optional params.

PR #865, opened 2017-07-26
🐛 Fix #837 - @turf-mask

### LINKED ISSUE #837: turf-mask fails
http://geojson.io/#id=gist:anonymous/59a73beffdced0e4f7d2d12eb271d380&map=16/-36.8815/174.8318

http://jsbin.com/joqeraloxe/edit?html,console,output

In the above data/example app, turf-mask fails with:

```
Uncaught Error: Each LinearRing of a Polygon must have 4 or more Positions.
    at Object.polygon (npmcdn.com/@turf/turf@4.5.2/turf.js:599)
    at buildMask (npmcdn.com/@turf/turf@4.5.2/turf.js:20252)
    at Object.module.exports [as mask] (npmcdn.com/@turf/turf@4.5.2/turf.js:20228)
    at e.<anonymous> (<anonymous>:516:23)
    at e.Evented.fire (mapbox-gl.js:411)
    at e._render (mapbox-gl.js:385)
```

From what I could tell it's the `unionPolygons` step where it returns a MultiPolygon but the spec at https://github.com/Turfjs/turf/blob/master/packages/turf-mask/index.js#L137 says it should only return Polygon: `@returns {FeatureCollection<Polygon>} polygons only apply union if they collide`

I don't have the time to investigate further and I'll workaround this issue so it's not a critical issues.

/cc @rowanwins just flagging you as you did the original implementation.

thiagoxvo:
Fix #837 
Changing from `featureEach` to `flattenEach` as suggested in the issue solved the problem.

I've also updated the test case for MultiPolygon to use the example that was broken before.


packages/turf-mask/index.js
```diff
@@ -84,11 +84,11 @@ function buildMask(maskPolygon, polygonOuters, polygonInners) {
 function separatePolygons(polygon) {
     var outers = [];
     var inners = [];
-    featureEach(polygon, function (multiFeature) {
+    flattenEach(polygon, function (multiFeature) {
         if (multiFeature.geometry.type === 'MultiPolygon') {
             multiFeature = flattenMultiPolygon(multiFeature);
         }
-        featureEach(multiFeature, function (feature) {
+        flattenEach(multiFeature, function (feature) {
```
DenisCarriere: The 2nd flattenEach is no longer required.

```js
if (multiFeature.geometry.type === 'MultiPolygon') {
    multiFeature = flattenMultiPolygon(multiFeature);
}
```
Can be dropped

rowanwins:
Strong emoji game on the pull request title as well @thiagoxvo , I approve!

Issue #853, opened 2017-07-20
Geometry null and bbox


fxi:
Hi,

Geometry null seems valid according to the geojson spec, but bbox does not like it. This error is thrown when the geometry is null with turf/bbox version 4.5.2;

`TypeError: Cannot read property 'type' of null`

```js
gJ = {
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": null
    }
  ]
}

turf.bbox(gJ);
```

Am I missing something ?

Thanks !

DenisCarriere:
> Geometry null seems valid according to the geojson spec

Where did you read that?

Also there's a newer [GeoJSON specification RFC 7946](https://tools.ietf.org/html/rfc7946#section-3.1) and the older deprecated [GeoJSON specification](http://geojson.org/geojson-spec.html#geometry-objects).

Both of them say nothing about `null` being a valid Geometry Object.

**New docs**
>A GeoJSON Geometry object of any type other than "GeometryCollection" has a member with the name "coordinates". The value of the "coordinates" member **is an array**.  The structure of the elements in this array is determined by the type of geometry.  GeoJSON processors MAY interpret Geometry objects with empty "coordinates" arrays as null objects.

**Older docs**
> A GeoJSON geometry object of any type other than "GeometryCollection" must have a member with the name "coordinates". The value of the coordinates member **is always an array**. The structure for the elements in this array is determined by the type of geometry.

fxi:
Ok, I did not see the deprecated notice, sorry, but it was there:

http://geojson.org/geojson-spec.html#feature-objects

> A feature object must have a member with the name "geometry". The value of the geometry member is a geometry object as defined above **or a JSON null value**.

Then in the newer version:
https://tools.ietf.org/html/rfc7946#section-3.2

> A Feature object has a member with the name "geometry".  The value of the geometry member SHALL be either a Geometry object as defined above or, in the case that the Feature is unlocated, a **JSON null value**.

And I use geojsonhint to do the validation, but it did not catch this:

https://jsfiddle.net/d9zob2c5/1/

But maybe there is something obvious I've missed..

DenisCarriere:
> A feature object must have a member with the name "geometry". The value of the geometry member is a geometry object as defined above or a JSON null value.

🤓 Heh... I guess it does seem valid... Learn something everyday :)

[...]

PR #866, opened 2017-07-26
Support Null geometries

DenisCarriere:
Adds `null` geometry support.

Ref: https://github.com/Turfjs/turf/issues/853

## Modified

- [x] [`@turf/invariant`](https://github.com/Turfjs/turf/tree/master/packages/turf-invariant)
  - [x] `getGeom`
  - [x] `getGeomType`
  - [x] `getCoord`
  - [x] `getCoords`
- [x] [`@turf/helpers`](https://github.com/Turfjs/turf/tree/master/packages/turf-helpers)
  - [x] `feature`
  - [x] `featureCollection`
- [x] [`@turf/meta`](https://github.com/Turfjs/turf/tree/master/packages/turf-meta)
  - [x] `geomEach`
  - [x] `geomReduce`
  - [x] `featureEach`
  - [x] `featureReduce`
  - [x] `flattenEach`
  - [x] `flattenReduce`
  - [x] `coordEach`
  - [x] `coordReduce`

## Implementation Discussions

**Callbacks Feature/Geometry with `null` geometry**
- [x] `geomEach`
- [x] `flattenEach`
- [x] `featureEach` 

**Does NOT return any Coordinate/Segment**
- [x] `coordEach`
- [x] `segmentEach`

## Tested

- [x] [`@turf/bbox`](https://github.com/Turfjs/turf/tree/master/packages/turf-bbox)
- [x] [`@turf/bbox-clip`](https://github.com/Turfjs/turf/tree/master/packages/turf-bbox-clilp)


packages/turf-invariant/test.js
```diff
@@ -208,3 +208,17 @@ test('invariant#getGeomType', t => {
...
 });
+
+// https://github.com/Turfjs/turf/issues/853
+test('null geometries', t => {
+    const nullFeature = {
+        type: 'Feature',
+        properties: {},
+        geometry: null
+    }
+    t.equal(invariant.getGeom(null), null);
```
stebogit: @DenisCarriere why supporting `null` values? Reading the [specs](https://tools.ietf.org/html/rfc7946#section-3) it seems the GeoJSON objects _is_ an object and _has_ to have a `type` member (which in my understanding is the only requirement to be a valid GeoJSON object)

packages/turf-meta/index.js
```diff
@@ -48,64 +48,73 @@ function coordEach(geojson, callback, excludeWrapCoord) {
...
+            var geomType = (geometry === null) ? null : geometry.type;
 
-            wrapShrink = (excludeWrapCoord &&
-                (geometry.type === 'Polygon' || geometry.type === 'MultiPolygon')) ?
-                1 : 0;
+            wrapShrink = (excludeWrapCoord && (geomType === 'Polygon' || geomType === 'MultiPolygon')) ? 1 : 0;
 
-            if (geometry.type === 'Point') {
+            switch (geomType) {
+            case null:
```
stebogit: @DenisCarriere same here, from the [specs](https://tools.ietf.org/html/rfc7946#section-3):
>  * A GeoJSON object has a member with the name "type".  The value of
      the member MUST be one of the GeoJSON types.

I guess the `geometry` attribute might be `null`, but the `type` must be a valid string

DenisCarriere:
Thanks for looking to this @stebogit , I wasn't too sure how to handle this for the `invariant` module.

>  I believe we should allow/hande only the `geometry: null` case; a type member must be defined to have a valid GeoJSON object (see specs), so null or an object with `type:null` are not valid inputs.

👍 Yep, we should only handle the `geometry: null` case.

packages/turf-meta/index.js
DenisCarriere: This section is more internal and doesn't need to reflect the exact spec. Only way to iterate over a FeatureCollection (containing valid `null` geometries) with `coordEach` is to not throw `Unknown Geometry Type` the error since we know that `null` is a valid geometry.

DenisCarriere:
@stebogit ❓ Question for you: When we're using `geomEach` or `flattenEach` and a FeatureCollection contains Features with `null` geometry, these should be excluded from the iterating process, correct?

stebogit:
> When we're using `geomEach` or `flattenEach` and a FeatureCollection contains Features with `null` geometry, these should be excluded from the iterating process

Good question @DenisCarriere... 🤔 
I would not exclude them, for we confirmed they are 100% valid GeoJSON features.
What if you actually need to iterate through them to actually check and populate the `geometry` field?

DenisCarriere:
> we confirmed they are 100% valid GeoJSON features.

Agreed they are valid GeoJSON features, however `geomEach` is meant to return valid GeometryObjects which `null` isn't since it doesn't have a `type` (that's the reason why we should exclude them).

We could return a `null` Feature for `flattenEach` since that is "valid".

🤔 still thinking how we should approach this one...

DenisCarriere:
🤔 `flattenEach` uses `geomEach`...

```js
function flattenEach(geojson, callback) {
    geomEach(geojson, function (geometry, index, properties) {
```
So the options are, we callback both `null` geometries, or we exclude both...

DenisCarriere:
@stebogit Chose to preserve `null` geometries for `flattenEach` & `geomEach`.

> What if you actually need to iterate through them to actually check and populate the geometry field?

👍 Good point.

DenisCarriere:
@stebogit Have a glance at this one last time, I'm 👍 to merge this.

packages/turf-helpers/index.js
```diff
@@ -248,6 +248,7 @@ function multiPolygon(coordinates, properties) {
  */
 function geometryCollection(geometries, properties) {
     if (!geometries) throw new Error('geometries is required');
```
stebogit: @DenisCarriere just being very fussy here, but from [specs](https://tools.ietf.org/html/rfc7946#section-3.1.8) geometryCollections can have an empty `geometries` array.
But `(!geometries)` would exclude that, right?

[Same goes](https://tools.ietf.org/html/rfc7946#section-3.3) actually for FeatureCollection, which I'm noticing has the same check.

packages/turf-meta/index.js
```diff
@@ -32,64 +32,73 @@ function coordEach(geojson, callback, excludeWrapCoord) {
         wrapShrink = 0,
         currentIndex = 0,
         isGeometryCollection,
-        isFeatureCollection = geojson.type === 'FeatureCollection',
-        isFeature = geojson.type === 'Feature',
+        type = (geojson) ? geojson.type : null,
```
stebogit: @DenisCarriere isn't geojson required here? I mean, shouldn't this actually throw an error if `!geojson`?
~~Then `type = geojson.type` any time, as `geojson.type` can be `null`~~
Sorry, `type` can NOT be `null`, it would be an invalid GeoJSON.

packages/turf-helpers/index.js
DenisCarriere: We can add those in the tests, but `!geometries` would not exclude empty geometries inside, it would only exclude `geometries: null` or `undefined.

lets add them to the tests and see what happens from there :) lol

packages/turf-meta/index.js
DenisCarriere: Well since a `Geometry` can be passed to `coordEach`, we need to handle `null` somehow.

Maybe its' best to add the `null` handling at the very top of the function.

```js
if (geojson === null) return;
```
That way the rest of the code can remain unchanged.

packages/turf-helpers/index.js
DenisCarriere: `(!geometries)` would not throw an error if you provide an empty Array.

```bash
$ node
> ![1,2,3]
false
> ![]
false
> !undefined
true
> !null
true
```

DenisCarriere: Added tests for this case: https://github.com/Turfjs/turf/pull/866/commits/c1edd39658d1f2dae611eb6106a4cb609aa9ab8d

stebogit: ah! 👍  sorry, PHP confusion! 😄

DenisCarriere: lol! Everything language has their quirks

PR #869, opened 2017-07-27
New module `@turf/boolean-equal`

stebogit:
Imported from [`turf-equal`](https://github.com/Turfjs/turf-equal).
Implements [DE-9IM](https://en.wikipedia.org/wiki/DE-9IM) "equal" [spatial relation](http://edndoc.esri.com/arcsde/9.0/general_topics/understand_spatial_relations.htm).

**TODOs:**
- [x] remove redundant coordinated from input features

Thanks @tmcw and @tcql for the original work! 👍


rowanwins:
Hey @stebogit 

Unfortunately this module isn't quite this simple. 

For example the following lines should pass
````
[[0, 0], [2, 2]]
and 
[[0, 0] [,1, 1], [2, 2]]
````
Shapely provides this helpful definition

> Returns True if the set-theoretic boundary, interior, and exterior of the object coincide with those of the other.

So points are easy, lines and polygons need  a bit more thought

**Comment by DenisCarriere** [comment]
@rowanwins 👍 Good catch, added that test in this module.

Shapely successfully passes (True) & this current library tests fail (False) on [`lines2.geojson`](https://gist.github.com/anonymous/b79147709cf09e2b129ecb2e0da1e750).

**Comment by stebogit** [comment]
> the following lines should pass
```js
[[0, 0], [2, 2]]
and 
[[0, 0] [,1, 1], [2, 2]]
```
@rowanwins @DenisCarriere the test now would pass if I could install the brand new `@turf/clean-coords` module; I tested it referencing the local module but I could not figure out how to install it from package.json. I tried `yarn`, `npm` and `lerna bootstrap`, but nothing. Can you please fix that?

BTW I had to [fix](https://github.com/Turfjs/turf/commit/0bb151fe12fdbd284312b1aece8a69fd0265e270) how `@turf/clean-coords` handled a single clean segment (like `[[0, 0], [2, 2]]`) cause it was failing. Hope it's fine.

packages/turf-boolean-equal/index.js
```diff
@@ -0,0 +1,34 @@
...
+ * turf.equal(pt2, pt3);
+ * //= false
+ */
+module.exports = function (feature1, feature2) {
+    // validation
+    if (!feature1) throw new Error('feature1 is required');
+    if (!feature2) throw new Error('feature2 is required');
+    var type1 = getGeomType(feature1);
+    var type2 = getGeomType(feature2);
+    if (type1 !== type2) throw new Error('features must be of the same type');
```
DenisCarriere: Why are we throwing an error, shouldn't this simply return `false`?

```js
if (type1 !== type2) return false;
```

rowanwins:
From my quick glance at the source and tests this doesn't support polys that have different starting vertices 
eg
````
[[0,0],[1,1],[1,0],[0,0]]
// should equal true
[[1,1],[1,0],[0,0][1,1]]
````
So we might need a bit more thought here.

packages/turf-boolean-equal/index.js
stebogit: ahahah! yes that's right @DenisCarriere! 😅

PR #871, opened 2017-07-27
modified CONTRIBUTING

stebogit:
Updated few lines.

@DenisCarriere Looking at the 'Code Style' paragraph though I was thinking wouldn't it be convenient to have a `lint` script (also for the @example) at the module level, maybe ran as `posttest`?
Since there is a global linting/testing to avoid unintentional external (from the module) changes, this would actually allow a faster check of your style/example before committing.


DenisCarriere:
> Updated few lines.

👍 Sweet, Looks good!

> wouldn't it be convenient to have a lint script (also for the @example) at the module level,

👎 Don't agree we should have any linting built-in to each module, the linting can easily be done at the root level and whenever someone sends a PR or a commit, the TravisCI automated builds will catch those errors if the user forgets to do `npm run lint`.

Maybe you can add an extra note in the [PULL_REQUEST_TEMPLATE.md](https://github.com/Turfjs/turf/blob/master/.github/PULL_REQUEST_TEMPLATE.md), however it already mentions about it the linting (might just need to expand on it).

DenisCarriere:
👍

PR #873, opened 2017-07-28
Doc updates

stebogit:
- [x] added missing function names
- [x] improved few @examples (they were quite slow in rendering on turf-www) 

@DenisCarriere I noticed the doc linting test we [added](https://github.com/Turfjs/turf/commit/edb624c99e9f9aff2ea5161c92e1161786ab2a6a) did not catch the missing `@name` in `clusters`, is it "normal"? Can/should we add such check?


stebogit:
I noticed also that if you run `packages/turf/test.js` the `documentations.js` test will only be applied to the modules listed in `index.js`. This way the new modules, which you can not add to `index.js` until you deploy the new version (right?), will not be tested.

Am I doing something wrong? Can we fix this?

DenisCarriere:
> `@name` in clusters, is it "normal"? Can/should we add such check?

This won't be caught in the linting because `@name` is optional, it will default to the function name. We could try to add this in the `/package/turf/test.js`.

> This way the new modules, which you can not add to index.js until you deploy the new version (right?), will not be tested.

Correct, those not listed in the `index.js` won't be tested, there was a reason for that... 🤔 ...

- [ ] add `@name` linting to tests
- [ ] investigate how to tackle linting new modules

Issue #874, opened 2017-07-29
New module proposal `@turf/clean-coords`


stebogit:
```js
/**
 * Removes redundant coordinates from a (Multi)LineString or (Multi)Polygon; 
 * ignores (Multi)Point.
 *
 * @name cleanCoords
 * @param {Geometry|Feature<any>} geojson Feature or Geometry
 * @param {boolean} [mutate=false] allows GeoJSON input to be mutated
 * @returns {Geometry|Feature<any>} the cleaned input Feature/Geometry
 * @example
 * var line = turf.lineString([[0, 0], [0, 2], [0, 5], [0, 8], [0, 8], [0, 10]]);
 *
 * turf.cleanCoords(line).geometry.coordinates;
 * //= [[0, 0], [0, 10]]
 */
module.exports = function (geojson, mutate) {
```

Don't know if these coordinates are actually redundant (as in useless), but in some cases are definitely in the way, e.g. #869 or the new `segmentEach`/`segmentReduce`.
We can have this as a module, handling the feature, or as a `helpers` function, simply returning the cleaned coordinates array.

Looking forward for any suggestion.

DenisCarriere:
Love it! Exactly what we needed! I'll dive into the code later on this weekend

stebogit:
Great! @DenisCarriere I'll work on the code in the weekend as well, for now was just a proposal to make sure it was acceptable

DenisCarriere:
I was going to propose the exact same thing at one point, sounds like a good module to help improve those boolean modules.

rowanwins:
My only comment @stebogit would be in relation to the name, im not sure that `clean-coords` is the best. My reading is that the module removes redundant vertices, I'd perhaps name it something more along those lines. IMHO cleaning implies other things might be done like trimming precision or removing coords outside a valid a range.

stebogit:
Implemented with #875

PR #875, opened 2017-07-30
New module `@turf/clean-coords` 

stebogit:
# New module `@turf/clean-coords`
Implements #874

```js
/**
 * Removes redundant coordinates from any GeoJSON Geometry.
 *
 * @name cleanCoords
 * @param {Geometry|Feature} geojson Feature or Geometry
 * @param {boolean} [mutate=false] allows GeoJSON input to be mutated
 * @returns {Geometry|Feature} the cleaned input Feature/Geometry
 * @example
 * var line = turf.lineString([[0, 0], [0, 2], [0, 5], [0, 8], [0, 8], [0, 10]]);
 * var multiPoint = turf.multiPoint([[0, 0], [0, 0], [2, 2]]);
 *
 * turf.cleanCoords(line).geometry.coordinates;
 * //= [[0, 0], [0, 10]]
 *
 * turf.cleanCoords(multiPoint).geometry.coordinates;
 * //= [[0, 0], [2, 2]]
 */
module.exports = function (geojson, mutate) {
```


DenisCarriere:
@stebogit Looking at this right now, I'll spend more time on it tomorrow, but would it be worth cleaning MultiPoint as well? points which have multiple points on top of each other should be cleaned?

packages/turf-clean-coords/index.js
```diff
@@ -0,0 +1,109 @@
+var clone = require('@turf/clone');
+var invariant = require('@turf/invariant');
+var getCoords = invariant.getCoords;
+var getGeomType = invariant.getGeomType;
+
+/**
+ * Removes redundant coordinates from a (Multi)LineString or (Multi)Polygon; ignores (Multi)Point.
+ *
+ * @name cleanCoords
+ * @param {Geometry|Feature<any>} geojson Feature or Geometry
```
DenisCarriere: I've been dropping the `<any>` in the recent JSDocs.
`{Geometry|Feature<any>}` => `{Geometry|Feature}`

packages/turf-clean-coords/index.js
```diff
@@ -0,0 +1,109 @@
...
+        break;
+    case 'Point':
+    case 'MultiPoint':
+        newPoints = coords;
+        break;
+    default:
+        throw new Error(type + ' geometry not supported');
+    }
+
+    var output = (mutate === true) ? geojson : clone(geojson, true);
```
DenisCarriere: It's a bit of "black magic" how clone works, but you don't need to define a new variable if you apply `clone` (aka. JSON.parse + stringify).

This below is valid:
```js
if (mutate !== true) geojson = clone(geojson, false);
```

DenisCarriere: 2nd - Since we aren't mutating `properties` we don't need to define `clone(geojson, true)` as TRUE, if we use the default or `FALSE` it will only clone the coordinates (which is a lot faster process).

stebogit:
@DenisCarriere we can definitely clean `MultiPoint`; at this point we can allow `Points` for completeness and just ignoring them.
I think a `MultiPoint` with overlapping `points` it's not necessarily wrong (or "dirt" 😄 ), but if the user is using this module on `MultiPoint` I guess it means he wants to clean it! 👍

packages/turf-clean-coords/index.js
```diff
@@ -0,0 +1,109 @@
+var clone = require('@turf/clone');
+var invariant = require('@turf/invariant');
+var getCoords = invariant.getCoords;
+var getGeomType = invariant.getGeomType;
+
+/**
+ * Removes redundant coordinates from a (Multi)LineString or (Multi)Polygon; ignores (Multi)Point.
```
DenisCarriere: I've added MultiPoint support, we can simply say this now:

```
* Removes redundant coordinates from any GeoJSON Geometry.
```

packages/turf-clean-coords/index.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="geojson" />
+
+type Feature = GeoJSON.Feature<any> | GeoJSON.GeometryObject;
+
+/**
+ * http://turfjs.org/docs/#cleancoords
+ */
+declare function cleanCoords(feature: Feature, mutate?: boolean): Feature;
```
DenisCarriere: @stebogit This one gets a bit tricky, one thing to keep in mind is trying to preserve the Input/Output type as much as possible.

At the moment, this definition is saying the input is any Feature or Geometry and it outputs either Geometry or a Feature.

Using the `extends` you validate the input, however the output will be the same as what the user enters, that way it preserves their Types.

```typescript
type GeometryObject = GeoJSON.GeometryObject
type Feature = GeoJSON.Feature<any>

/**
 * http://turfjs.org/docs/#cleancoords
 */
declare function cleanCoords<T extends GeometryObject|Feature>(feature: T, mutate?: boolean): T;
declare namespace cleanCoords {}
export = cleanCoords;
```

packages/turf-clean-coords/bench.js
```diff
@@ -0,0 +1,40 @@
...
+ * triangle x 93,747 ops/sec ±3.09% (76 runs sampled)
+ */
+const suite = new Benchmark.Suite('turf-clean-coords');
+glob.sync(path.join(__dirname, 'test', 'in', '*.geojson')).forEach(filepath => {
+    const {name} = path.parse(filepath);
+    const geojson = load.sync(filepath);
+    console.time(name);
+    cleanCoords(geojson);
+    console.timeEnd(name);
+    suite.add(name, () => cleanCoords(geojson));
```
DenisCarriere: For internal benchmark tests, we can add `mutate=true`, that way we get the "real" benchmarks results.

```
 * geometry x 3,053,905 ops/sec ±1.84% (87 runs sampled)
 * multiline x 3,943,840 ops/sec ±2.60% (85 runs sampled)
 * multipoint x 465,553 ops/sec ±1.04% (89 runs sampled)
 * multipolygon x 2,046,659 ops/sec ±0.77% (91 runs sampled)
 * point x 22,318,913 ops/sec ±1.20% (86 runs sampled)
 * polygon-with-hole x 3,109,098 ops/sec ±2.10% (91 runs sampled)
 * polygon x 4,603,124 ops/sec ±1.99% (84 runs sampled)
 * simple-line x 7,732,876 ops/sec ±1.18% (86 runs sampled)
 * triangle x 4,950,167 ops/sec ±1.36% (89 runs sampled)
```

DenisCarriere:
> but if the user is using this module on MultiPoint I guess it means he wants to clean it! 

Exactly.. :)

packages/turf-clean-coords/index.js
stebogit: @DenisCarriere that's right! Now that I looked better at it, `@turf/clone` does already clone (if defined) all the "standard" GeoJSON attributes (i.e. `properties`, `id` and `bbox`); that's what I was aiming for. Very well 👍

packages/turf-clean-coords/index.js
```diff
@@ -42,23 +43,31 @@ module.exports = function (geojson, mutate) {
...
         }
         break;
     case 'Point':
+        return geojson;
     case 'MultiPoint':
-        newPoints = coords;
+        var existing = {};
+        for (var i = 0; i < coords.length; i++) {
+            var point = coords[i];
+            var key = point.join('-');
```
stebogit: @DenisCarriere this is really smart! 🤓  👍

stebogit:
@DenisCarriere now that I cleaned the horrible for-loops I used initially I feel better. I can't stand them! 😄 
Good for me to merge this.

@rowanwins do you have any suggestion for an [alternative name](https://github.com/Turfjs/turf/issues/874#issuecomment-318889446), if you feel it should be different?
I am personally fine with this (of course 😄 ).

DenisCarriere:
@stebogit 👍 I'm on the same page with you for those for loops, there's some slight performance decrease (~8-10%) when using `forEach`, but it's a lot easier to read!

packages/turf-clean-coords/index.js
DenisCarriere: Thanks 🤓 !

rowanwins:
Happy to go with group consensus on the name - thoughts @DenisCarriere ? See [history](https://github.com/Turfjs/turf/issues/874#issuecomment-318889446)

DenisCarriere:
@stebogit Let me know what you think about this, I dropped `clone` entirely so that the `cleanCoords` module doesn't mutate input by adding `geometry` & `feature` methods.

> will need to update `@turf/helpers` to support `bbox` & `id` properties, also a `geometry` method would be useful.

DenisCarriere:
> @rowanwins Happy to go with group consensus on the name

I'm good with this name, struggling to find a better name for this.

stebogit:
@DenisCarriere 
> I dropped clone entirely

Why? is't this the perfect situation for using this module? It is basically exactly using `geometry` & `feature` helpers.

> so that the `cleanCoords` module doesn't mutate the input

Why? I would leave this option to the user. We have this useful option in quite a few other modules, like `truncate`, `transform`, `kmeans` and other

DenisCarriere:
@stebogit 
Well unlike `truncate` or those other modules, we aren't using `coordEach` which is impossible to prevent input mutation without using `clone`. Since we've already deconstructed the entire GeoJSON geometry we can easily put it back together with little performance loss (0.25x slower).

Reason why I changed it to `geometry/feature` is purely for performance since `mutate` will most likely always be defined as `false` (which will be the slowest option).

I ran some benchmarks using the different techniques.

## Considerations
[Foreign members](https://tools.ietf.org/html/rfc7946#section-6.1) are dropped using the `geometry`/`feature` approach, however if we are enforcing a "strict" GeoJSON implementation that might be a good thing, also this might be considered a positive +1 for this "clean" module.

👍 It's a good idea to keep the `mutate` options that way it allows the user to preserve foreign members and/or wants the input to be mutated.

## Summary
- `mutate=true` (0x - speed factors are compared to these benchmark results)
- `mutate=false`
  - using `geometry`/`feature` (0.25x slower)
  - using `@turf/clone` (8x slower)
  - using `JSON.parse` + `JSON.stringify` (16x slower)

```js
 * // mutate=false (using geometry/feature)
 * geometry x 1,524,640 ops/sec ±4.60% (74 runs sampled)
 * multiline x 1,511,608 ops/sec ±8.79% (72 runs sampled)
 * multipoint x 382,429 ops/sec ±3.56% (84 runs sampled)
 * multipolygon x 808,277 ops/sec ±2.84% (82 runs sampled)
 * point x 14,675,464 ops/sec ±4.42% (80 runs sampled)
 * polygon-with-hole x 1,493,507 ops/sec ±5.53% (72 runs sampled)
 * polygon x 2,386,278 ops/sec ±1.27% (86 runs sampled)
 * simple-line x 4,195,499 ops/sec ±2.88% (86 runs sampled)
 * triangle x 2,254,753 ops/sec ±1.10% (88 runs sampled)
 *
 * // mutate=false (using @turf/clone)
 * geometry x 202,410 ops/sec ±1.43% (88 runs sampled)
 * multiline x 235,421 ops/sec ±3.48% (86 runs sampled)
 * multipoint x 280,757 ops/sec ±1.59% (87 runs sampled)
 * multipolygon x 127,353 ops/sec ±1.35% (88 runs sampled)
 * point x 18,233,987 ops/sec ±1.34% (86 runs sampled)
 * polygon-with-hole x 199,203 ops/sec ±2.61% (84 runs sampled)
 * polygon x 355,616 ops/sec ±1.58% (86 runs sampled)
 * simple-line x 515,430 ops/sec ±2.40% (83 runs sampled)
 * triangle x 286,315 ops/sec ±1.64% (86 runs sampled)
 *
 * // mutate=false (using JSON.parse + JSON.stringify)
 * geometry x 93,681 ops/sec ±7.66% (75 runs sampled)
 * multiline x 112,836 ops/sec ±4.60% (82 runs sampled)
 * multipoint x 113,937 ops/sec ±1.09% (90 runs sampled)
 * multipolygon x 71,131 ops/sec ±1.32% (90 runs sampled)
 * point x 18,181,612 ops/sec ±1.36% (91 runs sampled)
 * polygon-with-hole x 100,011 ops/sec ±1.14% (85 runs sampled)
 * polygon x 154,331 ops/sec ±1.31% (89 runs sampled)
 * simple-line x 193,304 ops/sec ±1.33% (90 runs sampled)
 * triangle x 130,921 ops/sec ±3.37% (87 runs sampled)
 *
 * // mutate=true
 * geometry x 2,016,365 ops/sec ±1.83% (85 runs sampled)
 * multiline x 2,266,787 ops/sec ±3.69% (79 runs sampled)
 * multipoint x 411,246 ops/sec ±0.81% (89 runs sampled)
 * multipolygon x 1,011,846 ops/sec ±1.34% (85 runs sampled)
 * point x 17,635,961 ops/sec ±1.47% (89 runs sampled)
 * polygon-with-hole x 2,110,166 ops/sec ±1.59% (89 runs sampled)
 * polygon x 2,887,298 ops/sec ±1.75% (86 runs sampled)
 * simple-line x 7,109,682 ops/sec ±1.52% (87 runs sampled)
 * triangle x 3,116,940 ops/sec ±0.71% (87 runs sampled)
```

stebogit:
Great! I can't wait to use this in `@turf/boolean-equal` @DenisCarriere 😃

DenisCarriere:
Now all I need to do is find some time to push out a new minor release.

PR #882, opened 2017-08-02
Updates to 4.6.0-draft.md

stebogit:
@rowanwins @DenisCarriere please have a look, I'm sure I've missed something from [here](https://github.com/Turfjs/turf/milestone/10?closed=1&page=1)


DenisCarriere:
👍 Sweet!! 😍 So glad you do this!
I'll have a closer look at this.

One day we can automate this step using clever Titles & GitHub Tags, but the release notes will seem too "robotic" if we do it that way.

PR #886, opened 2017-08-02
Fix bbox-clip point intersection handling

tjdett:
In cases where only a single point touches the bounding box, it's possible for `lineclip` to return a line with multiple identical coordinates. For some cases it emits 4 coordinates, which is a zero-area polygon, and everything works fine. For the provided test case, it emits 3 coordinates.

This patch just sanity-checks the `lineclip` output so it only attempts to output polygons with at least 4 coordinates.


DenisCarriere:
Thanks Tim for sending over a "bug" & a solution (PR) at the same time 👍

PR #890, opened 2017-08-04
Fixes #820 - `@turf/intersect`

### LINKED ISSUE #820: turf.intersect raises "uncaught exception: [object Object]" on valid geometries
To reproduce this error message (which I believe should not be raised), just run this code:

`turf.intersect({"type": "Feature", "geometry":{"type":"Polygon","coordinates": [[[4869493.170354112,7598816.20365591],[4869512.279611183,7595796.941038646],[4873926.5179946525,7596007.14286643],[4873716.316166868,7598013.614858915],[4871671.62566024,7598758.875884696],[4869493.170354112,7598816.20365591]]]}},{"type": "Feature", "geometry":{"type":"Polygon","coordinates":[[[4894725.4051467,7624972.8484957],[4894725.4051469,7624972.8530041],[4894614.9242472,7624825.5926316],[4894398.2837739,7624987.6762962],[4894450.8149275,7625056.9040792],[4894668.7026646,7625096.7545718],[4894671.4728847,7625081.2756834],[4894674.9876194,7625066.9804281],[4894678.7586548,7625055.2998143],[4894682.9867902,7625043.5046946],[4894688.1291364,7625031.5950619],[4894694.8416008,7625017.8373286],[4894699.3280398,7625009.9516021],[4894705.0417681,7625000.4468024],[4894710.069852,7624992.8887655],[4894715.3264873,7624985.6742709],[4894721.1517534,7624977.9994939],[4894725.4051467,7624972.8484957]]]}})`

The whole problem is with the second feature and I'm not sure what may be wrong with it.

stebogit:
- [x] added truncate to inputs
- [x] added originally breaking test case
- [x] return `Feature<null>` if no geometry
- [x] remove extra `JSON.stringify`

Ref #820


packages/turf-intersect/test.js
```diff
@@ -3,6 +3,7 @@ const fs = require('fs');
 const test = require('tape');
 const load = require('load-json-file');
 const write = require('write-json-file');
+var truncate = require('@turf/truncate');
```
DenisCarriere: poor lonely `var`... 😢  he feels old next to all his younger `const` siblings.

packages/turf-intersect/index.js
```diff
@@ -34,22 +35,17 @@ var jsts = require('jsts');
...
-    if (poly2.type === 'Feature') geom2 = poly2.geometry;
-    else geom2 = poly2;
+    var geom1 = (poly1.type === 'Feature') ? poly1.geometry : poly1;
+    var geom2 = (poly2.type === 'Feature') ? poly2.geometry : poly2;
+
     var reader = new jsts.io.GeoJSONReader();
-    var a = reader.read(JSON.stringify(geom1));
-    var b = reader.read(JSON.stringify(geom2));
+    var a = reader.read(JSON.stringify(truncate(geom1)));
+    var b = reader.read(JSON.stringify(truncate(geom2)));
```
DenisCarriere: Since `truncate` does the `clone` operation, we can drop `JSON.stringify` from the mix.

packages/turf-intersect/index.js
```diff
@@ -34,22 +35,17 @@ var jsts = require('jsts');
...
     var intersection = a.intersection(b);
 
-    if (intersection.isEmpty()) {
-        return undefined;
-    }
+    if (intersection.isEmpty()) return null;
```
DenisCarriere: Weren't we going to return a `helpers.feature<null>` instead of `null`?

stebogit: Sorry I didn't noticed your suggestion in the issue comment.

DenisCarriere: No worries 👍 

Well I think we are all in agreement that if you have no results in collection of Features, then we would have an empty FeatureCollection, if the output is a single Feature which would be `undefined` or `null` then we would output a `Feature<null>` that way things don't blow up when you chain TurfJS methods by sending a `null` as input to another module.

DenisCarriere:
I like the new `bug-fix` label 😄

dpieri:
Shouldn't this have bumped the major version as it changed the API for turf.union? Any code that interpreted an undefined responses to mean there is no union breaks on this change.

DenisCarriere:
This fix actually got reverted, PR https://github.com/Turfjs/turf/pull/952 & Issue https://github.com/Turfjs/turf/issues/951

Issue #852, opened 2017-07-20
lineSplit does not split a lineString if splitter is very close to a vertex point


grundid:
The following test returns a feature collection with only one element.
I would expect two elements.

```javascript
const lineString = turf.lineString([[9.2202022, 49.1438226], [9.2199531, 49.1439048], [9.2196177, 49.1440264]]);
const resultFc = turf.lineSplit(lineString, turf.point([9.219953175634146, 49.14390480825327]));
t.equals(resultFc.features.length, 2);
```

When I reduce the precision of the splitter the result is correct:
```javascript
const lineString = turf.lineString([[9.2202022, 49.1438226], [9.2199531, 49.1439048], [9.2196177, 49.1440264]]);
const resultFc = turf.lineSplit(lineString, turf.point([9.2199531, 49.1439048]));
t.equals(resultFc.features.length, 2);
```

stebogit:
@grundid it seems to be a matter of decimals in the splitter (`point`) coordinates:
```js
const lineString = turf.lineString([[9.2202022, 49.1438226], [9.2199531, 49.1439048], [9.2196177, 49.1440264]]);
const resultFc = turf.lineSplit(lineString, turf.point([9.219953, 49.143905])); // coordinates truncated to 6 decimals
//={ "type": "FeatureCollection", "features": [ { "type": "Feature", "properties": {}, "geometry": { "type": "LineString", "coordinates": [ [ 9.2202022, 49.1438226 ], [ 9.2199531, 49.1439048 ], [ 9.219953, 49.143905 ] ] } }, { "type": "Feature", "properties": {}, "geometry": { "type": "LineString", "coordinates": [ [ 9.219953, 49.143905 ], [ 9.2196177, 49.1440264 ] ] } } ] }
```

Turf follows GeoJSON [recommendations](https://tools.ietf.org/html/rfc7946#section-11.2) by using 6 decimals to define coordinates.

@DenisCarriere we might want to `truncate` the coords of the splitter in`@turf/line-split` to avoid this issue?

grundid:
If Turf follows the GeoJSON recommendations then maybe turf.point should not create a GeoJSON object with to many decimals. ;)

DenisCarriere:
> then maybe turf.point should not create a GeoJSON object with to many decimals
Might not be a good idea to add `truncate` to `turf.point` since that would add unnecessary overhead if it's not required.

>  we might want to truncate the coords of the splitter in@turf/line-split to avoid this issue?

[...]

PR #892, opened 2017-08-07
added test cases for issue #852

grundid:
The tests fail as expected.


DenisCarriere:
Issue: https://github.com/Turfjs/turf/issues/852
image

stebogit:
Ooops I just wanted to add [my changes](https://github.com/Turfjs/turf/compare/line-split-fix) to @grundid's ones, but that (of course! 😞 ) actually merged his branch to mine. Sorry, I guess I should have merged with `--no-ff`...
This PR is still active, but the branch it refers doesn't exist anymore. Sorry again.
@DenisCarriere how can I fix this?

Anyway, I noticed increasing the precision of `truncate` to 7 or 8 (still above GeoJSON recommendations) solves 3 of the 4 new tests. The one still failing is the `precision end point`

DenisCarriere:
I would pull down his branch and make the changes from there, don't know of an easy way to do that, it's only 2 files so I would end up doing it manually.

DenisCarriere:
Not too sure if changing the decimal points of truncate would be the solution here, I think better handling of the exact start & end points would be a better approach.

Seems like it's having difficulties with exact coordinates, not too sure if we can exclude points from generating lines which are less then 10cm (0.000001 degrees) apart.

DenisCarriere:
@stebogit That's what happens when you reply to GitHub issues on the bus 😆

DenisCarriere:
@stebogit Add your commit to this branch

stebogit:
🤔  hmmm... so I need to clone `geohub-plattform` repo and commit to `master`, right?

DenisCarriere:
That's correct, ideally we would create a branch on TurfJS, that way it's a bit easier to switch between branches.

stebogit:
Ok, it should be fixed now. 
@grundid could you please check this?

DenisCarriere:
@grundid Just published a patch release for `@turf/line-split@4.6.1` with these fixes. That way it's a bit easier to test on your side.

Issue #895, opened 2017-08-10
transformScale doesn't always create similar shapes


DingoEatingFuzz:
Triangles become scalene and hexagons go concave.

I discovered this while working in Glitch. Here's the project link: https://glitch.com/edit/#!/nosy-flare?path=map.js:96:8

![image](https://user-images.githubusercontent.com/174740/29153333-d7b0e0a8-7d41-11e7-972b-0204ffacbf15.png)

rowanwins:
Thanks for the report @DingoEatingFuzz 

This is one for you @stebogit :)

DenisCarriere:
Very interesting... 🤔

stebogit:
@DingoEatingFuzz the issue has something to do with the `mutate` parameter, I'll dig more into it in the weekend.
For now you can create a new scaled grid:
```js
var scaledGrid = turf.featureCollection([]);
scaledGrid.features = processedGrid.features.map(f => {
  var factors = f.properties.list.map(ms => (ms - minDate) / dateDiff);
  var avgFactor = factors.reduce((sum, n) => sum + n, 0) / factors.length;
  return turf.transformScale(f, avgFactor * (maxFactor - minFactor) + minFactor);
})

map.getSource('grid').setData(scaledGrid);
```

![screen shot 2017-08-11 at 1 37 07 am](https://user-images.githubusercontent.com/12717225/29206182-a012270c-7e35-11e7-8311-7025805b6b2a.png)

DenisCarriere:
👍 @stebogit I'll look into this, I might know where the issue is on this one. Also we can add `@turf/clone` to the mix.

DenisCarriere:
Strange one... 🤔 

I've been able to get the same "expected" results as @stebogit using a slightly different approach.

```js
processedGrid.features = processedGrid.features.map(feature => {
  var factors = feature.properties.list.map(ms => (ms - minDate) / dateDiff);
  var avgFactor = factors.reduce((sum, n) => sum + n, 0) / factors.length;
  var factor = avgFactor * (maxFactor - minFactor) + minFactor;
  return turf.transformScale(feature, factor, 'centroid')
})
```

@stebogit This is the type of "strange" unexpected mutations that @morganherlocker has been advocating, hence why all of the early TurfJS modules don't have Mutate as param since it's hard to control the expected outputs.

I initially added the `mutate` param for testing purposes.

stebogit:
[...]

PR #897, opened 2017-08-11
Add FeatureCollection mutate tests (not failing)

DenisCarriere:
# Add test cases for `@turf/transform-scale`

- [x] Adding failing tests for https://github.com/Turfjs/turf/issues/895
- [x] Test mutating FeatureCollection (not the issue)
- [x] Fix issue

## Issue

On (Multi)Polygon the first & last coordinate gets repeated when using `coordEach` (not sure if this was intentional in `coordEach`).

## Before

image

## After

image


packages/turf-transform-scale/index.js
```diff
@@ -59,18 +61,24 @@ module.exports = function (geojson, factor, origin, mutate) {
...
     origin = defineOrigin(geojson, origin);
 
     // Shortcut no-scaling
     if (factor === 1 || isPoint) return geojson;
 
     // Scale each coordinate
-    coordEach(geojson, function (coord) {
+    coordEach(geojson, function (coord, coordIndex, featureIndex, featureSubIndex) {
+        // Ignore scaling on first coordinate of Polygons (Issue #895)
+        if (isPolygon && mutate === true && featureSubIndex === 0) return;
```
DenisCarriere: @stebogit Might want to add this type of behavior in a few of the `transform` modules.

packages/turf-transform-scale/index.js
```diff
@@ -59,18 +61,24 @@ module.exports = function (geojson, factor, origin, mutate) {
...
+    var isPoint = geomType === 'Point';
+    var isPolygon = geomType === 'Polygon' || geomType === 'MultiPolygon';
     origin = defineOrigin(geojson, origin);
 
     // Shortcut no-scaling
     if (factor === 1 || isPoint) return geojson;
 
     // Scale each coordinate
-    coordEach(geojson, function (coord) {
+    coordEach(geojson, function (coord, coordIndex, featureIndex, featureSubIndex) {
```
stebogit: I noticed `coordIndex` and `featureSubIndex` are always equals, throughout the entire test suite, and `featureIndex` is always `0`. 
Inside coordEach this never logs:
```js
if (coordIndex !== featureSubIndex || featureIndex > 0) console.log('Woah!');
```
Shouldn't the indices change at least for multi features?

packages/turf-transform-scale/index.js
stebogit: @DenisCarriere it looks ok, and clearly it does work. However I can not understand why we don't want to scale all the points of the ring(s) of a polygon, whether we clone it or not.
I really can't wrap my head around it... 🤔

packages/turf-transform-scale/index.js
```diff
@@ -59,18 +61,24 @@ module.exports = function (geojson, factor, origin, mutate) {
  * @param {Feature|Geometry} geojson GeoJSON Feature/Geometry
  * @param {number} factor of scaling, positive or negative values greater than 0
  * @param {string|Geometry|Feature<Point>|Array<number>} [origin="centroid"] Point from which the scaling will occur (string options: sw/se/nw/ne/center/centroid)
+ * @param {boolean} [mutate=false] allows GeoJSON input to be mutated
```
stebogit: Here `mutate` does not actually define if the feature is going to be mutated or not, that was handled before.
I can't find a good description though 😅

stebogit:
@DenisCarriere I found out the real issue: as you [said](https://github.com/Turfjs/turf/issues/895#issuecomment-321849798)
> The scaling happens twice on the first & last coordinate position (for some reason...)

because when `@turf/hex-grid` created the polygons the last vertex of the ring was set as reference to the first one, instead of cloning it.
screen shot 2017-08-13 at 1 58 58 am

I modified (locally) that line and all seems to be good with the original `transform-scale` code (but of course this branch still fails until we update `@turf/hex-grid`).

DenisCarriere:
> because when @turf/hex-grid created the polygons the last vertex of the ring was set as reference to the first one, instead of cloning it.

Ahh... that makes sense! Was wondering where that issue came from.

packages/turf-transform-scale/index.js
```diff
@@ -46,39 +46,33 @@ module.exports = function (geojson, factor, origin, mutate) {
...
-    return scale(geojson, factor, origin, mutate);
+    return scale(geojson, factor, origin);
 };
 
 /**
  * Scale Feature/Geometry
  *
  * @private
- * @param {Feature|Geometry} geojson GeoJSON Feature/Geometry
+ * @param {GeoJSON} geojson GeoJSON Feature/Geometry
```
DenisCarriere: GeoJSON means it handles FeatureCollection & GeometryCollection.

This doesn't?

packages/turf-transform-scale/index.js
```diff
@@ -56,14 +58,14 @@ module.exports = function (geojson, factor, origin, mutate) {
...
  * @param {number} factor of scaling, positive or negative values greater than 0
  * @param {string|Geometry|Feature<Point>|Array<number>} [origin="centroid"] Point from which the scaling will occur (string options: sw/se/nw/ne/center/centroid)
- * @returns {Feature|Geometry} scaled GeoJSON Feature/Geometry
+ * @returns {GeoJSON} scaled GeoJSON object
```
DenisCarriere: If the input is Geometry it outputs a scaled Geometry, we know the outputs/inputs can ONLY be Feature or Geometry.

Issue #819, opened 2017-06-27
turf.polygonize not creating polygons from grid of multilinestring lines


maphew:
_turf.polygonize_ isn't creating polygons from this input:

- https://gist.github.com/maphew/7bb77059574ea714977edae41e4b114c#file-w3w-grid-shwatka-4d-geojson
- Expected result similar to this output from 'Qgis >> Vector geomery tools >> polygonize': https://gist.github.com/maphew/7bb77059574ea714977edae41e4b114c#file-qgis-vector-polygonize-output-geojson

Code snippet:

    var grid = {"coordinates":[[[-135.025,6 ...snip.... 4092,60.6726],"type":"MultiLineString","properties"};
    var polygon = turf.polygonize(grid);

Full test case - https://jsfiddle.net/87cq6edb/

DenisCarriere:
@NickCis Can you have a look at this? You are the one most "qualified" to answer/fix this.

NickCis:
Current implementation of `polygonize` makes the following assumption (copying from docs):

> Edges must be correctly noded, i.e., they must only meet at their endpoints.

As far as i understand, in this test case the `lineString`s are not _correctly noded_, ie: they intersect between them, that's why polygonize isn't generating any polygons.

I must highlight that (according to the docs) this is the library's expected behavior.

At the moment, the user has to look for all the intersections and make the `lineString`s divitions. It was [the problem described at the end of the PR](https://github.com/Turfjs/turf/pull/767#issuecomment-305532903).

maphew:
Oh sorry. I read "I ended up porting this to javascript because I wanted the same functionality as Shapely's polygonize function" in the PR intro and assumed there was feature parity with Shapely (because that's what I wanted to believe ;-).

NickCis:
As far as i know Shapely is calling the GEOS's polygonizer class. This is a javascript port of that c++ class. I think that with Shapely you'll get the same problem.

If that's not the case, please tell me so (if possible, inform the Shapely version used and a working example). i could try to port that functionality.

DenisCarriere:
We can always attempt to port over Shapely's `polygonize` function to NodeJS and include it in the test cases.

We've done this so far with some of the `@turf/boolean` modules.

This test process is a bit unconventional, but if you want to check [`boolean-shapely`](https://github.com/DenisCarriere/boolean-shapely) (should only be used for testing purposes).

maphew:
[...]

PR #898, opened 2017-08-13
Adds doc note on `@turf/polygonize`

stebogit:
Ref. #819

cc: @NickCis @DenisCarriere

### DIFF
--- a/packages/turf-polygonize/README.md
+++ b/packages/turf-polygonize/README.md
@@ -6,8 +6,10 @@ Polygonizes [(Multi)LineString(s)](http://geojson.org/geojson-spec.html#linestri
 
 Implementation of GEOSPolygonize function (`geos::operation::polygonize::Polygonizer`).
 
-Polygonizes a set of lines that represents edges in a planar graph. Edges must be correctly
-noded, i.e., they must only meet at their endpoints.
+Polygonizes a set of lines that represents edges in a planar graph. Edges must be correctly noded, 
+i.e. they must only meet at their endpoints. (NOTE: while TurfJS has no specific function, an easy way 
+to ensure this is to use the [QGis](https://github.com/qgis/QGIS) library and call [`unaryUnion()`](https://github.com/qgis/QGIS/blob/master/src/core/geometry/qgsgeometry.cpp#L2057) 
+on the set of input geometries and then pass the result to `turf-polygonize`)
 
 The implementation correctly handles:
 
--- a/packages/turf-polygonize/index.js
+++ b/packages/turf-polygonize/index.js
@@ -6,7 +6,9 @@ var polygonize = require('polygonize');
  * Implementation of GEOSPolygonize function (`geos::operation::polygonize::Polygonizer`).
  *
  * Polygonizes a set of lines that represents edges in a planar graph. Edges must be correctly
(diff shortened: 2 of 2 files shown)


NickCis:
I'm currently working on an implementation of [unary union](https://github.com/nickcis/unary-union). Perhaps it's better to wait until I finish it in order to update the docs with a real solution.

stebogit:
@NickCis do you think you're going to have it done fairly soon? My goal was basically to close #819 to keep the list as short as possible.
We can easily modify again the docs whenever your module is done.

BTW were you thinking to add [`unary-union`](https://github.com/NickCis/unary-union) to `@turf/polygonize` to avoid the issue entirely?

DenisCarriere:
@NickCis Nice! Looks like you are doing some good progress on `unary-union`, looking forward to see everything 100% tested and good to go.

Feel free to create a TurfJS wrapper for your module, `@turf/unary-union` which we could then use internally for `polygonize`.

I'm happy to see that you're `dist` is a babel ES5 output 👍

DenisCarriere:
> @stebogit My goal was basically to close #819

It's good to update the docs, however this still wouldn't solve the issue.

I really like the idea of the `unary-union` module, I think it will be used a lot! (well I am personally going to use it! 😄)

DenisCarriere:
> BTW were you thinking to add unary-union to @turf/polygonize to avoid the issue entirely?

I'm pretty sure that would be the intent... I would suggest we do that 👍 .

stebogit:
> It's good to update the docs, however this still wouldn't solve the issue.

Isn't the current issue just a lack of info on a (currently) missing step to avoid the unexpected (but [not wrong](https://github.com/Turfjs/turf/issues/819#issuecomment-311839892) apparently) result? That's why the `docs` label.
Otherwise we might label it as `bug` and keep it open until `unary-union` is available.
However I would not consider this a bug.

NickCis:
From a developer point of view, `polygonize` works as expected, the docs clearly say:

> Polygonizes a set of lines that represents edges in a planar graph. **Edges must be correctly noded, i.e., they must only meet at their endpoints**.

But, from an user's perspective `polygonize` lacks of `unary-union` functionality.

I would suggest to keep the issue open until i finish the `unary-union` implementation. I can't give a estimated time since i'm only working on it on my free time.

DenisCarriere:
@NickCis Big thanks for those `unary-union` contributions. 👍 

@stebogit We can merge these `docs` notes and whenever we update `polygonize` with `unary-union` we can re-update the docs accordingly.

rowanwins:
Gday @NickCis

Nice job on the unary union stuff, that would certainly be super useful for our dissolve module. 

As an FYI we've also been trying to provide a bit of support to the `martinez` (repo)[https://github.com/w8r/martinez] by @w8r which does the full range of clip, difference, union etc. Not sure if you've seen that but with checking out.

And on a side note it would be great if someone could publish a sweepline module to stop people from having to reinvent that wheel as well, I'm sure it would be a handy module for a range of things.

w8r:
> And on a side note it would be great if someone could publish a sweepline module to stop people from having to reinvent that wheel as well, I'm sure it would be a handy module for a range of things.

@ggolikov is working on that right now, btw

Issue #333, opened 2016-03-15
turf-concave throws an exception with a group of GPS points


rnisthal:
Hi,

https://jsfiddle.net/rnisthal/7tsygLrs/

I have some stops from a GPS and I want to run concave on these points.
Why is this dataset causing an exception, and how can I prevent similar ones?

Thanks!

tmcw:
Specific error is

![2016-03-15 at 5 01 pm](https://cloud.githubusercontent.com/assets/32314/13793882/89037a7a-eacf-11e5-9666-5a9d1d3a4dd9.png)

I think we should hit #310 which should fix this issue as well.

rnisthal:
I like the output of the current method better than concaveman (it's better suited for my needs).
Maybe turf could provide both algorithms, the current implementation and concaveman?

To avoid the exception I first have to deduplicate the array of points.

https://jsfiddle.net/rnisthal/7tsygLrs/8/

For me the precision granted by Open Location Code is enough to fix my problem, at least with the test set. ;)
Any ideas for a better deduplication method?
In any case, maybe the deduplication should be made inside turf?

Thanks!

PS.
What I really need, which is ridiculously slow:

var p = [];

for (i = 0; i < points.features.length; i++) {
  var p1 = turf.buffer(points.features[i], 0.25, 'kilometers'));
  p.push(p1.features[0]);
}

var fc = turf.merge(turf.featurecollection(p));

1. Set a buffer of radius x in every point.
2. Merge all the buffers.
That would give me a multipolygon.
Group Points into each multipolygon
Concave hull of each group.

Is there an efficient method for this?

tmcw:
> I like the output of the current method better than concaveman (it's better suited for my needs).

Can you be much more specific? How is it better suited to what needs.

rnisthal:
The points represent the places a salesperson visits on a week, or any timeframe really. If you group all the stops several salespersons visit, then you have the region that group covers.
Then you can compare the region one group covers against the region other groups cover, and see if they overlap, for example.

turf.concave gives for a set of points several polygons. This would happen if you have two regions covered by the same sales group. If you define a maxEdge that is enough to group some stops together but not all.

concaveman groups everything together, at least I could not make it break into different regions for one set of points. It gives one outline even if there are edges longer than the maxEdge you give.

Sorry for my english.

Returning to turf.concave

it gave me a lot of Topology Errors, mostly like

JavaScript runtime error: TopologyError: found non-noded intersection between 
[...]

Issue #310, opened 2016-02-17
turf-concave should use concaveman


morganherlocker:
https://github.com/mapbox/concaveman

concaveman is drastically faster and more robust than the current implementation.

tmcw:
Downstream: https://github.com/Turfjs/turf-concave/issues/11

tmcw:
done: https://github.com/Turfjs/turf/blob/master/packages/turf-concave/index.js

PR #899, opened 2017-08-14
`@turf/concave` refactoring

stebogit:
I was working to fixing #333, but it seems it's not actually an issue anymore.
The [original jsfiddle](https://jsfiddle.net/rnisthal/7tsygLrs/) was using to Turf v2.0.
Now, at least after the refactoring, the error does not occur.
@DenisCarriere the [case](https://github.com/Turfjs/turf/issues/333#issuecomment-246165916) you added currently does not calculate a polygon because `maxEdge` is too small; so it now errors out with that message.

While refactoring the module I noticed two things:
- `@turf/concave` is currently using `@turf/union`, thus it means it's based on `jsts` 👎 
- #310, mentioned in #333, basically suggests `@turf/concave` should be now implemented with [`concaveman`](https://github.com/mapbox/concaveman) (👍 ), but it's not. It seems the original [`turf-concave`](https://github.com/turf-junkyard/turf-concave) is/was though, and I didn't get why the current module does not reflect the latest implementation of `turf-concave`.

Should we use `concaveman` then? I'd be happy to implement that.

cc @tmcw


DenisCarriere:
> @DenisCarriere the case you added currently does not calculate a polygon because maxEdge is too small; so it now errors out with that message.

Great! Good to know it now throws an Error that makes sense.

> Should we use concaveman then? I'd be happy to implement that.

💯%! Most definitely should!

packages/turf-concave/index.js
```diff
@@ -31,40 +32,41 @@ var distance = require('@turf/distance');
...
+function merge(features) {
+    var merged = clone(features[0]);
+    merged.properties = {};
 
     for (var i = 0, len = features.length; i < len; i++) {
         var poly = features[i];
-        if (poly.geometry) {
-            merged = union(merged, poly);
-        }
+        if (poly.geometry) merged = union(merged, poly);
```
DenisCarriere: Just an FYI, we are doing two `clone` operation here, one at the very top (which is fine) and a 2nd time internally with `union`. No need to change this, but we should one day revamp `union` (can be discussed separately).

stebogit:
Ok @DenisCarriere I'll implement `@turf/concave` with `concaveman`. Should I create a new PR for that?
I'd keep working on this branch however.

DenisCarriere:
@stebogit 👍 any branch is good

Issue #901, opened 2017-08-14
lineOverlap on two polygons giving no output


tremby:
I'm trying to run `lineOverlap` on the two polygons shown in the following geojson:

https://gist.github.com/anonymous/da2c20dc9f6cf1e6e384f1c05e93a3e6

The source data is the larger polygon, which describes an area of a longer pipe. I use `intersect` with the longer pipe and that area to produce the pipe polygon shown in the gist.

What I then want is the two end caps where the intersection actually took place, since I want to extend them and then draw some lines here. It looks from the documentation like `lineOverlap` is the method I want.

I expect that if I run that on the two polygons in the gist that I'll get out two short lines, one on each end, since the lines here are shared between both polygons.

`lineOverlap(areaPolygon, pipePolygon)`

However, I am returned an empty FeatureCollection.

Am I doing something wrong, or is this a bug?

I also tried running `lineOverlap` over the original longer pipe and the area, and I get an empty feature set here too. It's not clear to me from the documentation if this is expected -- I can't tell if it's only supposed to give me lines which are shared between the two (I would expect it to be empty in this case) or lines wherever there are "cuts" (this is what I want).

DenisCarriere:
Thanks for the feedback @tremby, you are not doing anything wrong and this would be considered a "bug". The current implementation of this module requires to have each line segment share nodes together, since the edges "do" intersect but the outer ring doesn't share nodes with those intersections it currently won't be considered as overlapping (even though it should).

I've updated the test cases for this issue https://github.com/Turfjs/turf/pull/902, however it does require a bit more work to figure out how to tackle this problem.

I'm pretty sure this would be great scenario for [`unary-union`](https://github.com/nickcis/unary-union) (currently in development by @NickCis) would be great to apply to both `source` & `target` before splitting the lineSegments.

> Image below: Markers indicate the lines do in fact intersect, however no line segments share nodes together so they are not added to the overlapping results.

![image](https://user-images.githubusercontent.com/550895/29300643-b3e43baa-8144-11e7-86f4-06b47bd69aee.png)

rowanwins:
Gday @DenisCarriere 

You could possibly use `@turf/pointOnLine` to do something in this space (he says without having actually looked at the source for the `lineOverlap` module)

tremby:
Thanks, @DenisCarriere; sounds like it's a difficult problem to solve. I guess you'd have to decide how much error to allow along the line when deciding whether two lines are following the same path or not.

[...]

PR #902, opened 2017-08-15
Match `@turf/line-overlap` based on proximity distance

DenisCarriere:
## Match `@turf/line-overlap` based on proximity distance

Primarily a fix for issue https://github.com/Turfjs/turf/issues/901.

[`@turf/line-overlap`](https://github.com/Turfjs/turf/tree/master/packages/turf-line-overlap) can now match lines within a proximity distance (in kilometers).

## JSDocs

```js
/**
 * Takes any LineString or Polygon and returns the overlapping lines between both features.
 *
 * @name lineOverlap
 * @param {Geometry|Feature<LineString|MultiLineString|Polygon|MultiPolygon>} line1 any LineString or Polygon
 * @param {Geometry|Feature<LineString|MultiLineString|Polygon|MultiPolygon>} line2 any LineString or Polygon
 * @param {number} [proximity=0] Proximity distance to match overlapping line segments (in kilometers)
 * @returns {FeatureCollection<LineString>} lines(s) that are overlapping between both features
 * @example
 * var line1 = turf.lineString([[115, -35], [125, -30], [135, -30], [145, -35]]);
 * var line2 = turf.lineString([[115, -25], [125, -30], [135, -30], [145, -25]]);
 *
 * var overlapping = turf.lineOverlap(line1, line2);
 *
 * //addToMap
 * var addToMap = [line1, line2, overlapping]
 */
module.exports = function (line1, line2, proximity) {
```

## To-Do's
- [x] Add test case for Issue #901
- [x] Support line segments which overlap which don't exactly share nodes.
- [x] Add `segmentEach` to iterate over line segments
- [x] ~Add [`unary-union`](https://github.com/nickcis/unary-union) to both source & target~ temporary fix using `point-on-line` has been added until `unary-union` is completed

## Simplified example (no sharing nodes)

- **Red:** Source Line
- **Blue:** Target Line
- **Green:** Overlapping lines (Results)

image


DenisCarriere:
@rowanwins @tremby Have a look at this PR, this seem to have solved both the simplified & the initial issue.

Since the line segments might not be exactly on the target line, I added a `proximity` parameter (in kilometers) used to improve/reduce the matches.

@tremby I used `proximity=0.05` (50 meters) for yours and it seems to have worked very well.

**Green:** Results => overlapping lines

image

**Simplified example**

image

DenisCarriere:
@rowanwins Tried to use `@turf/boolean-point-on-line`, it worked for the simplified test case, however it wasn't able to match the other since it has to be **exactly** on the line to return `true`.

Is there anyway to add a distance parameter in your boolean module? Is that even possible? Nothing too complicated but having this would help a lot for this type of scenario.

👍

tremby:
Looks like it does the trick; thank you! Lovely fast progress.

DenisCarriere:
> Lovely fast progress.

You caught me at a "bored" moment, usually it's not this fast.

Strange that the Travis CI keeps failing... 🤔 works fine locally.

tremby:
Questions about this fix:

- With proximity at `0` does it work on the simplified test case?
- By extension, is the reason it does not work on my sample data something to do with floating point error? As explained in the issue, the red shape (in your last screenshot) is the result of an intersection with the blue, so barring floating point error or a bug in either this method or `intersect` I don't see why they wouldn't still be on exactly the same path.

DenisCarriere:
> With proximity at 0 does it work on the simplified test case?

Yes with 0 it does work on the simplified case (added `@turf/boolean-point-on-line`).

> is the reason it does not work on my sample data something to do with floating point error

Pretty sure it's something like that, floating points, and when you dive down real close you can see that the points don't actually overlap but they cross each other, hence why we needed to add the `proximity` parameter to make that work.

It's a great example, I'm glad you brought it up!

> Strange that the Travis CI keeps failing... 🤔 works fine locally.

Ops, this broke something in `@turf/boolean-point-on-line`.
@rowanwins Might worth removing dependencies in your boolean modules, that way we don't have any circling dependencies (dependencies that depend on each other).

rowanwins:
Gday @DenisCarriere 

Yep we should be able to add some sort of proximity parameter (I'd perhaps call it tolerance) similar to any of the other distance style params that we offer for other modules. 

`point-to-line` doesn't have any dependencies, although some of the other boolean ones were probably calling `line-overlaps` I think.

Will see if I can bash it out quickly this evening.

DenisCarriere:
@rowanwins 👍 

`tolerance` is a better param name.

> point-on-line doesn't have any dependencies, although some of the other boolean ones were probably calling line-overlaps I think.

Most of the "core" functionality can be copy-pasted from the internal methods, they aren't very big modules. Close to zero dependencies for the boolean modules would be ideal, obviously there's some that can't which is fine.

PR #903, opened 2017-08-15
Simplify fix

### LINKED ISSUE #555: Simplify - process pending on particular geometry

- [x] GeoJSON data as a [gist file](https://gist.github.com/).
- [x] Provide an example.

When using `simplify`, this particular geometry doesn't return any results and keeps processing without returning any results nor throwing any errors.

**Polygons with issues**
- https://gist.github.com/DenisCarriere/075d2101f0def2e5c1ecb6cd83fd81f0
- https://gist.github.com/anonymous/f0ac3dcb8650654ab303d91d717e3e0f
- https://gist.github.com/anonymous/6d4dc397608de763fa333fae37848932
- https://gist.github.com/anonymous/a3cbea7097f532c34f011e3917bd8505
- https://gist.github.com/anonymous/d6d7bdf806383e2e8daa0c601780064a

```javascript
const turf = require('@turf/turf')

const polygon = {
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [
          -75.693254,
          45.431144
        ],
        [
          -75.693335,
          45.431109
        ],
        [
          -75.693335,
          45.431109
        ],
        [
          -75.693254,
          45.431144
        ]
      ]
    ]
  }
}

console.log(turf.simplify(polygon, 0.000001, true))
//= no results, no errors
```

**Solved  #by**

Seems `simplify-js` is only able to process Polygons with 3 unique coordinates or greater, a polygon with 5 coordinates but with only 2 unique coordinates seems to crash the process.

I created a simple hash table to sort the unique coordinates, then added an if statement to prevent the simplification to occur.

> This issue also occurs for if the Polygon has an inner with only 2 unique coordinates. 

Patch required for: https://github.com/Turfjs/turf/blob/master/packages/turf-simplify/index.js#L156-L181

```javascript
// Store unique coordinate hashes
const uniqueOuter = {}
const uniqueInner = {}

// Decimal precision defined by param
const precision = 5

// Create hash for outer polygon
feature.geometry.coordinates[0].map(coord => uniqueOuter[coord.map(x => Number(x.toFixed(precision))).join(',')] = true)

// Create hash for inner polygon
if (feature.geometry.coordinates[1]) {
  feature.geometry.coordinates[1].map(coord => uniqueInner[coord.map(x => Number(x.toFixed(precision))).join(',')] = true)
}
// Remove any inner polygon with two or less vertices
if (Object.keys(uniqueInner).length === 2 || Object.keys(uniqueInner).length === 1) {
  feature.geometry.coordinates = feature.geometry.coordinates.slice(0, 1)
}

// Do not simplify a polygon with two or less vertices
if (Object.keys(uniqueOuter).length > 2) {
    feature = turf.simplify(feature, 0.000001, true)
}
```

stebogit:
Fixes #555
- [x] used `@turf/clean-coords` (instead of [suggested patch](https://github.com/Turfjs/turf/issues/555#issue-202431557))
- [x] added (previously) failing tests.
- [x] extended support of any GeoJSON
- [x] refactored code, updated `bench.js` and `test.js` to current standard


packages/turf-simplify/index.d.ts
```diff
@@ -1,34 +1,10 @@
...
-    (feature: MultiPolygons, tolerance?: number, highQuality?: boolean): MultiPolygons;
-    (feature: GeometryCollection, tolerance?: number, highQuality?: boolean): GeometryCollection;
-    (feature: Feature, tolerance?: number, highQuality?: boolean): Feature;
-    (feature: Features, tolerance?: number, highQuality?: boolean): Features;
-}
-declare const simplify: SimplifyStatic;
+/**
+ * http://turfjs.org/docs/#simplify
+ */
+declare function simplify(geojson: Geoms, tolerance?: number, highQuality?: boolean): Geoms;
```
DenisCarriere: This isn't correct, the `): Geoms` is saying that the output is FeatureCollection/Feature/Geometry when in fact it should be the same as the Input defined in Geoms.

The previous definition was a mess! :)

But here's an updated one:

```
declare function simplify<T extends Geoms>(geojson: T, tolerance?: number, highQuality?: boolean): T;
```

Which means whatever you enter as `T` will be the same for the output.

packages/turf-simplify/index.js
```diff
@@ -1,19 +1,19 @@
...
+var feature = helpers.feature;
 
 /**
- * Takes a {@link LineString} or {@link Polygon} and returns a simplified version. Internally uses [simplify-js](http://mourner.github.io/simplify-js/) to perform simplification.
+ * Takes a {@link GeoJSON} object and returns a simplified version. Internally uses
+ * [simplify-js](http://mourner.github.io/simplify-js/) to perform simplification.
  *
  * @name simplify
- * @param {Feature<(LineString|Polygon|MultiLineString|MultiPolygon)>|FeatureCollection|GeometryCollection} feature feature to be simplified
+ * @param {GeoJSON} geojson object to be simplified
```
DenisCarriere: Way better to support all GeoJSON types 👍

packages/turf-simplify/index.js
```diff
@@ -37,104 +37,96 @@ var supportedTypes = ['LineString', 'MultiLineString', 'Polygon', 'MultiPolygon'
...
-                return simplifyPolygon(rings, tolerance, highQuality);
-            })
-        };
-    } else {
-        // unsupported geometry type supplied
-        return feature;
-    }
-}
+    var type = feature.geometry.type;
+    // "unsimplyfiable" geometry types
```
DenisCarriere: 👍 Agreed

packages/turf-simplify/package.json
```diff
@@ -22,16 +22,23 @@
     "simplify"
   ],
   "author": "Turf Authors",
+  "contributors": [
+    "Stefano Borghi <@stebogit>"
+  ],
```
DenisCarriere: Try including past contributors to this list when updating modules (@tmcw was heavily involved in the initial development of this module).

Click the **history** button on the `index.js` file.
https://github.com/Turfjs/turf/commits/b360f9beeec921c38468062fce893158e4a84988/packages/turf-simplify/index.js

packages/turf-simplify/test.js
```diff
@@ -1,112 +1,58 @@
...
+    };
+});
+
+test('simplify', t => {
+    for (const {filename, name, geojson} of fixtures) {
+        let {tolerance, highQuality} = geojson.properties || {};
+        tolerance = tolerance || 0.01;
+        highQuality = highQuality || false;
+
+        const simplified = simplify(geojson, tolerance, highQuality);
```
DenisCarriere: These tests look so much nicer now!! 😄

packages/turf-simplify/index.d.ts
DenisCarriere: Another catch would be to include more robust Typescript testing in `types.ts.`

> One day I'll write a guide for using Typescript with TurfJS

This would fail:

```ts
import {polygon, Polygon} from '@turf/helpers'
import * as simplify from './'

const poly = polygon([[[0, 0], [10, 10], [20, 20], [0, 0]]]);
const simplifiedPoly: Polygon = simplify(poly);
```
With the following error:
```
[ts]
Type 'Geoms' is not assignable to type 'Feature<Polygon>'.
  Type 'GeometryObject' is not assignable to type 'Feature<Polygon>'.
    Types of property 'type' are incompatible.
      Type 'string' is not assignable to type '"Feature"'.
```

DenisCarriere:
@stebogit Seems like this would benefit from having a `mutate` param since right now this module is stripping off any extra properties from the FeatureCollection or Features (ex: `id` or any custom properties).

We could include `clone` at the beginning and leverage `geomEach` to handle the simplify transformations.

DenisCarriere:
Added test cases for `bbox` & `id` removal.

Failing: 
 - `bbox`
 - `id`

```js
test('simplify -- removes ID & BBox from properties', t => {
    const properties = {foo: 'bar'};
    const id = 12345;
    const bbox = [0, 0, 2, 2];
    const poly = polygon([[[0, 0], [2, 2], [2, 0], [0, 0]]], properties, bbox, id);
    const simple = simplify(poly, 0.1);

    t.equal(simple.properties.id, id);
    t.deepEqual(simple.properties.bbox, bbox);
    t.deepEqual(simple.properties, properties);
    t.end();
});
```

DenisCarriere:
@stebogit You'll have to take a lot at it again to see if everything is ok, I've done some pretty drastic changes.

stebogit:
It seems you got it all done @DenisCarriere 😜 👍

DenisCarriere:
Woot! 🤓

PR #905, opened 2017-08-15
Bump `polygonize` dependency in @turf/polygonize

globetro:
This allows pulling in this fix https://github.com/NickCis/polygonize/pull/1 which would allow us to use @turfjs in React Native apps.

### DIFF
--- a/packages/turf-polygonize/package.json
+++ b/packages/turf-polygonize/package.json
@@ -41,6 +41,6 @@
     "write-json-file": "2.2.0"
   },
   "dependencies": {
-    "polygonize": "1.0.1"
+    "polygonize": "1.0.2"
   }
 }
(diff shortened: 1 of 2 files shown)


DenisCarriere:
All external dependencies should be using an exact version, it's very difficult to figure out what's wrong with a module when a patch fix changes the outputs drastically.

Feel free to bump the version to `1.0.2` instead.

In the next TurfJS releases everything will (should) be an exact version.

globetro:
Sorry, I realized the fix to polygonize still isn't fully fixing the issue, closing this PR until I fix that.

DenisCarriere:
@globetro 👍 Sounds good, yea I wouldn't know what the issue is with React Native... seems like an odd one.

Does React Native work with the other TurfJS modules? is it just Polygonize that's not working?

globetro:
Yeah, the gist of the issue is that React Native's preprocessor seems to try to honor any babel configurations in any package dependency: https://github.com/facebook/react-native/issues/8296

It seems that it's just the `polygonize` package that is including its babel configuration in its npm package, so I've opened a PR to not include it: https://github.com/NickCis/polygonize/pull/2

Once that is merged and published as v1.0.3, then we can bump that package version here and it should be fixed. I can't be certain if all other turfjs package dependencies also have this issue.

DenisCarriere:
@globetro Sounds like a plan! 👍

PR #907, opened 2017-08-17
Additional `@turf/concave` refactoring

stebogit:
Following #899:
- [x] replaced `@turf/union` with [`geojson-dissolve`](https://github.com/digidem/geojson-dissolve), thus removed `jsts` dependency 🎊 
- [x] ⬆️ ⬆️ ⬆️ ⬆️  speed 👍 
- [x] Support `null` geometries

The reason why I preferred `geojson-dissolve` to [`concaveman`](https://github.com/turf-junkyard/turf-concave) (as [previously suggested](https://github.com/Turfjs/turf/pull/899#issuecomment-322212476)) is because the latter returns always a single concave Polygon embracing all the input points, while `geojson-dissolve` allows to return (Multi)Polygons even with holes if appropriate (as the current `@turf/concave` implementation does).
However, for what I've seen online after some research, I'm not sure that this is the right/expected behaviour of a `concave` module. This is mainly due to the `maxEdge` parameter: if it was set to `Infinity` `@turf/concave` would always return a single concave hull Polygon (like `concaveman`).
In the next major release I would set `maxEdge` to `Infinity` by default.

Also I noticed this module throws an `Error` if there is no hull to return ("too few polygons found to compute concave hull"), likely due to a small `maxEdge`. I would rather return an empty `Feature`.


rowanwins:
Hey @stebogit 

That `geojson-dissolve` module looks interesting, a few q's, 
- what is performance of it like compared to our dissolve module
- whats is it like size wise once it includes its dependencies

Im sure either the geojson-dissolve or concaveman approaches will be a big improvement!

DenisCarriere:
>  thus removed jsts dependency

🎉 Woot! 👍 

> while geojson-dissolve allows to return (Multi)Polygons even with holes if appropriate (as the current @turf/concave implementation does).

👍 Better to keep it similar to the same implementation, if one wants to use `concaveman` then it's pretty straightforward to use as an external library.

> In the next major release I would set maxEdge to Infinity by default.

👍 Sounds like a plan!

> Also I noticed this module throws an Error if there is no hull to return ("too few polygons found to compute concave hull"), likely due to a small maxEdge. I would rather return an empty Feature.

👍 Sounds good, this shouldn't be an error. An error should be caused when you are misusing the module (which isn't the case).

## Question/Proposal

- [ ] Improve/refactor `@turf/dissolve` using `geojson-dissolve`
- [ ] Drop `@turf/union` in next major release in favor of using `@turf/dissolve` (both modules serve the same purpose).

If we are replacing `@turf/union` with `geojson-dissolve`, couldn't we refactor [`@turf/dissolve`](https://github.com/Turfjs/turf/tree/master/packages/turf-dissolve) to use that module (currently it's also using `@turf/union`).

Also, what's the difference between `@turf/union` & `@turf/dissolve`, seems like they would serve the same purpose (`@turf/dissolve` has a bit more features, but it it's essentially the same for the geometry).

packages/turf-concave/package.json
```diff
@@ -24,25 +24,29 @@
...
   "author": "Turf Authors",
   "contributors": [
-    "Stefano Borghi <@stebogit>"
+    "Tom MacWright <@tmcw>",
+    "Lyzi Diamond <@lyzidiamond>",
+    "Denis Carriere <@DenisCarriere>",
+    "Stefano Borghi <@stebogit>",
+    "Rowan Winsemius <@rowanwins>",
+    "Daniel Pulido <@dpmcmlxxvi>"
   ],
```
DenisCarriere: I would also include the Authors/Contributors of `geojson-dissolve`.

- Stephen Whitmore @noffle
- Gregor MacLennan @gmaclennan

noffle:
Nice, someone else is using our code. :heart:

It's worth noting that `geojson-dissolve` is just a thin wrapper around [topojson](https://github.com/topojson/topojson-client/blob/master/README.md#merge) for the Polygon case, and deserve lots of credit for doing the Hard Work.

packages/turf-concave/index.js
```diff
@@ -51,23 +53,32 @@ module.exports = function (points, maxEdge, units) {
...
-
-    for (var i = 0, len = features.length; i < len; i++) {
-        var poly = features[i];
-        if (poly.geometry) merged = union(merged, poly);
-    }
-    return merged;
+function removeDuplicates(points) {
+    var cleaned = [];
+    var existing = {};
+    points.features.forEach(function (pt) {
```
DenisCarriere: 👍 This syntax looked so familiar that I thought I wrote this part :)

https://github.com/Turfjs/turf/blob/master/packages/turf-clean-coords/index.js#L52-L58

DenisCarriere:
@noffle You're code is awesome! 👍 You've got a new fan boy for you're many awesome repos in [Digital Democracy](https://github.com/digidem).

> thin wrapper around topojson for the Polygon case, and deserve lots of credit for doing the Hard Work.

We will add Mike Bostock @mbostock to the list of contributors for his great work in `topojson`.

DenisCarriere:
> @rowanwins what is performance of it like compared to our dissolve module

Humm.. I personally worked on `@turf/dissolve`, one of the main differences is the use of `rbush` to quickly create an index of the geometries that intersects each other and only perform the `union/dissolve` operation on those geometries.

Looking at the [source code of `geojson-dissolve`](https://github.com/digidem/geojson-dissolve/blob/master/index.js) there's a few optimizing things that can speed up the performance:
- [ ] [`flattenEach`](https://github.com/Turfjs/turf/tree/master/packages/turf-meta#flatteneachcallback) in `@turf/meta` (instead of using `flatten` + `geomEach`)
- [ ] [`@turf/clone`](https://github.com/Turfjs/turf/tree/master/packages/turf-clone) instead of using `JSON.parse + JSON.stringify`
- [ ] [`geojson-rbush`](https://github.com/DenisCarriere/geojson-rbush) to reduce the amount of dissolve operations (GeoJSON => TopoJSON => Dissolve => TopoJSON => GeoJSON).
- [ ] Would it be possible to translate [`topojson.merge`](https://github.com/topojson/topojson-client/blob/master/src/merge.js) to pure GeoJSON? It would be worth doing for performance, but might take some time to figure out 🤔..

packages/turf-concave/index.js
```diff
@@ -51,23 +53,32 @@ module.exports = function (points, maxEdge, units) {
...
+        var key = pt.geometry.coordinates.join('-');
+        if (!existing.hasOwnProperty(key)) {
+            cleaned.push(pt);
+            existing[key] = true;
+        }
+    });
+    return featureCollection(cleaned, points.properties);
```
DenisCarriere: FeatureCollection doesn't have `properties` as a param.

noffle:
Very happy to accept those PRs. :grin:

DenisCarriere:
> Very happy to accept those PRs.

Me too! 👍 haha

DenisCarriere:
+10 commits later, I think this last commit will "pass" lol 😫

rowanwins:
Gday @DenisCarriere 

I wrote the bulk of the dissolve module so yep I remember using `rbush` :) 

I'd think it's worth keeping both `union` and `dissolve`. While geometry-wise they might have similar inputs and outputs their application is a bit different. 

In regards to `geojson-dissolve` it would be good to do some performance testing against our existing dissolve module. It would be good to see how performant it is in all those conversions of format etc.

DenisCarriere:
@rowanwins 👍 For sure, we can do that on a separate PR, I'm also interested in seeing how `jsts` performs compared to `topojson.merge`.

DenisCarriere:
> I wrote the bulk of the dissolve module so yep I remember using rbush :)

Yes team work 👍 🥇

packages/turf-concave/index.js
stebogit: Yes! I told you it was sooo smart! 😄

packages/turf-concave/index.js
stebogit: ah... ok! 👍

stebogit:
@rowanwins @DenisCarriere `geojson-dissolve` merges only contiguous features, i.e. sharing a side or segment -- being it based on TopoJSON it make sense. If the features don't touch or overlap it just returns the original inputs .
Thus it was perfect for `@turf/concave`, which uses `@turf/tin`, but I don't see it particularly useful for `@turf/union` or `@turf/dissolve`.
I've been searching online for union/merge algorithms and tools to replace `jsts`, but no luck so far... 😞 
But we'll find a way, eventually! 😃

stebogit:
> Drop `@turf/union` in next major release in favor of using `@turf/dissolve` (both modules serve the same purpose).

...and `union` does not use `jsts`! 👍 
However I'd still call the final module `@turf/union`, a quite common term for boolean operations on polygons, and I'd define the `propertyName` (now a `@turf/dissolve` argument) to optional, merging all the features by default.

stebogit:
>I'd think it's worth keeping both union and dissolve. While geometry-wise they might have similar inputs and outputs their application is a bit different.

@rowanwins what kind of differences do you have in mind? It looks to me they could be integrated in a single module that can handle a "filter/selection" parameter.

rowanwins:
My understanding is that traditionally the output of a `union` operation won't have any overlapping features, whereas with a `dissolve` you can still have overlapping features. 

This is a [helpful article](https://blogs.esri.com/esri/supportcenter/2014/12/02/combining-spatial-data-in-arcmap/), by and large I think it would be good to retain consistent language with what everyone else is doing.

DenisCarriere:
👍 Thanks for the reference @rowanwins.

Helpful link, got me thinking of a few new easy modules `@turf/merge` & `@turf/append` (we can open up new issues for that - if you feel those are useful).

I understand a bit better now the difference between dissolve/union, the main difference with dissolve is spatial operation based on a common attribute (which it's currently doing 👍 ).

stebogit:
> the output of a union operation won't have any overlapping features, whereas with a dissolve you can still have overlapping features.

@rowanwins so basically `dissolve` is a "filtered" `union` on the collection of features? We might have one single module that accept an optional `filter` function/parameter that define which feature should be merged. If well thought the `filter` could actually give much more flexibility to the module, and so to the user.

DenisCarriere:
👍 `@turf/filter` module, I also did this type of filter behavior in the [`@turf/cluster`](https://github.com/Turfjs/turf/blob/master/packages/turf-clusters/index.js#L228-L244) module, but it would better if this was extracted out of the clusters module and reside in it's own repo.

PR #909, opened 2017-08-17
Fix for boolean properties not being translated using @turf/clone

DenisCarriere:
## Quickfix to `@turf/clone`

Properties with `typeof==="boolean"` wasn't being cloned 🤦‍♂️ .

- [x] Handles properties with `boolean` values
- [x] Added test cases for issue


DenisCarriere:
Merging since tests will fail, need to apply this fix to `@turf/simplify`.

Issue #742, opened 2017-05-13
@turf/dissolve doesn't completely dissolve all geometries


DenisCarriere:
Notice in the example below the two strangling polygons, whenever a geometry is completely within the dissolved outer ring some geometries seem to be excluded.

@rowanwins With many recent changes, we might want to revisit the source code for [`@turf/dissolve`](https://github.com/Turfjs/turf/blob/master/packages/turf-dissolve/index.js).

**Buffered points**
https://gist.github.com/490c0e00536ecc3ef1c98b74c40d2c29

![image](https://cloud.githubusercontent.com/assets/550895/26029213/25cc50bc-37fe-11e7-994f-5d566b6da482.png)

**Dissolved points**
https://gist.github.com/7ad5ef7316f2a0a4a2578e9696902886
![image](https://cloud.githubusercontent.com/assets/550895/26029229/4da6d9ea-37fe-11e7-82b8-ef52caca269c.png)

rowanwins:
Roger that @DenisCarriere, I'll take a look

andrewharvey:
@DenisCarriere did you want to try this again with dissolve from v7 alpha? I had a similar issue which was fixed in v7.

PR #911, opened 2017-08-21
Dissolve fix attempt

stebogit:
Ref. #742 
@rowanwins I tried to fix the issue, but for some reason the current code breaks with an error instead of returning a wrong output.
I also refactored the code trying to improve readability, but somehow the speed dropped... 👎  😕  

Anyway the problem now seems to be `@turf/union` that outputs a `MultiPolygon` for this case in [this line](https://github.com/Turfjs/turf/compare/dissolve-fix?expand=1#diff-a4826b5fe059b642dd812835b6f3eae8R78):
https://gist.github.com/anonymous/b86d5eba2fa0a033407297ff6fe825c3
screen shot 2017-08-21 at 8 32 56 am

The two polygons shares only one point (actually the same point several times since it's repeated in their coords). I'm not sure if [this check](https://github.com/Turfjs/turf/compare/dissolve-fix?expand=1#diff-a4826b5fe059b642dd812835b6f3eae8R76) should actually define this case as `overlapping` or `intersect`, or not.

cc: @DenisCarriere


rowanwins:
Hey @stebogit 

I think I'd possibly leave refactoring this module until we've got a new jsts-free `union` module. 

Re the shared co-ord Im not sure is `boolean-overlap` the correct check here, I think `boolean-intersects` is a better check, but we don't yet have this module complete, another one on my _to-finalise_ list *sigh* 

Re the performance drop a few things come to mind
- My understanding is that `forEach` loops are generally a bit slower than traditional loops
- I'm not sure whats going on under the hood with `geojson-rbush` so there may be a performance hit there...

DenisCarriere:
🤔 Looks like this one is a thinker...

@stebogit Do you want to send another PR with just the refactoring? That way we can merge that and then focus on fixing the issue afterwards, that way we are only seeing a few lines of code changed instead of 50-80%.

stebogit:
@DenisCarriere let's try that 👍

stebogit:
Actually @DenisCarriere this _is_ just a refactoring. The two original tests pass, while the issue#742 featureCollection is erring out just as it does with the current (v4.6.1) version:
https://jsfiddle.net/jStefano/f7g65set/

So basically my changes _should_ not have changed the output.

With few tests I found out what causes the reeeally slow computation is the `clone(featureCollection)` I introduced to avoid input mutation. It is unexpectedly slow compared with a feature cloning. We could remove it but I whould add to the docs a note to warn the user about the input mutation.

rowanwins:
Hey @stebogit 

Good pickup about the clone slowing things down. We could perhaps pass in an optional param called `mutateInput` (or something more friendly), defaulting it allow mutating as that's the most performant. In fact now that I think about it perhaps that could be another of one those `v5` changes, it could get applied to a whole bunch of modules....

DenisCarriere:
Hey @stebogit, I've been diving into refactoring `dissolve` module, I think what would be useful for this approach would be to create sub dissolve modules such as `@turf/line-dissolve` & `@turf/polygon-dissolve` that way it's easier for testing and easier to create the source code.

Started a discussion here: https://github.com/Turfjs/turf/issues/1027

DenisCarriere:
@stebogit this PR "should" pass, however the outputs aren't ideal, I think it was broken even before you attempted to refactor it.

I've got an idea for the next refactoring, using [`clusterEach`](https://github.com/Turfjs/turf/tree/master/packages/turf-clusters#clustereach) will help separate the features based on properties.

DenisCarriere:
`@turf/dissolves` still need more ❤️, even after PR merge.

Issue #906, opened 2017-08-16
Adding tolerance to point-on-line


rowanwins:
A couple of days ago @DenisCarriere [raised](https://github.com/Turfjs/turf/pull/902#issuecomment-322385233) the possibility of adding a tolerance threshold to `point-on-line`. Currently the point has to be exactly on the line, there is 0 margin for error.

A few finds from my investigations
- As far as I can work out the current algorithm doesn't really support adding in a tolerance check
- I did find a pretty neat algorithm for finding the distance between a point and a line [here](https://stackoverflow.com/a/1501725)
  - we don't really have a module for this currently, is this something we'd like to package up @DenisCarriere ?
  - this is quite a bit slower than the current algorithm (eg half the number of ops per sec) but if it only got called when the tolerance threshold was set that might be acceptable...
- there are a couple of other potential options that I'd like to explore still that would be an entire replacement
  - eg [here](http://forums.codeguru.com/showthread.php?419763-Check-a-Point-lies-in-a-Line-segment) and [here](https://stackoverflow.com/a/25689069)

I knew I should've paid more attention in maths!

rowanwins:
Both my other options are a no go, they were more about using an epsilon value to round float point precision which isn't really what we're wanting from this tolerance type setting. 

So I'm thinking the distance measurement is the best bet at the moment...

Will keep mulling it over

DenisCarriere:
👍 @rowanwins great findings, I wouldn't sweat it if we don't get this implemented, the `pointOnLine` does seem to do the job at the moment.

Very interesting links you provided, I also should of paid more attention to math 😄

stebogit:
> we don't really have a module for this ['finding the distance between a point and a line'] currently

@rowanwins `@turf/point-on-line` does actually calculate that distance (probably not in the fastest way) and returns the point on the line at the closest distance (from the input point).
It might be useful to create a module for that calculation, and maybe make sure we use the fastest algorithm in there.

BTW, in this issue we are talking about `@turf/boolean-point-on-line` right?
Quite confusing that we have two different modules almost with the same name 😅

DenisCarriere:
> maybe make sure we use the fastest algorithm in there.

👍 That would be handy!

> Quite confusing that we have two different modules almost with the same name

Agreed, it can get confusing at times 😢

rowanwins:
[...]

Issue #141, opened 2014-09-08
wgs84-to-mercator


morganherlocker:
Map a feature or featurecollection from wgs84 to mercator

stebogit:
Implemented in #927

morganherlocker:
Thanks @stebogit, quite the old issue!

DenisCarriere:
👍

PR #925, opened 2017-08-30
New module `@turf/point-to-line-distance`

stebogit:
Following up https://github.com/Turfjs/turf/issues/906#issuecomment-323276237 

```js
/**
 * Returns the minimum distance between a {@link Point} and a {@link LineString}, being the distance from a line the
 * minimum distance between the point and any segment of the `LineString`.
 *
 * @name pointToLineDistance
 * @param {Feature<Point>|Array<number>} point Feature or Geometry
 * @param {Feature<LineString>|Array<Array<number>>} line GeoJSON Feature or Geometry
 * @param {string} [units=kilometers] can be degrees, radians, miles, or kilometers
 * @param {boolean} [mercator=false] if distance should be on Mercator or WGS84 projection
 * @returns {number} distance between point and line
 * @example
 * var pt = turf.point([0, 0]);
 * var line = turf.lineString([[1, 1],[-1, 1]);
 *
 * var d = pointToLineDistance(pt, line, 'miles');
 * //=69.11854715938406
 */
module.exports = function (point, line, units, mercator) {
```

**Note:** This module, like probably all the other distance related ones, calculates distance with a little error, due to the current Earth radius value (see discussion https://github.com/Turfjs/turf/pull/715).

This module might allow a faster computation of [`@turf/point-on-line`](https://github.com/Turfjs/turf/tree/master/packages/turf-point-on-line). I'll try work on that to test any improvement.

Also, we can also use the following functions (thanks @tmcw) to implement the [`@turf/wgs84-to-mercator`](https://github.com/Turfjs/turf/issues/141) module:
```js
/**
 * Convert lon/lat values to 900913 x/y.
 * from https://github.com/mapbox/sphericalmercator
 *
 * @private
 * @param {Array<number>} lonLat WGS84 point
 * @returns {Array<number>} Mercator [x, y] point
 */
function toMercator(lonLat) { ...
```
```js
/**
 * Convert 900913 x/y values to lon/lat.
 * from https://github.com/mapbox/sphericalmercator
 *
 * @private
 * @param {Array<number>} xy Mercator [x, y] point
 * @returns {Array<number>} WGS84 [lon, lat] point
 */
function toWGS84(xy) { ...
```


stebogit:
I don't get why this is failing... 😞 
It has something to do with `eslint`, but I don't understand where
screen shot 2017-08-30 at 12 35 56 am

rowanwins:
You broke it @stebogit - we're all doomed!

stebogit:
Now, I fixed several issues updating external turf modules (don't know why that was necessary...), but now this fails on Travis (not on my computer) in few tests due to decimal approximation.
How come the result is different? What might be different?

DenisCarriere:
@stebogit New error, looks like you might need to add a `round()` or `truncate()` to the test cases:

```
    not ok 8 city-segment-projected1
      ---
        operator: deepEqual
        expected: 3.589782772179696
        actual:   3.589782772179697
```

stebogit:
@DenisCarriere Could the approximation in the calculation result depend on the nose version? Cause as I said in my computer they all pass.

stebogit:
🎉

DenisCarriere:
Published under `v4.0.0` will be published in `v4.7.0` when I do the full publish

PR #927, opened 2017-09-01
New module `@turf/projection`

stebogit:
This is a first draft of the module(s) to convert any GeoJSON from WGS84 to Mercator and vice versa, where the conversion functions are taken from [`sphericalmercator`](https://github.com/mapbox/sphericalmercator).
Example:
```js
var pt = turf.point([4, 7.5])
var mercator = turf.toMercator(pt);
var wgs84 = turf.toWgs84(mercator);
```

Do we actually want two separate modules for the conversions or a single one with a parameter?
We might have something like:
```js
var pt = turf.point([4, 7.5])
var mercator = turf.convert(pt, 'wgs84', 'mercator');
var wgs84 = turf.convert(pt, 'mercator', 'wgs84');
```

I decided to use [`proj4`](https://github.com/proj4js/proj4js) as a reference to create the test cases, is that ok?

Ref: https://github.com/Turfjs/turf/issues/141, https://github.com/Turfjs/turf/pull/925


DenisCarriere:
👍 Nice! Looking forward to play with this module 😄 

I'm good for using `toWgs84` and `toMercator`, these coordinate conversions should only be used for WGS84 => Mercator => WGS84, anything else more should be handle with custom scripts and not be built-in for TurfJS. The last thing we want is handling projection issues.

> I decided to use proj4 as a reference to create the test cases, is that ok?

Proj4 is perfect! I would've used the same.

DenisCarriere:
Significant recommendation, I would group these methods (`toWgs84` & `toMercator`) in a single module (ex: `@turf/projection`) and have this module export those two or more methods.

Or we use your 2nd approach, except change `convert` to `projection`.

```
var projection = require('@turf/projection');
var pt = turf.point([4, 7.5])
var mercator = turf.projection(pt, 'wgs84', 'mercator');
var wgs84 = turf.projection(pt, 'mercator', 'wgs84');
```

stebogit:
> I would group these methods (toWgs84 & toMercator) in a single module (ex: @turf/projection) and have this module export those two or more methods.

This is what I really wanted to do, but I couldn't figure out how to do it, cause I didn't think about that!! 🎉 
Excellent, will do! 😃 👍

DenisCarriere:
Awesome! 👍 Just need to look at `@turf/helpers` or `@turf/meta` for multi-function modules.

stebogit:
@DenisCarriere I am not sure how to handle the 180th meridian issue here: `proj4` returns values quite different than `sphericalmercator`, thus the test fails.
What should be the output for the test case?
Is there anything like [`geojson.io`](http://geojson.io) for Mercator coordinates?

DenisCarriere:
🎉 @stebogit module published `@turf/projection`

Issue #881, opened 2017-08-02
ringEach Proposal


rowanwins:
So just wanted to continue the discussion from [this pull request](https://github.com/Turfjs/turf/pull/876).

I'd find it quite handy to have a module that looped through interior and exterior rings of polys easily.

````
turf.ringEach(feature, function (ringCoords, ringIndex, isExterior) {
  //=ringCoords = [[0,0], [1,1], [1, 0], [0, 0]]
  //=ringIndex = 0
  //=isExterior = true
});
````
@DenisCarriere you proposed splitting the module into interior and exterior versions, I'm just trying to think through whether it would be better to have them together in a single module and how I'd use them....

In an ideal world I'd then love to be able to pass a coord array into one of the other meta functions, eg `coordEach` without having to first turn into a geojson feature or geometry.

Anyway just starting the discussion rolling in a more trackable place than a vaguely related pull request :)

DenisCarriere:
> you proposed splitting the module into interior and exterior versions

👍 I like your approach better, combining both of them prevents doing a 2nd iteration over the GeoJSON.

> In an ideal world I'd then love to be able to pass a coord array into one of the other meta functions, eg coordEach without having to first turn into a geojson feature or geometry.

Agreed, the only thing that we would need to figure out is the depth of the coord Array. Using a Feature or Geometry the depth can be figured out by the Geometry Type, but with an Array it could be a bit trickier (but not impossible).

> ringIndex

- [ ] We might want to include `ringSubIndex` to capture MultiPolygon.

stebogit:
@rowanwins @DenisCarriere based on GeoJSON specs the `Polygon` outer ring has to be the first one. Should we follow and enforce that rule (assuming the input is a valid GeoJSON)? We wouldn't need the `isExterior` index.

DenisCarriere:
👍 Agreed, since the OuterRing will always be `ringIndex=0`

@stebogit How should we handle `MultiPolygon` Geometries?

Also we should add `LineString` & `MultiLineString` as valid inputs (might as well throw in Points as well to support all Geometry types).

DenisCarriere:
> How should we handle MultiPolygon Geometries?

- `ringIndex` will always be the "inner/outer" rings of the Polygons

- `ringSubIndex` will be the index for the MultiParts, if Feature only has 1 part then `ringSubIndex=0` (maybe call it `multiPartIndex`?)

rowanwins:
[...]

PR #930, opened 2017-09-04
New `@turf/meta` methods `lineEach` and `lineReduce`

stebogit:
Following https://github.com/Turfjs/turf/issues/881
@rowanwins @DenisCarriere not sure if the tests I set up are enough though.
Should these methods support also collections?

```js
/**
 * Iterate over line or ring coordinates in LineString, Polygon, MultiLineString, MultiPolygon Features or Geometries,
 * similar to Array.forEach.
 *
 * @name lineEach
 * @param {Geometry|Feature<LineString|Polygon|MultiLineString|MultiPolygon>} geojson object
 * @param {Function} callback a method that takes (currentLine, lineIndex, lineSubIndex)
 * @example
 * var mtLn = turf.multiLine([
 *   turf.lineString([[26, 37], [35, 45]]),
 *   turf.lineString([[36, 53], [38, 50], [41, 55]])
 * ]);
 *
 * turf.lineEach(mtLn, function (currentLine, lineIndex) {
 *   //=currentLine
 *   //=lineIndex
 * });
 */
function lineEach(geojson, callback) {
```

```js
/**
 * Reduce features in any GeoJSON object, similar to Array.reduce().
 *
 * @name lineReduce
 * @param {Geometry|Feature<LineString|Polygon|MultiLineString|MultiPolygon>} geojson object
 * @param {Function} callback a method that takes (previousValue, currentFeature, featureIndex)
 * @returns {*} The value that results from the reduction.
 * @example
 * var mtp = turf.multiPolygon([
 *   turf.polygon([[[12,48],[2,41],[24,38],[12,48]], [[9,44],[13,41],[13,45],[9,44]]]),
 *   turf.polygon([[[5, 5], [0, 0], [2, 2], [4, 4], [5, 5]]])
 * ]);
 *
 * turf.lineReduce(mtp, function (previousValue, currentLine, lineIndex, lineSubIndex) {
 *   //=previousValue
 *   //=currentLine
 *   //=lineIndex
 *   //=lineSubIndex
 *   return currentLine
 * }, 2);
 */
function lineReduce(geojson, callback, initialValue) {
```

## `lineEach` & `lineReduce` supports the following:
- ✅ `lineString` based Geometries
- ❌  throws error at `Point` & `MultiPoint`
- ✅ `Feature` & `Geometry`
- ❌ throws error at `FeatureCollection` & `GeometryCollection`
- ❌ throws error at `Feature<null>`


stebogit:
@DenisCarriere this is new... 🤔 
Did you modified the `types` test? I don't understand though why would it fail here since this branch is straight out of master.

screen shot 2017-09-04 at 2 08 21 am

DenisCarriere:
@stebogit Ops, yes I did change types... it was kinda experimental at first... I'll look into that.

DenisCarriere:
> not sure if the tests I set up are enough though. Should these methods support also collections?

So far all the `@turf/meta` methods support Feature & FeatureCollection with all geometry types, might be worth doing the same for this, even if it's a Geometry Point, it would simply return no callbacks but still be a valid input.

DenisCarriere:
## `lineEach` & `lineReduce` supports the following:
- ✅ Only `lineString` based Geometries (throws errors at `Point` & `MultiPoint`)
- ✅ Only `Feature` & `Geometry` (throws an error at `FeatureCollection` & `GeometryCollection`)

Issue #916, opened 2017-08-24
Buffer zone transform


ArtCodeKarine:
Hello everybody, 
I wish to know how i can resolve my problem :
when the buffer zone touches itself, it makes only one area instead of making a "hole" in the middle. Even if the zone is not close. Do you have a solution for it to overlap and it does not merge?
![buffer1](https://user-images.githubusercontent.com/17497949/29664758-5a52ba6e-88d1-11e7-9ded-4ed3847429a5.PNG)
![buffer2](https://user-images.githubusercontent.com/17497949/29664760-5b6170a8-88d1-11e7-9d54-de866e38b68b.PNG)

rowanwins:
Hi @ArtCodeKarine 

Short answer is there is probably no a answer at the moment.

Unfortunately the buffer module currently uses the `jsts` library under the hood so we had limited ability to apply fixes for bugs like these. We're currently attempting to remove that dependency but as Im sure you can see it's quite complicated so it's still a little way off sorry :(

ArtCodeKarine:
@rowanwins  Thank's for your answer :D

stebogit:
@ArtCodeKarine have you tried with [`@turf/transform-scale`](https://github.com/Turfjs/turf/tree/master/packages/turf-transform-scale)? It might work for you, just know you'll have to calculate your `factor` and define the `origin` (see #849).

DenisCarriere:
Thanks for reporting this @ArtCodeKarine , never knew about this strange behavior, this is definitely a bug in `jsts` which we can't control 😢.

stebogit:
Fixed with #938

Issue #900, opened 2017-08-14
buffer issue


sc-keyzo:
Hi 

Using the following lines of code with @turf/turf 4.6.0 results in a buffer with kinks

`var buffer = turf.buffer(route, 100,  'miles');`

'route' in the above code is the linestring defined in the attached geojson file.  This geojson file also shows the result of the turf.buffer call.

[buffer-error.geojson.zip](https://github.com/Turfjs/turf/files/1222713/buffer-error.geojson.zip)

Have i specified something incorrectly?

regards
Simon

DenisCarriere:
The kinks are definitely causing some issues in the Buffer operations, both in `jsts` & the newly proposed https://github.com/Turfjs/turf/pull/876.

Wouldn't know what to do for temporary solution, I've tried `@turf/clean-coords` & `@turf/simplify` and nothing seems to fix that LineString.

## JSTS

![image](https://user-images.githubusercontent.com/550895/29280937-1398944c-80eb-11e7-8410-1a69a79a480c.png)

stebogit:
Fixed with #938

sc-keyzo:
Hi guys, thanks for your work on this, would it be possible to provide me with an example of how i would use turf/projection to fix my issue?

thanks
Simon

DenisCarriere:
I'll publish a patch release with this fix

Issue #815, opened 2017-06-22
Buffer for lineString not consistent width


stephanbogner:
I recently had the need to outline a lineString and it seems that using 'buffer' is not doing it as I expected.
I expected it to be like a stroke, so width is everywhere the same, which it clearly isn't:
<img width="670" alt="bildschirmfoto 2017-06-22 um 17 13 32" src="https://user-images.githubusercontent.com/7897006/27441562-32fc7aa2-576e-11e7-982f-25a2f5b6d198.png">

You can see [here](https://gist.github.com/anonymous/5a9d5be2204451e4001b75890406c70e) my implementation vs. buffer.

- Red is created using 'buffer'
- Blue by [my own implementation](https://gist.github.com/stephanbogner/8a2216817ad143b6c455311355733041)
- Line is original geometry

→ So my question is if that is intended behaviour and if so for what reason?

rowanwins:
Thanks @stephanbogner for highlighting the issue, it does indeed look like something is a bit off... 

Currently the buffer functionality is provided by the `jsts` library so we have limited control over that. However it's been a longstanding task/dream to remove that dependency. This is another helpful reason to keep us pursue that so thank you :)

stephanbogner:
Thank you very much for explaining this dependency – I was not aware.

DenisCarriere:
The main issue with `buffer` was applying a buffer which is equidistant (ref issue https://github.com/Turfjs/turf/issues/660).

Before it use to do this:
![image](https://user-images.githubusercontent.com/550895/27871275-2d541da8-6173-11e7-9eff-5733b9bacdf2.png)

To fix this, we applied a project & reproject to the decimal degrees coordinates.

https://github.com/Turfjs/turf/blob/master/packages/turf-buffer/index.js#L105-L110

There's definitely room for improvement in the `@turf/buffer`.

stebogit:
Fixed with #938

PR #938, opened 2017-09-11
Buffer fixes

stebogit:
I was working on solving #900, and I fixed it simply using the new `@turf/projection`, but then find out that seems to fix also #815 and #916. 😃 

Still for some reason the actual size of the buffer is slightly off, i.e. smaller than the expected one.

cc: @sc-keyzo @ArtCodeKarine @stephanbogner


DenisCarriere:
@stebogit Wow 🎉 Really cool! The `@turf/projection` module is going to be VERY useful! (we all knew that 😉 )

packages/turf-buffer/index.js
```diff
@@ -94,16 +94,24 @@ function buffer(geojson, radius, units, steps) {
...
-    var projected = {
-        type: geometry.type,
-        coordinates: projectCoords(geometry.coordinates, projection)
-    };
+    var projected;
+    var bbox = turfBbox(geojson);
+    var needsTransverseMercator = bbox[1] > 50 && bbox[3] > 50;
+
+    if (needsTransverseMercator) {
+        var projection = defineProjection(geometry);
```
DenisCarriere: @stebogit Did you want to move over this `defineProjection` to `@turf/projection` (as `define`).

```js
/**
 * Define Transverse Mercator projection
 *
 * @private
 * @param {Geometry|Feature<any>} geojson Base projection on center of GeoJSON
 * @returns {GeoProjection} D3 Geo Transverse Mercator Projection
 */
function defineProjection(geojson) {
    var coords = center(geojson).geometry.coordinates.reverse();
    var rotate = coords.map(function (coord) { return -coord; });
    var projection = d3.geoTransverseMercator()
        .center(coords)
        .rotate(rotate)
        .scale(6373000);

    return projection;
}
```

stebogit:
Ooops... 😨  I was convinced this branch had been merged already... 😅

packages/turf-buffer/index.js
stebogit: @DenisCarriere wouldn't it be even more useful having a `@turf/projection.toTransverse` method for the actual conversion? I mean, why would you define the (Transverse in this case) projection but for converting the coordinates?
 
In that regard, at this point we might need to re-think the structure of the `projection` module to leave it open to accept maybe other projections in the future, if ever needed, like:
- `@turf/projection.toMercator`
- `@turf/projection.fromMercator`
- `@turf/projection.toTransverse`
- `@turf/projection.fromTransverse`
- ...

knowing the base projection for Turf is `WGS84`

DenisCarriere:
Lol it can be, I only looked at it real quick, when I have more time I'll dive into it more, unless you feel it's good to merge.

stebogit:
At this point I guess we could merge since all tests pass and they actually look better (less distortion); we can always reopen any issue if necessary.
May I have the honor? 😃

DenisCarriere:
@stebogit merge away! 👍

stebogit:
Woo-hoo!! :bowtie:

ArtCodeKarine:
Hello @stebogit, 
I try to install your new module with "npm install @turf/turf", but it doesn't work. Your module isn't display in my project.
So is it normal that your package is not in the file : turf/packages/turf/package.json ?
Thank's for your answer.

stebogit:
@ArtCodeKarine what module are you talking about? Do you mean the updated `@turf/buffer` code?
@DenisCarriere is going to publish a new [patch release](https://github.com/Turfjs/turf/issues/900#issuecomment-329179539) (I guess v4.7.4) with those changes as soon as possible, so stay tuned! 😜

DenisCarriere:
Just published a patch release for `@turf/buffer`, try out:

```bash
$ npm install --save @turf/buffer@4.7.4
```

ArtCodeKarine:
@DenisCarriere @stebogit thanks, I will try it :D

ArtCodeKarine:
I think i have to wait the new version of turf because my application is running with nodejs and bower. Bower uses the @turf/turf folder and this doesn't contain the patch version of the buffer.

Issue #198, opened 2015-01-17
turf.nearest for linestrings


peterqliu:
For example, when I want to find the closest gas station to the _stretch of freeway_ where I'll likely be low on gas.

This is better than doing turf.nearest on a single point, for two reasons:
- Turf.nearest on a linestring is likely to return nearer results than on a point, because it's a larger search area.
- when looking for gas stations, I want to find one that's not too far out of my way, and I'm willing to go a couple more exits before turning off if that one is more convenient. I don't really care _where_ the station is, so I don't have a specific point in mind to search around.

nevi-me:
The challenge with nearest for linestrings is that a linestring could have only 2 points, making it difficult to run nearest through each sets of coordinates in that point. I solved this problem by taking linestring and 'complexifying' it by adding points on the linestring within a certain distance of each other. Then I can search for nearest in that list of coordinates on the linestring.

I did this in rwt-to/GeoJSON-Tools (https://github.com/rwt-to/GeoJSON-Tools/commit/35abcf081fc4d13d9552ff69193236b49ee34da6), but I'm not promoting that you use the tool. I just saw Turfjs today, and if someone would like to port the 'complexify' method from GeoJSON-Tools, you're welcome to. 
We use GeoJSON-Tools internally, and there's lots of work to do, so I haven't committed anything to it in some time now.

rowanwins:
Gday @peterqliu 

I've had a shot at creating this [module](https://github.com/rowanwins/turf-nearestPointToLine), basically you input a collection of points and a linestring, and it returns the nearest point as well as the distance that point is from the line, this is included as a property on the feature. The user can also specify the distance units as a parameter (as per the existing turf distance module) which defaults to kilometers.

It was actually quite simple to do as it builds on a number of existing turf modules, namely [point-on-line](http://turfjs.org/static/docs/module-turf_point-on-line.html) and [distance](http://turfjs.org/static/docs/module-turf_distance.html).

cc @tcql & @morganherlocker . I still need to write some tests but if you're interested in transferring ownership at some stage let me know.

Cheers
Rowan

peterqliu:
@rowanwins thank you so much for jumping in here. hoping to see this in the not-too-distance turf future.

stebogit:
@rowanwins @DenisCarriere it seems to me that `@turf/point-on-line` is exactly what this issue refers to, isn't it?
[...]

PR #939, opened 2017-09-12
New module `@turf/nearest-point-to-line`

stebogit:
```js
/**
 * Returns the closest {@link Point|point}, of a {@link FeatureCollection|collection} of points, to a {@link LineString|line}.
 * The returned point has a `dist` property indicating its distance to the line.
 *
 * @name nearestPointToLine
 * @param {FeatureCollection|GeometryCollection<Point>} points collection
 * @param {Geometry|Feature<LineString>} line Feature
 * @param {string} [units=kilometers] can be degrees, radians, miles, or kilometers
 * @returns {Feature<Point>} the closest point
 * @example
 * var points = featureCollection([point([0, 0]), point([0.5, 0.5])]);
 * var line = lineString([[1,1], [-1,1]]);
 *
 * var nearest = turf.nearestPointToLine(points, line);
 *
 * //addToMap
 * var addToMap = [nearest, line];
 */
module.exports = function (points, line, units) {
```

Ref #198

Not sure though how to interpret [this](https://github.com/Turfjs/turf/issues/198#issuecomment-309587582) discussion:
> How would the case of a parallel line be handled? There is no single closest point. 

> If in doubt @morganherlocker the answer is c - we'll just place something artibitralily 😜!

I don't see the case of parallel lines here @morganherlocker and @rowanwins.

image


packages/turf-nearest-point-to-line/index.js
```diff
@@ -0,0 +1,45 @@
...
+var collectionOf = invariant.collectionOf;
+
+/**
+ * Returns the closest {@link Point|point}, of a {@link FeatureCollection|collection} of points, to a {@link LineString|line}.
+ * The returned point has a `dist` property indicating its distance to the line.
+ *
+ * @name nearestPointToLine
+ * @param {FeatureCollection<Point>} points collection
+ * @param {Feature<LineString>} line Feature
+ * @param {string} [units=kilometers] can be degrees, radians, miles, or kilometers
```
rowanwins: > @param {string} [units=kilometers] can be degrees, radians, miles, or kilometer

My only comment would be saying that this is for the output property. So perhaps
`@param {string} [units=kilometers] unit of the output distance property, can be degrees, radians, miles, or kilometer`

packages/turf-nearest-point-to-line/index.js
```diff
@@ -0,0 +1,45 @@
...
+    var pt = null;
+    featureEach(points, function (point) {
+        var d = pointToLineDistance(point, line, units);
+        if (d < dist) {
+            dist = d;
+            pt = point;
+        }
+    });
+
+    pt.properties.dist = dist;
```
rowanwins: I'd call the output property `distance` (rather than `dist`) because it's slightly more explicit/obvious.

Perhaps this should be an upstream change in `@turf/pointOnLine` as well...

Just my personal preference

stebogit:
@rowanwins I found some issues with (I think) the `@turf/point-to-line-distance` when testing for two points at the same distance, so I'll be working on that.
Thanks for the suggestion so far. 👍

stebogit:
@rowanwins @DenisCarriere 
Guys I need your input here, cause I can't figure it out myself... 🤔 
I am convinced that in both the following two examples the points should be at the same distance from the line in the middle, but for the second case `@turf/point-to-line-distance` gives like 0.000011m difference and I don't understand if they are truly not at the same distance or is the new module that is actually miscalculating (for the first example with the line N-S it returns equal distance).
Suggestions? 

https://gist.github.com/anonymous/491d71542965d44b07f19a14b20a94d5?short_path=000ca83
screen shot 2017-09-15 at 8 49 52 pm

https://gist.github.com/anonymous/dae55177159c3275f141b2accb96d1f4?short_path=5a48b33
screen shot 2017-09-15 at 8 50 53 pm

DenisCarriere:
> I am convinced that in both the following two examples the points should be at the same distance from the line in the middle

I wouldn't worry too much about the 0.000011m difference, we can definitely add them to the test cases and look into this later.

> The only other thought is what we if we have 2 points at exactly the same distance? Could we pass back an array of points?

I don't think we should return an Array of points (or FeatureCollection of Points), the expected output would be a single point and the chances that a point would fall exactly within a 0.000001 distance would be very slim. Any points which fall exactly on the same distance would simply be ignored and the first closest point would be returned.

> I'd call the output property distance (rather than dist) because it's slightly more explicit/obvious.

We've been using `dist` in the past modules, I would stick with that naming convention instead of a more explicit name (which I normally do prefer).

packages/turf-nearest-point-to-line/index.d.ts
```diff
@@ -0,0 +1,12 @@
+/// <reference types="geojson" />
+
+type Points = GeoJSON.FeatureCollection<GeoJSON.Point>;
+type LineString = GeoJSON.LineString;
```
DenisCarriere: This would only allow Geometry LineString and not a `Feature<LineString>`:

**Correct definition for both Geometry & Feature support:**
```ts
type LineString = GeoJSON.Feature<GeoJSON.LineString> | GeoJSON.LineString;
```

packages/turf-nearest-point-to-line/index.js
```diff
@@ -0,0 +1,45 @@
...
+var featureEach = meta.featureEach;
+var collectionOf = invariant.collectionOf;
+
+/**
+ * Returns the closest {@link Point|point}, of a {@link FeatureCollection|collection} of points, to a {@link LineString|line}.
+ * The returned point has a `dist` property indicating its distance to the line.
+ *
+ * @name nearestPointToLine
+ * @param {FeatureCollection<Point>} points collection
+ * @param {Feature<LineString>} line Feature
```
DenisCarriere: Should we also support LineString as a Geometry?

```js
/**
 * @param {Feature<LineString>|LineString} line Line Feature
```

packages/turf-nearest-point-to-line/index.d.ts
```diff
@@ -0,0 +1,12 @@
...
+type Point = GeoJSON.Feature<GeoJSON.Point> | GeoJSON.Point;
+
+/**
+ * http://turfjs.org/docs/#nearestpointtoline
+ */
+declare function nearestPointToLine(points: Points, line: LineString): Point;
```
DenisCarriere: Missing `units` as an optional param

packages/turf-nearest-point-to-line/index.js
```diff
@@ -0,0 +1,47 @@
...
+var featureOf = invariant.featureOf;
+var featureEach = meta.featureEach;
+var collectionOf = invariant.collectionOf;
+
+/**
+ * Returns the closest {@link Point|point}, of a {@link FeatureCollection|collection} of points, to a {@link LineString|line}.
+ * The returned point has a `dist` property indicating its distance to the line.
+ *
+ * @name nearestPointToLine
+ * @param {FeatureCollection<Point>} points Point Collection
```
DenisCarriere: Could we also support `GeometryCollection`

DenisCarriere:
Updated a few things:
- [x] GeometryCollection support
- [x] Geometry LineString support (also for `@turf/point-to-line-distance`)
- [x] Add tests for TypeScript
- [x] Update JSDocs

CC: @stebogit

DenisCarriere:
Easier to preview the test outputs when using GeometryCollection:

image

stebogit:
Yay!! 🎉

Issue #735, opened 2017-05-12
Propose isParallel as new module


DenisCarriere:
# Propose `@turf/boolean-parallel` module

To complement the `boolean` themed modules, this would only input **(Multi)LineString** geometry types and should be able to detect if the two inputs are parallel or not.

The majority of the code is already built in `@turf/line-offset` in [`isParallel()`](https://github.com/Turfjs/turf/blob/d2a9dcf8df0519b11e736e0e6d69f8ef5cbf2ed9/packages/turf-line-offset/index.js#L179) which was inspired by [`node-intersection`](https://github.com/rook2pawn/node-intersection).

The only trick with this module would be to support LineString inputs with more than 2 vertices, `@turf/line-segment` can help iterate over both inputs.

DenisCarriere:
I don't know if this definition is accurate or not, however I've seen this type of **line parallel** method in a few modules so far and it would be nice to extract that into a single boolean method (@psalaets's [line-intersect](https://github.com/psalaets/line-intersect/blob/master/lib/check-intersection.js#L19-L29) is another library that has a 2-vertex parallel boolean method).

This would definitely be a bit trickier to support in 3D space and would most likely only be implemented in a 2D plane.

stebogit:
@DenisCarriere couldn't we simply calculate the slope/bearing of the segments and check if they are equal? I believe that is a correct definition for parallel segments on a plane. 
This would also make easy calculation for a `boolean-perpendicular` (`abs(bearing1 - bearing2) == 90`), if of any interest.
We could use `@turf-bearing` or `@turf-rhumb-bearing`, or give the user the option to chose between the two.

The definition of 'parallel' should be well defined for multi-segment lines, as it generally applies to straight lines or segments.
**Example**: would you define these two multi-segment lines as parallel?
![screen shot 2017-06-11 at 10 55 53 pm](https://user-images.githubusercontent.com/12717225/27020849-2e4b4f2e-4ef9-11e7-98a2-5c6a6a0e26fb.png)

For more complex geometries, how would you define if two `MultiLineString`s are parallel? All `LineString`s should be parallel in pairs maybe? But how can/would you pair the lines though?
I would restrict this module to `LineString` inputs, and leave the user to iterate through `MultiLineString`s as he considers appropriate.

DenisCarriere:
👍 I agree that it might be best to restrict it to only `LineString`, it can get very messy to support `MultiLineString`.

dpmcmlxxvi:
[...]

PR #941, opened 2017-09-14
New module `@turf/boolean-parallel`

stebogit:
Ref. https://github.com/Turfjs/turf/issues/735
```js
/**
 * Boolean-Parallel returns True if each segment of `line1` is parallel to the correspondent
 * segment of `line2`
 *
 * @name booleanParallel
 * @param {Geometry|Feature<LineString>} line1 GeoJSON Feature or Geometry
 * @param {Geometry|Feature<LineString>} line2 GeoJSON Feature or Geometry
 * @returns {Boolean} true/false if the lines are parallel
 * @example
 * var line1 = turf.lineString([[0, 0], [0, 1]]);
 * var line2 = turf.lineString([[1, 0], [1, 1]]);
 *
 * turf.booleanParallel(line1, line2);
 * //=true
 */
module.exports = function (line1, line2) {
```

The module supports only `LineStrings` (feature or geometry), leaving the user to decide how to apply the comparison among multi lines.

**Note:** the definition of comparison is based on segment having the same slope on the map plain, i.e. equal `rhumb-bearing`.
I don't know which definition has, if existing, parallelism on a sphere, but I feel it would be beyond Turf applications.


packages/turf-boolean-parallel/index.js
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Returns Feature's type
+ *
+ * @private
+ * @param {Geometry|Feature<any>} geojson Geometry or Feature
+ * @param {string} name of the variable
+ * @returns {string} Feature's type
+ */
+function getType(geojson, name) {
```
DenisCarriere: @stebogit This private looks great, we should add this to [`@turf/invariant`](https://github.com/Turfjs/turf/blob/master/packages/turf-invariant).

Similar to [`getGeomType`](https://github.com/Turfjs/turf/blob/master/packages/turf-invariant/index.js#L192)

stebogit: Yeah, I always wondered why we didn't have that there :smile:

DenisCarriere: 👍 I'll add it

packages/turf-boolean-parallel/test.js
```diff
@@ -13,7 +13,6 @@ test('turf-boolean-parallel', t => {
         const [line1, line2] = geojson.features;
         const result = booleanParallel(line1, line2);
 
-        if (process.env.SHAPELY) shapely.contains(line1, line2).then(result => t.true(result, '[true] shapely - ' + name));
```
stebogit: @DenisCarriere why did you remove this?
Many other boolean modules have that, some of them also have
```js
if (process.env.JSTS) t.false(booleanJSTS('within', feature1, feature2), '[false] JSTS - ' + name);
```
What's their use exactly?

stebogit:
Woo-hoo!! 😄:+1:

PR #942, opened 2017-09-16
Add getType to `@turf/invariant`

DenisCarriere:
Ref: https://github.com/Turfjs/turf/pull/941#discussion_r139294209

```js
/**
 * Get GeoJSON object's type, Geometry type is prioritize.
 *
 * @param {GeoJSON} geojson GeoJSON object
 * @param {string} [name] of the variable
 * @returns {string} GeoJSON type
 * @example
 * var point = {
 *   "type": "Feature",
 *   "properties": {},
 *   "geometry": {
 *     "type": "Point",
 *     "coordinates": [110, 40]
 *   }
 * }
 * var geom = turf.getType(point)
 * //="Point"
 */
function getType(geojson, name) {}
```


packages/turf-invariant/index.js
```diff
@@ -195,6 +195,34 @@ function getGeomType(geojson) {
...
+ * var geom = turf.getType(point)
+ * //="Point"
+ */
+function getType(geojson, name) {
+    if (!geojson) throw new Error('geojson is required');
+    // GeoJSON Feature & GeometryCollection
+    if (geojson.geometry && geojson.geometry.type) return geojson.geometry.type;
+    // GeoJSON Geometry & FeatureCollection
+    if (geojson.type) return geojson.type;
+    if (name) throw new Error('Invalid GeoJSON object for ' + name);
```
stebogit: 👍

stebogit: 👍

packages/turf-invariant/index.js
```diff
@@ -195,6 +195,34 @@ function getGeomType(geojson) {
...
+ *   "geometry": {
+ *     "type": "Point",
+ *     "coordinates": [110, 40]
+ *   }
+ * }
+ * var geom = turf.getType(point)
+ * //="Point"
+ */
+function getType(geojson, name) {
+    if (!geojson) throw new Error('geojson is required');
```
stebogit: Also here it should probably use `name` if available. Like maybe:
```js
if (!geojson) throw new Error((name || 'geojson') + ' is required');
```

DenisCarriere: That looks good to me!

DenisCarriere:
> Don't you think getType could actually replace getGeomType in v5.0.0?

Agreed, `getType` would be able to replace `getGeomType`, we could drop it in the next major release.

packages/turf-invariant/index.js
```diff
@@ -189,10 +191,36 @@ function getGeom(geojson) {
  * var geom = turf.getGeomType(point)
  * //="Point"
  */
-function getGeomType(geojson) {
-    if (!geojson) throw new Error('geojson is required');
-    var geom = getGeom(geojson);
-    if (geom) return geom.type;
+function getGeomType(geojson, name) {
+    console.warn('getGeomType will be deprecated in v5.0.0 in favor of getType');
```
DenisCarriere: @stebogit added deprecated message to this method, I don't see any issues simply wrapping it with `getType` at the moment, the output is practically the same with the exception of raised errors.

Issue #940, opened 2017-09-12
updating @turf/inside or @turf/within from v4.1 to v4.2 causes "coordinates must only contain numbers" error


barbalex:
This happens in all versions between 4.2.0 and 4.6.0 (it seems that in npm other versions exist than on github).

It happens:

- if I only update within
- if I only update inside
- when I update both

This is the module that runs when the error happens: https://github.com/barbalex/apf2/blob/75ad2a1efc2c57d006e52d93fda8eee1b6d6d9b7/src/modules/tpopIdsInsideFeatureCollection.js

I have checked the data generated for this action several times. There are multiple checks that make sure only valid numbers are passed.

This is the console message:

    index.js:81 Uncaught Error: coordinates must only contain numbers
        at containsNumber (index.js:81)
        at getCoords (index.js:58)
        at ./node_modules/@turf/within/node_modules/@turf/inside/index.js.module.exports (index.js:37)
        at ./node_modules/@turf/within/index.js.module.exports (index.js:44)
        at ./src/modules/tpopIdsInsideFeatureCollection.js.__webpack_exports__.a (tpopIdsInsideFeatureCollection.js:50)
        at Object../src/store/extend/map.js.__webpack_exports__.a.Object.mapFilter.tpop.Object.name (map.js:168)
        at trackDerivedFunction (mobx.module.js:2812)
        at ComputedValue../node_modules/mobx/lib/mobx.module.js.ComputedValue.computeValue (mobx.module.js:1407)
        at ComputedValue../node_modules/mobx/lib/mobx.module.js.ComputedValue.trackAndCompute (mobx.module.js:1397)
        at ComputedValue../node_modules/mobx/lib/mobx.module.js.ComputedValue.get (mobx.module.js:1360)
        at shouldCompute (mobx.module.js:2763)
        at ComputedValue../node_modules/mobx/lib/mobx.module.js.ComputedValue.get (mobx.module.js:1359)
        at shouldCompute (mobx.module.js:2763)
        at ComputedValue../node_modules/mobx/lib/mobx.module.js.ComputedValue.get (mobx.module.js:1359)
        at Object.get (mobx.module.js:1584)
        at myChildren (index.js:135)
        at _class.Projekte (index.js:176)
        at _class.render (index.js:753)
        at Object.allowStateChanges (mobx.module.js:921)
        at index.js:645
        at trackDerivedFunction (mobx.module.js:2812)
[...]

PR #947, opened 2017-09-18
Implement allowNull options to @turf/meta fixes #946

### LINKED ISSUE #946: include `allowNull` optional param to featureEach & featureReduce
## include `allowNull` optional param to `featureEach` & `featureReduce`

Ref: https://github.com/Turfjs/turf/issues/940

Add an extra optional property `allowNull=false` to `featureEach` / `featureReduce` in [`@turf/meta`](https://github.com/Turfjs/turf/tree/master/packages/turf-meta) to exclude or include null geometries from being returned in the callback.

This would solve a lot of problems from modules which depend on `@turf/meta`.

The new `options` param syntax should be used for this new feature.

## JSDocs

```js
/**
 * Iterate over features in any GeoJSON object, similar to
 * Array.forEach.
 *
 * @name featureEach
 * @param {(FeatureCollection|Feature|Geometry)} geojson any GeoJSON object
 * @param {Object} [options] Extra options, if none are provided, this param defaults as callback.
 * @param {boolean} [options.allowNull=false] exclude or include null geometries from callback
 * @param {Function} callback a method that takes (currentFeature, featureIndex)
 * @example
 * var features = turf.featureCollection([
 *   turf.point([26, 37], {foo: 'bar'}),
 *   turf.point([36, 53], {hello: 'world'})
 * ]);
 *
 * turf.featureEach(features, function (currentFeature, featureIndex) {
 *   //=currentFeature
 *   //=featureIndex
 * });
 */
function featureEach(geojson, options, callback) {
```

## To-Do

- [x] Add `allowNull` optional param:
  - [x] `featureEach` 
  - [x] `featureReduce` 
  - [x] `geomEach`
  - [x] `geomReduce` 
  - [x] `flattenEach` 
  - [x] `flattenReduce`

DenisCarriere:
Implements: #946
Ref Issues: #940 

## JS Example

```js
var features = turf.featureCollection([
  turf.point([26, 37], {foo: 'bar'}),
  turf.point([36, 53], {hello: 'world'}),
  turf.feature(null)
]);

turf.featureEach(features, {allowNull: false}, feature => {
  // Callbacks 2 features
});

turf.featureEach(features, {allowNull: true}, feature => {
  // Callbacks 3 features
});

// Backwards compatible with TurfJS v4.0
// allowNull defaults to `false`
// most common errors in TurfJS are related to null geometries being contained in FeatureCollections
turf.featureEach(features, feature => {
  // Callbacks 2 features
});
```

## Up for debate

`allowNull` currently defaults to `false`, there could be an argument to set this default as `true`. My personal preference is `allowNull=false`.

### DIFF
--- a/packages/turf-concave/test/out/support-null-geometry.geojson
+++ b/packages/turf-concave/test/out/support-null-geometry.geojson
@@ -3,10 +3,7 @@
 	"features": [
 		{
 			"type": "Feature",
-			"properties": {
-				"marker-color": "#f0f",
-				"marker-size": "small"
-			},
+			"properties": {},
 			"geometry": null
 		},
 		{
--- a/packages/turf-meta/README.md
+++ b/packages/turf-meta/README.md
@@ -138,6 +138,34 @@ Callback for propEach
 -   `geojson`  
 -   `callback`  
 
+# propReduceCallback
+
+Callback for propReduce
+
+The first time the callback function is called, the values provided as arguments depend
+on whether the reduce method has an initialValue argument.
+
+If an initialValue is provided to the reduce method:
+
+-   The previousValue argument is initialValue.
+-   The currentValue argument is the value of the first element present in the array.
+
+If an initialValue is not provided:
+
+-   The previousValue argument is the value of the first element present in the array.
+-   The currentValue argument is the value of the second element present in the array.
+
+**Parameters**
+
+-   `previousValue` **Any** The accumulated value previously returned in the last invocation
+    of the callback, or initialValue, if supplied.
+-   `currentProperties` **Any** The current properties being processed.
+-   `featureIndex` **[number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number)** The index of the current element being processed in the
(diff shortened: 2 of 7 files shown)


stebogit:
@DenisCarriere I'll just comment only on this one but it applies to the other method as well.
The tests you put in place will certainly prove me wrong, but how can this
```js
turf.featureEach(features, {allowNull: false}, feature => {
  // Callbacks 2 features
});
```
be backward compatible and not breaking code based on the current version?

I thought this would have been the right order:
```js
turf.featureEach(features, feature => {
  // Callbacks 2 features
}, {allowNull: false});
```
I'm probably missing something here.

DenisCarriere:
@stebogit Agreed it's a bit confusing having options, it's even more confusing when handling `featureReduce` since if the `options` are defined after the callback it's in the same position as the `initialValue` which is defined as `any` value.

🤔 I think I'm just going to drop this PR... no `options` in all of these `Array.map` & `Array.reduce` typed methods.

DenisCarriere:
Going to close this PR, doing this I found a few issues in the tests, I'm going a new PR related to only changes related to the `test.js`

PR #952, opened 2017-09-18
Revert @turf/intersect back to null fixes #951

### LINKED ISSUE #951: Intersect documentation is outdated
The documentation at http://turfjs.org/docs/#intersect says that `intersect` returns `undefined` for empty intersections, but it seems that it now returns an empty `Feature` instead, since `4.7.1`.

I guess that should have been a major version update, since that is a breaking change. It breaks what is explicitly documented (so `intersect(a, b) === undefined` does not work anymore), and related idioms (like `intersect(a, b) || c`).

DenisCarriere:
Ref: https://github.com/Turfjs/turf/issues/951

### DIFF
--- a/packages/turf-intersect/bench.js
+++ b/packages/turf-intersect/bench.js
@@ -1,21 +1,14 @@
-var intersect = require('./');
-var Benchmark = require('benchmark');
-var fs = require('fs');
+const fs = require('fs');
+const path = require('path');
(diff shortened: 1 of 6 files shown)


DenisCarriere:
If we decide to change it back to `undefined` I'm ok with it, but `null` value makes the most sense to me and this is what @stebogit used as the last edit.

Issue #788, opened 2017-06-14
New module @turf/shortestPath


dhivehi:
Requesting for a feature to find shortest path from point A to point B without colliding with any polygon
![screenshot-dhivehi mv-2017-06-14-10-52-11](https://user-images.githubusercontent.com/18715293/27117444-0b02d9a8-50f0-11e7-9151-4dd8b6055bec.png)

rowanwins:
Gday @dhivehi 

That would be a nifty module. There is an interesting answer here that might help get someone started.
https://stackoverflow.com/a/6793562

DenisCarriere:
👍 That's awesome! I think the module name sounds appropriate `@turf/shortestPath`

dhivehi:
Hope this link would also be useful: https://github.com/TaipanRex/pyvisgraph

rowanwins:
Thanks for the link @dhivehi that is helpful.

My only other question is should the person be able to pass on a buffer distance as a parameter? Or alternatively they could pass in buffered polygons...

If I was a betting man I'd say your interest is shipping given this module request and the tangents request as well  :)

DenisCarriere:
> My only other question is should the person be able to pass on a buffer distance as a parameter?

👎 I don't think we should mix another module into this, the user can easily apply the buffer to the obstacle polygons before using `shortestPath`.

dhivehi:
I think adding a 4th parameter as buffer distance would be nice.

dpmcmlxxvi:
@rowanwins Depending on which algorithm gets implemented there might be a "buffer" like parameter needed. Some shortest path algorithms use a gridding approach, so the size of the grid cells is a buffer like parameter. Some algorithms don't. Pros and cons to both approaches.

dhivehi:
I think adding buffer parameter is useful.

DenisCarriere:
That works for me if you feel it would be useful to have, you would have two extra params (distance & units)

bulutkartal:
from a ship captain perspective; distance to polygons parameter would be great. So they can use it for boat/ship routing. They can get the lands as polygons, set a safe distance and create their routes by using this module.

DenisCarriere:
@bulutkartal 👍 Great feedback, if you have some sample data you'd like to provide for Ship examples, feel free to send a PR with your GeoJSON examples.

You can simply add them to the `turf/packages/turf-shortest-path/test/in` folder.

This module will be published in the next Major release `v5.0`, (very soon 🤞).
[...]

PR #956, opened 2017-09-21
New module `@turf/shortest-path`

stebogit:
Ref. https://github.com/Turfjs/turf/issues/788

```js
/**
 * Returns the shortest {@link LineString|path} from {@link Point|start} to {@link Point|end} without colliding with
 * any {@link Feature} in {@link FeatureCollection<Polygon>| obstacles}
 *
 * @name shortestPath
 * @param {Geometry|Feature<Point>} start point
 * @param {Geometry|Feature<Point>} end point
 * @param {GeometryCollection|FeatureCollection<Polygon>} obstacles polygons
 * @param {Object} [options={}] optional parameters
 * @param {string} [options.units="kilometers"] can be degrees, radians, miles, kilometers, ...
 * @param {number} [options.resolution=100] distance between matrix points on which the path will be calculated
 * @returns {Feature<LineString>} shortest path between start and end
 * @example
 * var start = turf.point([-5, -6]);
 * var end = turf.point([9, -6]);
 * var obstacles = turf.featureCollection([turf.polygon([[0,-7],[5,-7],[5,-3],[0,-3],[0,-7]]));
 *
 * var path = shortestPath(start, end, obstacles);
 *
 * //addToMap
 * var addToMap = [start, end, obstacles, path];
 */
module.exports = function (start, end, obstacles, options) {
```

There is a lot of room for improvements, but what do you think of this first draft @DenisCarriere @rowanwins ? Any suggestion to improve speed and/or final result is more than welcome.

screen shot 2017-09-21 at 1 14 46 am


rowanwins:
Hey @stebogit

There a bunch of known algorithms for this sort of thing. For example try googling "dijkstra's algorithm", I think this is a semi-famous one that might help. Might be easier and more robust to follow a known pattern.

DenisCarriere:
👍 Love it! @stebogit 🎩 off to you for trying to implement this directly with only TurfJS modules.

> @rowanwins There a bunch of known algorithms for this sort of thing.

Agreed, would be neat to have a comparable to see if @stebogit's implementation is faster & accurate to those other algorithms. I'm 👍 for any direction we decide to go to, as long as it's fast and accurate.

packages/turf-shortest-path/index.js
```diff
@@ -0,0 +1,125 @@
...
+ * //addToMap
+ * var addToMap = [start, end, obstacles, path];
+ */
+module.exports = function (start, end, obstacles, options) {
+    // validation
+    if (!start) throw new Error('start point is required');
+    if (getType(start) !== 'Point') throw new Error('start must be Point');
+    if (!end) throw new Error('end point is required');
+    if (getType(end) !== 'Point') throw new Error('end must be Point');
+    if (!obstacles) throw new Error('obstacles collection is required');
```
DenisCarriere: Could we possibly have obstacles as an optional parameter? In case you want to use this module over an iterative process and maybe in some scenarios you would have 2 points with or without an obstacle, we could simply just return a lineString from both points (same as an empty `featureCollection`).

packages/turf-shortest-path/index.js
```diff
@@ -0,0 +1,125 @@
...
+    options = options || {};
+    var units = options.units || 'kilometers';
+    var resolution = options.resolution;
+    if (resolution && !isNumber(resolution) || resolution <= 0) throw new Error('resolution must be a number, greater than 0');
+
+    var collection = clone(obstacles);
+    collection.features.push(start);
+    collection.features.push(end);
+
+    var box = bbox(scale(bboxPolygon(bbox(collection)), 1.15)); // extend 15%
```
DenisCarriere: Couldn't `extend` also be an optional parameter instead of hardcoded as "extend 15%"?

packages/turf-shortest-path/package-lock.json
```diff
@@ -0,0 +1,1981 @@
+{
```
DenisCarriere: I've recently just been removing/excluding the `package-lock.json` files, I'm not too sure if we should be having two `lock` files (yarn + npm).

I've added `package-lock.json` to the `.gitignore`, we could always add it back, but it seems like two lock files is a bit overkill (we only need to stick with one).

https://github.com/Turfjs/turf/commit/593cb914494ae24161895db652be4519aeba3907

packages/turf-shortest-path/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "load-json-file": "^2.0.0",
+    "tape": "^4.6.3"
+  },
+  "dependencies": {
+    "@turf/turf": "4.7.3",
+    "easystarjs": "0.4.1",
+    "grid-to-matrix": "1.3.0",
+    "javascript-astar": "0.4.1",
+    "pathfinding": "0.4.18"
+  }
```
DenisCarriere: There's definitely missing some dependencies in the `package.json`, this is one of the reasons why the TravisCI tests will fail.

DenisCarriere:
@stebogit Do you have an ESLint plugin installed in your current text editor?

**TravisCI Error**

```
/home/travis/build/Turfjs/turf/packages/turf-shortest-path/index.js
  116:37  error  'i' is defined but never used  no-unused-vars
✖ 1 problem (1 error, 0 warnings)
```

packages/turf-shortest-path/package.json
stebogit: All this is absolutely **"work-in-progress"**, I added the whole `@turf/turf` for convenience while developing. 
Sorry, I should have titled this more clearly as "proposal" (like the label); I was just interested at this point in you suggestions and check with you if my approach was reasonably sound so far.

packages/turf-shortest-path/index.js
stebogit: Yes! Great idea 👍

packages/turf-shortest-path/index.js
stebogit: This in my mind was just to allow some space to allow the path around any object, like to avoid this (just a silly example):
screen shot 2017-09-21 at 8 20 57 pm
I don't see any other use for this parameter.

stebogit:
@rowanwins thanks for the tip.
I did some quick research and it turns out [A*](https://en.wikipedia.org/wiki/A*_search_algorithm) is the best choice here, since it's faster and (as Dijkstra) is guaranteed to return the shortest path:
http://theory.stanford.edu/~amitp/GameProgramming/AStarComparison.html
Now, I'll try to compare the few implementations I found on npm to pick the best/fastest one.

Thanks guys!

rowanwins:
good luck with it @stebogit ! I find it really hard reading some of those maths/computer science papers and interpreting what they're saying, but if we can follow them they should lead us to decent performant implementations, and I think they teach us some good skills in queues and sweeplines etc that we should be able to apply elsewhere.

packages/turf-shortest-path/index.js
DenisCarriere: Fair enough... that is a pretty silly scenario 😝

packages/turf-shortest-path/package.json
DenisCarriere: 👍 Looks great to me so far! Was just trying to get the tests to pass

DenisCarriere:
`A *` looks like a great approach, @stebogit good luck in your npm quest to find one!

I like this image:

image

DenisCarriere:
@stebogit Other possible module name would be `@turf/least-cost-path`, however I prefer your original proposed name `@turf/shortest-path`.

bulutkartal:
I don't if you guys would consider but shortest sea routes may be killer, either with reverse altitude or based on depth colors on maps. [here](https://www.searates.com/reference/portdistance/) is an example based on color.

rowanwins:
Hi @bulutkartal 

Are you able to elaborate on your idea, are you saying that altitude or depth have a bearing on the path calculation? This module is for the actual calculation of the shortest path, it doesn't take into account how to visualise the result.

bulutkartal:
I was trying to mean the reserve altitude can be used to calculate shortest sea routes(only water based), but using elevation services may be quite expensive and will I don't think it is very suitable for turf.  But instead of elevation services , map background colors can be used to calculate shortest sea routes. Lines will not pass through the land, only will stay in water colors. The link I sent above as an example works that way. It calculates the shortest distance between two ports and the line script creates stays in water. I saw the approach when I opened their bundled js. It was kinda cool what they did. This wouldn't be beneficial for land routes. I just saw the example was created at sea and idea just came out and found that link :)

rowanwins:
Hi @bulutkartal 

Using the proposed module you will be able to pass in your start and end point (presumably ports), as well as a bunch of polygons (which represent land borders), the output line will show you the route between the points avoiding the land. So it wont have anything to do with elevation. I hope that helps clear things up.

stebogit:
Just a quick note, I'm not really sure this implementation is the best one, 'cause the path should not move only on a grid...
I believe the right output should look more like this (compared to the one above):
screen shot 2017-09-27 at 12 01 07 am

I'll keep looking for an alternative though, probably after we are done with `v5.0`.

bulutkartal:
using elevation was just an another idea to prevent line going through land since elevation will be negative in water. Was just an idea, nothing more :)

DenisCarriere:
> I believe the right output should look more like this

@stebogit Agreed, the 2nd map looks like it would be a better solution.

👍 to push this at the next minor release `v5.1`, big push for now would be to implement the ES Modules (we can add more modules afterwards).

DenisCarriere:
🎉 Merged & Published `@turf/shortest-path`

PR #957, opened 2017-09-21
Set `@turf/difference` to return null with complete overlap

stebogit:
Ref. https://github.com/Turfjs/turf/issues/951#issuecomment-330326745

**New output:**
```js
var poly = polygon([[[121, -31], [144, -31], [144, -15], [121, -15], [121, -31]]]);
var result = difference(poly, poly);
//= null
```


DenisCarriere:
Woot! 🎉 

```js
var result = difference(poly, poly) || 'foobar';
//='foobar'
```

PR #968, opened 2017-09-26
Fix segmentReduce & segmentEach `@turf/meta`

DenisCarriere:
## Fix `segmentReduce` & `segmentEach` `@turf/meta`

Was using `segmentReduce` and it didn't return anything in the callback, looks like https://github.com/Turfjs/turf/pull/959 had do something with it.

Going to attempt a patch release for `@turf/meta` 🤞

```js
const geojson = featureCollection([
    point([0, 1]), // ignored
    lineString([[0, 0], [2, 2], [4, 4]]),
    polygon([[[5, 5], [0, 0], [2, 2], [4, 4], [5, 5]]]),
    point([0, 1]), // ignored
    multiLineString([
        [[0, 0], [2, 2], [4, 4]],
        [[0, 0], [2, 2], [4, 4]]
    ])
]);
segmentEach(geojson, (segment, segmentIndex, featureIndex, featureSubIndex) => {
  //=segmentIndexes => [0, 1, 0, 1, 2, 3, 0, 1, 0, 1]
  //=featureIndexes => [1, 1, 2, 2, 2, 2, 4, 4, 4, 4]
  //=featureSubIndexes => [0, 0, 0, 0, 0, 0, 0, 0, 1, 1]
});
```
It's previous state wasn't usable, so I highly doubt this would "break" anyone's code since I personally wasn't able to use the index numbers properly before this fix.

### DIFF
--- a/packages/turf-meta/README.md
+++ b/packages/turf-meta/README.md
@@ -510,9 +510,9 @@ Callback for segmentEach
 **Parameters**
 
 -   `currentSegment` **[Feature](http://geojson.org/geojson-spec.html#feature-objects)&lt;[LineString](http://geojson.org/geojson-spec.html#linestring)>** The current segment being processed.
--   `featureIndex` **[number](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number)** The index of the current element being processed in the array, starts at index 0.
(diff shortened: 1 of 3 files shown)


DenisCarriere:
image

Damn those early merges 🤔 Might want to disable that...

Issue #971, opened 2017-09-26
Remove Geodesy as dependency (rhumb modules)


DenisCarriere:
# Remove Geodesy as dependency

Looking at our no-ES6 libraries, noticed [Geodesy](https://github.com/chrisveness/geodesy) is only used in the [`@turf/rhumb-bearing`](https://github.com/Turfjs/turf/tree/master/packages/turf-rhumb-bearing) & other rhumb modules.

This seems like it would be easy enough to [include directly](https://github.com/chrisveness/geodesy/blob/d78c067211b35a4c755ba122a5df63ebbdbe364c/latlon-spherical.js#L452-L477) in TurfJS, this would eliminate a 268KB dependency (minified would be smaller).

CC: @stebogit

stebogit:
Agreed!! 👍  ✂️ 📄

stebogit:
Man! I want you coffee machine @DenisCarriere! ☕️  🤣

DenisCarriere:
🎉 Implemented

PR #953, opened 2017-09-18
Adopt options to all modules (⚠️ Work in progress 15%)

DenisCarriere:
# Adopt `options` object for optional parameters

Ref: https://github.com/Turfjs/turf/issues/944
 
Adding the following syntax to the modules:

## Single optional parameter (Backwards compatible with v4.0)

```js
/**
 * @param {Feature<LineString>} line input line
 * @param {number} distance distance along the line
 * @param {Object} [options] Optional parameters
 * @param {string} [options.units="kilometers"] can be degrees, radians, miles, or kilometers
 */
export default function (line, distance, options) {
    // Backwards compatible with v4.0
    var units = (typeof options === 'object') ? options.units : options;
```

**index.d.ts**

```ts
interface Options {
    units?: Units;
}
export default function along(line: LineString, distance: number, options?: Options): Point;
```

## Multiple optional parameters (❗️ breaking change)

```js
/**
 * @param {Feature<LineString>} line input LineString
 * @param {Object} [options] Optional parameters
 * @param {number} [options.resolution=10000] time in milliseconds between points
 * @param {number} [options.sharpness=0.85] a measure of how curvy the path should be between splines
*/
export default function (line, options) {
    // Optional params
    options = options || {};
    var resolution = options.resolution || 10000;
    var sharpness = options.sharpness || 0.85;

    // validation
    if (typeof options !== 'object') throw new Error('options must be an object');
    if (typeof resolution !== 'number') throw new Error('options must be an number');
    if (typeof sharpness !== 'number') throw new Error('sharpness must be an number');
```

**index.d.ts**

```ts
interface Options {
    resolution?: number;
    sharpness?: number;
}
export default function bezier(line: LineString, options?: Options): LineString;
```

## Apply to the following modules:

- [x] [@turf/along](https://github.com/Turfjs/turf/tree/master/packages/turf-along) (@DenisCarriere)
- [x] [@turf/bearing](https://github.com/Turfjs/turf/tree/master/packages/turf-bearing) (@DenisCarriere)
- [x] [@turf/bezier](https://github.com/Turfjs/turf/tree/master/packages/turf-bezier) (@DenisCarriere)
- [x] [@turf/boolean-point-on-line](https://github.com/Turfjs/turf/tree/master/packages/turf-boolean-point-on-line) (@DenisCarriere)
- [x] [@turf/buffer](https://github.com/Turfjs/turf/tree/master/packages/turf-buffer) (@DenisCarriere)
[...]

PR #973, opened 2017-09-26
Removes Geodesy dependency from Rhumb modules + adds options parameter

stebogit:
Ref. https://github.com/Turfjs/turf/issues/971 (Geodesy) + https://github.com/Turfjs/turf/pull/953 (options)

 - [x] `@turf/rhumb-bearing`
 - [ ] `@turf/rhumb-distance`
 - [ ] `@turf/rhumb-destination`

> This seems like it would be easy enough to include directly in TurfJS

@DenisCarriere is this how you would do it?


packages/turf-rhumb-bearing/index.js
```diff
@@ -28,16 +27,69 @@ export default function (start, end, final) {
...
-    var coordsStart = getCoord(start);
-    var coordsEnd = getCoord(end);
-    var origin = new LatLonSpherical(coordsStart[1], coordsStart[0]);
-    var destination = new LatLonSpherical(coordsEnd[1], coordsEnd[0]);
-    var bear360;
+    const coordsStart = getCoord(start);
+    const coordsEnd = getCoord(end);
+    const origin = new LatLon(coordsStart[1], coordsStart[0]);
+    const destination = new LatLon(coordsEnd[1], coordsEnd[0]);
+    let bear360;
```
DenisCarriere: Got to remove that ES6 sugar (`let` & `const`)

packages/turf-rhumb-bearing/index.js
```diff
@@ -28,16 +27,69 @@ export default function (start, end, final) {
...
+ *
+ * @private
+ * @constructor
+ * @param {number} lat - Latitude in degrees.
+ * @param {number} lon - Longitude in degrees.
+ *
+ * @example
+ *     var p1 = new LatLon(52.205, 0.119);
+ */
+function LatLon(lat, lon) {
```
DenisCarriere: 🤔 I wonder if we could remove the Class Object entirely and simply just make a `rhumbBearingTo` method?

packages/turf-rhumb-bearing/index.js
stebogit: Meaning we should not use them? Why?

DenisCarriere: Good question... Well the only thing we added from ES6 was the modules, Rollup will only bundle the file but it currently doesn't compile the code to ES5 (We would need to include Babel for that) is that something that's worth doing?

DenisCarriere:
👍 Nice update!

packages/turf-rhumb-bearing/index.js
```diff
@@ -27,16 +27,16 @@ export default function (start, end, final) {
     if (!start) throw new Error('start point is required');
     if (!end) throw new Error('end point is required');
 
-    const coordsStart = getCoord(start);
-    const coordsEnd = getCoord(end);
-    const origin = new LatLon(coordsStart[1], coordsStart[0]);
-    const destination = new LatLon(coordsEnd[1], coordsEnd[0]);
-    let bear360;
+    var coordsStart = getCoord(start);
```
DenisCarriere: @stebogit Let me see if we could compile our `index.js` from ES6 to ES5... (without too much hassle)

packages/turf-rhumb-bearing/index.js
DenisCarriere: @stebogit Going to look into this... if we are already compiling the code, might as well go all the way to ES6 🚀

stebogit: @DenisCarriere I thought this was the idea, to justify all the work we (you) are doing for this change.

Also, only now I realize with this new setup I can't debug the code in PHPStorm (or at least not as easily as before), cause `import/export` break in node... that's a bummer 😞

DenisCarriere: I'm starting to see the same downfalls. 😢 

Well we could always convert the `test.js` back to NodeJS 4 or 6 compatible, that way there's 1 less file that has import/export statements.

We could always explore integrating [`@std/esm`](https://github.com/standard-things/esm) somewhere, maybe that will help?

packages/turf-rhumb-bearing/index.js
```diff
@@ -1,7 +1,6 @@
 // https://en.wikipedia.org/wiki/Rhumb_line
-// http://www.movable-type.co.uk/scripts/latlong.html#rhumblines
-import { getCoord } from '@turf/invariant';
-var LatLonSpherical = require('geodesy').LatLonSpherical;
+import {getCoord} from '@turf/inconstiant';
```
DenisCarriere: @stebogit 🤦‍♂️ 🤣

packages/turf-rhumb-bearing/index.js
stebogit: Going wild here: isn't there a way to compile **to** ES6, at least the `import/export` syntax?
That way we could develop (and debug) in node using `require` and the compilation, pre rollup, would take care of that additional feature.

OR, you could just implement `import/export` in Node.js v9.0. I bet with your coffee machine it shouldn't take you that long. 🤣

DenisCarriere: > import/export in Node.js v9.0.

Ideally this would be the ultimate solution.

The goal with this push to ES6 modules is to get ready for that, it obviously requires some HEAVY refactoring (which we are noticing).

Having a `pretest` global script is pretty handy... we can work with that! 👍

DenisCarriere: Currently there's 3 main environments:
- Web UMD (Compile to ES5 UMD format)
- ES6 Modules (and all the other new ES6 features)
- ES5 CommonJS (NodeJS 4 support)

Attempting to support all 3 is difficult... lol! I'm sure we can figure out a low-impact Rollup solution, it's just going to take a bit more tinkering.

DenisCarriere:
@stebogit FYI: for any new PR's try to use the v5.0.0-alpha for `meta`, `helpers` & `invariant`.

Weird things happen when the dependencies get mixed.

DenisCarriere:
@stebogit Going to merge this now since I'm going to do some significant changes again (ref: https://github.com/Turfjs/turf/issues/974)

PR #977, opened 2017-09-27
Removes Geodesy dependency from Rhumb modules 2

stebogit:
Ref: https://github.com/Turfjs/turf/issues/971
https://github.com/Turfjs/turf/pull/973 sequel...

### DIFF
--- a/packages/turf-rhumb-bearing/index.js
+++ b/packages/turf-rhumb-bearing/index.js
@@ -27,14 +27,10 @@ export default function (start, end, final) {
     if (!start) throw new Error('start point is required');
     if (!end) throw new Error('end point is required');
 
-    const coordsStart = getCoord(start);
-    const coordsEnd = getCoord(end);
-    const origin = new LatLon(coordsStart[1], coordsStart[0]);
-    const destination = new LatLon(coordsEnd[1], coordsEnd[0]);
     let bear360;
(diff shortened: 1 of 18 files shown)


DenisCarriere:
Woohoo! 🎉

DenisCarriere:
@stebogit Resolved conflicts, 🤞 should pass...

I'll have a look at this again tomorrow, so far looks great! 👍

packages/turf-rhumb-bearing/index.js
```diff
@@ -52,7 +52,7 @@ export default function (start, end, options) {
  * const p2 = new LatLon(50.964, 1.853);
  * const d = p1.rhumbBearingTo(p2); // 116.7 m
  */
-function rhumbBearing(from, to) {
+function calculateRhumbBearing(from, to) {
```
DenisCarriere: @stebogit I've been adding names to the export default function, in theory this shouldn't have any impact, however when bundling multiple modules sometimes the compile `dist/index.js` gets confused if another function has the same name as the main functions (`@turf/rhumbBearing`).

Have a look at the `dist/index.js` and you'll see how Rollup compiles the code.

stebogit: 👍

Issue #978, opened 2017-09-27
Correct earth radius value in `@turf/helpers`


stebogit:
@DenisCarriere @rowanwins  I noticed `@turf/helpers` still uses an [incorrect earth radius](https://github.com/Turfjs/turf/blob/master/packages/turf-helpers/index.js#L334), and I don't know why exactly since it was decided to change it: https://github.com/Turfjs/turf/issues/635#issuecomment-292011500, https://github.com/Turfjs/turf/pull/645

I think we should modify it, knowing it will probably be very tricky because impacting a lot of modules.

I'd also suggest we add an `earthRadius` function (just returning `6371008.8`, meters) in `@turf/helpers` so we can refer to it in all the modules that right now need that value (I see the Rhumb modules and `@turf/buffer` for example), or in future modules, to avoid mismatching.

DenisCarriere:
Yep agreed, we don't necessarily need a function, we just need to export variables like this library: https://github.com/mapbox/wgs84/blob/master/index.js

We should expose those 4 variables in turf helpers

stebogit:
Oh cool, didn't know about exposing variables like this, but it makes sense. :+1:

stebogit:
Fixed with  #980

PR #980, opened 2017-09-27
Add WGS84 to @turf/helpers

DenisCarriere:
## Add WGS84 to `@turf/helpers`

Pulled from @tmcw's https://github.com/mapbox/wgs84

Built-in variables for WGS84 related things.

Ref: https://github.com/Turfjs/turf/issues/978

## How to use

```js
import { wgs84 } from '@turf/helpers';

wgs84.RADIUS // => 6378137
```

CC: @stebogit

### DIFF
--- a/packages/turf-helpers/index.d.ts
+++ b/packages/turf-helpers/index.d.ts
@@ -139,3 +139,13 @@ export function isNumber(num: any): boolean
  * http://turfjs.org/docs/#isobject
  */
 export function isObject(input: any): boolean
+
+/**
+ * http://turfjs.org/docs/#wgs84
+ */
+export const wgs84: {
+    RADIUS: number
+    FLATTENING_DENOM: number
+    FLATTENING: number
+    POLAR_RADIUS: number
+}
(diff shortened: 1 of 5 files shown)


DenisCarriere:
@stebogit Weird errors like this happen... 🤔 

https://travis-ci.org/Turfjs/turf/builds/280554132#L1495

```
lerna ERR! execute     if (mutate !== true) geojson = clone(geojson);
lerna ERR! execute                                    ^
lerna ERR! execute 
lerna ERR! execute TypeError: clone is not a function
```

I think these errors are caused by `export default function()` and the module doesn't have a name and for some reason can't figure out it's name.. 🤔 

https://github.com/Turfjs/turf/blob/master/packages/turf-clone/index.js#L13

Going to merge this since 1/2 of the tests passed.

stebogit:
@DenisCarriere the earth radius is not what [here](https://github.com/Turfjs/turf/issues/635#issuecomment-292013737) was agreed upon.
In my understanding it should be `6371008.8` meters.

DenisCarriere:
Yep, I changed it to the value you provided, the earth radius value we need is for the Haversine area/distance calculations.

DenisCarriere:
`earthRadius` => 6371008.8

PR #981, opened 2017-09-27
Refactor index.js to ES5 + ES Modules

DenisCarriere:
An easy way to test for ES5 syntax is to use UglifyJS as a `posttest`.

## Catching errors like this:

**unicode charactors**
```
Parse error at ./dist/index.js:60,8
    let Δλ = helpers.degrees2radians((to[0] - from[0]));
        ^
ERROR: Unexpected token: name (Δλ)
```

**const & let**
```
Parse error at ./dist/index.js:32,4
    const final = (typeof options === 'object') ? options.final : options;
    ^
ERROR: Unexpected token: keyword (const)
```


packages/turf-helpers/index.js
```diff
@@ -535,19 +534,7 @@ export function isObject(input) {
...
- * http://earth-info.nga.mil/GandG/wgs84/gravitymod/egm96/egm96.html
+ * Earth Radius used with the Harvesine formula and approximates using a spherical (non-ellipsoid) Earth.
  */
-export const wgs84 = {
-    RADIUS,
-    FLATTENING_DENOM,
-    FLATTENING,
-    POLAR_RADIUS,
-};
+export var earthRadius = 6371008.8;
```
DenisCarriere: @stebogit Added `earthRadius` to `@turf/helpers`. (was suppose to be different PR).

Removed `wgs84` info since we aren't doing those types of calculations.

Ref: https://github.com/Turfjs/turf/issues/978

DenisCarriere:
@stebogit @rowanwins I've enabled strict ES5 syntax checker. You should now have errors when you add `const` & `let` to your code.

stebogit:
@DenisCarriere does this mean we still can't use ES6 syntax in the `index.js`files (as usual)?
What's not working with it?
Sorry I haven't been able to really follow all your recent tests and changes.

DenisCarriere:
For this next major release, let's not include any ES6 syntax, it's very difficult to troubleshoot ES5 syntax errors when depending on compilers to convert from ES6 to ES5.

Doing the switch from `import/export` is already a big enough task on it's own, let's only focus on that right now and then maybe.... 🤞 for the next major release (v6.0) we could consider using pure ES6 (at least NodeJS 4 compatible).

Issue #944, opened 2017-09-17
Adopt `options` object for optional parameters


stebogit:
Formalizing @DenisCarriere's elegant [proposal](https://github.com/Turfjs/turf/pull/876#issuecomment-319070336), which I really like. 😄 
I would actually include **all** the optional parameters in the `options` object, making it more consistent throughout the library.

```js
/**
 * @param {Object} [options={}] Optional parameters
 * @param {string} [options.units="kilometers"] can be degrees, radians, miles, kilometers, ...
 * @param {number} [options.steps=64] number of steps on a rounded corner
 */
module.exports = function (feature, distance, options) {
    options = options || {};
    var units = options.units;
    var steps = options.steps;
```
As suggested, we might formally adopt this for new modules and refactor the existing modules in the new major release.

cc: @rowanwins 

## Single optional parameter (Backwards compatible with v4.0)

```js
/**
 * @param {Feature<LineString>} line input line
 * @param {number} distance distance along the line
 * @param {Object} [options] Optional parameters
 * @param {string} [options.units="kilometers"] can be degrees, radians, miles, or kilometers
 */
export default function (line, distance, options) {
    // Backwards compatible with v4.0
    var units = (typeof options === 'object') ? options.units : options;
```

**index.d.ts**

```ts
interface Options {
    units?: Units;
}
export default function along(line: LineString, distance: number, options?: Options): Point;
```

## Multiple optional parameters (❗️ breaking change)

```js
/**
 * @param {Feature<LineString>} line input LineString
 * @param {Object} [options] Optional parameters
 * @param {number} [options.resolution=10000] time in milliseconds between points
 * @param {number} [options.sharpness=0.85] a measure of how curvy the path should be between splines
*/
export default function (line, options) {
    // Optional params
    options = options || {};
    var resolution = options.resolution || 10000;
    var sharpness = options.sharpness || 0.85;

    // validation
    if (typeof options !== 'object') throw new Error('options must be an object');
    if (typeof resolution !== 'number') throw new Error('options must be an number');
    if (typeof sharpness !== 'number') throw new Error('sharpness must be an number');
```

**index.d.ts**

```ts
interface Options {
    resolution?: number;
    sharpness?: number;
}
export default function bezier(line: LineString, options?: Options): LineString;
```

## Apply to the following modules:

- [x] [@turf/along](https://github.com/Turfjs/turf/tree/master/packages/turf-along) (@DenisCarriere)
[...]

Issue #834, opened 2017-07-11
Remove @turf/point-grid `centered` parameter


stebogit:
Ref #833

>  there's really no benefit of having an inconsistent offset grid.

## JSDocs

```javascript
/**
 * Creates a {@link Point} grid from a bounding box, {@link FeatureCollection} or {@link Feature}.
 *
 * @name pointGrid
 * @param {Array<number>|FeatureCollection|Feature<any>} bbox extent in [minX, minY, maxX, maxY] order
 * @param {number} cellSide the distance between points
 * @param {string} [units=kilometers] used in calculating cellSide, can be degrees, radians, miles, or kilometers
 * @param {boolean} [bboxIsMask=false] if true, and bbox is a Polygon or MultiPolygon, the grid Point will be created
 * only if inside the bbox Polygon(s)
 * @returns {FeatureCollection<Point>} grid of points
 */
module.exports = function (bbox, cellSide, units, bboxIsMask) {
```

stebogit:
Implemented with #982

PR #982, opened 2017-09-28
Refactoring @turf/point-grid 

stebogit:
- [x] removes `centered` parameter (https://github.com/Turfjs/turf/issues/834)
- [x] adds `options` for optional parameters (https://github.com/Turfjs/turf/issues/944)
- [x] updates types (hopefully)

**Extra:**
- [x] adds `properties` option, to be passed to each point of the grid

```js
/**
 * Creates a {@link Point} grid from a bounding box, {@link FeatureCollection} or {@link Feature}.
 *
 * @name pointGrid
 * @param {BBox|FeatureCollection|Feature} bbox extent in [minX, minY, maxX, maxY] order
 * @param {number} cellSide the distance between points
 * @param {string} [options.units="kilometers"] used in calculating cellSide, can be degrees, radians, miles, or kilometers
 * @param {number} [options.bboxIsMask=false] if true, and bbox is a Polygon or MultiPolygon, the grid Point will be created
 * @param {Object} [options.properties={}] passed to each point of the grid
 * @returns {FeatureCollection<Point>} grid of points
 * @example
 * var extent = [-70.823364, -33.553984, -70.473175, -33.302986];
 * var cellSide = 3;
 *
 * var grid = turf.pointGrid(extent, cellSide, {units: 'miles'});
 *
 * //addToMap
 * var addToMap = [grid];
 */
export default function pointGrid(bbox, cellSide, options) {
```


stebogit:
I don't understand why Travis fails: my branch is in sync with `master`, but still `@turf/interpolate` errors out for some reason:

screen shot 2017-09-28 at 1 24 03 am

DenisCarriere:
🤔 looking into that now.. I'm getting the same error as Travis.

```
# turf-interpolate
ok 1 data-1km
ok 2 data-500m
ok 3 data-weight-2
ok 4 hex-zValue
/Users/mac/Github/turf/packages/turf-helpers/dist/index.js:374
    if (num === undefined || num === null || isNaN(num)) throw new Error('num is required');
                                                         ^

Error: num is required
```

DenisCarriere:
The issue is in one of the `hex-zValue` is returning `NaN` as elevation value.

Seems like all the values are the same (value, min, max) which makes percent = to NaN:

```
percent 8.333333333333332
value 7
min 2
max 14
percent -8.333333333333332
ok 4 hex-zValue
value 77
min 77
max 77
percent NaN
```

DenisCarriere:
It's doing something like this, 🤔 ... Issue is somewhere in the middle of that point hex grid.

image

DenisCarriere:
@stebogit Seems like it's the `point-random` that's causing the issue with `@turf/interpolate`... 🤔 ... The elevation properties are not being "interpolated".

```
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
{ elevation: 76.90974403809834 }
```

image

stebogit:
@DenisCarriere the weirdest thing is that this error does not occur on `master` (just tested it)... 🤔

DenisCarriere:
@stebogit This does seem weird... no clue! I'll keep looking at it when I have more time, it's an odd one...

DenisCarriere:
@stebogit Added `@turf/clone` and fixed `@turf/point-grid` to the new Options format.

packages/turf-interpolate/index.js
```diff
@@ -49,7 +51,7 @@ export default function (points, cellSize, gridType, property, units, weight) {
     switch (gridType) {
     case 'point':
     case 'points':
-        grid = poinGrid(box, cellSize, units, true);
+        grid = poinGrid(box, cellSize, {units: units, bboxIsMask: true});
```
DenisCarriere: @stebogit We should enforce strict error catching to prevent forgetting adding extra optional params.

packages/turf-point-grid/index.js
```diff
@@ -10,60 +10,59 @@ import { getType } from '@turf/invariant';
...
  * var cellSide = 3;
- * var units = 'miles';
  *
- * var grid = turf.pointGrid(extent, cellSide, units);
+ * var grid = turf.pointGrid(extent, cellSide, {units: 'miles'});
  *
  * //addToMap
  * var addToMap = [grid];
  */
-export default function (bbox, cellSide, units, centered, bboxIsMask) {
```
DenisCarriere: We need to throw an error if `options` is not an object.

```js
if (!isObject(options)) throw new Error('options is invalid')
```

DenisCarriere:
🎉 
image

packages/turf-interpolate/index.js
stebogit: aaaaah, that's why! great job! 👍

DenisCarriere: Sooooo annoying because there was no errors causing this issue

packages/turf-point-grid/index.js
stebogit: well @DenisCarriere, it can be `undefined` though

DenisCarriere: Yes but the default value will kick in and be an object. That's why defining the default values should be before the validation, easier validation syntax

PR #989, opened 2017-10-02
added a test to check feature order split location precision

alexgleith:
- Modified tests for `turf-line-split`

This PR adds a test for an issue with the precision of split coordinates when splitting feature one with feature two and feature two with feature one, i.e., this split location _should_ be the same, and it's not.


DenisCarriere:
🤔 .. Don't think you are using `@turf/line-split` correctly.

**1st param:** Line to split
**2nd param:** Splitter (Any GeoJSON which acts as the "knife" to split the line)
**returns:** 1st param with modified line splits

```js
/**
 * Split a LineString by another GeoJSON Feature.
 *
 * @name lineSplit
 * @param {Feature<LineString>} line LineString Feature to split
 * @param {Feature} splitter Feature used to split line
 * @returns {FeatureCollection<LineString>} Split LineStrings
 */
function lineSplit(line, splitter) {
```
@alexgleith The tests you provided will always fail since `resultsOne.geojson` !== `resultsTwo`

packages/turf-line-split/test.js
```diff
@@ -33,6 +33,16 @@ test('turf-line-split', t => {
...
 });
 
+test('turf-line-split -- lines should split the same feature 1 with 2 as 2 with 1', t => {
+    const geojson = load.sync(directories.in + 'linestrings.geojson');
+    const featureOne = geojson.features[0];
+    const featureTwo = geojson.features[1];
+
+    const resultsOne = lineSplit(featureOne, featureTwo);
+    const resultsTwo = lineSplit(featureTwo, featureOne)
+    t.deepEquals(resultsOne, resultsTwo, 'Splits were made in different locations')
```
DenisCarriere: `resultsOne` will never equal to `resultsTwo`

packages/turf-line-split/test.js
```diff
@@ -33,6 +33,16 @@ test('turf-line-split', t => {
...
+})
```
DenisCarriere: Note: Add `t.end();` when using Tape tests.

alexgleith: Oops, sorry. Will sort it.

packages/turf-line-split/test.js
alexgleith: Ah, you're right... 

I need to be comparing the intersection points. Sorry, that was foolish of me! In my head I was comparing the points that the lines were split at, but of course this isn't doing that!!

alexgleith:
Hey @DenisCarriere, I've changed the test so that it is checking the coordinates specifically.

Let me know what you think of this one.

packages/turf-line-split/test.js
DenisCarriere: :) No worries, your test case's intent is valid, the intersections should be compared.

packages/turf-line-split/test.js
```diff
@@ -20,18 +21,31 @@ const fixtures = fs.readdirSync(directories.in).map(filename => {
...
-    for (const {filename, name, geojson}  of fixtures) {
-        const line = geojson.features[0];
-        const splitter = geojson.features[1];
-        const results = colorize(lineSplit(line, splitter));
-        featureEach(geojson, feature => results.features.push(feature));
-
-        if (process.env.REGEN) write.sync(directories.out + filename, results);
-        t.deepEquals(results, load.sync(directories.out + filename), name);
-    }
+// test('turf-line-split', t => {
```
DenisCarriere: Don't comment out this test case... 😯

packages/turf-line-split/test.js
```diff
@@ -20,18 +21,31 @@ const fixtures = fs.readdirSync(directories.in).map(filename => {
...
+test('turf-line-split -- lines should split the same feature 1 with 2 as 2 with 1', t => {
+    const featureOne = lineString([[114.58215786065353825, -14.82470576519326144], [137.21678649707752129, -16.71692980416107588]]);
+    const featureTwo = lineString([[119.1412061636556956, -19.83670052919270788], [133.06640625, -12.64033830684678961]]);
+
+    const resultsOne = lineSplit(featureOne, featureTwo);
+    const resultsTwo = lineSplit(featureTwo, featureOne);
+
+    const midCoordOne = getCoords(resultsOne.features[0])[1];
+    const midCoordTwo = getCoords(resultsTwo.features[1])[0];
+    t.deepEquals(midCoordOne, midCoordTwo, 'Splits were made in different locations')
```
DenisCarriere: Bit better, however kinda hard to troubleshoot if there isn't a GeoJSON output... typically we save all our outputs in `test/out` that way we can easily preview the expected or NOT expected GeoJSON outputs.

DenisCarriere:
Doesn't seem like there's an issue, it's only a precision error and it's less than < 1cm (0.0000001 degrees off)
```
[ 126.85435645043746, -15.850644817612286 ] => Expecting
[ 126.85435648027894, -15.850644808066447 ] => Results
```

If you actually truncate these values at 6 decimals then there's no error (6 decimal precision is 10cm accurate).

DenisCarriere:
Added `@turf/truncate` and no errors happened, that means the error is less than 10cm.

```js
import truncate from '@turf/truncate'

const midCoordOne = truncate(point(getCoords(resultsOne.features[0])[1]));
const midCoordTwo = truncate(point(getCoords(resultsTwo.features[1])[0]));
t.deepEquals(midCoordOne, midCoordTwo, 'Splits were made in different locations');
```

alexgleith:
Hey @DenisCarriere 

One more thing, I have a fix for the `polygon-split` function using `truncate` in my local code, but it breaks on one of the test cases. I see you've got `round` in here.. does this work for all the tests for `polygon-split`?'

I felt that using round (which is per coordinate) was a bit heavy, and that the root cause might be able to be fixed.

DenisCarriere:
@alexgleith any issues which has an error of less than 0.000001 degrees isn't considered an error, it's just a precision difference. We use `@turf/truncate` + `round()` in many test cases.

DenisCarriere:
Going to merge this now, you can always update this test case later on if you find another scenario which has a greater precision error.

PR #993, opened 2017-10-03
Adopt options for inside/isobands/isolines

DenisCarriere:
# Adopt options for:

- inside
- isobands
- isolines

Ref: #944


DenisCarriere:
🤞 This last one should pass...

PR #1001, opened 2017-10-04
Bundle Valid ES5 syntax checker in Rollup

DenisCarriere:
## Bundle Valid ES5 syntax checker in Rollup

@rowanwins This one is for you 😄 

This is a VERY simplified ES5 syntax checker, previously using UglifyJS to validate ES5 syntax, but it was overkill for our particular purpose.

Using 3rd party plugins seemed to be problematic with our Root & Sub module setup, if a rollup dependency was added you would need to install those devDependencies at the root level & module level (which is irritating).


DenisCarriere:
1 of 2 tests passed 🤔 So weird... (once uglifyjs is removed the tests might run smoother)

Issue #463, opened 2016-08-24
Uncaught TopologyException: found non-noded intersection between LINESTRING and LINESTRING


codeofsumit:
I'm getting this error when doing a lot of intersect and difference calls on two polygons.

```
Uncaught TopologyException: found non-noded intersection between LINESTRING ( -0.06986182563364439 51.51023450894928, -0.07004291534423829 51.51005341923869 ) and LINESTRING ( -0.07004291534423829 51.51010683860267, -0.06992667398091246 51.51016495928433 ) [ (-0.06993607661627625, 51.510160257966646, undefined) ]
```

![image](https://cloud.githubusercontent.com/assets/2399810/17927378/b3f551fe-69f5-11e6-97c9-5fea25892e7e.png)

Any ideas how I could prevent this or what I'm doing wrong?

codeofsumit:
ok I think it's a self intersection. Seems like I have to disallow self-intersections as turf doesn't seem to support it.

codeofsumit:
Turf definitely throws self-intersection errors even if there are no self intersection sometimes.

rowanwins:
Sorry for the late reply but are you able to post a jsfiddle or something so that we can see the data you're working with? 

Thanks

morganherlocker:
> Turf definitely throws self-intersection errors even if there are no self intersection sometimes.

I have not seen any cases where this has been true, but would love a test case.

DenisCarriere:
For any issues related to `@turf/intersect` submit a PR with a new GeoJSON Feature Collection with two polygons that should or should not intersect.

**Step 1:** Add polygon to [test/in](https://github.com/Turfjs/turf/tree/master/packages/turf-intersect/test/in)

**Step 2:** Run tests `npm test`

**Step 3:** If an error occurs, submit PR with new GeoJSON causing the error.

1994sunshine:
Maybe this answer will solve your problem（https://stackoverflow.com/questions/61926747/topologyexception-when-using-turf-intersect）。
Here's how I managed to make it work:

var options = {tolerance: 0.0001, highQuality: true};
poly = turf.cleanCoords(turf.simplify(poly, options));

I also checked for kinks, and divided the polygon if it returned true:

var kinks = turf.kinks(poly);
if (kinks.features.length){
  polys = turf.unkinkPolygon(poly);
...

Issue #297, opened 2015-12-10
found non-noded intersection between LINESTRING 


ajmas:
I ran into the following error, when calling `turf.intersect(poly, point1)`

    "found non-noded intersection between LINESTRING ( 180 -84.71338, -179.942499 -84.721443 ) and LINESTRING ( -143.107718 -85.040752, -142.892279 -84.570497 ) [ (-142.96105271343765, -84.720614586011) ]"

A bit of background:

I am using a combination of turf and turf-multipolygon, with the geojson outline at http://openlayers.org/en/v3.8.2/examples/data/geojson/countries.geojson

My code looks something as follows:
```
features = countryOutlines.features;
for (i=0; i<features.length; i++) {

    if ( features[i].geometry.type === 'MultiPolygon' ) {
        for (j=0; j<features[i].geometry.coordinates.length; j++) {
            poly = turf.polygon(features[i].geometry.coordinates[j]);
            try {
                if (turf.intersect(poly, point1)) {
                    return features[i].properties.name;
                }                             
            } catch (e) {
                console.error(e, features[i].id, j);
            }
        }
     } else {
        poly = turf.polygon(features[i].geometry.coordinates[0]); 
	 
        if (turf.intersect(poly, point1)) {
            return features[i].properties.name;
        }        
     }
}
```

The issue happens with the entry for Antartica (ATA). Should the intersect function be dealing with situation or is there a function that would 'correct' the data for this situation?

BTW I did modify turf-multipolygon for use in browser based JS:

```
turf.multipolygon = function(coordinates, properties) {
  if (!coordinates) {
    throw new Error('No coordinates passed');
  }
  return {
    "type": "Feature",
    "geometry": {
      "type": "MultiPolygon",
      "coordinates": coordinates
    },
    "properties": properties || {}
  };
};
```

rowanwins:
Hi @ajmas 

Turf.intersect is designed to work with two polygons so I'm a bit confused why you're calling turf.intersect(poly, point1)  ...

It's a bit hard to work out what else might be going wrong with your code from your description, you might be able to use something like [Turfjs.party](http://turfjs.party/) or jsfiddle to provide  a better idea of what you've got so far.

Thanks,
Rowan

sainsb:
I'm experiencing the same error with 2 polygons. Fairly simple ones at that...
https://jsfiddle.net/hc5v73tq/2/

ajmas:
[...]

Issue #357, opened 2016-04-24
Trim unnecessary coordinate precision?


tmcw:
from @mourner 's comment at 

https://github.com/Turfjs/turf-hex-grid/issues/19

---

The current code generates unnecessarily detailed coordinates because there's no rounding:

[-95.59408416187462, 31.27105676807148],
[-95.80518167741707, 31.58431182228841]
We should round to at least 7-8 digits.

---

I'm not sure that this makes sense... why should we limit coordinate precision?

mourner:
Because 7th digit has a maximum margin of error that equals 11mm. We have 14 digits, half of them are simply unnecessary, leading to 2 times bigger sizes of GeoJSON when saved to a file. Although maybe this should be done as a post-processing step with a separate JS module, not as a part of each individual module, most users are unaware of the unnecessary digits.

tmcw:
Yeah, I think this should be in a separate module - the space savings only happens when serializing out geojson, and often turf-using applications don't serialize at all/save to disk, or have lots of temporary results that aren't ever saved.

wboykinm:
I'm hitting file transfer limits when I run my 6-points-of-precision geojson through turf and get 14 points back. Even a separate module for this would be fabulous.

mourner:
@wboykinm just realized that you can probably do without a separate module:

```js
var meta = require('turf.meta');

meta.coordEach(geojson, function (p) {
    p[0] = Math.round(p[0] * 1e6) / 1e6;
    p[1] = Math.round(p[1] * 1e6) / 1e6;
});
```

wboykinm:
SO EASY

PR #1004, opened 2017-10-06
turf-intersect: new test data with small intersection

zbyte64:
Had an error in my app: 'Uncaught TopologyException: found non-noded intersection between LINESTRING and LINESTRING', issue #463 recommended creating a test case with the failure and that is what this pull request does. On this branch the intersect tests fail with REGEN=true . The added geojson file defines an intersect geometry only 60 inches wide. https://github.com/Turfjs/turf/issues/297#issuecomment-327484653 indicates that JSTS has issues with resolution and may be the cause.


zbyte64:
Attempted to normalize the precision as described: https://github.com/Turfjs/turf/issues/357 ; issue still persists. Maybe the larger geometry has too many points spanning the small box?

DenisCarriere:
Looks like there's two polygon outers in the Polygon Feature. The inner ring is outside the outer ring.

image

Moved vertices of the "inner" ring.

image

zbyte64:
Is there a function for checking this, like kinks?

DenisCarriere:
There's unlink polygon that might be useful. https://github.com/Turfjs/turf/blob/master/packages/turf-unkink-polygon/README.md

zbyte64:
`kink` doesn't report any kinks. The geometry of the lot is represented as one array in the geojson; how were you able to construct/detect the inner ring?

DenisCarriere:
Oh wait... the Inner & Outer was two polygon shapes in a FeatureCollection 🤦‍♂️  my bad.

It does seem to be a valid GeoJSON according to https://www.mapbox.com/geojsonhint

🤔 ... looking into this.

stebogit:
sorry guys, that typo was driving me crazy 🤣

DenisCarriere:
@stebogit hilarious!

DenisCarriere:
@zbyte64 Made a quick patch fix for this issue, seems like the small inner polygon was throwing a topology error in JSTS, when truncating the precision to 4 and removing any duplicate nodes using `clean-coords` the smaller polygon only had 2 remaining nodes and a valid Polygon needs a minimum of 4 coordinates.

Not an ideal result, but at least it stopped crashing... 👍 

🤔 I couldn't send commits to your PR, so I opened a new one here: https://github.com/Turfjs/turf/pull/1011

ilijaz:
Terrible. Its break all my code, cause i'm try to use turf in precision farming.
Some micro polygons will to be a points. I'm very excited about  #88

PR #1005, opened 2017-10-06
Refactor bench.js to ES modules

DenisCarriere:
## Refactor `bench.js` to ES modules

Used [`5to6-codemode`](https://github.com/5to6/5to6-codemod) for the majority and then fixed a few broken ones.


packages/turf-isolines/test/out/bigMatrix.geojson
```diff
@@ -12,804 +12,7 @@
 			},
 			"geometry": {
 				"type": "MultiLineString",
-				"coordinates": [
```
DenisCarriere: 🤔 ... what happened here... 🤔

DenisCarriere: Reverted this change

PR #1007, opened 2017-10-06
Add properties option to @turf/rhumb-destination

DenisCarriere:
## Add properties option to `@turf/rhumb-destination`

```js
const destination = rhumbDestination([-110, 50], 0, 45, {properties: {foo: 'bar'}});
//=destination.properties => {foo: 'bar'}
```

Might help clean up the following code:
https://github.com/stebogit/matrix-to-grid/blob/master/index.js#L62-L66

### DIFF
--- a/packages/turf-rhumb-destination/index.js
+++ b/packages/turf-rhumb-destination/index.js
@@ -1,5 +1,5 @@
 // https://en.wikipedia.org/wiki/Rhumb_line
-import { earthRadius, point, convertDistance, degrees2radians } from '@turf/helpers';
+import { earthRadius, point, convertDistance, degrees2radians, isObject } from '@turf/helpers';
 import { getCoord } from '@turf/invariant';
 
 /**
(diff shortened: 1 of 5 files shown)


packages/turf-rhumb-destination/index.js
```diff
@@ -39,7 +45,7 @@ function rhumbDestination(origin, distance, bearing, options) {
     // compensate the crossing of the 180th meridian (https://macwright.org/2016/09/26/the-180th-meridian.html)
     // solution from https://github.com/mapbox/mapbox-gl-js/issues/3250#issuecomment-294887678
     destination[0] += (destination[0] - coords[0] > 180) ? -360 : (coords[0] - destination[0] > 180) ? 360 : 0;
-    return point(destination);
+    return point(destination, properties);
```
DenisCarriere: Proper properties handling will be done in `@turf/helpers`.

Something like this will be added

```js
for (var prop in properties) {
    pt.properties[prop] = properties[prop];
}
```

PR #1012, opened 2017-10-11
Update earth radius

DenisCarriere:
# Update earth radius

### Modules modified:

- `@turf/buffer`
- `@turf/point-on-line`
- `@turf/distance`
- `@turf/helpers`

Main changes comes from `@turf/helpers`, the rest (278 files) are simply fixture updates.

https://github.com/Turfjs/turf/blob/13a8bed8f7e62510be2332489cbe3b318983e70e/packages/turf-helpers/index.js#L1-L25


Isaac-Leonard:
Does the assumption of earth being spherical and not ellipsoid have a major effect when not working near the poles?

stebogit:
@Isaac-Leonard if the effect is "major" or significant - or not - depends on your specific use case and the dimensions of your features compared to the earth dimensions

Issue #910, opened 2017-08-20
`@turf/concave` improvements 


stebogit:
- [x] set `maxEdge` parameter optional and by default to `Infinity`.
A common expected output of a `concave` function (like [`concaveman`](https://github.com/mapbox/concaveman)) is always to return a single concave hull `Polygon`. `@turf/concave` has the capability to return a multipolygon (with holes! 👍 ), but this should not be the default behaviour.

- [x] remove Error output if there is no hull to return, it should return an empty `Feature` (`helpers.feature(null)`). (done [here](https://github.com/Turfjs/turf/commit/acbe77d44cb4427d2773802a26b4c681cbc6b617#diff-0da3d3c7e07d2320e0a04705049ec3b0L53)).
Currently the Error "too few polygons found to compute concave hull" is likely due to a small maxEdge. This shouldn't be considered an error, because an error should mean a misuse of the module (which isn't the case here).

Ref. #907.

DenisCarriere:
👍

stebogit:
@DenisCarriere are we sure we should return a `helpers.feature(null)` if no concave hull is possible?
I'd say it should return `null`, for the same reasons why we have now [`null` from `@turf/difference` and `@turf/intersect`](https://github.com/Turfjs/turf/issues/951)

DenisCarriere:
Also seems like the `@turf/great-circle` is doing the same:
https://github.com/Turfjs/turf/blob/fa7acb5cbd75a61353c988395c7c258ae5c7ce20/packages/turf-great-circle/arc.js#L38

stebogit:
Implemented with #1020 and https://github.com/Turfjs/turf/pull/991/commits/acbe77d44cb4427d2773802a26b4c681cbc6b617

PR #1020, opened 2017-10-13
sets maxEdge default to Infinity

stebogit:
Ref "`@turf/concave` improvements" #910

### DIFF
--- a/packages/turf-concave/bench.js
+++ b/packages/turf-concave/bench.js
@@ -24,9 +24,8 @@ const fixtures = fs.readdirSync(directory).map(filename => {
  * pts3: 0.375ms
  */
 for (const {name, geojson} of fixtures) {
-    const {maxEdge, units} = geojson.properties || {maxEdge: 1};
     console.time(name);
-    concave(geojson, maxEdge, {units});
+    concave(geojson, geojson.properties);
     console.timeEnd(name);
 }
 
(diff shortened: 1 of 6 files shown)


packages/turf-concave/index.d.ts
```diff
@@ -5,8 +5,8 @@ import { Point, Feature, FeatureCollection, MultiPolygon, Polygon, Units} from '
  */
 export default function concave(
     points: FeatureCollection<Point>,
-    maxEdge: number,
     options?: {
-      units?: Units
+        maxEdge: number,
```
DenisCarriere: using the `?` means it's optional, at the moment if someone would enter options, then `maxEdge` would be required.

Fix to:
```ts
options?: {
  maxEdge?: number,
  units?: Units
}

File: packages/turf-concave/bench.js [hunk 2]
```diff
@@ -24,9 +24,8 @@ const fixtures = fs.readdirSync(directory).map(filename => {
  * pts3: 0.375ms
  */
 for (const {name, geojson} of fixtures) {
-    const {maxEdge, units} = geojson.properties || {maxEdge: 1};
     console.time(name);
-    concave(geojson, maxEdge, {units});
+    concave(geojson, geojson.properties);
```
DenisCarriere: 👍 I'm liking this Object `options`, you can simply give it the `geojson.properties` like you did, love it!

DenisCarriere:
👍 Looks good!

packages/turf-concave/bench.js
stebogit: Thanks Sir 😄

Issue #962, opened 2017-09-23
Major release pre-announcement (TurfJS v5.0)


DenisCarriere:
## Major release pre-announcement (TurfJS v5.0) ⭐️ 📢 

Due to some drastic changes coming up, the next TurfJS release will be `v5.0` (Major release).

If anyone has any suggestions for breaking changes, the time would be now to do it.

## Timeline

Approximately a 1-2 weeks (end of September/ early October, 2017).

## Major To-Dos before publish

- [ ] Finish adding new optional parameters (`options`) (PR https://github.com/Turfjs/turf/pull/953)
- [x] Convert to ES6 modules using Rollup (PR https://github.com/Turfjs/turf/pull/960)
- [ ] [v5.0 issues](https://github.com/Turfjs/turf/labels/major-release)

## Not happening for next publish

- [x] `jsts` won't be entirely removed

@stebogit @rowanwins Any additional suggestions for `v5.0` changes?

Don't hesitate to send some breaking PR changes (within reason & approved).

CC: @Turfjs/ownership

stebogit:
There's all this stuff waiting for the next major release:
https://github.com/Turfjs/turf/labels/major-release
They should be easy improvements. I'll start working on it as soon as I can

rowanwins:
Aoowww I so wanted to remove jsts for v5 but I think your right, it probably won't happen unfortunately :(

DenisCarriere:
> @rowanwins wanted to remove jsts for v5

Maybe this will help solve our `jsts` large bundle size issue:
https://github.com/bjornharrtell/jsts/pull/317

@rowanwins Do you want to attempt to see if you can make it work with the ES6 module `jsts` bundle?

**JSTS as a ES6 module Rollup bundle**
https://gist.github.com/DenisCarriere/7b268a90f341cf9da2574a9f845b64b3

DenisCarriere:
> @stebogit There's all this stuff waiting for the next major release:

😄 👍 Awesome, I'll tackle a few of those as well.

bjornharrtell:
JSTS - the undead legacy library ported from lovely Java 4 (!) compatible code that just... wont.. die. :smiling_imp:

DenisCarriere:
@bjornharrtell Very 👍

amenadiel:
I believe `@turf/meta` needs a default export in the form

```js
var meta = {
    coordEach: coordEach,
    coordReduce: coordReduce,
    propEach: propEach,
[...]

PR #1022, opened 2017-10-16
Update geojson-rbush to ES modules

DenisCarriere:
## Update [`geojson-rbush`](https://github.com/DenisCarriere/geojson-rbush) to ES modules

`geojson-rbush` now uses Rollup to create the CommonJS & ES Modules.

Ref: https://github.com/Turfjs/turf/issues/962#issuecomment-336920324


DenisCarriere:
@amenadiel Try using the newest release of `geojson-rbush`

Issue #1021, opened 2017-10-16
isobands documentation clarity


jpoynton:
Isoband documentation states that it takes a "grid FeatureCollection of Point features with z-values and an array of value breaks and generates filled contour isobands.". Can you clarify docs to be clear that it requires a square or rectangular grid in order to work.  

The confusion came about because turf.pointGrid states that it "Creates a Point grid from a bounding box, FeatureCollection or Feature". And is close enough that implies that should be valid input to isobands. However, a pointGrid created from an irregular polygon mask will not work in isoband/isoline as the 2D matrix created is not valid. Turf.Isobands unfortunately doesn't tell you this through documentation or error, it just produces an isoband with no geometry. 

See https://jsfiddle.net/rzj3ge87/1/ for an example.

stebogit:
@jpoynton thanks for the feedback.
You're absolutely right, we need to clarify the `@turf/isobands` docs and maybe also throw an error if the input is not a square matrix. Also `@turf/point-grid` docs might have a note about `isobands` compatibility.

BTW @DenisCarriere I find confusing though that `@turf/point-grid`
> "Creates a Point grid from a bounding box, FeatureCollection or Feature"

Why FeatureCollection or Feature are included? I'd say a bounding box should be the (only) correct input. Same for all the other grid modules.

jpoynton:
@stebogit thanks. As for your question to @DenisCarriere I'll have a go at answering. I believe point grid will calculate the bounding box for you if you give it a featurecollection/feature. The other option of a masked point grid would also require a feature collection feature as it's input. And that's useful to me.  I'm generating isochrones using OSRM and use the grid points as my target destinations. If I use a polygon as a mask to point-grid then I can clip the points to a coastline or administrative boundary so that I don't bother looking for journey times that I know are not routeable.

stebogit:
@jpoynton you are right, for `@turf/point-grid` a `Polygon` or `FeatureCollection` (of `Polygon`s only) make sense due to the mask option, however I'd say not for the other grid modules, which don't have the mask option.

DenisCarriere:
@stebogit Good point, what if we make the input `bbox` only and use the `mask` option as `Feature<Polygon>` instead of boolean. That way it's easier to understand the process and have to guess "what happens when I enter a Polygon" as input.

DenisCarriere:
Since we're already pushing for a breaking change, we should do this change now before we officially release `v5.0`

stebogit:
[...]

Issue #1019, opened 2017-10-13
Remove @turf/square-grid "completelyWithin"


stebogit:
@DenisCarriere I am working on [renaming](https://github.com/Turfjs/turf/issues/460#issuecomment-310360602) the `cellSide` parameter of `@turf/hex-grid` and I noticed we mentioned to add also a `completelyWithin` option, already [implemented](https://github.com/Turfjs/turf/pull/612) in `@turf/square-grid`.
However I was wondering if it makes sense to have that parameter, or if actually would be better to have, for both `@turf/hex-grid` and `@turf/square-grid`, always a centered output like for `@turf/point-grid`, where we just [removed](https://github.com/Turfjs/turf/issues/613) the `centered` parameter.
This would make the output of all the grid modules consistent.

DenisCarriere:
👍 Also noticed the same, we should try to revamp all the `<type>-grid` modules to make them behave the same way, they should all have the same Options (or at least the same base options).

stebogit:
All done in #1029

Issue #460, opened 2016-08-19
Question: what does 'cellSize' geometrically measure in the resulting hex_grid


mikejacktzen:
I've used the hexgrid function and i'm confused at the resulting hexagon properties

https://github.com/Turfjs/turf/tree/master/packages/turf-hex-grid

I can't figure out what the cellSize corresponds to geometrically.

```
cellSize number dimension of cell in specified units
```

'number dimension' is pretty ambiguous

Is it the "long diameter", "short diameter", "side" of a hexagon, or none of the above.

I suspect it's none of the above, since  I've used the turf.area() command for a single hexagon (pulled from the grid of hexagons) and it drastically disagrees with back of the envelope hexagon area calculations using "long/short diameter" and "side

For context, my goal is to get a hexagon grid where each hexagon inscribes a circle of radius 10.

morganherlocker:
cc @jseppi

jseppi:
Hrm, good question. I agree that the description of that argument is ambiguous.

It used to be the "radius" (center to vertex) of a regular hex (in degrees), but that was changed back in https://github.com/turf-junkyard/turf-hex-grid/commit/a2e601ac9c6434bfd415827cc1f3db1738d61721#diff-168726dbe96b3ce427e7fedce31bb0bcR36 (part of this PR https://github.com/turf-junkyard/turf-hex-grid/pull/13). Now it looks like the `cellSize` is a kind of meaningless ratio value :\ Perhaps the hex-grid module should be refactored back to using radius but with the specified units instead of always being degrees.

mikejacktzen:
making the cellSize argument correspond to the radius (in specified units) of the center to the vertex would be perfect.

When I had this question, that was the natural first interpretation that jumped to mind.

It would make my use case (inscribing a hexagon into a circle) automatic.

I'm new to using turf, so for me, it's a good idea. However, i realize maybe the weird hexagon ratio was some systematic choice to align the standards with the other set of grid functions.

DenisCarriere:
@mikejacktzen @jseppi I Agree with both your comments, `cellSize` is a bit vague. Using `diameter` or `radius` would make it easier to understand.

Another thing I noticed is `triangle=true` translates to `radius` for the `cellSize` and `triangle=false` (hexagon) translates to `diameter` for the `cellSize`.

If you measure the diameter of the very bottom left hexagon it's perfectly 50 miles.
[...]

PR #1029, opened 2017-10-18
Revamp grid modules

stebogit:
Ref https://github.com/Turfjs/turf/issues/1019

- [x] Removed `@turf/square-grid` `completelyWithin` parameter
- [x] Renamed `cellDiameter` to `cellSide` in all grid modules (https://github.com/Turfjs/turf/issues/460)
- [x] Centered output of all grid modules (https://github.com/Turfjs/turf/issues/613)
- [x] Added input validation to `@turf/square-grid` and `@turf/hex-grid`
- [x] Added `properties` parameter (passed to the output) to `@turf/square-grid` and `@turf/hex-grid`
- [x] refactored test for `@turf/hex-grid`
- [x] improve mask option for `@turf/point-grid` (https://github.com/Turfjs/turf/issues/1021#issuecomment-337293877)
- [ ] create Mercator friendly grid features

**New** ⚠️ 
- [x] Modules which input Point data can simply use an `Array<number>`.
- [x] Optional parameters should be declared initially `options.mask` => `mask`
- [x] `options.mask` should have (Multi)Polygon geometry type validation.
- [x] Add `@turf/truncate` to output test results to prevent any precision decimal errors.

**New - Improvements** ⚠️ 

- [x] Add translate properties to grid features (`options.properties`)
- [x] Add mask to each grid module (`options.mask`)
- [x] Use `@turf/boolean-within` or `@turf/intersect` for `options.mask`

**Update fixtures for**

- [x] `@turf/interpolate`
- [x] `@turf/transform-scale`


stebogit:
@DenisCarriere I noticed all the grid have projection distortion because of the way the grid elements are created:
screen shot 2017-10-17 at 10 45 24 pm

screen shot 2017-10-17 at 10 47 39 pm

screen shot 2017-10-18 at 12 17 33 am

Should we modify that and have Mercator-projected grids or we consider these correct outputs?

DenisCarriere:
> Should we modify that and have Mercator-projected grids or we consider these correct outputs
Yes I think we should make it Mercator friendly since 98% of users will be using a Leaflet type map to display their geojson.

DenisCarriere:
🤔 ... @stebogit might be best to use a new branch.. lots of conflicts and might be just easier to start a new PR based on the master branch.

I think these conflicts are caused from the `@turf/distance` earth radius update.

stebogit:
@DenisCarriere I'm not sure what's the correct type definition for `mask` here, kind of confused between `Polygon` and `Feature<Polygon>` 🤔 :

screen shot 2017-10-19 at 12 17 33 am

DenisCarriere:
@stebogit Sorry been busy the past days, I'm in Boulder at SOTM now.

The difference between `Polygon` & `Feature<Polygon>`:

- `Polygon` is the same as `Geometry<Polygon>` which would look like this:

```js
{
  type: "Polygon",
  coordinates: Array<number>
}
```
- `Feature<Polygon>` is the Feature wrapped with the Polygon Geometry.

You can always look at the `@types/geojson` file to decipher how the `GeoJSON.Polygon` & `GeoJSON.Feature<GeoJSON.Polygon>` have been created:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/geojson/index.d.ts#L68-L71

`@turf/helpers` tries to replicate an exact match of the GeoJSON type definition, except you don't have to add `GeoJSON.<type>` you can simply just import it from `@turf/helpers`.

```js
import {Polygon, Feature} from '@turf/helpers'
```

packages/turf-point-grid/types.ts
```diff
@@ -1,10 +1,11 @@
-import { BBox } from '@turf/helpers';
+import { BBox, polygon, Polygon} from '@turf/helpers';
 import pointGrid from './';
 
 const cellSide = 50;
 const bbox: BBox = [-95, 30, -85, 40];
+const poly: Polygon = polygon([[[20, 30], [10, 10], [20, 20], [20, 30]]]);
```
DenisCarriere: @stebogit This is my bad... `Polygon` was short for `Feature<Polygon>`, however now `Polygon` from `@turf/helpers` is now equivalent to `Geometry<Polygon>`.

The reason for the error is because `mask` doesn't handle adding `Geometry<Polygon>` (which should be a valid input, both Geometry & Feature).

packages/turf-point-grid/index.d.ts
```diff
@@ -1,14 +1,14 @@
...
  */
 export default function pointGrid(
-    bbox: BBox | AllGeoJSON,
+    bbox: BBox,
     cellSide: number,
     options?: {
         units?: Units,
         properties?: Properties,
-        bboxIsMask?: boolean;
+        mask?: Polygon|MultiPolygon;
```
DenisCarriere: The correct definition would be:

> To support both `Geometry<Polygon|MultiPolygon>` & `Feature<Polygon|MultiPolygon>`

```ts
export default function pointGrid(
    bbox: BBox,
    cellSide: number,
    options?: {
        units?: Units,
        properties?: Properties,
        mask?: Feature<Polygon | MultiPolygon> | Polygon | MultiPolygon;
    }
): FeatureCollection<Point>;
```

packages/turf-point-grid/index.js
```diff
@@ -1,18 +1,17 @@
...
  *
  * @name pointGrid
- * @param {BBox|GeoJSON} bbox extent in [minX, minY, maxX, maxY] order
- * @param {number} cellSide the distance between points
+ * @param {Array<number>} bbox extent in [minX, minY, maxX, maxY] order
+ * @param {number} cellSide the distance between points, in units
  * @param {Object} [options={}] Optional parameters
  * @param {string} [options.units="kilometers"] used in calculating cellSide, can be degrees, radians, miles, or kilometers
- * @param {number} [options.bboxIsMask=false] if true, and bbox is a Polygon or MultiPolygon, the grid Point will be created
+ * @param {Feature<Polygon|MultiPolygon>} [options.mask=undefined] if passed a Polygon or MultiPolygon, the grid Points will be created only inside it
```
DenisCarriere: No need to define the option as `undefined`, `[options.mask]` is perfectly fine.

packages/turf-point-grid/index.js
```diff
@@ -1,73 +1,72 @@
...
+    var west = bbox[0];
+    var south = bbox[1];
+    var east = bbox[2];
+    var north = bbox[3];
 
-    var  xFraction = cellSide / (distance(point([west, south]), point([east, south]), options));
-    var  cellWidth = xFraction * (east - west);
-    var  yFraction = cellSide / (distance(point([west, south]), point([west, north]), options));
-    var  cellHeight = yFraction * (north - south);
+    var xFraction = cellSide / (distance(point([west, south]), point([east, south]), units));
```
DenisCarriere: Distance no longer requires the input data to be Point, it can be an Array of numbers.

```js
distance([west, south], [east, south], options)
```

Also, all modules have changed to the `options` param syntax, even the ones with 1 param.

packages/turf-point-grid/index.js
```diff
@@ -1,73 +1,72 @@
...
  */
 function pointGrid(bbox, cellSide, options) {
     // Optional parameters
     options = options || {};
-    if (typeof options !== 'object') throw new Error('options is invalid');
-    var bboxIsMask = options.bboxIsMask || false;
-    var properties = options.properties || {};
+    if (!isObject(options)) throw new Error('options is invalid');
+    var units = options.units;
+    var maskIsPoly = options.mask && (getType(options.mask) === 'Polygon' || getType(options.mask) === 'MultiPolygon');
```
DenisCarriere: I wouldn't use a `maskIsPoly` approach, instead add some type validation for the input and call it `mask`.

packages/turf-point-grid/index.js
```diff
@@ -1,73 +1,72 @@
...
     while (currentX <= east) {
-        var  currentY = south + deltaY;
+        var currentY = south + deltaY;
         while (currentY <= north) {
-            var  pt = point([currentX, currentY], properties);
-            if (bboxIsMask === true && bboxIsPoly) {
-                if (inside(pt, bboxMask)) {
+            var pt = point([currentX, currentY], properties);
+            if (maskIsPoly) {
+                if (inside(pt, options.mask)) {
```
DenisCarriere: For all optional parameters, you should use the param name `mask` instead of `options.mask`.

DenisCarriere:
- [ ] Modules which input Point data can simply use an `Array<number>`.
- [ ] Optional parameters should be declared initially `options.mask` => `mask`
- [ ] Optional parameters can have geometry type validation.

stebogit:
Thanks @DenisCarriere!
I'll work on Mercator friendly output

DenisCarriere:
> at SOTM now.
@stebogit Definitely awesome talks and great people, you should definitely come to the next US one.

DenisCarriere:
> I'll work on Mercator friendly outputs 

👍 Nice!

DenisCarriere:
- [ ] Add translate properties to grid features (`options.properties`)
- [ ] Add mask to each grid module (`options.mask`)
- [ ] Use `@turf/boolean-inside` and/or `@turf/boolean-touches` for `options.mask`

DenisCarriere:
@stebogit Woot 🎉 First passing commit.

Next thing to implement is the Mercator outputs and then I think we are ready to merge this PR.

kusamochi:
Hi, and sorry with my limited English skills.
I am using pointGrid and I found features coordinates are missing when unit is set to degrees. doesn't occur when close to 0,0, the further away from 0,0 the more it is missing. is it something to do with projection distortion? thanks

PR #1033, opened 2017-10-20
test `@turf/transformRotate` against samples with mutate turned on

zbyte64:
Noticed some strange behavior when rotating polygons in-place. Decided to modify the unit tests to also do a mutate in place against the samples and then to a deep compare; this resulted in tests failing.


stebogit:
Hi @zbyte64, what kind of strange behavior are you noticing? Would you please provide some sample of the data or a snippet of code to reproduce the issue?

Passing `mutate = true` to any `transform` module allows the library to process the actual input data, i.e. mutating the input (this way is faster, skipping the cloning step).
Thus the current tests against features would clearly fail, as they expect two different features as output (original + rotated), while with mutation they would end up being equals, both rotated.

DenisCarriere:
Going to close this PR, the `mutate` option is already being tested here:
https://github.com/Turfjs/turf/blob/master/packages/turf-transform-rotate/test.js#L48-L58

Issue #1040, opened 2017-10-25
@turf/shortest-path used for maritime industry


DenisCarriere:
## [`@turf/shortest-path`](https://github.com/Turfjs/turf/tree/master/packages/turf-shortest-path) used for maritime industry

>  @bulutkartal Mine was just an idea to be able to make it useful for maritime industry. [Land polygons](http://openstreetmapdata.com/data/land-polygons) can be implemented on maps and with min safe distance to polygons can be used for marine routing. This way they will be able to set how far their boat/ship will be from the closest land. Very useful for marine industry.

The `shortest-path` module provides a few Optional parameters.

- [x] `options.obstacles` => Can be used for land
- [ ] `options.minDistance` => Can be used for safe distance ( ⚠️ NOT IMPLEMENTED YET)

Ref issue: https://github.com/Turfjs/turf/issues/788 PR: https://github.com/Turfjs/turf/pull/956

## To-Do

- [ ] Add `options.minDistance` feature
- [ ] Add more complex test cases

bulutkartal:
@DenisCarriere here is the test example can be used. basically it is Bermuda Triangle with start and end points are out of it. I also noticed this can be use for aviation as well. so they can set restricted areas for flight, set a min distance and set their flight routes out of those areas with min safe distance. So can be used for different industries as well.

{
    "start": [-60.925, 22.335],
        "end": [-79.991, 29.965],
            "obstacles": {
        "type": "Feature",
            "properties": { },
        "geometry": {
            "type": "Polygon",
                "coordinates": [
                    [
                        [
                            -80.190,
                            25.774
                        ],
                        [
                            -66.118,
                            18.466
                        ],
                        [
                            -64.757,
                            32.321
                        ]
                    ]
                ]
        }
    }
}

rowanwins:
Now available as a first cut thanks to @stebogit 
http://turfjs.org/Docs/#shortestPath

ngoluan:
Do you have any tips on how to use this with a land geojson? I'm trying to plot a course between two points and avoid land masses, which is represented by a GeometryCollection of Polygons. But, I keep encountering a "Uncaught Error: coordinates must contain numbers" error. Alternatively, I tried using a multipolygon as an obstacle but it throws an "invalid obstacle error".

I've attached geojson file that I use here. 
[land_sm.zip](https://github.com/Turfjs/turf/files/1753998/land_sm.zip)

Thanks!

PR #1041, opened 2017-10-25
Add Bermuda Triangle as test case

DenisCarriere:
## Add Bermuda Triangle as test case

Ref: https://github.com/Turfjs/turf/issues/1040#issuecomment-339474346

Test case provided by @bulutkartal:

image

## Note
@stebogit I changed the input test case to be a pure GeoJSON Feature Collection, that way it's easier to preview the input.

Also, GitHub now displays `.json` as a GeoJSON map, the only reason why we had the data as `.geojson` is because it wouldn't display as a map in GitHub, now that's fixed we can switch over the input/output datasets to `.json` which is a tad easier to read by simply using `require('data.json')`


stebogit:
> Also, GitHub now displays .json as a GeoJSON map

Cool! 👍

PR #1046, opened 2017-10-26
fix bug when the number of segments is integer(line-chunk)

xiongjiabin:
if numberOfSegments is an integer, like line length is 100, and segmentLength is 10, it's just 10 segments, no need to plus 1 to 11, otherwise will throw an exception later.


DenisCarriere:
@xiongjiabin Why did you close this PR? this looked fine...? just needed a few linting fixes

DenisCarriere:
@xiongjiabin You should need to fix the linting issues:

This should be the code:

```js
    var numberOfSegments = lineLength / segmentLength;

    // If numberOfSegments is integer, no need to plus 1
    if (!Number.isInteger(numberOfSegments)) {
        numberOfSegments = Math.floor(numberOfSegments) + 1;
    }
```

xiongjiabin:
I found failed in lint, so close the PR, thanks!

DenisCarriere:
This is why we have automated testing, to catch minor issues like Linting, no need to close PR, just commit a new fix with the provided error from TravisCI.

PR #1051, opened 2017-10-29
Improve turf v5 Rollup build

DenisCarriere:
## `@turf/turf` v5.0 Rollup build

Improvements / testing for v5.0 release Rollup build.

CC: @rowanwins


DenisCarriere:
More to follow on on v5 final build

Issue #1050, opened 2017-10-28
polygonTangents gives Error "No coordinates passed"


bluepup:
I'm not sure if there is something invalid in the coordinates of my Polygon, it seems OK.
With the following code I get "No coordinates passed" in Firefox 52.4.0 and newest  minified file from npmcdn. Even with successfully truncated coordinates (turf.truncate) the same error occurs in polygonTangents.

```javascript
var point = {"type":"Feature","properties":{},"geometry":{"type":"Point","coordinates":[8.725,51.57]}};

var efe = {
  "type":"Feature",
  "properties":{},
  "geometry":{
	"type":"Polygon",
	"coordinates":[
	  [[8.788482103824089,51.56063487730164],
	  [8.788583,51.561554],[8.78839,51.562241],
	  [8.78705,51.563616],[8.785483,51.564445],
	  [8.785481,51.564446],[8.785479,51.564447],
	  [8.785479,51.564449],[8.785478,51.56445],
	  [8.785478,51.564452],[8.785479,51.564454],
	  [8.78548,51.564455],[8.785482,51.564457],
	  [8.786358,51.565053],[8.787022,51.565767],
	  [8.787024,51.565768],[8.787026,51.565769],
	  [8.787028,51.56577],[8.787031,51.565771],
	  [8.787033,51.565771],[8.789951649580397,51.56585502173034],
	  [8.789734,51.563604],[8.788482103824089,51.56063487730164]]
	]
  }
};

turf.polygonTangents(point, efe);

//Error: No coordinates passed
```

[turftangenttsError.txt](https://github.com/Turfjs/turf/files/1424514/turftangenttsError.txt)

DenisCarriere:
Thanks @bluepup for reporting this issue, your input data seems correct.

I've created a PR with the sample data you provided, maybe @rowanwins will have a better idea what the issue is.

rowanwins:
Resolved by #1058

PR #1052, opened 2017-10-29
Add test case for issue #1050

DenisCarriere:
# Add test case for issue #1050

The input data seems correct, `rtan` is not being handled correctly in `processPolygon()`.

image

CC: @rowanwins


DenisCarriere:
Merging, however this still remains to be an issue, this PR simply adds the test case.

PR #1056, opened 2017-10-29
Drop @turf/isobands JSDocs example

DenisCarriere:
## Drop [`@turf/isobands`](https://github.com/Turfjs/turf/tree/master/packages/turf-isobands) JSDocs example

The current example of isobands is not stable and causes sporadic TravisCI errors due to lack of memory.

We can always re-add the JSDocs `@example`, except we need to make sure it's stable and simple enough to handle on TravisCI testing VM's.


stebogit:
So sorry to see `@turf/isobands` failing so badly... 😞 
I will work on fixing it as soon as I find the time

Issue #70, opened 2014-01-09
smooth for polygons


morganherlocker:
Uses the bezier curve algo on all arcs within a polygon or set of polygons, then transforms the results into valid topology.

andrewharvey:
@MarkItzero added a PR at https://github.com/turf-junkyard/turf-bezier/pull/7 which looks like it's trying to solve the same problem as this.

stebogit:
Implemented with [`@turf/bezier`](https://github.com/Turfjs/turf/tree/master/packages/turf-bezier)

Ref #582, #360 and [`turf-junkyard/turf-bezier#7`](https://github.com/turf-junkyard/turf-bezier/pull/7)

andrewharvey:
@stebogit I'm not sure I understand why this was closed, https://github.com/Turfjs/turf/tree/master/packages/turf-bezier still says "Takes a line".

stebogit:
@andrewharvey you're right, my bad, I didn't notice this was actually referring to polygons! Sorry

dhivehi:
Hope this might be helpful: http://viglino.github.io/ol3-ext/examples/map.geom.cspline.html

DenisCarriere:
👍 Very cool example @dhivehi Thanks for sharing

rowanwins:
Implemented by #1077

PR #1077, opened 2017-11-05
New module - smooth polygon 

rowanwins:

First crack at the `polygonSmooth` module (or should it be `smoothPolygon` !?) in an attempt to progress #70 at long last!

This uses the [chaiken algorithm](http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html) which QGIS offers. Apparently it's not super-performant on larger polygons but it's certainly better than nothing and suitable for most smaller polys.

A smidge of tidying required to make sure it supports the full range of features vs geoms vs polys vs multipolys etc but it's not far off.


rowanwins:
Oh and ps we can leave this till v5.0.1

packages/turf-polygon-smooth/index.js
```diff
@@ -0,0 +1,55 @@
...
+ * var polygon = turf.polygon([[[11, 0], [22, 4], [31, 0], [31, 11], [21, 15], [11, 11], [11, 0]]]);
+ *
+ * var smoothed = turf.polygonSmooth(polygon)
+ *
+ * //addToMap
+ * var addToMap = [smoothed, polygon];
+ */
+function polygonSmooth(inputPolys, options) {
+    var outPolys = [];
+    options = options || {iterations: 1};
```
DenisCarriere: 🤔 Don't think this would be the best way to define default settings, in the case of options given as an empty `Object` then the default wouldn't be defined as `1`.

```js
turf.polygonSmooth(polygon, {})
// options.iterations => undefined
```

packages/turf-polygon-smooth/index.js
```diff
@@ -0,0 +1,55 @@
...
+import { featureCollection, polygon, isObject } from '@turf/helpers';
+
+/**
+ * Smooths a {@link Polygon}. Based on [Chaikin's algorithm](http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html).
+ * Warning: may create degenerate polygons.
+ *
+ * @name polygonSmooth
+ * @param {FeatureCollection<Polygon>} inputPolys to smooth
+ * @param {Object} [options] Optional parameters
+ * @param {string} [options.iterations] THe number of times to smooth the polygon. A higher value means a smoother polygon.
```
DenisCarriere: Define default values in JSDocs `[options.iterations=1]`

packages/turf-polygon-smooth/index.js
```diff
@@ -0,0 +1,55 @@
+import { coordEach, geomEach } from '@turf/meta';
+import { featureCollection, polygon, isObject } from '@turf/helpers';
+
+/**
+ * Smooths a {@link Polygon}. Based on [Chaikin's algorithm](http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html).
+ * Warning: may create degenerate polygons.
+ *
+ * @name polygonSmooth
+ * @param {FeatureCollection<Polygon>} inputPolys to smooth
+ * @param {Object} [options] Optional parameters
```
DenisCarriere: Optional, I've been been giving an empty Object as the default value `[options={}]`

packages/turf-polygon-smooth/test.js
```diff
@@ -0,0 +1,30 @@
...
+};
+
+const fixtures = fs.readdirSync(directories.in).map(filename => {
+    return {
+        filename,
+        name: path.parse(filename).name,
+        geojson: load.sync(directories.in + filename)
+    };
+});
+process.env.REGEN = true
```
DenisCarriere: Having a hard time define environment variables on Windows 😉

rowanwins: yes - stupid stinkin environment variables, always forget how to set them because it has to be so super specific!

rowanwins:
Yeah it's a crazy simple module like this. The [ArcGIS docs](http://desktop.arcgis.com/en/arcmap/10.3/manage-data/editing-existing-features/smoothing-a-feature-smooth-.htm) provide a couple of other approaches that look a bit nicer but I can't find any literature on those approaches.

packages/turf-polygon-smooth/test.js
DenisCarriere: True that, but once the fixtures are defined, "in theory"  they shouldn't be changing.

I think in Windows you can use:

```
setx DEBUG true
```

https://technet.microsoft.com/en-us/library/cc755104(v=ws.11).aspx

DenisCarriere:
👍 I like it!! I'll have a glance at it again once v5.0 is released

rowanwins:
Urgh supporting multipolys, polys with holes, and collections stinks, lineEach() is an improvement but man it still feels trickier than it needs to be...

DenisCarriere:
🤔 .. sounds painful.. @rowanwins do your best, we can always improve it later on when we have more time.

Maybe throw some errors on the Geometry types that are not yet implemented and that way it will force us to implement them down the road.

rowanwins:
I'm going to hold off on progressing this until the improved `@turf/meta` is released @DenisCarriere - that should make this task much simpler to finish off.

rowanwins:
Latest commit supports polys, poly with holes, multipolys.

Only thing not supported after the latest commit is multipolys with holes due to further bug with #1092 . Once the bug is resolved however I dont think this will actually need any further code changes, just another test.

DenisCarriere:
@rowanwins The bug should be patched & fixed now 👍

rowanwins:
This is good to go whenever @DenisCarriere

DenisCarriere:
👍 @rowanwins Nice

dbauszus:
👍 Super. I have been waiting for this feature since the issue was first opened. Will this be available in the Turf 5.2.0 release? I would drop the feature into [this jsfiddle](https://jsfiddle.net/goldrydigital/8deoap46) which I use to demo the generation of drivetime isobands with turfjs.

PR #1080, opened 2017-11-08
Fix link to Wikipedia page

aemkei:
The old markdown syntax was in the wrong order. URL and text are now working correctly.


rowanwins:
Thanks for the contribution @aemkei - that had been bugging me too but kept forgetting to fix it :)

PR #1087, opened 2017-11-11
Create an Ellipse submodule based on Circle

muziejus:
Comments:

* I haven't changed the `index.d.ts` file because I don't know what it does.

Please fill in this template.


DenisCarriere:
Nice! Thanks for the contributions! Going to have a look at this more closely later on.

I'll help out to make the TravisCI tests pass.

packages/turf-ellipse/index.js
```diff
@@ -0,0 +1,68 @@
...
+ * const center = [-75.9975, 40.730833];
+ * const xRadius = 0.5;
+ * const yRadius = 0.1;
+ * const options = {steps: 10, properties: {foo: 'bar'}};
+ * const ellipse = turf.ellipse(center, xRadius, yRadius, options);
+ *
+ * //addToMap
+ * const addToMap = [turf.point(center), ellipse]
+ */
+module.exports = function (center, xRadius, yRadius, options) {
```
DenisCarriere: We are using ES Modules now + this type of module should also be using `export default`.

```js
function ellipse(center, xRadius, yRadius, options) {
   ...
}
export default ellipse
```

packages/turf-ellipse/index.js
```diff
@@ -0,0 +1,68 @@
+const polygon = require('@turf/helpers').polygon;
+const lengthToDegrees = require('@turf/helpers').lengthToDegrees;
+const getCoord = require('@turf/invariant').getCoord;
+
+/**
+ * Takes a {@link Point} and calculates the ellipse polygon given two axes in degrees and steps for precision.
+ *
```
DenisCarriere: Most modules will contain a `@name` JSDocs tag.

```js
/**
 * @name ellipse
 */
```

packages/turf-ellipse/index.js
```diff
@@ -0,0 +1,68 @@
+const polygon = require('@turf/helpers').polygon;
+const lengthToDegrees = require('@turf/helpers').lengthToDegrees;
+const getCoord = require('@turf/invariant').getCoord;
```
DenisCarriere: Since Turf now uses ES Modules, you should be using the `import` syntax.

```js
import { polygon, lengthToDegrees } from '@turf/helpers';
import { getCoord } from '@turf/invariant';
```

packages/turf-ellipse/index.d.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="geojson" />
+
+type Feature = GeoJSON.Feature<any> | GeoJSON.GeometryObject;
+
+/**
+ * http://turfjs.org/docs/#ellipse
+ */
+declare function ellipse(feature1: Feature, feature2: Feature): boolean;
```
DenisCarriere: ... 🤔 I know Typescript is new for a lot of folks, but you could simply copy the Typescript definition from `@turf/circle` which will be very similar to the new ellipse module.

https://github.com/Turfjs/turf/blob/master/packages/turf-circle/index.d.ts

DenisCarriere:
When creating an Ellipse, would it also be possible points (Coordinates) to define `xRadius` & `yRadius`... 🤔  might be tricky, but I think that would be possible using `@turf/distance`.

packages/turf-ellipse/index.js
```diff
@@ -0,0 +1,68 @@
...
+    const getTanDeg = function (deg) {
+        const rad = deg * Math.PI / 180;
+        return Math.tan(rad);
+    };
+
+    // validation
+    if (!center) throw new Error('center is required');
+    if (!xRadius) throw new Error('xRadius is required');
+    if (!yRadius) throw new Error('yRadius is required');
+    if (typeof options !== 'object') throw new Error('options must be an object');
```
DenisCarriere: There's a helper function `isObject` in `@turf/helpers` instead of using `typeof`.

packages/turf-ellipse/index.js
```diff
@@ -0,0 +1,68 @@
...
+    if (typeof steps !== 'number') throw new Error('steps must be a number');
```
DenisCarriere: There's a helper function `isNumber` in `@turf/helpers` instead of using `typeof`.

packages/turf-ellipse/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+    const centerCoords = getCoord(center);
```
DenisCarriere: At the moment TurfJS is only using ES5 code, ES6 is not being compiled to ES5 for the final build (at the moment).

- Replace `const` with `var`

packages/turf-ellipse/index.js
```diff
@@ -0,0 +1,68 @@
...
+    xRadius = lengthToDegrees(xRadius, units);
+    yRadius = lengthToDegrees(yRadius, units);
+
+    let coordinates = [];
```
DenisCarriere: Same as above, replace `let` with `var`

packages/turf-ellipse/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@turf/ellipse",
+  "version": "5.0.4",
```
DenisCarriere: Could you change the version to `5.0.0` for the first publish, Lerna will decide the patch version.

packages/turf-ellipse/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "description": "turf ellipse module",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "files": [
+    "index.js",
+    "index.d.ts"
```
DenisCarriere: Missing `main.js` that will be the CommonJS distributed file from Rollup

packages/turf-ellipse/package.json
```diff
@@ -0,0 +1,43 @@
...
+  ],
+  "scripts": {
+    "test": "node test.js",
```
DenisCarriere: Use the same scripts as all the other modules.

```json
  "scripts": {
    "pretest": "rollup -c ../../rollup.config.js",
    "test": "node -r @std/esm test.js",
    "bench": "node -r @std/esm bench.js"
  },
```

DenisCarriere: You'll also need to install `@std/esm` with the following config:

```json
  "@std/esm": {
    "esm": "js",
    "cjs": true
  }
```

DenisCarriere:
I'm looking at the ellipse code base, why don't we simply add `xRadius` & `yRadius` as options to `@turf/circle`?

DenisCarriere:
> My goal, if it's not obvious, is to add Standard Deviational Ellipse functionality to turf. The potential for geostatistical packages in interactive maps (in leaflet, say)…

👍 Super! 🚀 Lets definitely get this done, I think expanding the `@turf/circle` module will be our best bet. I'll set up a PR to get us started.

muziejus:
@DenisCarriere Thanks for making these changes. A few notes:

Abstractly:
1. I don't think adding a second radius to `circle` will do the trick, because `circle` uses `distance` to plot the points on the circle, but the distance is variable in an ellipse, relying on the interplay between the two radii. I think the more traditional means of measuring w/ tangents works great. Whether that should be abstracted to its own helper, on the other hand…
1. I don't think we can use points instead of distances for the radii because the radii have to be perpendicular to each other. 

Concretely:
1. Some of the lack of packages, etc., was a function of how the default new submodule script works. I'm not very familiar w/ node, so this required some guess work.
1. I stayed away from using `import`, etc., because they were simply breaking my tests:
```
/Users/moacir/Projects/turf/packages/turf-ellipse/test.js:1
(function (exports, require, module, __filename, __dirname) { import test from 'tape';
                                                              ^^^^^^

SyntaxError: Unexpected token import
```
Switching to `require()` syntax fixed that.

stebogit:
@muziejus @DenisCarriere nice addition here. 👍 

Are the radii required by the module actually the semi-minor and the semi-major axes of the ellipse? If so I'd describe that way in the docs.

May I suggest to include also an optional `tilt` parameter related to the major axis of the ellipse for example? You could probably just apply `@turf/transform-rotate` at the end if the parameter (angle) is defined.

I don't think having this as a separate module would be a terrible idea, it would make it "easier to find" or like more explicit that Turf has that functionality. Just a thought.

muziejus:
@stebogit You're right; they are the semi-minor and semi-major axes. My terminology is bad. But we can't use terms like "major" and "minor" because they actually relate to the x and y--it's very easy to  send, say, `ellipse(place, 2, 10);` and end up with an ellipse where the semi-major axis is along the y-axis. 

Maybe `xSemiAxis` and `ySemiAxis`?

muziejus:
@stebogit As for tilting, I considered that, since rotation is a vital aspect of the standard deviational ellipse, but given there's already a diff utility, I don't think adding another parameter is necessary.

If I were to add a `tilt` option, why not also add a `translate` or `scale` option, etc.?

stebogit:
@muziejus looking at the picture in my understanding we are passing `O` (as a point), `a` and `b` (as lengths); with those and a `tilt` angle (`𝛼`), you can describe the ellipse in the picture. In fact the center defines the position (no need for a `translate` parameter) and the axes define the dimensions (no need for a `scale` param). However without a `tilt` you would actually define the gray ellipse in the picture (where `a=c` and `b=d`).
Unless we always want to create ellipses with axes parallel to the `x` and `y` axes (is this actually the case?), I'd say we need a `tilt`.

ellipse

> Maybe xSemiAxis and ySemiAxis?

We could easily use `major` and `minor` axes and internally divide them by 2

> there's already a diff utility

What do you mean?

muziejus:
@stebogit You're right. I meant "separate utility" in that the user could always use `transformRotate()` on their own after the fact, but I don't see why I should be adamant about not adding the functionality into this submodule.

DenisCarriere:
✅ 👍 for adding `tilt` as an option to this module.

✅ 👍 using `major` & `minor`.

✅ 👍 Creating it's own module `@turf/ellipse`

❌ 👎 add a translate or scale option (doesn't seem necessary given there's already 3 parameters that will allow to control that)

It is best to create it's own dedicated module `@turf/ellipse` like @stebogit mentioned, it would be easier to find and it would make it easier to understand the context of the `major` & `minor` options given it belongs to a module called ellipse.

DenisCarriere:
🤔 Seems like the build is failing because of a linting issue.

```
/home/travis/build/Turfjs/turf/packages/turf-ellipse/index.js
  50:13  error  'angle' is already defined                    no-redeclare
  66:1   error  Expected indentation of 4 spaces but found 0  indent
✖ 2 problems (2 errors, 0 warnings)
  1 error, 0 warnings potentially fixable with the `--fix` option.
```

DenisCarriere:
**Cannot find module @turf/transform-rotate**

Missing dependency 😄 The joys... Thank you TravisCI for being there for us.

```
lerna ERR! Error [ERR_MISSING_MODULE]: Cannot find module @turf/transform-rotate
lerna ERR!     at Object.<anonymous> (/home/travis/build/Turfjs/turf/packages/turf-ellipse/main.js:5:23)
lerna ERR!     at Object.<anonymous> (/home/travis/build/Turfjs/turf/packages/turf-ellipse/main.js)
lerna ERR!     at Module._compile (module.js:641:30)
lerna ERR! npm ERR! code ELIFECYCLE
```

DenisCarriere:
👍 Awesome stuff @muziejus!

Issue #1082, opened 2017-11-09
New major release! Turf 5.0 🎉


DenisCarriere:
# New major release! Turf 5.0 🎉

TurfJS releases are being tracked using [Milestones](https://github.com/Turfjs/turf/milestone/4?closed=1).

## Contributors

- @DenisCarriere
- @rowanwins
- @stebogit
- @stevage

## ⭐️ Major Updates

- TurfJS now supports ES Modules ([Related PR's](https://github.com/Turfjs/turf/pulls?q=is%3Apr+modules+is%3Aclosed+label%3Aes-module))
- Optional parameters are now defined as an `Object`.

### [`@turf/voronoi`][voronoi]
Takes a FeatureCollection of points, and a bounding box, and returns a FeatureCollection of Voronoi polygons.
(PR https://github.com/Turfjs/turf/pull/1043 - Author @stevage)

### [`@turf/shortest-path`][shortest-path]
Returns the shortest path from start to end without colliding with any feature in obstacles
(PR https://github.com/Turfjs/turf/pull/956 - Author @stebogit)

### [`@turf/boolean-parallel`][boolean-parallel]
Boolean-Parallel returns True if each segment of `line1` is parallel to the correspondent segment of `line2`
(PR https://github.com/Turfjs/turf/pull/941 - Author @stebogit)

### [`@turf/nearest-point-on-line`][nearest-point-on-line]
Takes a {@link Point} and a {@link LineString} and calculates the closest Point on the (Multi)LineString.
(PR https://github.com/Turfjs/turf/pull/939 - Author @stebogit)

## 🏅 New Features/Enhancements
- Updates [`@turf/unkink-polygon`][unkink-polygon] testing & added `flattenEach` instead of using `flatten`.
(PR https://github.com/Turfjs/turf/pull/889)

- Add multi-geomtry support to [`@turf/line-split`](line-split)
(PR https://github.com/Turfjs/turf/pull/1078)

- Improve [`@turf/meta`](meta) `lineEach` method to provide properties, id and bbox 
(PR https://github.com/Turfjs/turf/pull/1010)

## 🐛 Bug Fixes
- Fixes [`@turf/helpers`](helpers) earth radius variable 
(PR https://github.com/Turfjs/turf/pull/1012)

- Fixes [`@turf/polygon-tangents`](polygon-tangents) bug 
(PR https://github.com/Turfjs/turf/pull/1058)

[...]

PR #1088, opened 2017-11-11
Update CHANGELOG.md

DenisCarriere:
Update Changelog using Release logs.

Also mentioned via https://github.com/Turfjs/turf/issues/1082#issuecomment-343274461


DenisCarriere:
@ka7eh Thanks for mentioning this. 👍

Issue #1092, opened 2017-11-15
coordEach indexing bug


rowanwins:
I noticed this evening that when using `coordEach` that it doesn't really deal with holes properly.

So for example I pass in a poly with a hole like
````
[ [ [ 100, 0 ], [ 101, 0 ], [ 101, 1 ], [ 100, 1 ], [ 100, 0 ] ],
  [ [ 100.2, 0.2 ], [ 100.8, 0.2 ], [ 100.8, 0.8 ],[ 100.2, 0.8 ],[ 100.2, 0.2 ] ]
]
````
if I do something like
```js
    coordEach(poly, function (currentCoord, currentIndex, featureIndex, featureSubIndex) {
        console.log(currentCoord)
        console.log(currentIndex, featureIndex, featureSubIndex)
```
I get 
````
[ 100, 0 ]
0 0 0
[ 101, 0 ]
1 0 0
[ 101, 1 ]
2 0 0
[ 100, 1 ]
3 0 0
[ 100.2, 0.2 ]
4 0 0
````
So the hole get's the same featureIndex and featureSubIndex as the rest of the poly which means I can't detect that the coord belongs to a hole.

Would be great to add some polys with holes to the test fixtures to those `meta` modules.

**Comment by DenisCarriere** [comment]
👍 Agreed, I've added a 3rd index (`geometryIndex`) which indicates the index when iterating over the geometry (used in Polygon & MultiPolygon geometries).

There are 3 major index tiers which are broken down like so:
-  `featureIndex` => `multiFeatureIndex` => `geometryIndex`

Using those 3 indexes you should be able to find your way back to the exact geometry part of the feature.

```js
turf.coordEach(features, function (currentCoord, coordIndex, featureIndex, multiFeatureIndex, geometryIndex) {
  //=currentCoord
  //=coordIndex
  //=featureIndex
  //=multiFeatureIndex
  //=geometryIndex
});
```

@rowanwins Have a go at it and let me know if you are having any issues with the new update.

rowanwins:
Thanks @DenisCarriere yeah I thought we meet need a new level. I'll try and get smooth close off this weekend which enable me to test it

DenisCarriere:
👍 Sounds like a plan!

I think the NEXT major release we can re-organize those optional indexes using an Object.

Something like this =>

```js
coordEach(features, (currentCoord, indexes) => {
  //=currentCoord
  //=indexes.coordIndex
  //=indexes.featureIndex
[...]

PR #1093, opened 2017-11-15
Support for polygon holes for @turf/meta

rowanwins:

Related to #1092 

Added a test fixture for a polygon with a hole and made a slight amendment to coordEach. Would be good to check that you're happy with this before proceeding to fix the other meta modules... Particularly we need to check out the implications with multipolys and if they have holes.

### DIFF
--- a/packages/turf-meta/index.js
+++ b/packages/turf-meta/index.js
@@ -98,7 +98,7 @@ export function coordEach(geojson, callback, excludeWrapCoord) {
                         callback(coords[j][k], coordIndex, featureIndex, featureSubIndex);
                         coordIndex++;
                     }
-                    if (geomType === 'MultiLineString') featureSubIndex++;
+                    featureSubIndex++;
                 }
                 if (geomType === 'Polygon') featureSubIndex++;
                 break;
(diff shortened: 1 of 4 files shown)


packages/turf-meta/index.js
```diff
@@ -98,7 +98,7 @@ export function coordEach(geojson, callback, excludeWrapCoord) {
                         callback(coords[j][k], coordIndex, featureIndex, featureSubIndex);
                         coordIndex++;
                     }
-                    if (geomType === 'MultiLineString') featureSubIndex++;
```
DenisCarriere: Strange line 🤔 Don't know why this was here in the first place...

There's a bunch more like that...

DenisCarriere: Actually this is there for a reason, `featureSubIndex` is used to find the MultiPart Index.

If you wanted to know which Ring from the Polygon we would need 3 tiers.

- FeatureIndex
- FeatureSubIndex
- GeometryIndex (currently not implemented)

DenisCarriere:
Without the 3rd tier of index it would be impossible to find the "hole" in a MultiPolygon.

PR #1097, opened 2017-11-17
Add new module @turf/center-mean

DenisCarriere:
## Add new module `@turf/center-mean`

Thanks to @muziejus's initial effort for implementing mean center using weights.

### JSDocs

```js
/**
 * Takes a {@link Feature} or {@link FeatureCollection} and returns the mean center. Can be weighted.
 *
 * @name centerMean
 * @param {GeoJSON} geojson GeoJSON to be centered
 * @param {Object} [options.properties={}] an Object that is used as the {@link Feature}'s properties
 * @param {number} [options.weight] the property name used to weight the center
 * @returns {Feature<Point>} a Point feature at the mean center point of all input features
 * @example
 * var features = turf.featureCollection([
 *   turf.point([-97.522259, 35.4691], {weight: 10}),
 *   turf.point([-97.502754, 35.463455], {weight: 3}),
 *   turf.point([-97.508269, 35.463245], {weight: 5})
 * ]);
 *
 * var options = {weight: "weight"}
 * var mean = turf.centerMean(features, options);
```

## To-Do

- [ ] Support negative weights
- [ ] Handle combined negative + positive weights when total equals zero (-10, -10, +10, +10 => 0) which causes an error.

> @muziejus Can you help with the To-Do's... having a hard time with those negative weights.

## Examples

This current implementation works on all geometry types 🎉 

## Points with Weights

image

## Polygon with Weights

image

## Polygon without Weights

image


muziejus:
@DenisCarriere Thanks! This is great. I have a bunch of things to mention, but I'll start with why I created a new module instead of refactoring `@turf/center`: namely that `@turf/center` gives us the center of the extent of a collection, but the center of the extent isn't the mean center, and the mean center is what I (at least) am more interested in. To wit:

```
const a = [[0, 0], [9.5, 0], [10, 0]];
const center = turf.center(turf.featureCollection(a.map(function(coord){ return turf.point(coord); })));
turf.getCoord(center);
// => [5, 0]
```
This strikes me as confusing, because I expect a "mean center" (`[6.5, 0]`) to describe this distribution, not the midpoint of the extent. I want the average value.

So I wonder if it would not make sense to have different center modules or at least some kind of switch that lets the user choose between extent center, mean center, and, hopefully someday, median center.

DenisCarriere:
Ok I get it... have you tried `center-of-mass`? Does that do the same effect as your `mean-center`?

If we do chose to go towards the `mean-center` dedicated module, I would rename it to `center-mean` that way we have all the "center" modules at the same place.

- `center-of-mass`
- `center-mean`
- `center-median` (in the future)

I'll close this PR and re-use for the renaming of the new module. 👍

muziejus:
When I realized `center` wasn't doing what I wanted, I looked at `center-of-mass`, but that's also not what I need. For starters, it's measuring filled space in a polygon, whereas mean center is among the means by which we can measure the tendency of a distribution of observations in 2d space. 

So what's the next step?

DenisCarriere:
@muziejus I re-created the module `@turf/center-mean` using  what I pulled from `center` and I reverted the changes from the previous state of `@turf/center` (should be back to normal now).

## `@turf/center`

image


## `@turf/center-mean`

> Blue dot is where `@turf/center` is positioned

image

@muziejus Test that out and let me know what you come up with up, after that I'll merge it down and publish it soon.

DenisCarriere:
Also works with the `weight` options, the black dot has `15` as weight.

image

muziejus:
@DenisCarriere I'm not sure what to do w/ the weights. The [_CrimeStat_ manual](https://www.nij.gov/topics/technology/maps/documents/crimestat-files/CrimeStat%20IV%20Chapter%204.pdf) explains:

> The weights have to be a positive number greater than or equal to 1.

But no justification is given. It seems like throwing errors in case of the sum of weights' being 0 should be sufficient, no? I don't quite know what a "negative weight" might be, but that shouldn't preclude their use… In short, I'm also a little stumped.

muziejus:
@DenisCarriere Also, I don't know how to help fix the PR now that you're running it. The test was failing for me because `@turf/meta` was defined both as a dependency and as a dev dependency.

DenisCarriere:
> I don't quite know what a "negative weight" might be, but that shouldn't preclude their use… In short, I'm also a little stumped.

We should definitely try to include them, however when I attempted to add them it gave me some strange results.. maybe I'll have a stab at it again, could you provide an example in the `test/in` folder that would look like viable dataset with negative weights.

>  I don't know how to help fix the PR now that you're running it. 

@muziejus I've added you as a contributor with permissions to create/edit branches, for any major changes please submit a branch (like you've done in the past), for any minor edits you can push directly to master if you are 100% confident with the change.

muziejus:
So this is interesting:

screen shot 2017-11-20 at 00 01 45

ArcMap seems to just throw away the first point, as that mean center is what you get from the other three by themselves. Wikipedia asserts that [weights cannot be negative](https://en.wikipedia.org/wiki/Weighted_arithmetic_mean#Mathematical_definition), in line w/ the _CrimeStat_ manual.

Perhaps weights that are 0 or lower mean that that point should be dropped entirely?

muziejus:
I've changed it to:

```
        if (weightValue > 0) {
            coordEach(geom, function (coord) {
                sumXs += coord[0] * weightValue;
                sumYs += coord[1] * weightValue;
                sumNs += weightValue;
            });
        }
```
But I don't know how to push my change. see https://github.com/muziejus/turf/commit/1b9e312d12fc7e67b63b16430df47c9c1800f9fd

muziejus:
I guess logically this means the test for `sumNs === 0` can also be binned.

DenisCarriere:
@muziejus Heh... I guess that makes sense... 🤔 Dropping weights with less than 1.

Would be interesting to know what other tools such as QGIS would spit out as a result with your test example.

This PR is now using a branch from Turf, you'll need to clone Turf again using the `center-weight` branch.

```bash
$ git clone git@github.com:Turfjs/turf.git
$ cd turf
$ git checkout center-weight
```
For these two changes I'll just add them directly from the GitHub website.

DenisCarriere:
@muziejus On your next PR, try to create a branch directly from TurfJS. That way it's easier for both of us to work on the same branch.

👍 Very cool stuff about the CrimeStat, I'd be interested in attempting to replicate a few of the outputs from these diagrams such as this one:

image

muziejus:
The code for the standard deviational ellipse is already written and sitting on my computer. I'll create a proper PR for it soon.

muziejus:
> This PR is now using a branch from Turf, you'll need to clone Turf again using the center-weight branch.

Ah, I see. Now that I'm a contributor, I can write straight to the upstream repo and don't have to go through my fork anymore.

DenisCarriere:
👍 @muziejus Awesome stuff, once you push that we could publish an official minor release of Turf v5.1, that way all of these new modules will be on the official `@turf/turf` repo.

Looking forward to it!

I'll also have a last glance at the new modules we merged before we publish it (seems good to me now)

muziejus:
Incidentally, qGIS just rolls with the negative weight:
screen shot 2017-11-20 at 04 45 52

so that's [20/-7, 20/-7] (or: (0 + 10 + 0 + 10) / (1 + 1 + 1 -10))

muziejus:
@DenisCarriere Super 👍 . I've already written two more (three total) modules while waiting for there to be a mean-center module that gets into turf.

I also want to look at your tests for this module a bit more carefully to learn from them, because they are far more complex than I expected them to be.

DenisCarriere:
> I also want to look at your tests for this module a bit more carefully to learn from them, because they are far more complex than I expected them to be.

Well initially the tests don't have to be completely complex, however the initial tests need to output our expected results, so for the negative weight example we need to decide on the ArcMap way or the QGIS way... once we decide on whichever approach we would create a test case which would fail or pass based on the approach we decide take.

DenisCarriere:
>  I've already written two more (three total) modules while waiting for there to be a mean-center module that gets into turf.

OK I'll publish the next release sometime this week, I want to add a few extra modules to `@turf/helpers` which will be `points()`, `lineStrings()`, `polygons()`..etc... the plural of the singular helper methods.

PR #1105, opened 2017-11-20
bbox: Reflect possible values for bbox function in types and docs

woutervh-:
bbox only allowed Feature and FeatureCollection objects in the types and docs, but the implementation can actually handle any GeoJSON. Updated the types and docs to reflect this.

Please fill in this template.


woutervh-:
Just an FYI: doing `npm run docs` changed more than what I changed, seems like current master revision is behind with the docs tool.

DenisCarriere:
👍 Thanks @woutervh- for improving the docs!

Agreed that this module does support all GeoJSON types.

> doing `npm run docs` changed more than what I changed

Correct... 😢 I was thinking of changing the `npm run docs` if ran inside a local module it would only populate the README of that module, if `npm run docs` is ran at the root level it would run the update on all modules (not implemented yet).

stebogit:
> `npm run docs` if ran inside a local module

I like this 👍 

Couldn't this be ran automatically at root level at each push by Travis so we don't ever forget to update the READMEs?

DenisCarriere:
> Couldn't this be ran automatically at root

We could... I've done some hooks like this in the past, it can get a bit messy but we can have a look at this.

PR #1115, opened 2017-11-22
Use rhumb lines to calculate @turf/ellipse axes, except when units are "degrees"

muziejus:
There was a mistake in `@turf/ellipse`:

screen shot 2017-11-21 at 22 23 46
screen shot 2017-11-21 at 22 24 10

Red is an ellipse with 50km major and minor semi axes. Blue is an ellipse w/ a 50km semimajor axis and 20km semiminor axis. Green is the blue ellipse, rotated 90 degrees.

As you can see, `@turf/transform-rotate` does not compensate for distortion at high latitudes. 
As such, I have disabled and removed `@turf/ellipse`'s dependency on `@turf/transform-rotate`. Now it uses the `angle` property to rotate each point as it is being drawn. I've added a test to confirm this and the drawings come out like this:

screen shot 2017-11-21 at 22 25 17
screen shot 2017-11-21 at 22 24 50


DenisCarriere:
> As you can see, `@turf/transform-rotate` does not compensate for distortion at high latitudes.

@stebogit Would that be something we should fix or ignore in `transform-rotate`? 🤔 ... I don't know the answer.

👍 @muziejus Great fix and explanation!

muziejus:
cheers. I was trying to figure out why my ellipses were so different looking from the ones drawn by ArcMap and qGIS for the SDE, even though the centers, axes, and angles were all equivalent (for ArcMap--slightly off for qGIS). They still don't look quite right, so I think there's a little something going on in my conversions from those numbers into something `@turf/ellipse` can handle.

muziejus:
> They still don't look quite right, so I think there's a little something going on in my conversions from those numbers into something @turf/ellipse can handle.

Solved it 😂

DenisCarriere:
@muziejus A did a few modifications to the `test.js`, noticed you were modifying by hand the `test/out` results, those should be generated by `npm t` whenever you define the environment variable `REGEN=true`. There's a few ways to define environment variables, check based on your operating system, the easiest way for a Mac or Linux is to set them using the following:

```
export REGEN=true
npm t
```

Then when you are done to prevent the test outputs from being generated.

```
export REGEN=''
```

DenisCarriere:
I've created some test input data `simple.json` with some basic params, however when I create a circle & an ellipse using +90 degrees it doesn't match to the extent of the Circle.. 🤔 

@muziejus Any thoughts? How did you create your circles that align with the ellipses?

### Params
**center**: [-73.9975, 40.730833]
**xSemiAxis:** 5
**ySemiAxis:** 1
**units**: kilometers


**Blue:** unmodified params
**Green:** +90 Degree angle
**Red:** Circle based on the maximum axis value

image

DenisCarriere:
@muziejus Added some more "geo-visual" aids to figure out the ellipses.

I used `@turf/rhumb-destination` to create a new point from the center towards a 0 & +90 degree angle using the maxAxis value.

It seems like the Circle (red) lines up perfectly where the 5 kilometer distance lands, seems like your ellipses are not actually 5 km `xSemiAxis`.

**Purple**: Destination point/lineString created from same param values used in Ellipse

image

DenisCarriere:
Even more problems in Northern latitudes, this one was using 50km & 10km as the axis.

image

muziejus:
Thanks for the tips on testing, @DenisCarriere. How do you make your images? I'm always stuck between using leaflet on the browser, which means I'm always using a not-quite-right version of turf.

DenisCarriere:
> Solved it 😂

@muziejus 👍 Sweet! You've now got tons of test data that will help you visualize your results easier.

I didn't touch the `index.js` file since your last commit.

DenisCarriere:
@muziejus Go to http://geojson.io/ and copy-paste the `test/out` data

I just do a screen capture with my computer. The entire screen will look like this:

image

muziejus:
>@muziejus Any thoughts? How did you create your circles that align with the ellipses?

In leaflet:
```javascript
let nyu = [-73.99250799849149,40.73046499853991];
L.geoJson(ellipse(nyu, 5, 2), {style: { weight: 7, color: "#00ff00", fillOpacity: 0}}).addTo(map);
L.geoJson(ellipse(nyu, 5, 2, {angle: 45}), {style: { weight: 7, color: "#0000ff", fillOpacity: 0}}).addTo(map);
L.geoJson(ellipse(nyu, 5, 2, {angle: 90}), {style: { weight: 7, color: "#000000", fillOpacity: 0}}).addTo(map);
L.geoJson(ellipse(nyu, 5, 5), {style: { weight: 7, color: "#FF0000", fillOpacity: 0}}).addTo(map);
L.geoJson(ellipse(nyu, 2, 5)).addTo(map);
var line = L.geoJson(turf.lineString([turf.getCoords(nyu), turf.getCoords(turf.rhumbDestination(nyu, 5, 0))])).addTo(map);
var line = L.geoJson(turf.lineString([turf.getCoords(nyu), turf.getCoords(turf.rhumbDestination(nyu, 5, 90))])).addTo(map);
```

gives me:

screen shot 2017-11-22 at 02 23 04

I'm basically positive the issue is converting degrees to length.

muziejus:
>@muziejus 👍 Sweet! You've now got tons of test data that will help you visualize your results easier.

Haha... sorry, I meant I solved the problem I was having with the standard deviational ellipse. That's the funny thing--the ellipse that turf draws, using the version of ellipse I have here, is exactly the same as the one drawn by qgis or arcmap--I export the shapefiles as geojson, and there it is. But that's a different issue...

muziejus:
This is 2 and 1 mile increments, using Chicago as a grid, where every block is 1/8mi. `@turf/ellipse` is surely misdrawing...

screen shot 2017-11-22 at 02 44 11

DenisCarriere:
Yea something is off... don't know what exactly.

But don't you find it strange that:

```js
L.geoJson(ellipse(nyu, 5, 5), {style: { weight: 7, color: "#FF0000", fillOpacity: 0}}).addTo(map);
```
Is not a Circle?... with both axis 5 kilometers it should be perfectly round, correct? (like the `@turf/circle` module)

DenisCarriere:
Can you also include in your leaflet example `@turf/circle`, since we know that circle creates an equidistance circle (it's been tested a lot!)

```
L.geoJson(circle(nyu, 5), {style: { weight: 7, color: "#FF0000", fillOpacity: 0}}).addTo(map);
```

muziejus:
@DenisCarriere I should never have doubted `@turf/transform-rotate`. I was miscalculating lengths by relying on `lengthToDegrees`.

muziejus:
@DenisCarriere Thanks for holding my hand for this and for explaining the `REGEN` feature on the tests. I had just skipped that part of the tests, since I didn't understand it (and, yes, I was editing the out files by hand). I assumed we externally generated the out file and then just matched it to the function at hand. But for this, I've done the tests properly.

stebogit:
> Switch distance conversions to rhumb lines.

👍  @muziejus, `@turf/transform-rotate` [does not distort](https://github.com/Turfjs/turf/blob/master/packages/turf-transform-rotate/test/out/polygon-resolute-bay.geojson) at any latitude cause it uses Rhumb functions.
When drawing geometric shapes it's always good using such functions to avoid distortion, although I just noticed [`@turf/circle` does not use them](https://github.com/Turfjs/turf/blob/master/packages/turf-circle/index.js#L42)... 🤔

DenisCarriere:
> Thanks for explaining the REGEN feature on the tests.

👍 You'll notice a lot of our tests are going to be "dynamic" which means all we need to do to add more tests is to include additional GeoJSON (.json) files to the `test/in` folder and repopulate the `test/out` by using `npm test`.

This approach makes it a LOT easier to add more tests without adding extra lines of code to `test.js`.

DenisCarriere:
> @stebogit `@turf/transform-rotate` does not distort at any latitude cause it uses Rhumb functions.

👍 That's good, that's what I thought so too... nice to have a 2nd yes on that one.

muziejus:
@stebogit @DenisCarriere I'm sorry for doubting `@turf/transform-rotate`… I'm so used to seeing distorted "circles" on maps, etc. The problem here was clear--I was always converting to degrees at some point in the process, which has use cases (more to come), but if the user is giving a linear distance  from the beginning (radian, km), the distance should remain the fundamental mode of measurement.  

Now that distance is respected. 😞  I am chastened.

muziejus:
I've added a separate case for degrees and accompanying tests, which all pass. The module I'm working on now is unit agnostic, but it gives its response in the same units as the [x, y] coordinates are given in. Because in geoJSON that's degrees, there needs to be a way to draw ellipses w/ degrees that is also true to that mode of measurement. The excuse is that the ellipse is drawn because of what it contains and doesn't contain; its actual "size" (as in, the values of axes or area) isn't important. (What would _n_ degrees squared mean anyway?)

DenisCarriere:
@muziejus Tweaked the tests a bit more 👍 Ready to be merged, a few more tweaks to the Ellipse, but for now it's "good" for an initial publish.

image

PR #1122, opened 2017-11-25
New module: Center median

muziejus:
This module algorithimically calculates the median center of a set of points. This is less sensitive to outliers. It can take a weight, like `@turf/center-mean`, and it also takes a `tolerance`, which is the minimum difference between candidate median centers that the algorithm will tolerate. The weighted tests are weird, and the Ukraine problem in Kuhn and Kuenne has at least one mistake in the data input (the latitude--at least--for Xarxiv is wrong). The solution to the Steiner problem is also not quite correct because it looks like Kuhn and Kuenne truncate their candidate's coordinates with every iteration. But the green marker ends up in the right place, more or less, all the same. Below, red is the mean center and green is the median center.

screen shot 2017-11-25 at 12 50 11
Square coordinates

screen shot 2017-11-25 at 12 49 43
Steiner's problem

screen shot 2017-11-25 at 13 03 53
Steiner's problem w/ v. low tolerance

screen shot 2017-11-25 at 12 50 33
Ukraine problem.


packages/turf-center-median/index.js
```diff
@@ -0,0 +1,71 @@
...
+ * Numerical Solution of the Generalized Weber Problem in Spatial
+ * Economics,” _Journal of Regional Science_ 4, no. 2 (1962): 21–33,
+ * doi:{@link https://doi.org/10.1111/j.1467-9787.1962.tb00902.x}.
+ *
+ * James E. Burt, Gerald M. Barber, and David L. Rigby, _Elementary
+ * Statistics for Geographers_, 3rd ed., New York: The Guilford
+ * Press, 2009, 150–151.
+ *
+ * @name centerMedian
+ * @param {FeatureCollection<Point>} features GeoJSON Feature or Geometry
```
DenisCarriere: Might want to call the parameter `points` instead of `features`, that way it's explicitly indicating that you need to provide Points.

Also I would replace the `GeoJSON Feature or Geometry` as the description

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {Object} [options={}] Optional parameters
+ * @param {string} [options.weight] the property name used to weight the center
+ * @param {number} [options.tolerance=0.001] the difference in distance between candidate medians at which point the algorighim stops iterating.
+ * @returns {Feature<Point>} The median center of the collection
+ * @example
+ * var points = turf.featureCollection([turf.point([0, 0], turf.point([1, 0]), turf.point([0, 1]), turf.point([5, 8])]);
```
DenisCarriere: We've got a FANCY ⚡️ new module called `points()` which allows you to do the following:

```js
cons points = turf.points([
  [0, 0],
  [1, 0],
  [0, 1],
  [1, 1]
]);

File: packages/turf-center-median/index.js [hunk 3]
```diff
@@ -0,0 +1,71 @@
...
+ * var medianCenter = turf.centerMedian(points);
+ *
+ * //addToMap
+ * var addToMap = [points, medianCenter]
+ */
+function centerMedian(features, options) {
```
DenisCarriere: Replace `feature` => `points`

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,71 @@
...
+    // Calculate mean center & number of features:
+    var meanCenter = centerMean(features, {weight: weightTerm});
+    var medianCandidates = [];
+
+    function findMedian(candidateMedian, prevCandidate) {
+        var candidateXsum = 0;
+        var candidateYsum = 0;
+        var kSum = 0;
+        featureEach(features, function (feature) {
+            var weight = feature.properties[weightTerm] || 1;
```
DenisCarriere: ⚠️ If the weight is 0, this would default to `1` (must handled `0`, `undefined` & `null` separately if zero is allowed)

You can do the following:

```js
var weightValue = feature.properties[weightTerm];
var weight = (weightValue === undefined || weightValue === null) ? 1 : weightValue;
```

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,71 @@
...
+    var tolerance = options.tolerance || 0.001;
+
+    // Validation:
+    if (!isNumber(tolerance)) throw new Error('tolerance must be a number');
+
+    // Calculate mean center & number of features:
+    var meanCenter = centerMean(features, {weight: weightTerm});
+    var medianCandidates = [];
+
+    function findMedian(candidateMedian, prevCandidate) {
```
DenisCarriere: I would block out this `findMedian` method outside of the `centerMedian` function scope, that way it's clear what the inputs/outputs are.

packages/turf-center-median/test.js
```diff
@@ -0,0 +1,37 @@
...
+    glob.sync(path.join(__dirname, 'test', 'in', '*.json')).forEach(filepath => {
+        // Define params
+        const {name} = path.parse(filepath);
+        const geojson = load.sync(filepath);
+        const options = {
+          weight: geojson.properties.weight,
+          tolerance: geojson.properties.tolerance
+        };
+        let meanCenter = centerMean(geojson, options);
+        meanCenter.properties['marker-color'] = '#a00';
```
DenisCarriere: Try to use a custom helper function `colorize()` which removes any "noise" from this test case, it makes it easier to read this code block.

muziejus:
@DenisCarriere Give me a second here on the `features` / `points` thing. It's some functionality I initially included then took out but am now putting back in.

stebogit:
@muziejus may I suggest to find a more meaningful example for both `@turf/center-mean` and `@turf/center-median`?
I'd create a good input example, probably not trivial, to use in both docs that would highlight the difference between the two modules (not really clear to me) when compared, and maybe using it also in `@turf/center` to make it even clearer in a quick output/map comparison.
Although I'm pretty sure we are dealing with quite specific functionalities, this might make easier for non-geographers to understand what they exactly do and maybe allow them to take advantage of these new modules.

muziejus:
@stepogit No problem.

muziejus:
All, weighting isn't working properly. I have to think about this some more.

muziejus:
@stebogit I've updated the README to discuss the four different center functions in turf. If we all (@DenisCarriere?) settle on a kind of boilerplate, I think we can include it in all four of the functions. IMO, this image does a good job of showing the four different functions.

blue: `@turf/center`
gray: `@turf/center-of-mass`
red: `@turf/center-mean` (unweighted)
green: `@turf/center-median` (unweighted)
screen shot 2017-11-26 at 00 57 21

DenisCarriere:
@muziejus Try to use `@turf/truncate` as much as possible for the test results, at the moment TurfJS only uses the latest NodeJS release to test, however many of us will be using NodeJS 6/8/9 and each one of these releases have a slight precision difference which makes the tests only pass on a specific Node release (Node 9 in this case).

image

DenisCarriere:
> I've updated the README to discuss the four different center functions in turf.

👍 I'm good with that for now until we find a better documentation solution.

packages/turf-center-median/test.js
```diff
@@ -0,0 +1,46 @@
...
+import centerOfMass from '@turf/center-of-mass';
+import centerMean from '@turf/center-mean';
+import centerMedian from '.';
+
+test('turf-center-median', t => {
+    glob.sync(path.join(__dirname, 'test', 'in', '*.json')).forEach(filepath => {
+        // Define params
+        const {name} = path.parse(filepath);
+        const geojson = load.sync(filepath);
+        const properties = geojson.properties || {}
```
DenisCarriere: The great thing now with the Optional parameters is you can simply just declare `options` as the GeoJSON Properties instead of declaring each parameter.

## Valid Options

```js
const options = geojson.properties;
```

packages/turf-center-median/test.js
```diff
@@ -0,0 +1,46 @@
...
+        const options = {
+          weight: properties.weight,
+          tolerance: properties.tolerance
+        };
+        const meanCenter = colorize(centerMean(geojson, options), '#a00');
+        const medianCenter = colorize(centerMedian(geojson, options), '#0a0');
```
DenisCarriere: Needed to truncate `medianCenter`'s candidates by doing the following:

```js
// Truncate median properties
medianCenter.properties.medianCandidates.forEach((candidate, index) => {
    medianCenter.properties.medianCandidates[index] = [round(candidate[0], 6), round(candidate[1], 6)];
});
```

packages/turf-center-median/index.d.ts
```diff
@@ -0,0 +1,22 @@
+import { Feature, Point, Coord, FeatureCollection } from '@turf/helpers'
+ 
+/**
+ * http://turfjs.org/docs/#centermedian
+ */
+
+export interface medianProps {
+    medianCandidates: Array<Coord>,
```
DenisCarriere: 😛 Did it again, `Coord` is used to define all 3 Point Types.

`Position[]` is the one you are looking for, the following are all equivalent:
- `[number, number][]`
- `Array<[number, number]>`
- `Array<Array<number>>`

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,122 @@
...
+ * dataset, not the geographical. It can also be weighted, meaning certain
+ * points are more important than others.
+ *
+ * `@turf/center-median` takes the mean center and tries to find, iteratively,
+ * a new point that requires the least amount of travel from all the points in
+ * the dataset. It is not as sensitive to outliers as `@turf/center`, but it is
+ * attracted to clustered data. It, too, can be weighted.
+ *
+ * @name centerMedian
+ * @param {FeatureCollection<Any>} features GeoJSON Feature or Geometry
```
DenisCarriere: `FeatureCollection<Any>` means Polygon & LineString would also be accepted.. Do we just want to stick with Points for this one, the Typescript definition says otherwise so maybe this is a typo.

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,122 @@
...
+ * @param {Object} [options={}] Optional parameters
+ * @param {string} [options.weight] the property name used to weight the center
+ * @param {number} [options.tolerance=0.001] the difference in distance between candidate medians at which point the algorighim stops iterating.
+ * @param {number} [options.counter=10] how many attempts to find the median, should the tolerance be insufficient.
+ * @returns {Feature<Point>} The median center of the collection
+ * @example
+ * var points = turf.featureCollection([turf.points[[0, 0], [1, 0], [0, 1], [5, 8]]);
```
DenisCarriere: The output of `points()` is already a FeatureCollection, you don't need to wrap it with a FeatureCollection.

```js
var points = turf.points([[0, 0], [1, 0], [0, 1], [5, 8]]);
```

Your JSDocs examples are currently not being tested since they are not in the TurfJS Core library, this would of raised an error (not your fault 👍 )

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,124 @@
...
+        medianCandidates: []
+    };
+    return findMedian(meanCenter, [0, 0], centroids, counter);
+}
+
+/**
+ * Recursive function to find new candidate medians.
+ *
+ * @private
+ * @param {Feature<Point>} candidateMedian current candidate median
```
DenisCarriere: I changed these params to `Feature<Point>` however, one param you are giving a `Feature<Point>` and the other (previousCandidate) you are giving an `Array<number>`...

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,124 @@
...
+ * @param {number} counter how many attempts to try before quitting.
+ * @returns {Feature<Point>} the median center of the dataset.
+ */
+function findMedian(candidateMedian, previousCandidate, centroids, counter) {
+    var tolerance = centroids.properties.tolerance || 0.001;
+    var candidateXsum = 0;
+    var candidateYsum = 0;
+    var kSum = 0;
+    var centroidCount = 0;
+    featureEach(centroids, function (theCentroid) {
```
DenisCarriere: Instead of calling it `theCentroid` you can simply use a singular variable name like `centroid` if the input is plural (`centroids`)

packages/turf-center-median/index.js
```diff
@@ -0,0 +1,124 @@
...
+        var weightValue = theCentroid.properties.weight;
+        var weight = (weightValue === undefined || weightValue === null) ? 1 : weightValue;
+        weight = Number(weight);
+        if (!isNumber(weight)) throw new Error('weight value must be a number');
+        if (weight > 0) {
+            centroidCount += 1;
+            var distanceFromCandidate = weight * distance(theCentroid, candidateMedian);
+            if (distanceFromCandidate === 0) distanceFromCandidate = 1;
+            var k = weight / distanceFromCandidate;
+            candidateXsum += getCoord(theCentroid)[0] * k;
```
DenisCarriere: @muziejus Using `getCoord` does "look" good, however it does do some unnecessary processing.

This method is mostly used to clean the input data, you don't need to use `getCoord` if you are handling the output of a TurfJS method since the output will always be the same type.

Since you are iterating over a FeatureCollection of Point using `featureEach`, then `theCentroid` will ALWAYS be a `Feature<any>` which means if we access `theCentroid.geometry.coordinates` it will always be present without throwing any errors.

Everytime `getCoord` is fired, the following needs to be executed (which slows down the overall performance).

https://github.com/Turfjs/turf/blob/master/packages/turf-invariant/index.js#L15-L26

```js
export function getCoord(obj) {
    if (!obj) throw new Error('obj is required');

    var coordinates = getCoords(obj);

    // getCoord() must contain at least two numbers (Point)
    if (coordinates.length > 1 && isNumber(coordinates[0]) && isNumber(coordinates[1])) {
        return coordinates;
    } else {
        throw new Error('Coordinate is not a valid Point');
    }
}
```

packages/turf-center-median/test.js
```diff
@@ -0,0 +1,49 @@
...
+        const medianCenter = centerMedian(geojson, options);
+        const extentCenter = center(geojson);
+        const massCenter = centerOfMass(geojson);
+
+        // Truncate median properties
+        medianCenter.properties.medianCandidates.forEach((candidate, index) => {
+            medianCenter.properties.medianCandidates[index] = [round(candidate[0], 6), round(candidate[1], 6)];
+        });
+        const results = featureCollection([
+            ...geojson.features,
```
DenisCarriere: **Note:** When trying to include a FeatureCollection inside another FeatureCollection, you can use the ES6 Deconstructing syntax `...` to explode (easiest syntax to solve this).

packages/turf-center-median/index.js
muziejus: No, it works for any. This is intentional 😂 . I think you see this by the end of the comments, though?

packages/turf-center-median/index.js
muziejus: OK thanks.

packages/turf-center-median/index.d.ts
muziejus: Rats. I thought for sure I'd changed that. I'm all about that `Position` now.

muziejus:
@DenisCarriere Thanks for the changes! I wondered why you were always using adding `@turf/truncate`; I didn't think that node would be so picky w/ its precision.

packages/turf-center-median/index.js
DenisCarriere: 👍 we should then apply the same description to the Typescript definition since that is the only file that will actually enforce the Input Types, the JSDocs is simply for the documentation and not used for enforcing types.

DenisCarriere: I'll change it back to `any`

packages/turf-center-median/index.d.ts
DenisCarriere: Haha nice! 🤣 

FYI: `Position` is defined here in the GeoJSON specification:

https://tools.ietf.org/html/rfc7946#section-3.1.1

Also defined the same way in the GeoJSON Typescript definition:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/geojson/index.d.ts#L22

muziejus: Nice. That's super useful. I've started learning some of the basics of TypeScript, so I think this kind of typing will come more naturally to me once I have a stronger sense of what a type is and where to find their definitions.

PR #1127, opened 2017-11-26
Take care of zero handling for weight in @turf/center-mean

muziejus:
I've tried to standardize how zeroes are handled in functions w/ weighting.

### DIFF
--- a/packages/turf-center-mean/index.js
+++ b/packages/turf-center-mean/index.js
@@ -30,7 +30,7 @@ function centerMean(geojson, options) {
     options = options || {};
     if (!isObject(options)) throw new Error('options is invalid');
     var properties = options.properties;
-    var weight = options.weight;
+    var weightTerm = options.weight;
 
     // Input validation
     if (!geojson) throw new Error('geojson is required');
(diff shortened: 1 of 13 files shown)


packages/turf-center-mean/index.js
```diff
@@ -39,14 +39,15 @@ function centerMean(geojson, options) {
     var sumYs = 0;
     var sumNs = 0;
     geomEach(geojson, function (geom, featureIndex, properties) {
-        var weightValue = properties[weight] || 1;
-        if (!isNumber(weightValue)) throw new Error('weight value must be a number for feature index ' + featureIndex);
-        weightValue = Number(weightValue);
-        if (weightValue > 0) {
+        var weightValue = properties[weightTerm];
+        var weight = (weightValue === undefined || weightValue === null) ? 1 : weightValue;
+        weight = Number(weight);
```
DenisCarriere: `weight = Number(weight)` should be after `isNumber` type checking since any invalid String given to Number() will convert it to `NaN` which will now be a number.

```js
weight = Number('foo');
//= weight => NaN
typeof weight // => 'number'
```

packages/turf-center-mean/index.js
```diff
@@ -39,14 +39,15 @@ function centerMean(geojson, options) {
     var sumYs = 0;
     var sumNs = 0;
     geomEach(geojson, function (geom, featureIndex, properties) {
-        var weightValue = properties[weight] || 1;
-        if (!isNumber(weightValue)) throw new Error('weight value must be a number for feature index ' + featureIndex);
-        weightValue = Number(weightValue);
-        if (weightValue > 0) {
+        var weightValue = properties[weightTerm];
```
DenisCarriere: You don't need to define both `weight` & `weightValue`, changing `weightValue` to Number won't mutate the original input, so you will be ok to do this:

```js
var weight = properties[weightTerm];
weight = (weight === undefined || weight === null) ? 1 : weight;
```
The only reason why you would need to have both `weight` & `weightValue` is if you are comparing both of those values down the road (which isn't the case).

DenisCarriere:
👍 Seems to work just fine 😄 

**Black Point:** 13 weight
**Grey Point:** 3 weight
**Grey Point without Star:** 0 or negative values

**Red Point:** Mean Center
**Blue Point:** Center

image

muziejus:
Super. the tests w/ the bad weights here and in center-median were working fine for me, but thanks for the corrections! 🎉

PR #1131, opened 2017-11-27
Add Babel to Rollup publishing

DenisCarriere:
## Add Babel to Rollup publishing

First stab at trying to improve the published TurfJS bundle.

The ideal goal would be to "Tree Shake" JSTS's excess code and reduce the minified bundle of TurfJS.

## ✅ Success

- Able to reduce total bundle size by ~30%~ (~200 KB~) 40% (317 KB)

**Bundle Before**

```
1.86 MB    turf.js
765.94K   turf.min.js
```

**Bundle Now**

```
1.2M    turf.js
448K    turf.min.js
```

## ❌ Current Issues

- When applying `babel()` all the Tree Shaking is disabled and filesize is increased the current size.
- Babel + Rollup might have an issue when using [`export *` syntax from `turf-jsts`](https://github.com/DenisCarriere/turf-jsts/blob/master/index.js)


DenisCarriere:
@stebogit If we get this right we can now introduce ES6 code to any dependencies & internally. That means Classes, Set(), Map() would be fine to include. As long as the ES6 syntax is supported with NodeJS.

More and more dependencies will slowly start having ES6 only code, I'd personally rather use real ES6 Classes than fake Functions that act like Classes.

DenisCarriere:
## JSTS-free bundle filesize

```js
// JSTS Modules
export {default as difference} from '@turf/difference';
export {default as buffer} from '@turf/buffer';
export {default as union} from '@turf/union';
export {default as intersect} from '@turf/intersect';

// JSTS Sub-Modules
export {default as dissolve} from '@turf/dissolve';
export {default as hexGrid} from '@turf/hex-grid';
export {default as mask} from '@turf/mask';
export {default as squareGrid} from '@turf/square-grid';
export {default as triangleGrid} from '@turf/triangle-grid';
export {default as interpolate} from '@turf/interpolate';
```

When commenting out the JSTS dependent modules from the index and the file size significantly gets reduced to:

```
668K    turf.js
168K    turf.min.js
```

## With JSTS bundle size

The difference in filesize between both is about 372KB which means JSTS inflates the TurfJS repo by 3x.

```
1.4M    turf.js
540K    turf.min.js
```

Issue #1104, opened 2017-11-20
How to cite sources for various equations, etc?


muziejus:
The modules I'm working on now demand (in my opinion) some kind of reference to the scholarly literature from which they emerged. This is I think useful both as a justification for why a module uses the equations it does and also as a point of interest for GIS students who might, like me, be attracted to turf because of its flexibility, freeness, and so on, but who also want to know more of where these ideas came from.

This feels like it should go in every module's README, but I get the sense that those READMEs are supposed to be quick and to the point, without a short bibliography at the end, say.

The issue is also a bit more relevant now because @DenisCarriere (and I?) has/have decided that `@turf/center-mean` will ignore weights that are <= 0. This is in line with how Esri measures the mean center in ArcMap, but _unlike_ how qGIS does it. I don't know…

DenisCarriere:
@muziejus Great topic, most of the TurfJS descriptions are pretty straight forward for simplicity, however you can always add more documentation & citations if you decide to include it.

For example [`@turf/tin`](https://github.com/Turfjs/turf/blob/master/packages/turf-tin/index.js#L5-L14) has 7 lines of code dedicated in the description with a Wikipedia link for more reading.

Feel free to add as much documentation & citations as possible 👍

muziejus:
Super. I looked at the tags for JSDoc and it looks like I could do something like:

```
For more on the standard deviational ellipse, @see {@link http://www.jstor.org/stable/490885|Yuill 1971, “The Standard Deviational Ellipse; An Updated Tool for Spatial Description.”}
```

DenisCarriere:
👍 Sounds good to me!

rowanwins:
Another alternative here might be to use the `@see` [tag ](http://usejsdoc.org/tags-see.html)in jsdocs, then we could put those references somewhere outside of the main description, for instance, in a modal or something? 

The main benefit I can see in this would be that we can split out the really technical references which limited ppl are probably interested in, from the general description (which is probably what most ppl want). Just another way to reducing cognitive overload in the docs.

smallsaucepan:
Closing this issue with the intent to start using @see in JSDocs and update turf-www to render any new content appropriately.

PR #1135, opened 2017-11-27
Add **Bibliography** header to modules.

muziejus:
Pursuant to #1104, I’ve started adding bibliographies where appropriate. They should be, it seems, marked as such.

Note: this is breaking tests and lints because npm complains that it cannot import these two modules?

### DIFF
--- a/packages/turf-standard-deviational-ellipse/index.js
+++ b/packages/turf-standard-deviational-ellipse/index.js
@@ -12,7 +12,13 @@ import ellipse from '@turf/ellipse';
  * an ellipse that contains about one standard deviation’s worth (~ 70%) of the
  * data.
  *
- * This module was created in consultation with the following articles:
+ * This module mirrors the functionality of {@link
+ * http://desktop.arcgis.com/en/arcmap/10.3/tools/spatial-statistics-toolbox/directional-distribution.htm|Directional
+ * Distribution} in ArcGIS and the {@link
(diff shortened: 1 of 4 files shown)


DenisCarriere:
>  this is breaking tests and lints because npm complains that it cannot import these two modules?

What breaks exactly? Is it the `@link` tag? You could use Markdown instead to identify the doi links, might be a safer option:

```js
/**
  * [doi](https://doi.org/10.1111/j.1467-9787.1962.tb00902.x)
```

packages/turf-standard-deviational-ellipse/index.js
```diff
@@ -12,7 +12,13 @@ import ellipse from '@turf/ellipse';
  * an ellipse that contains about one standard deviation’s worth (~ 70%) of the
  * data.
  *
- * This module was created in consultation with the following articles:
+ * This module mirrors the functionality of {@link
```
DenisCarriere: 🤔 ... I wouldn't split this up, maybe that's the cause of breaking?

If you are to create a link, keep it on the same line of code, not too fussy on a maximum character of lines.

Also I think we should be using Markdown to identify the links instead of `{@link`.

Back story for `{@link`: It's currently used as shortcut links using DocumentationJS, instead of inserting the hardcoded URL all the time, we can simple create a `{@link Point}` and the URL will be added from the following paths: https://github.com/Turfjs/turf/blob/master/scripts/generate-readmes#L22-L49

For all other links you can use Markdown to create links `[label](url)`

PR #1138, opened 2017-11-29
Internalize @types/geojson

DenisCarriere:
# Internalize `@types/geojson`

Due to the latest breaking changes of `@types/geojson`, decided it would be best internalize the Typescript definition (still keeping the same syntax as `@types/geojson`).

Read more related to this topic on this PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21590

Thankfully we've done some heavy TypeScript refactoring over the past year and all of the TurfJS modules rely solely on `@turf/helpers` which makes it easy to manage all Typescript definitions if each of other Turf modules import the definitions via `@turf/helpers`.

🤞 There should be very little impact for this change, we will roll this out at the next minor release `v5.1`.


packages/turf-helpers/geojson.d.ts
```diff
@@ -0,0 +1,238 @@
...
+ * Developers should use "&" type in TypeScript or extend the interface to add these foreign members.
+ */
+export interface GeoJSONObject {
+    // Don't include foreign members directly into this type def.
+    // in order to preserve type safety.
+    // [key: string]: any;
+    /**
+     * Specifies the type of GeoJSON object.
+     */
+    type: string;
```
DenisCarriere: @zakjan To solve your issue https://github.com/DefinitelyTyped/DefinitelyTyped/pull/21590#issuecomment-348153458 I actually converted `GeoJSONObject` as `type: string`.

Maybe you should suggest a change to the new `@types/geojson` definition.

PR #1139, opened 2017-11-29
tests for issue 1132

Knight219:
Please fill in this template.

- [ ] Use a meaningful title for the pull request. Include the name of the package modified.


packages/turf-intersect/test/in/issue-1132-line.geojson
```diff
@@ -0,0 +1,241 @@
...
+  "type": "FeatureCollection",
+  "features": [
+    {
+      "type": "Feature",
+      "geometry": {
+        "type": "Polygon",
+        "coordinates": [
+          [
+            [
+              -8479843.24714559,
```
DenisCarriere: @Knight219 You're coordinates are not using Decimal Degrees.

Coordinates that are not using a WGS84 projection is not going to be supported by TurfJS or the GeoJSON specification.

Could you replace your test data using Decimal Degrees instead.

Knight219: Intresting:). We use turf 4.0.0 with this format and all works well.

DenisCarriere: Humm... 🤔 No clue then, right now the output of your test data is returning a `Null` geometry.

https://travis-ci.org/Turfjs/turf/builds/308889447#L3074-L3076

Regardless, we won't be supporting no-decimal degree coordinates in TurfJS.

Knight219: I'll update our code and tests

Knight219: Looks that my data also in WGS84 with EPSG3857 but not EPSG4236

DenisCarriere: @Knight219 🤔 ? Does it also fail using WGS84? If so then send a commit with the change of test data.

If we don't come up with a solution we can skip it for now but at least the test case will be present for the future.

DenisCarriere:
Not fixed, but the test case is now manually skipped in the module.

Issue #1144, opened 2017-11-30
@turf/simplify - freeze window


dabajabaza:
Just notice that **turf/simplify** freeze browser window in Google Chrome and Firefox (but not in IE11) if parameter **options** is undefined.

Tested in 

1. Google Chrome Version 62.0.3202.94 (Official Build) (64-bit) 
2. Firefox Quantum 57.0
3. Internet Explorer 11

In [this example](https://codepen.io/ckpunmep/pen/LOBwVj?editors=1010) I comment out a freeze causing line.

DenisCarriere:
Thanks @dabajabaza for reporting this issue & creating a web map example, this behavior has now stopped in `v5.1.x`.

PR #1145, opened 2017-12-01
Prevent Window Freeze when using @turf/simplify

DenisCarriere:
# Prevent Window Freeze when using `@turf/simplify`

CC: @dabajabaza
Ref: Issue #1144 

- [x] I'm not too sure if this will solve it, but maybe internalizing `simplify-js` might help... 🤷‍♂️ 
- [x] I also defined some default values in the `options`, however I don't think that was the cause of the Window Freeze (the freeze was also occurring with my browser as well, strange...) 
- [x] Added the Polygon that "freezes the window" as test data.


DenisCarriere:
Might be an issue with publishing, lets revisit this issue at the next minor release.

Issue #1140, opened 2017-11-29
Add DE-9IM graphics to Web Documentation


DenisCarriere:
## Add DE-9IM graphics to Web Documentation

I was attempting to explain what `@turf/boolean-disjoint` was to someone and even I got confused on what that module was meant to do, I then decided to go to ESRI's docs and look at the **disjoint** graphic for more help.

We should try to include at least 1 info graphic to each module (optional) which helps better understand what the module does in a conceptual way.

http://edndoc.esri.com/arcsde/9.0/general_topics/understand_spatial_relations.htm

These infographics are very useful

**equal**

![image](https://user-images.githubusercontent.com/550895/33394746-2a97d3f4-d4f8-11e7-9ee0-870666c41527.png)

**disjoint**

![image](https://user-images.githubusercontent.com/550895/33394751-2f8e2f16-d4f8-11e7-8e12-ae33eac26b36.png)

rowanwins:
Yeah that would be nice @DenisCarriere , there may be some newer graphics somewhere on the ESRI site so might be worth a quick dig around.

stebogit:
👍 👍

DenisCarriere:
👍 We just need to find a consistent naming convention for these infographics.

Maybe we can include a collection of images to create a sort of Photo Gallery of maps/graphics which derive from that particular Module.

## Image Folder Structure

- `images/*.png`
- `images/*.jpg`

**Example:** `./packages/turf-center/images/infographic.jpg`

rowanwins:
Do we put in the images in the turf-www repo ?

DenisCarriere:
No we should keep them in the module repo, that way we can also include those images when we are using [`scripts/generate-readmes`](https://github.com/Turfjs/turf/blob/master/scripts/generate-readmes)

The `turf-www` repo already has `turf` built inside, so we can reach those images through the main Turf folder.

I've done the same thing for the [`documentation.yml` config](https://github.com/Turfjs/turf-www/pull/128/files#diff-66599dd3d735c957a4e336cc8211f6edR19).

rowanwins:
yeah ok, no worries. Well if that's the case then I reckon your proposed above looks good.

Slightly newer Esri image here
http://desktop.arcgis.com/en/arcmap/10.3/manage-data/using-sql-with-gdbs/relational-functions-for-st-geometry.htm
Although not sure if we can steal them due to licensing...

The postgis images are a bit more complicated because they are split into lots of portions.
[...]

PR #1152, opened 2017-12-05
Add images to generate-readmes

DenisCarriere:
## Add images to generate-readmes

This adds `images/*` to a `Diagrams` section.

Ref: https://github.com/Turfjs/turf/issues/1140

🤔 Maybe we should call the image folder `diagrams/*` instead of using `images/*`.


DenisCarriere:
@muziejus @stebogit I've renamed the images folder to `diagrams/`.

Using `npm run docs` should now add those images to the `README.md`.

@rowanwins Once we have a few more modules with diagrams we can try to figure out an easy way to implement this in the website.

PR #1158, opened 2017-12-06
Release notes v5.1

DenisCarriere:
## Release notes v5.1

@rowanwins @muziejus Anything to add to the release notes before I publish the release?

### DIFF
--- a/releases/5.1.0.md
+++ b/releases/5.1.0.md
@@ -61,7 +61,7 @@ Takes a Point and calculates the ellipse polygon given two semi-axes expressed i
   - This uses nuxt to generate the pages as static html which should be better for search engines
   - The docs are now all on a single page
   - Built sizes are much better than the current
-  - The map moves to a module when it's description comes into view, I think this functions reasonably nicely.
+  - The map moves to a module when its description comes into view, I think this functions reasonably nicely.
 
 ## 🏅 New Features/Enhancements
 
@@ -76,6 +76,7 @@ Takes a Point and calculates the ellipse polygon given two semi-axes expressed i
 - Add input validation to [@turf/bbox-polygon][bbox-polygon] (https://github.com/Turfjs/turf/pull/1121)
 - Use rhumb lines to calculate [@turf/ellipse][ellipse] axes, except when units are "degrees" (https://github.com/Turfjs/turf/pull/1115)
 - [coordEach][meta] indexing bug (https://github.com/Turfjs/turf/issues/1092)
+- Add null handling for property cloning on @turf/clone. Fixes issue #1095 (https://github.com/Turfjs/turf/pull/1096)
 
 Twitter: <Tweet>
 CC: @Turfjs/ownership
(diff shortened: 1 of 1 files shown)


muziejus:
I was thinking about maybe we should rename `@turf/center-mean` to `@turf/center-weighted-mean` and farm out its (unweighted) mean-finding to `@turf/centroid`?

releases/5.1.0.md
```diff
@@ -0,0 +1,182 @@
...
+
+(PR https://github.com/Turfjs/turf/pull/1087 - Author @muziejus)
+
+## 📖 Documentation
+
+- TurfJS's website has a fresh new look! 🚀 http://turfjs.org (Author @rowanwins PR https://github.com/Turfjs/turf-www/pull/127)
+  - This uses nuxt to generate the pages as static html which should be better for search engines
+  - The docs are now all on a single page
+  - Built sizes are much better than the current
+  - The map moves to a module when it's description comes into view, I think this functions reasonably nicely.
```
muziejus: "when its description"

releases/5.1.0.md
```diff
@@ -0,0 +1,182 @@
...
+- Add extra geometry support to [@turf/points-within-polygon][points-within-polygon] (https://github.com/Turfjs/turf/pull/1117)
+
+## 🐛 Bug Fixes
+
+- Prevent Window Freeze when using [@turf/simplify][simplify] (https://github.com/Turfjs/turf/pull/1145)
+- Take care of zero handling for weight in [@turf/center-mean][center-mean] (https://github.com/Turfjs/turf/pull/1127)
+- Add input validation to [@turf/bbox-polygon][bbox-polygon] (https://github.com/Turfjs/turf/pull/1121)
+- Use rhumb lines to calculate [@turf/ellipse][ellipse] axes, except when units are "degrees" (https://github.com/Turfjs/turf/pull/1115)
+- [coordEach][meta] indexing bug (https://github.com/Turfjs/turf/issues/1092)
+
```
muziejus: there was the bug in `@turf/clone`, too.

muziejus:
I'm putting together a pr rn for center-weighted-mean

muziejus:
my tests are failing because I don't know how to tell turf there's a new name for the module.

releases/5.1.0.md
DenisCarriere: When you scroll down to the specific module, then the map appears, instead of loading every single map (that was the previous behavior)

DenisCarriere:
> @muziejus @turf/center-mean to @turf/center-weighted-mean

We can simply deprecate `@turf/centroid` in favor of `@turf/center-mean` since that would be the most "accurate" meaning for centroid.

Weighted or unweighted shouldn't be two modules, it should simply remain to be an option in `@turf/center-mean`.

Deprecating `@turf/centroid` would only display a message saying "next major release it will be removed", we've done this for a few modules already.

releases/5.1.0.md
muziejus: I just meant that the apostrophe isn't needed 😆

muziejus:
@DenisCarriere cheers. Deprecating `@turf/centroid` sounds like the better solution.

releases/5.1.0.md
DenisCarriere: Oh :) lol

Issue #1169, opened 2017-12-11
pointGrid fails with multi-polygon mask (5.1.6)


jpoynton:
Docs and validation for pointGrid state that can pass a multi-polygon as a mask, however multi-polygon fails as pointGrid relies on boolean-within to check if the point is within the mask. Boolean-within does not accept multi-polygon as valid input.

Simple recreation follow

```js
var multiPoly = turf.multiPolygon([
  [
    [
      [0, 50],
      [0, 10],
      [10, 10],
      [10, 0],
      [0, 0]
    ]
  ]
]);

var options = {
  units: 'miles',
  mask: multiPoly
}

var bbox = turf.bbox(multiPoly);
var pointGrid = turf.pointGrid(bbox, 50, options);
```

DenisCarriere:
Thanks @jpoynton for reporting this, this is definitely a bug, should be relatively easy to fix so this can be patched soon.

DenisCarriere:
@rowanwins This PR https://github.com/Turfjs/turf/pull/1171 should help do the trick.

Issue #785, opened 2017-06-11
Turf tangents fails in some cases again


dhivehi:
Sample data for test:
![screenshot-dhivehi mv-2017-06-11-15-14-10](https://user-images.githubusercontent.com/18715293/27010274-c9451d92-4eb9-11e7-9612-fe799db30139.png)

```javascript
var point = {"type":"Feature","geometry":{"type":"Point","coordinates":[73.5755476513307,0.4323399237340624]},"properties":null};

var poly  = {"type":"Feature","geometry":{"type":"Polygon","coordinates":[[[73.57502193836378,0.43239356638694915],[73.57526333717512,0.43281197906664204],[73.57561738876508,0.43281197906664204],[73.57560129551099,0.4325384015479159],[73.57543499855207,0.4324686661003909],[73.57551010040449,0.43223263842709514],[73.57579441456006,0.43239893065221224],[73.57609482196972,0.4324364805090113],[73.57643278030561,0.4323184666727826],[73.57594998268293,0.4322380026924719],[73.5754779138963,0.4318464113095217],[73.5751184978883,0.4320448891362787],[73.57516141323254,0.4322862810807635],[73.57500048069166,0.43227018828469],[73.57502193836378,0.43239356638694915]]]},"properties":null};

tangents = turf.polygonTangents(point,poly);
```

DenisCarriere:
@rowanwins any thoughts?

rowanwins:
Gday @dhivehi 

Thanks for the test case, I'll take a look into it and see what I can find.

DenisCarriere:
👍 @rowanwins's PR fixes this issue.

![image](https://user-images.githubusercontent.com/550895/27254873-e125e338-535f-11e7-9c10-d49394c1fe37.png)

dhivehi:
![image](https://user-images.githubusercontent.com/18715293/31824608-f6b05492-b5c8-11e7-9d0b-519033b9d5b0.png)

Again failed in this case also:

```javascript
var point = {"type":"Feature","geometry":{"type":"Point","coordinates":[73.2258944382542,0.6673910301706059]},"properties":null};
[...]

PR #1171, opened 2017-12-11
Add MultiPolygon support for feature2 to @turf/boolean-within

DenisCarriere:
## Add MultiPolygon support for `feature2` to `@turf/boolean-within`

Add support for:

- Point => MultiPolygon
- MultiPoint => MultiPolygon
- LineString => MultiPolygon
- Polygon => MultiPolygon

This would also fix https://github.com/Turfjs/turf/issues/1169 by allowing the `mask` to be a MultiPolygon geometry type.


DenisCarriere:
@rowanwins have a glance at these changes, they are pretty straightforward since `@turf/boolean-point-in-polygon` already supports Polygon & MultiPolygon

packages/turf-boolean-within/index.js
```diff
@@ -34,6 +34,7 @@ function booleanWithin(feature1, feature2) {
         case 'LineString':
             return booleanPointOnLine(geom1, geom2, {ignoreEndVertices: true});
         case 'Polygon':
+        case 'MultiPolygon':
             return booleanPointInPolygon(geom1, geom2, {ignoreBoundary: true});
```
rowanwins: Don't you love easy fixes :)

rowanwins:
Just found that this change is going to be helpful for my polygon-tangents bug #785 . Have the bug pretty much resolved, just need this to get published and it should be good to go!

packages/turf-boolean-within/index.js
DenisCarriere: Very easy fix 👍 Hence why I didn't tackle `MultiLineStrings` or anything more complicated than that, lol

DenisCarriere:
👍 Merging

rowanwins:
Turns out I didn't need the fix for `polygonTangents` after all but Im still glad you did it :)

Issue #1172, opened 2017-12-12
New module proposal @turf/angle


DenisCarriere:
## New module proposal `@turf/angle`

A  module to find a specific angle between 3 points

```js
/**
 * Finds the angle between 3 points.
 *
 * @param {Coord} startPoint Start Point Coordinates
 * @param {Coord} midPoint Mid Point Coordinates
 * @param {Coord} endPoint End Point Coordinates
 * @return {number} angle
 * @example
 * turf.angle([5, 5], [5, 6], [3, 4])
 * //=45
 */
export default function angle (startPoint, midPoint, endPoint) {
```

rowanwins:
ooo this would be handy, I've found myself doing things like this a bit with things like lineOffset.

stebogit:
Nice suggestion @DenisCarriere 👍 
We could also replace [these lines](https://github.com/Turfjs/turf/blob/master/packages/turf-point-to-line-distance/index.js#L82-L84) with this new module.

DenisCarriere:
👍 Perfect! I'll add this module, I've already got the implementation and the tests ready to go.

DenisCarriere:
Implemented

PR #1173, opened 2017-12-12
Implement new module @turf/angle 🚀

DenisCarriere:
## Implement new module `@turf/angle` 🚀 

Ref: https://github.com/Turfjs/turf/issues/1172

Implement `@turf/angle` module

```js
/**
 * Finds the angle between 3 points.
 *
 * @param {Coord} startPoint Start Point Coordinates
 * @param {Coord} midPoint Mid Point Coordinates
 * @param {Coord} endPoint End Point Coordinates
 * @returns {number} angle
 * @example
 * turf.angle([5, 5], [5, 6], [3, 4])
 * //=45
 */
export default function (startPoint, midPoint, endPoint) {
```


stebogit:
@DenisCarriere just to clarify, which angle will the module return?
untitled drawing 1

DenisCarriere:
@stebogit Good question.. lol! 😄 Also, it might be worth adding an "inverse" option.

DenisCarriere:
@stebogit This module always calculates the "Inner Angle", regardless if you swap Start/End points.

I've added an options called `options.exterior` to allow you to retrieve the outer angle (360 - angle).

turf-angle

PR #1174, opened 2017-12-12
Fix for concave polys where point lies inside concave edge

rowanwins:

This fixes issue #785 where a point lies inside the convex hull of a polygon and it is more southern than the first coordinate (or something crazy like that). I've had to digress from the [original algorithm here](http://geomalgorithms.com/a15-_tangents.html) to get this working so hopefully no further issues. I gather the orig algorithm was primarily designed for convex polys hence why we've had some issues.


packages/turf-nearest-point/main.es.js
```diff
@@ -0,0 +1,50 @@
+import clone from '@turf/clone';
```
DenisCarriere: @rowanwins `main.es.js` accidentally got added to your PR, I'm assuming the `.gitignore` wasn't updated when you created this branch.

packages/turf-polygon-tangents/main.es.js
```diff
@@ -0,0 +1,119 @@
+import { getCoords, getType } from '@turf/invariant';
```
DenisCarriere: Another `.es.js` file got added.

packages/turf-polygon-tangents/test.js
```diff
@@ -21,7 +21,6 @@ const fixtures = fs.readdirSync(directories.in).map(filename => {
 
 test('turf-polygon-tangents', t => {
     for (const {name, filename, geojson} of fixtures) {
-        if (name === 'issue#785') continue;
```
DenisCarriere: Woot! 🚀 🎉

packages/turf-nearest-point/main.es.js
rowanwins: Yeah you're probably right sorry.

PR #1175, opened 2017-12-12
Performance improvement for nearest-point

rowanwins:

Was looking at this module quickly while on my polygon-tangents fix and saw a bit of an inefficiency of potential unnecessary cloning. Also rejigged the tests which appeared to be a bit broken.

### DIFF
--- a/packages/turf-nearest-point/bench.js
+++ b/packages/turf-nearest-point/bench.js
@@ -1,14 +1,19 @@
 import fs from 'fs';
 import path from 'path';
 import Benchmark from 'benchmark';
+import { point } from '@turf/helpers';
 import nearestPoint from './';
 
-var pt = JSON.parse(fs.readFileSync(path.join(__dirname, 'test', 'pt.geojson')));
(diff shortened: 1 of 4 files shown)


rowanwins:
Also worth noting that I did try [knn-rbush](https://github.com/mourner/rbush-knn) however for our benchmark it was slower, however for other use cases that might be a good option for users, eg get me the 3 closest points

packages/turf-nearest-point/bench.js
```diff
@@ -1,14 +1,19 @@
...
+ *
+ * turf-nearest-point x 72,623 ops/sec ±9.23% (73 runs sampled)
+ */
 var suite = new Benchmark.Suite('turf-nearest-point');
 suite
-  .add('turf-nearest-point', () => nearestPoint(pt, pts))
-  .on('cycle', e => console.log(String(e.target)))
-  .on('complete', () => {})
-  .run();
+    .add('turf-nearest-point', () => nearestPoint(point(pts.properties.targetPoint), pts))
```
DenisCarriere: Every module that has point as input "should" also be able to input only a pair of coords (without the `point()`)

Issue #1180, opened 2017-12-17
id == 0 option ignored. 


Spown:
``turf.point([1,2], {}, {id: 0})`` doesn't seem to apply id.

![](https://drive.google.com/uc?id=1DdpiivfT67cIMssV7zJ7cfAa4Buq1z8z)

I deduce there is simple troothy/falsy check (ie ``!options.id``) in place before the id application.

stebogit:
Thanks @Spown, I created PR #1184 for this

stebogit:
Fixed with #1184

PR #1184, opened 2017-12-17
fixes issue #1180, helpers.point id validation

stebogit:
@DenisCarriere I fixed the issue, but I got this error and I'm not sure if it's me or what

screen shot 2017-12-17 at 10 32 04 am


DenisCarriere:
👍 Looks good, as for the `rollup-plugin-node-resolve` module missing, you might just need to update the branch and install the new `devDependencies` at the TurfJS root level.

I've added those recently: https://github.com/Turfjs/turf/blob/master/package.json#L71-L74

Issue #1182, opened 2017-12-17
breaking iterations


Spown:
It is kinda a common knowledge by now for many libraries that provide some iterating helpers that by returning false in the iteration callback you can exit the loop prematurely. ~~And it seems that is also the case for turf, at least for ``.featureEach()`` (dunno about the rest of the methods). But there is no mention of it in the docs. I think it still should be mentioned.~~

Thank you.

rowanwins:
Hi @Spown 

Hmmm can't say I've heard of that looping functionality but that certainly would be a neat addition, can you provide examples of which libraries do that?

I've tried doing that [here](https://runkit.com/rowanwins/5a370531998e5a00115c2d2b) with `featureEach` but I can't get it working so I'm interested in how it's working for you, are you able to show us an example?

Cheers

Spown:
Yeah, my mistake, turf's loops don't do that - I just was inspecting already filtered out results so it looked like there were no superfluous iterations after ``return false``.

Well, both underscore and [lodash do that](https://lodash.com/docs/4.17.4#forEach) and [jQuery](http://api.jquery.com/each/). Would be nice to have it here too, GeoObject collections can be quite large and every resource-saving measure would be appreciated. Consider it a feature request.

Thank you.

rowanwins:
Ah yeah ok, I'd had a quick browser on the lodash page but didn't see it referenced anywhere so thanks for those specifics. 

I'd certainly be in favor of adding this in as I've had to resort to generating my own looping systems where I've needed to break out early (eg many of the boolean style operations).

stebogit:
implemented with #1185 and  #1186

PR #1185, opened 2017-12-18
Add breaking iterators to @turf/meta

DenisCarriere:
## Add breaking iterators to `@turf/meta`

Ref: https://github.com/Turfjs/turf/issues/1182

This should add the ability to break a TurfJS iterator (eventually we can introduce real [Iterators & Generators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators) which will be for another day)

```js
const lines = lineStrings([
    [[10, 10], [50, 30], [30, 40]],
    [[-10, -10], [-50, -30], [-30, -40]]
]);

let count = 0;
meta.featureEach(lines, () => {
    count += 1;
    return false;
});
// count => 1 (instead of 2)
```

## To-Do

- [x] Add documentation about this new behavior (@Spown care to help?)
- [ ] Add more testing (current one does the job.. but not very robust)
- [ ] Issues implementing `meta.segmentEach`

CC: @rowanwins @Spown


Spown:
> Add documentation about this new behavior (@Spown care to help?)

#1186

PR #1192, opened 2017-12-21
Clarification about @turf-angle

stebogit:
Currently the new `@turf-angle` is calculating the angle defined by three points (`A`, `o` and `B` in the picture) is defined/calculated as the angle (< 180˚) between the lines extending `Ao` and `oB`; as a result the angles `AoB` and `BoA` are equals.
**Note:** passing the option `exterior=true` will return the explementary angle (360 - `angle`).
angle2

Now, I believe a more useful calculation (maybe in addition to the current one) would be defining the angle as the clockwise (or counterclockwise) angle between `A` (`startPoint`) and `B` (`endPoint`) with center in `o` (`midPoint`). This would allow comparison between angles due to the reference `Ao` segment.
angle

In the picture above the current implementation of the module would return the red angle for `AoB` and the blue dotted angle for `AoB'`.  I believe it would be more useful getting `AoB'`  as the solid blue line, otherwise there is no way to properly compare the two angles.

In this PR I just added a few test cases to confirm and highlight the current behavior.

@DenisCarriere @rowanwins I propose to modify the output as suggested ([here a possible implementation](https://github.com/Turfjs/turf/blob/angle-tests/packages/turf-point-to-line-distance/index.js#L82-L84), but I'm sure we can find a better one) and adding an option (`absolute`?) to provide the current result.
I'd also rename the `external` option to `explementary`.

If the current output is fine I'd probably add some notes to the docs to clarify the output.


DenisCarriere:
> adding an option (absolute?) to provide the current result.
👍 

> I'd also rename the external option to explementary.
👍 

@stebogit I agree with all your comments, this was more a "draft" module, it's still not implemented in `@turf/turf` so we can still do some changes as patches.

DenisCarriere:
@stebogit I implemented your `bearingToAzimuth` change to `@turf/angle` (with minor changes).

Have a look at it to see if that's what you were thinking.

DenisCarriere:
@stebogit I've added some color the test outputs, a lot easier to visualize the outputs now.

## Obtuse Angle

- **Red**: Interior Angle
- **Blue**: Explementary Angle

image

## Acute Angle

- **Red**: Interior Angle
- **Blue**: Explementary Angle

image

stebogit:
@DenisCarriere I tried to add a short description in the jdocs, but I believe it can be explained better. I also added few tests.

This looks ok to me, you did all the work! 😄  👍

DenisCarriere:
> This looks ok to me, you did all the work!

You did the initial work 👍 😄 

Good stuff, i'll merge

packages/turf-angle/index.js
```diff
@@ -21,20 +22,18 @@ function angle(startPoint, midPoint, endPoint, options) {
...
 
-    var azimuthAO = bearingToAzimuth((mercator !== true) ? bearing(A, O) : rhumbBearing(A, O));
-    var azimuthBO = bearingToAzimuth((mercator !== true) ? bearing(B, O) : rhumbBearing(B, O));
+    var azimuthAO = bearingToAzimuth((options.mercator !== true) ? bearing(A, O) : rhumbBearing(A, O));
+    var azimuthBO = bearingToAzimuth((options.mercator !== true) ? bearing(B, O) : rhumbBearing(B, O));
     var angleAO = Math.abs(azimuthAO - azimuthBO);
 
     // Explementary angle
-    if (explementary === true) return 360 - angleAO;
+    if (options.explementary === true) return 360 - angleAO;
```
DenisCarriere: 👍 I'm also starting to like using `options.*` instead of defining them at the beginning, it's easier to tell which variables are derived from the options.

packages/turf-angle/test.js
```diff
@@ -66,3 +67,15 @@ test('turf-angle -- issues', t => {
     t.false(isNaN(a), 'result is not NaN');
     t.end();
 });
+
+test('turf-angle -- throws', t => {
+    const pt1 = point([-10, -30]);
+    const pt2 = point([-11, -33]);
+    const pt3 = point([-12, -36]);
+    t.throws(() => angle(null, pt2, pt3), /obj is required/, 'missing startPoint');
```
DenisCarriere: Maybe we should throw an Error that says `missing startPoint` instead of `obj is required`

packages/turf-angle/index.js
stebogit: 👍

packages/turf-angle/test.js
stebogit: Yes, I was concerned about the performance, but it should not be a big deal.
Have you checked the benchmark by any chance?

DenisCarriere: I wouldn't worry about performance for those types of checks, it doesn't slow anything down.

The biggest performance hits is reiterating over a FeatureCollection twice when 1 pass can achieve the same result.

I'd say all the input type checking is always valid to add, the "heavy" performance hit will be the main logic and not the input checking.

Issue #1157, opened 2017-12-06
Getting fill property not supported on IE11


estradamarkie:
Hi,

I'm getting this error on IE11 using https://npmcdn.com/@turf/turf@5.0.4/turf.min.js

> Object doesn't support property or method 'fill'

Just wondering if this issue has been sorted for IE11 or anyone else is having this issue?

Thanks

DenisCarriere:
🤔 ... I believe the `v5.0` isn't being compiled with Babel, it's using Browserify + Uglify to compile the final bundle.

@estradamarkie Can you try to pinpoint which module is causing this, we've got LOTS of dependencies and sub-modules.

Next release will 100% be compiled with Babel, so maybe we just have to wait out until the next release to have this fixed.

DenisCarriere:
@estradamarkie You are correct that `Array.fill` is not supported by IE 11

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill

Of course, our lovely `JSTS` repo has this 😓 ...
https://github.com/DenisCarriere/turf-jsts/search?utf8=%E2%9C%93&q=fill&type=

estradamarkie:
@DenisCarriere Yeah `Array.fill()` is not compatible with IE, thought this would have been seen by other people who's using turf.js. 

Guessing this will be a major bug for the plugin? Are we expecting a fix soon?

DenisCarriere:
Well `v5.0` wasn't using Babel to compile, I assumed that all dependencies didn't use any Polyfills.

By compiling the next minor release with Babel this "should" fix this.

DenisCarriere:
Hopefully we won't have to include this Polyfill

https://github.com/bjornharrtell/jsts/blob/master/src/Array.js

rowanwins:
Or maybe, just maybe, perhaps Martinez will be ready!

DenisCarriere:
lol indeed! We can make that a major release, it's going to take sometime to get that released, I'm sure we will have few bugs to fix before we swap out JSTS for Martinez

estradamarkie:
Is there a work-around for now while waiting for the official fix?

DenisCarriere:
You can include the following Polyfills in your code before importing Turf.

- [Array.js](https://github.com/bjornharrtell/jsts/blob/master/src/Array.js)
- [Math.js](https://github.com/bjornharrtell/jsts/blob/master/src/Math.js)
[...]

PR #1193, opened 2017-12-21
Ponyfill Object.assign - fix for #1157

DenisCarriere:
Ref: https://github.com/Turfjs/turf/issues/1157#issuecomment-353005807


rowanwins:
Always thought we needed a ponyfill @DenisCarriere , turf is much better for it

Issue #914, opened 2017-08-24
Type Checking for Coordinates and NaN


markmeyerphoto:
Creating points doesn't coerce strings into numbers but instead gives an error:

    var lat = '61',
    var lon = '-149'
    turf.point([lon, lat])
    // Error: Coordinates must contain numbers

   That's a little unjavascript-like, but reasonable enough. A common way of dealing with this is to coerce numbers with `parseInt()` or:

    turf.point([+lon, +lat])

This works, but creates a false sense of security. Because turf's error checking depends on `typeof`,
this works with no error:

    var lat = 'Foo',
    var lon = 'Bar'
    turf.point([+lon, +lat])

 `+"Foo"` returns `NaN` which is `typeof number`. I wonder if it would be possible for turf to fail here when given `NaN` to make it easier to pass in converted strings without type checking. If Turf is going to do type checking, it would be great if it was a little more thorough.

stebogit:
@DenisCarriere I noticed you actually removed that check [here](https://github.com/Turfjs/turf/commit/61861a4602b80ec4c77d709c7b939db24940a85b#diff-f048fd878f5c2b1cdbcfba91ab39b51cL95) (@markmeyerphoto this change is not in the current/latest release, as you can see).
If we actually had the type validation on `point` coordinates I guess we should have it in all the similar `helpers` functions, which would probably slow down particularly the multi feature creation functions. Is this why you removed it?

If we did want to have a reliable check for numbers, I tested this function: https://repl.it/KXcu/4

DenisCarriere:
>  I noticed you actually removed that check

👍 Looks like it was removed by accident 🤦‍♂️ , that was a big commit and I might of missed that when I committed that change.

As for more advanced type checking, we shouldn't be doing anything more than the basic `typeof` validation checks.

If one wants advanced Type checking then use `Typescript` which will handle all your type checking.

> we should have it in all the similar helpers functions

Points should be the only helper function to do this, all the other helper functions would be overkill and would significantly slow down the performance if the geometry is very large.

I would recommend we stick with the current style of basic validation:
- typeof geojson === `string`
- geojson === `undefined`
- geojson === `null`

DenisCarriere:
@stebogit 👍 Sweet method! 👏  Maybe we can add that to `@turf/helpers`?

```js
function isNumber(x) {
    return !isNaN(x) && x !== null && !Array.isArray(x);
}
```

DenisCarriere:
# To-Do list

[...]

PR #1197, opened 2017-12-23
Increase 5x performance of invariant.getCoord()

DenisCarriere:
## Increase 5x performance of `invariant.getCoord()`

Ref: https://github.com/Turfjs/turf/issues/914#issuecomment-351806286

Tests remained unchanged, however performance is 5x faster than previous implementation.

Assumes when giving a `Feature<Point>` it already contains valid coordinates.

## Benchmark Results

```
getCoord (BEFORE) x 14,586,933 ops/sec ±0.92% (88 runs sampled)
getCoord x 49,812,538 ops/sec ±1.00% (87 runs sampled)
```

### DIFF
--- a/packages/turf-invariant/index.d.ts
+++ b/packages/turf-invariant/index.d.ts
@@ -6,14 +6,16 @@ import {
     GeometryTypes,
     CollectionTypes,
     Types,
+    Point,
     AllGeoJSON,
-    Geometries
+    Geometries,
+    Position
 } from '@turf/helpers'
 
 /**
  * http://turfjs.org/docs/#getcoords
  */
-export function getCoord(obj: Feature<any> | GeometryObject | any[]): number[];
+export function getCoord<P extends Position>(coord: Feature<Point> | Point | P): P;
 
 /**
  * http://turfjs.org/docs/#getcoords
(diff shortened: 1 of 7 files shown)


packages/turf-invariant/index.js
```diff
@@ -4,32 +4,28 @@ import { isNumber } from '@turf/helpers';
...
  */
-export function getCoord(obj) {
-    if (!obj) throw new Error('obj is required');
-
-    var coordinates = getCoords(obj);
+export function getCoord(coord) {
+    if (!coord) throw new Error('coord is required');
+    if (coord.type === 'Feature' && coord.geometry !== null && coord.geometry.type === 'Point') return coord.geometry.coordinates;
+    if (coord.type === 'Point') return coord.coordinates;
+    if (coord.length >= 2 && coord[0].length === undefined && coord[1].length === undefined) return coord;
```
DenisCarriere: @rowanwins & @stebogit This seems to be a 99% light & performant solution to test if the coordinates is correctly an Array of numbers.

packages/turf-invariant/index.d.ts
```diff
@@ -6,14 +6,16 @@ import {
...
-    Geometries
+    Geometries,
+    Position
 } from '@turf/helpers'
 
 /**
  * http://turfjs.org/docs/#getcoords
  */
-export function getCoord(obj: Feature<any> | GeometryObject | any[]): number[];
+export function getCoord<P extends Position>(coord: Feature<Point> | Point | P): P;
```
DenisCarriere: Extends Array output to match the Array input, if `[number, number, number]` is entered, then the output will be the same type.

Issue #1202, opened 2017-12-25
@turf/invariant proposal methods (first & last => coord & segment)


DenisCarriere:
## ~`@turf/meta`~ `@turf/invariant` proposal methods

A few helpful methods would be useful to have in the [`@turf/invariant`](https://github.com/Turfjs/turf/tree/master/packages/turf-invariant) collection.

### Proposed new methods

- [x] `invariant.firstCoord`
> Retrieve the first `Coordinate` of the GeoJSON as an `Array<number>`
- [x] `invariant.lastCoord`
> Retrieve the last `Coordinate` of the GeoJSON as an `Array<number>`
- [ ] `invariant.firstSegment`
> Retrieve the first `Segment` of the GeoJSON as a 2 vertex `Feature<LineString>`
- [ ] `invariant.lastSegment`
> Retrieve the last `Segment` of the GeoJSON as a 2 vertex `Feature<LineString>`

### Example

```js
const line = lineString([[10, 10], [50, 30], [30, 40]]);

invariant.firstCoord(line) // => [10, 10]
invariant.lastCoord(line) // => [30, 40]
invariant.firstSegment(line) // => Feature<LineString<[[10, 10], [50, 30]]>>
invariant.lastSegment(line) // => Feature<LineString<[[50, 30], [30, 40]]>>
```

DenisCarriere:
## 1. Change method names

Or keep the naming schema consistent with the other `@turf/meta` methods names.

- `meta.coordFirst(geojson)`
- `meta.coordLast(geojson)`
- `meta.segmentFirst(geojson)`
- `meta.segmentLast(geojson)`

## 2. Change method parent to `@turf/invariant`

Possibly these methods would be more appropriate under `@turf/invariant` since they don't perform any iterations, this would sync well with other invariant methods like `getCoord()` & `getCoords()`.

stebogit:
@DenisCarriere I'm sure you found a use case for these modules but I can't think of one at the moment, particularly regarding getting the first point of the first line of a `MultiLine` 🤔 .
Anyway, trying to contribute to your idea, I'd create:
- `invariant.start` and `invariant.end`, applied only to lines returning the corresponding point
- `invariant.first` and `invariant.last`, applied to lines and multilines, returning the first/last segment or line
- `invariant.lat` and `invariant.long` only for points, returning the named coordinate

If we made them right we could probably chain them like:
```js
var latitude = invariant.lat(invariant.start(invariant.first(multiline))); 
```
[...]

PR #1204, opened 2017-12-26
Implement @turf/invariant firstCoord & lastCoord

DenisCarriere:
## Implement `@turf/invariant`'s `firstCoord` & `lastCoord`

Ref: https://github.com/Turfjs/turf/issues/1202

### `invariant.firstCoord()`

```js
/**
 * Retrieve the first Coordinate of the GeoJSON as an Array<number>
 *
 * @param {Feature|Geometry} geojson Any GeoJSON Feature or Geometry
 * @param {string} [name="geojson"] name of the variable to display in error message
 * @returns {Array<number>} GeoJSON Position
 */
export function firstCoord(geojson, name) {
```

### `invariant.lastCoord()`

```js
/**
 * Retrieve the last Coordinate of the GeoJSON as an Array<number>
 *
 * @param {Feature|Geometry} geojson Any GeoJSON Feautre or Geometry
 * @param {string} [name="geojson"] name of the variable to display in error message
 * @returns {Array<number>} GeoJSON Position
 */
export function lastCoord(geojson, name) {
```

### Example

```js
const line = lineString([[10, 10], [50, 30], [30, 40]]);
invariant.firstCoord(line) // => [10, 10]
invariant.lastCoord(line) // => [30, 40]
```

### Other use cases

**Retrieve the first Coord of each Ring from a MultiPolygon**

```js
meta.lineEach(multiPoly, line => {
  invariant.firstCoord(line) // => Coord
})
```

## Up for discussion

- How to handle Polygon's lastCoord? Do we chose the last coordinate of the outer ring or the lastCoord of the entire coordinates (which could be an inner ring).


DenisCarriere:
Going to merge this for now since I'm going to be working on `firstSegment` & `lastSegment` on a separate PR which will extend from this PR.

PR #1212, opened 2018-01-05
Experimental ES Modules (*.mjs)

DenisCarriere:
## Experimental ES Modules (`*.mjs`)

Using Node 9.3 you can use the new ES Modules (import & export) syntax by using `--experimental-modules` flag.

Downside is it's very strict on the file extension, it must be `*.mjs`.

Going to publish 3 main TurfJS core modules that experiment when importing it from different libraries, internally it seems to be fine, when publishing to npm it's a different story.

```
$ node --experimental-modules index.mjs
```
 
## References

https://github.com/nodejs/node-eps/blob/master/002-es-modules.md#44-shipping-both-esm-and-cjs

### 4.4. Shipping both ESM and CJS
When a package.json main is encountered, file extension searches are used to provide a means to ship both ESM and CJS variants of packages. If we have two entry points index.mjs and index.js setting "main":"./index" in package.json will make Node pick up either, depending on what is supported.

### 4.4.1. Excluding main
Since main in package.json is entirely optional even inside of npm packages, some people may prefer to exclude main entirely in the case of using ./index as that is still in the Node module search algorithm.


rowanwins:
Hey @DenisCarriere I was wondering just the other day if we were at a stage we're individual modules could be made available via a CDN, if I understand correctly that could be an outcome of this work?

DenisCarriere:
@rowanwins You should be able to use these now, I've published `helpers`, `meta` & `invariant` as under the `6.0.0-beta.3` version.

- https://unpkg.com/@turf/meta@6.0.0-beta.3?module
- https://unpkg.com/@turf/invariant@6.0.0-beta.3?module
- https://unpkg.com/@turf/helpers@6.0.0-beta.3?module

You can do something like this with your HTML:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>ES Modules - Single Module</title>
  </head>
  <body>
  <script type='module'>
    import {point} from "https://unpkg.com/@turf/helpers?module"
    console.log(point([100, 0]));
  </script>
  </body>
</html>
```

rowanwins:
That is cool @DenisCarriere , support for that[ approach is a bit patchy ](https://caniuse.com/#feat=es6-module) but will presumably continue to improve.

I was probably thinking something more like traditional turf CDN usage where you included a script and then call `turf.whatever` , so I guess essentially a rolledup version of every module exposing a turf global...

Anyway I'll keep thinking about it, I hadn't put loads of thought in, it was just a 'wouldnt that be nice' kinda thought the other day.

DenisCarriere:
> @rowanwins I was probably thinking something more like traditional turf CDN usage where you included a script and then call turf.whatever

Yep, you can already do that in the current Turf release, same as if you would with the global Turf CDN except using `<script type='module'>`.

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>ES Modules - Single Module</title>
  </head>
  <body>
  <script type='module'>
    import * as turf from "https://unpkg.com/@turf/turf?module"
    console.log(turf.point([100, 0]));
  </script>
  </body>
</html>
```

The only reason why I didn't mention this is because it's really heavy in file size (264KB).

image

DenisCarriere:
> @rowanwins support for that approach is a bit patchy but will presumably continue to improve.

Agreed, I don't think this is going to be widely adopted (just yet)...

Another approach I'm thinking of going is simply convert repo to Typescript, I've had a lot of success in my other libraries maintaining an "ES6+" source code and easily/safely transpile to ES5 friendly code.

I'll most likely close this PR and retry with Typescript (let's see where that goes...).

PR #1231, opened 2018-01-22
Remove jsts from turf/intersect

rowanwins:

First cut of replacing jsts for turf/intersect

What I've noticed (which I never really knew) however is that jsts could generate line and point outputs.

So for example 
- passing in a polygon and a point would return the point (eg see this [test](http://bl.ocks.org/d/7155591e35c40590df491e6e6bb7121d))
- or a polygon and line would return the line (eg [this](http://bl.ocks.org/d/67e300d18fd8d5ecfcc72dc512deff5f))
- or even more interesting two polygons with a shared segment would return that segment as a linestring (eg [this](http://bl.ocks.org/d/20f95b1f2f2aced5495c1ca4d5f9b235))

So we'll need to work out if we want to support these feature combinations and results. 
- We've already got point-in-polygon supported so we could easily drag that in 
- line-intersection of polygons is a bit trickier, i not aware of anything that tackles that in a robust way...

Anwyay  bit more thought and work required for this module


DenisCarriere:
> @rowanwins What I've noticed (which I never really knew) however is that jsts could generate line and point outputs.

I never really liked the fact that the jsts outputs could be semi-random (point & lineString).

It might be best to support (`Polygon`, `Polygon`) input => `(Multi)Polygon` | `null` as output. That way it's easier to handle the output data since it would be consistent or at least always in the Polygon family.

PR #1233, opened 2018-01-24
Expand other languages section of readme

1ec5:
Updated the readme to reflect the fact that the Swift port supports five platforms as of [v0.1.0](https://github.com/mapbox/turf-swift/releases/tag/v0.1.0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,10 @@ TypeScript is supported internally within each module, no installs required.
 
 ### Other languages
 
-Ports of Turf.js are underway [in Swift](https://github.com/mapbox/turf-swift/) (for iOS and macOS) and [in Java](https://github.com/mapbox/mapbox-java/blob/master/docs/turf-port.md) (for Android and Java SE).
+Ports of Turf.js are available in:
+
+* [Java](https://github.com/mapbox/mapbox-java/blob/master/docs/turf-port.md) (Android, Java SE)
+* [Swift](https://github.com/mapbox/turf-swift/) (iOS, macOS, tvOS, watchOS, Linux)
 
 - - -
 
--- a/README.md
+++ b/README.md
@@ -64,8 +64,9 @@ TypeScript is supported internally within each module, no installs required.
 
 Ports of Turf.js are available in:
 
-* [Java](https://github.com/mapbox/mapbox-java/blob/master/docs/turf-port.md) (Android, Java SE)
-* [Swift](https://github.com/mapbox/turf-swift/) (iOS, macOS, tvOS, watchOS, Linux)
+- [Java](https://github.com/mapbox/mapbox-java/blob/master/docs/turf-port.md) (Android, Java SE)
+- [Swift](https://github.com/mapbox/turf-swift/) (iOS, macOS, tvOS, watchOS, Linux)
+> Turf for Swift is **experimental** and its public API is subject to change. Please use with care.
 
 - - -
 


DenisCarriere:
@1ec5 TurfJS **is** is written in a functional programming paradigm without any state. At the moment the Swift port is not aligning with TurfJS, here are a few issues I'm seeing:

- Class based (instead of Functional)
- Method names are different
- Module names are different
- No documentation on the Swift port
- Mono file ([`Turf.swift`](https://github.com/mapbox/turf-swift/blob/master/Sources/Turf/Turf.swift) should be divided into modules)
- Mono test cases, also should be divided into separate test cases
- Not following the GeoJSON naming structure (Polyline, should be called LineString instead)
- `@turf/meta` & `@turf/invariant` are not implemented in Swift which are CORE modules which are used in 80% of all recent TurfJS modules.

If the Turf Swift port continues to separate itself from the TurfJS core values, I would highly recommend removing the name "Turf" from the Mapbox Swift library and call it something else since this port is not a reflection of TurfJS.

On a positive note, the Java port seems to be in the right TurfJS direction.

CC: @cammace

1ec5:
@DenisCarriere, thanks for taking a look at this PR, and I agree with keeping a note about the Swift library being a work in progress. There’s a lot more work to do!

I find it unfortunate that some of the differences mentioned above are considered a departure from TurfJS’s core values, considering that Swift is a very different language than JavaScript and Java. If all of these values are non-negotiable for a port to another language, then it is all but guaranteed that there will never be a Turf in Swift. That would be unfortunate, because there is a clear demand for simple spatial analysis tools in that language, and Turf is a very good foundation to base such tools on. I hope you’ll reconsider based on the reasoning below, but – speaking for myself – I’m not fundamentally opposed to finding a different name for the library.

> Class based (instead of Functional)

Structs aren’t classes; they’re the standard way to namespace functionality in Swift. It isn’t straightforward to separate the library into individual modules along the lines of Turf’s modules: the entire library must be imported en masse, so making all the methods into standalone functions would severely pollute the global namespace, hindering adoption of the library. It also isn’t feasible to separate each port of a module into its own library, because there’s a very low limit to the number of libraries an iOS application can link to dynamically.

We did opt to use instance methods and computed properties instead of static methods. Having everything be a static method would feel foreign in Swift, as in C code that bridges poorly to Swift. Yes, these methods are a little more reminiscent of object-oriented programming than functional programming, but that shouldn’t stop anyone from using these methods within otherwise [functional code](https://github.com/mapbox/turf-swift/blob/9cd6e9038bc69127c43481ff151f746d67db3de0/Sources/Turf/Turf.swift#L177).

If the concern is that a particular operation would need to be duplicated across multiple structs in order to support multiple geometry types, note that once we start supporting a given operation on multiple geometry types, we’ll probably end up moving that functionality to a protocol that can be shared among multiple structs.

> Method names are different

The renaming of methods may seem severe, but it’s important for the library to follow [Swift’s API design guidelines](https://swift.org/documentation/api-design-guidelines/), particularly as they relate to the naming of argument labels. For example, a function name has to have multiple labels (the things that end with colons) in order to take multiple arguments.

> No documentation on the Swift port

There is documentation in the form of a documentation comment on each method. These comments automatically appear within Xcode’s code completion window and Quick Help. Certainly, there’s room for improvement, and mapbox/turf-swift#44 tracks publishing a copy of that documentation in HTML format.

> Mono file (Turf.swift should be divided into modules)
> Mono test cases, also should be divided into separate test cases

I agree, it should be divided into separate files (not modules): mapbox/turf-swift#42.

> Not following the GeoJSON naming structure (Polyline, should be called LineString instead)

Renaming Polyline to LineString is certainly a possibility. We were originally going for naming consistency with other iOS libraries like MapKit and trying to avoid confusion with String. On the other hand, GeoJSON is a central aspect of working with Turf, so a rename along those lines might make sense in the context of mapbox/turf-swift#43.

> `@turf/meta` & `@turf/invariant` are not implemented in Swift which are CORE modules which are used in 80% of all recent TurfJS modules.

Thanks for the context. We’re tracking functional parity with the JS library with [this label](https://github.com/mapbox/turf-swift/labels/JS%20parity) and would certainly be open to either porting all the functionality in those modules (at least any functionality that isn’t already contained in the Swift Standard Library).

PR #1234, opened 2018-01-26
Convert *.js => *.mjs

DenisCarriere:
## Convert *.js => *.mjs

One step closer to supporting ES Modules.

TurfJS packages would support 3 Javascript formats:

- CommonJS => `index.js`
- ES Modules => `index.mjs`
- Typescript => `index.d.ts`

The code base still needs to be written in ES5 with the exception of the ES Module syntax (`import` & `export`).

This approach involves minimal refactoring and running tests require little amounts of compile tile (takes 0.4s to compile `*.mjs` => `*.js`).

## ⚠️ Breaking Changes

Very minor breaking changes will occur, however some of the older Webpack configurations might need to be updated to reflect the `*.mjs` file extension. Most users that are still using CommonJS shouldn't see any different.

> Note: Using Browserify should not be affected by this change.

## Modules Converted (for experimental purposes)

- `@turf/helpers`
- `@turf/invariant`
- `@turf/meta`


Turbo87:
any reason why you're not using https://github.com/standard-things/esm instead?

DenisCarriere:
@Turbo87 Using `esm` is still only a patch, defining the ES Modules files as `*.mjs` would help at least identify which files are using `export/import` and which file is CommonJS (`*.js`).

As an ideal world, I'd rather convert everything to Typescript, that way we can finally write ES6+ code and easily/safely compile it down to CommonJS, but hoping to Typescript now would require too much efforts in refactoring the codebase.

Node 9 can now use ES Modules, however the downside is the file extension must be defiend as `*.mjs` and not `*.js` for it to work.

https://nodejs.org/api/esm.html

Turbo87:
> but hoping to Typescript now would require too much efforts in refactoring the codebase

can you elaborate on that? what would be needed to do that? it doesn't seem like such a big task to me as this could be done iteratively once the basic machinery is in place 🤔

DenisCarriere:
@Turbo87 Well some of the modules like `@turf/helpers` & `@turf/meta` would take some time to rewrite entirely in Typescript, the current code is kinda "hacky" and Typescript would throw a lot of errors, but it's definitely doable.

Let me see if can do 1 or 2 smaller Turf modules as a Test bed for this.

Another "annoying" thing is that Typescript doesn't support outputting `*.mjs` file extensions from `*.ts` files. We might still need to use Rollup to create the local bundles.

https://github.com/Microsoft/TypeScript/issues/18442

milafrerichs:
hello, first of all, thanks for turf it is awesome.
but I wanted to use a 6.0 package because there is a fix for the square grids.
but I use webpack 3.x and can't import the `.mjs` modules or at least they don't get imported correctly.
I tried to research that and found a lot of articles but none of them made it any clearer for me what to do.
can you help me and point in the right direction how I would use turf@6.0 with webpack 3.x?

thanks a lot.

DenisCarriere:
@milafrerichs Thanks for this comment, I agree that the `*.mjs` file extensions is more of a hassle at this point with today's current Javascript systems.

Currently looking into remove the 3-5 modules that are using `*.mjs` and replacing it with either Typescript or (Types + Javascript).

milafrerichs:
thanks. I would love to use 6.x modules since they fix certain issues I'm facing. so would be great if they're published with backwards compatibility or at least a How-To guide to use `.mjs`

DenisCarriere:
Well they are backwards compatible, a `.js` is also provided as the `main`.

For example `@turf/helpers` v6 => https://unpkg.com/@turf/helpers@6.0.1/

Includes support for:
- CommonJS => main => `.js`
- Typescript => types => `.d.ts`
- ES Modules => module => `.mjs`

The issue is in Webpack and how to configure Webpack to use `.mjs` (not TurfJS)

milafrerichs:
yes, you're right. sorry. 
the problem is that webpack 3 tries to load the ES modules and not the CommonJS modules. 
do you have a good source where it is explained in more detail and how I would fix it?
thanks a lot.

jsauca:
Any solution to this issue "The issue is in Webpack and how to configure Webpack to use .mjs" ? @DenisCarriere Thanks a lot

PR #1240, opened 2018-01-30
Converting TurfJS to 100% Typescript + Rollup?

DenisCarriere:
## Converting TurfJS to 100% Typescript + Rollup?

This approach actually has **many** benefits and we will **FINALLY** allow us to write in ES6+ without having any transpiling issues to make it backwards compatible to ES5 CommonJS. The reason why Rollup is included is the create the `*.mjs` ES Modules bundle & `*.js` CommonJS bundle (typescript is only be able to create 1 output "at the moment").

## Benefits

- Transpiles to CommonJS ES5 with confidence.
- Outputs both `*.mjs` & `*.js` bundles using a single Rollup command
- Can write in ES6 (`const`, `let`, `=>` syntax would now be allowed)
- Strict enforcement of GeoJSON Types (annoying at times, but worth it to have a stable & consistent code base)
- No need for `@std/esm` (very good library, however no longer required if we move to Typescript).
- Relatively fast & easy to convert a simple TurfJS package to Typescript (~5-10 minutes per package).
- ⭐️ **(newly added)**: JSDocs will get transferred to both Typescript, CommonJS & ES Modules (before this proposal the Typescript types didn't contain the full JSDocs from the CommonJS file)
- ⭐️ **(newly added)**: Supports 3 main formats using a single index file:
  - Typescript (`index.ts` "default")
  - ES Modules (`index.mjs`)
  - CommonJS (`index.js`)

## Pitfalls

- Dependency libraries that don't yet support Typescript will need to be included to [DefinitelyTyped](github.com/DefinitelyTyped/DefinitelyTyped). Most of the "geo" related libraries I've added myself over the past year, so most of the ones we need are present (even [Martinez](https://github.com/w8r/martinez/blob/master/index.d.ts) includes Typescript support).
- Some of the more complex TurfJS libraries might need some special attention (@DenisCariere can tackle those ones).

## How to convert Turf packages to Typescript

- [ ] Rename `index.js` => `index.ts`
- [ ] Add any Typescript types to the input & outputs (can have a peak at the `index.d.ts`)
- [ ] Delete `index.d.ts` since `index.ts` will now be used as the Typescript main file.
- [ ] Convert `test.js` & `bench.js` to CommonJS (replace `import` with `require()`). No need to convert our test files to Typescript, that's a bit too much for now.
- [ ] Update `package.json` to reflect the following:

```json
  "main": "index",
  "module": "index.mjs",
  "files": [
    "index.js",
    "index.mjs",
    "index.ts"
  ],
```
- [ ] Update `scripts` in the `package.json` to reflect the following:

```json
  "scripts": {
    "pretest": "rollup -c ../../rollup.config.typescript.js",
    "test": "node test.js",
    "bench": "node bench.js",
    "docs": "node ../../scripts/generate-readmes"
  },
```
- [ ] Drop `@std/esm` from `devDependencies` & extra `@std/esm` configs from `package.json`
- [ ] Drop `types.ts` files since the Typescript testing will now be achieved via the `index.ts`

## Approved Votes

- 👍 @DenisCarriere
- ❓ @rowanwins 
- ❓ @stebogit 
- ❓ @morganherlocker
- ❓ @tmcw 
- ❓ @muziejus

@Turfjs/ownership Thoughts?


rowanwins:
Hmmmm I think it's a👎 from me sorry - primarily because it's a barrier to entry for contributors, its just another thing someone has to be aware of before they can contribute to turf.

For the timebeing I'd rather us focus on getting some bug fixes rather than another change to tooling (we already had a substantial overhaul <6 months ago)

bradjonesca:
I'm not a member but I concur with @DenisCarriere here ... Turfjs is getting more sophisticated now and Typescript has advanced and matured (including tooling support) to the point where combined they are highly complementary.  Besides if not now (6.0 and beyond) then when?

DenisCarriere:
>  primarily because it's a barrier to entry for contributors

I think it's the opposite, Typescript tells you when something doesn't make sense in your source code.

Also, there's very little change in the "before" (ES Modules + CommonJS) & "after" (Typescript).

https://github.com/Turfjs/turf/pull/1240/files#diff-904189f4658913b6a914b708f56e7a27

Including Typescript allows us to drop `@std/esm` which is still "odd" tooling for some folks.

> rather than another change to tooling 

That tooling was for the ES Modules and it's pretty "clunky" using the `main.es.js` builds. To fix that, we can switch to `*.mjs` (which is an entirely new tooling)

Finally, creating consistent TurfJS CommonJS bundles has been a nightmare in making sure everything is using CommonJS ES5 syntax. Introducing Typescript will finally allow us to write in ES6 without the hassles & worries of not compiling correctly to CommonJS.

tmcw:
I'm into it. The replacement of JSTS still seems like the number 1 turf thing, the cause of the most unfixable bugs and the majority of bundle size. But that's not something you can just 'throw more people at' and fix, and the tooling is certainly worth some more evolution, given that it's still kind of rube-goldbergian.

rowanwins:
googling 'rube-goldbergian' :) - _Achieving a simple objective by absurdly complicated means; over-complicated, elaborate and impractical_

DenisCarriere:
> The replacement of JSTS still seems like the number 1 turf thing,

Agreed, @rowanwins & @w8r have been working hard on being able to replace some of the JSTS dependent libraries by improving the [Martinez](https://github.com/w8r/martinez) library.

Google'd 'rube-goldbergian'

image

stebogit:
Guys about Typescript I literally only know that was originally developed by Microsoft cause I read it somewhere by chance (so that might even be a wrong... 😅 ), so I can't really contribute to this decision.
Typescript seems becoming more and more popular, however @rowanwins has a good point, it might make Turf less appealing or even accessible for possible contributors.
On the other hand I do trust @DenisCarriere and @tmcw opinion that it promising and mature enough to become our new standard.
Having said that, Turf has been for me a great learning tool since the beginning, so I would consider this change as a new learning opportunity (when time permits 😞).

DenisCarriere:
> @stebogit Turf has been for me a great learning tool since the beginning, so I would consider this change just a new learning opportunity

Thanks for your input @stebogit, Typescript is really not so bad, it's that or we switch to [Facebook's Flow](https://github.com/facebook/flow) (however I still can't wrap my head around Flow's GeoJSON typing definitions).

> it might make Turf less appealing or even accessible for possible contributors

Many very popular repos have switched their codebase to Typescript:
- [Angular](https://github.com/angular/angular)
- [Rollup](https://github.com/rollup/rollup)
- [Mobx](https://github.com/mobxjs/mobx)
- [Apollo (GraphQL Client)](https://github.com/apollographql/apollo-client)
- [Microsoft's VSCode](https://github.com/Microsoft/vscode)
- [Ant Design (UI Design implemented in React)](https://github.com/ant-design/ant-design)
- [Ionic (Mobile App Framework)](https://github.com/ionic-team/ionic)

A bit too many examples 😄 ... but there's a lot of Top projects that have done the switch to pure Typescript.

tcql:
I'm not sure I have a solid opinion either way at the moment (other than the grumpy part of me that hates change is complaining and the rest of me is telling him to pipe down), but I figured this bit of perspective might be useful to think about:

> Many very popular repos have switched their codebase to Typescript:

I think that turf isn't the same as a project like angular or rollup or something because we just aren't that big. Our pool of contributors isn't huge, so, just because a project with a much larger footprint did this doesn't make it right _for us_. If the community here is in to it (which it sounds like there *is* some interest,), great! Then we can all learn to improve together and that's that.

> Typescript tells you when something doesn't make sense in your source code.

I think thought that just because it's state of the art or simpler by some metric doesn't necessarily mean it's _easier_ to pull contributors – a simple thing that's unfamiliar is still friction. It could be a deterrent to some otherwise good contributors, but then again, so might a lot of things be. 

Just things to consider. All-in-all, I think it's an interesting idea. Strongly opinionated tools generally agree with me well 😅. And I guess that on one hand, having a smaller contributor base is a plus in favor of making changes, because it's easier to onboard everybody who is active in the community.

stebogit:
In light of @tcql considerations, I wonder, due to my ignorance, is Typescript one of those things that everybody will need to learn sooner or later because it will become the industry standard, like ES6 for instance?
Or is it "just" another tool to "make it easier" to use bleeding-edge technologies (like again ES6 or ES7 now), which will no longer be needed once that technology will be the new standard?
Or maybe it's just another language, that is going to stick anway for some particular reason?

I could probably google this, but it might as well be food for thoughts here (says my lazy side that prevents me to google it 😅 ...).

DenisCarriere:
> @tcql I think that turf isn't the same as a project like angular or rollup or something because we just aren't that big. Our pool of contributors isn't huge

Looking at our contribution graphs, I'd say we're at par or above Rollup & Mobx (even though you might not believe it).

- https://github.com/Turfjs/turf/graphs/contributors (6 users with >50 commits)
- https://github.com/rollup/rollup/graphs/contributors (4 users with >50 commits)
- https://github.com/mobxjs/mobx/graphs/contributors (1 user with >50 commits - original Author)

> because we just aren't that big

Maybe this is due to lack of funding by larger geospatial organizations?? Hint hint... (https://opencollective.com/turf)

> @stebogit is Typescript one of those things that everybody will need to learn sooner or later because it will become the industry standard, like ES6 for instance?

Typescript (supported by Microsoft) has taken a big role in allowing people to write in ES6,ES7 code but still remain backwards compatible with ES5 CommonJS (which is still needed).

The issue that's happening right now with the current TurfJS codebase is we introduced `import` & `export` syntaxes (which is ES6), however we still need to write the rest of the code in ES5 in combination with Rollup. I think written mixed ES5 / ES6 code is even more confusing to "new" folks, on top of all that we are defining the Typescript definition files (`index.d.ts`) as a temporary Typescript support (instead we could just write in Typescript).

The whole reason why I introduced Typescript to Turf's project (about a year ago) was that GeoJSON types can get VERY confusing when using Turf modules, figuring out if this method allows for Points or Polygons, when you do X method does it output a LineString or a Feature Collection of LineString, all of those questions were pretty abstract unless we have strong documentation (which took a lot of effort to include all the JSDocs). Even with strong documentation, that doesn't mean the output/input types are being enforced (we try to make the methods do what we say).


## One liner

Typescript will helps enforce strict & consistent GeoJSON types across all TurfJS modules.

DenisCarriere:
Going to do the rest of the work on a different PR, this one has a lot of configuration & setup changes.

PR #1247, opened 2018-02-03
Yarn + Lerna workspace environment

DenisCarriere:
## Yarn + Lerna workspace environment

@tmcw had mentioned this a long time ago, using `yarn`'s workspace with `lerna` seemed like a great idea, however the configurations for this setup was never implemented in Turf.

Configuration pulled from:

https://github.com/0xProject/0x.js

## Install dependencies

If you don't have yarn workspaces enabled (Yarn < v1.0) - enable them:

```
$ yarn config set workspaces-experimental true
```

Using this approach essentially pulls all `packages/turf-*` dependencies into a single folder, thus making the install procedure very fast since it doesn't need to install all the dependencies into each package folder.

> FYI: Secondly, dropped some "useless" metadata in the main `package.json`.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,8 +1,9 @@
 {
-  "lerna": "2.0.0-beta.34",
-  "npmClient": "yarn",
-  "version": "5.1.5",
+  "lerna": "2.5.1",
   "packages": [
     "packages/*"
-  ]
+  ],
+  "version": "independent",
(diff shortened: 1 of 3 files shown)


DenisCarriere:
Continued setting up lerna workspace in PR https://github.com/Turfjs/turf/pull/1240

PR #1256, opened 2018-02-09
【update】1、Ensure the result has a same cellWidth and cellHeight.

DLCJianyf:
1、Ensure the result has a same cellWidth and cellHeight
2、Don't use turf.distance() function, to avoid when the angle between two points is obtuse, it will get wrong distance

edit by wangmy


DLCJianyf:
thank you i will check it

DLCJianyf:
fixed test error in turf/turf-square-grid

DenisCarriere:
Thanks @DLCJianyf for the initial effort.

I believe @rowanwins has resolved the issue here: https://github.com/Turfjs/turf/pull/1293

> PR's are welcomed (sorry it took a while for merging)

Issue #1265, opened 2018-02-15
Type Definitions seem broken on 5.16


pwilczynski:
Looking at the typings, they seem totally messed up for both booleanWithin and intersect:

https://github.com/Turfjs/turf/blob/v5.1.6/packages/turf-boolean-within/index.d.ts
https://github.com/Turfjs/turf/blob/v5.1.6/packages/turf-intersect/index.d.ts

Intersect seems better on master, but boolean-within still seems wrong. I'm happy to fix them, but I don't know how to tag a release. Any thoughts?

rowanwins:
Hi @pwilczynski 

@DenisCarriere is currently in the midst of refactoring some of those modules as he'd like to refactor the codebase to being typescript, previously we just had some typescript bindings. So @DenisCarriere is on it - he might be able to provide some direction if you'd like to help.

DenisCarriere:
@pwilczynski Hopefully the last two PR's would fix any Typescript related issues.

https://github.com/Turfjs/turf/pull/1263
https://github.com/Turfjs/turf/pull/1250

These changes will not be published in any `v5.x` release, these will be published under `v6.x`

pwilczynski:
Great - when does v6.x get published?

pwilczynski:
@DenisCarriere -> how do I pick up the v6.x version?

DenisCarriere:
Depends on which module, you would need to install them separately.

```bash
$ yarn add @turf/helpers
```

Will download v6.0.1 => https://unpkg.com/@turf/helpers@6.0.1/ 

```js
// CommonJS
const point = require('@turf/helpers').point;

// or ES6/Typescript
import {point} from '@turf/helpers';
```

pwilczynski:
OK interesting ->  booleanWithin is grabbing 5.1.5, is that expected?

DenisCarriere:
`@turf/boolean-within` hasn't be published as v6 yet, so it's still use `v5`.

pwilczynski:
Okdoke! Thanks a bunch!

DenisCarriere:
@pwilczynski just published v6 modules to npm.

Ref: https://github.com/Turfjs/turf/pull/1279

Re-open if you have any issues

pwilczynski:
@DenisCarriere -> thanks so much, really appreciate the fast turn-around here.

stebogit:
> @DenisCarriere -> thanks so much, really appreciate the fast turn-around here.

[...]

PR #1266, opened 2018-02-15
Update typings for boolean within

pwilczynski:
How do I depend on this nicer version. Are there 6.0 pre-releases yet?

https://github.com/Turfjs/turf/issues/1265

### DIFF
--- a/packages/turf-boolean-within/index.d.ts
+++ b/packages/turf-boolean-within/index.d.ts
@@ -3,7 +3,7 @@ import { Feature, GeometryObject } from '@turf/helpers'
 /**
  * http://turfjs.org/docs/#booleanwithin
  */
-export default function (
+export default function booleanWithin(
     feature1: Feature<any> | GeometryObject,
     feature2: Feature<any> | GeometryObject
 ): boolean;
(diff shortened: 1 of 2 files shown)


DenisCarriere:
> Are there 6.0 pre-releases yet

Going to be pre-releasing the `boolean` modules once they are all converted to Typescript.

Instead of doing one giant release with everything breaking, I've decided to slowly trickle out the `v6.x` package releases, that way we don't have a massive break in the build.

FYI: `@turf/turf` will be the last one to be published.

packages/turf-boolean-within/index.d.ts
```diff
@@ -3,7 +3,7 @@ import { Feature, GeometryObject } from '@turf/helpers'
 /**
  * http://turfjs.org/docs/#booleanwithin
  */
-export default function (
+export default function booleanWithin(
```
DenisCarriere: Those this actually prevent the Typings from working?

Since it's using `export default` the method name is irrelevant (IMO).

Also, this module will be moving entirely to Typescript, check out https://github.com/Turfjs/turf/pull/1263 for examples.

PR #1267, opened 2018-02-15
Resolves #1021

### LINKED ISSUE #1021: isobands documentation clarity

rowanwins:
simple docs fix for #1021

### DIFF
--- a/packages/turf-isobands/index.js
+++ b/packages/turf-isobands/index.js
@@ -9,11 +9,11 @@ import gridToMatrix from './lib/grid-to-matrix';
 import isoBands from './lib/marchingsquares-isobands';
 
 /**
- * Takes a grid {@link FeatureCollection} of {@link Point} features with z-values and an array of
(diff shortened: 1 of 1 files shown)


DenisCarriere:
👍 No need for PR :) just push directly to the master branch for doc related commits

PR #1268, opened 2018-02-15
Add new modules to docs

rowanwins:
adding polygonSmooth and booleanConcave to docs. Surely these tests will pass :)


DenisCarriere:
> Surely these tests will pass :)

Indeed! FYI, @rowanwins you don't need to make a PR for these types of changes.

Issue #1273, opened 2018-02-20
segmentEach Bug


rowanwins:
`segmentEach` is currently buggy as it creates a segment between an interior and exterior ring

https://runkit.com/rowanwins/segmenteach-bug

Basically we need to watch for an increase in the geometryIndex.

DenisCarriere:
👍 Agreed, we did the geometryIndex, it's strange that `segmentEach` passed the Travis Tests, I guess we don't have good enough tests for that use case.

rowanwins:
Took a bit of a look at the tests and yeah they seem a little sparse.

rowanwins:
Resolved by #1278

PR #1278, opened 2018-02-21
Add test case for segmentEach Polygon bug

DenisCarriere:
## Add test case for segmentEach Polygon bug

CC: @rowanwins 

Ref: https://github.com/Turfjs/turf/issues/1273

## To-Do

- [x] Add test case for `segmentEach` bug
- [ ] Fix segments being joined from inner/outer rings


DenisCarriere:
👍

Issue #64, opened 2013-12-31
 valid


morganherlocker:
function for checking if a given geojson object is valid. This might be possible with jsts, or with some sort of linter module.

vlasvlasvlas:
this is nice.

some sort of st_makevalid would be even nicer

blackrez:
Hello, I implemented the ST_isvalid of postgis : https://github.com/blackrez/turf-isvalid. 

How I can add in turf?

stebogit:
Is this geojson validator still of interest @morganherlocker ?
I know `jsts` is a goner, so we might want to implement this with [geojsonhint](https://github.com/mapbox/geojsonhint) instead.

@DenisCarriere @rowanwins is this material for `booleans`?

rowanwins:
It's still interesting from my perspective @stebogit 

I think this module probably requires more than `geojsonhint` and yep would probably fit in the booleans collection. 

Off the top of my head it's probably a combination of 
- geojsonhint
- @turf/kinks for id'ing self-intersections
- check for duplicate points  
- plus a bunch of other checks!

Some good [background reading here](https://postgis.net/docs/using_postgis_dbmanagement.html#OGC_Validity) from the postgis world.

DenisCarriere:
> I will give this a crack tomorrow. I will probably plug into geojsonhint.

@morganherlocker That's a long tomorrow 🤣

DenisCarriere:
> is this material for booleans? (@stebogit)

Not particularly, this would simply be a module which would throw an error if the GeoJSON is invalid.

DenisCarriere:
> I know jsts is a goner

Yes, however it might be worth looking into how `jsts` handles it's `isValid` method. Or at least add it to the `test.js` cases (same as shapely).

rowanwins:
FYI I've made a good start on this. Basically I'm looking at using the [OGC simple feature spec](https://portal.opengeospatial.org/files/?artifact_id=829).

I do however have a question about the most helpful API. So here are my options
1. I could do it as a boolean style module (eg simply return true or false for a feature validity). 
2. I could add properties to a feature showing which validity checks have passed or failed. The is obviously more useful in providing info (eg did the poly fail because start and end points weren't the same or because the polygon hole touched the exterior ring twice).
3. I could throw an error on the first error encountered with the details

Any other ideas on this one @DenisCarriere , @stebogit , @morganherlocker

stebogit:
[...]

PR #1302, opened 2018-03-08
Initial commit isValid module

rowanwins:
Please fill in this template.

Have made a start on an isValid module which basically check geometries against the [OGC simple feature spec](https://portal.opengeospatial.org/files/?artifact_id=829)

## To Do
- [ ] Multipolys that touch should pass - but if they overlap if should fail, I just need to work out the best way to run this check still
- [ ] Would be good to validate our results against [something like shapely](https://toblerity.org/shapely/manual.html#object.is_valid)
- [ ] Finalise desired API - see [discussion here](https://github.com/Turfjs/turf/issues/64#issuecomment-370112607)
- [ ] I'm also only tackling geometry validity, we could extend the module to check for things effective geojson validity like `properties`, or we could rename the module to `isGeometryValid`
- [x] Fix Typescript issues
- [x] Change Module name `boolean-is-valid` => `boolean-valid`


rowanwins:
Oh and I'm not entirely sure how we're supposed to be structuring modules anymore in terms of index.js vs index.ts vs whatever @DenisCarriere :)

DenisCarriere:
@rowanwins You're structuring correctly, you've just got errors.

packages/turf-boolean-is-valid/bench.js
```diff
@@ -0,0 +1,29 @@
+import path from 'path';
```
DenisCarriere: For now, we would need to convert the `bench.js` & `test.js` into CommonJS.

We "could" convert them to Typescript, however lets just focus on the `index.js` file for now and we can worry about the Bench & Test files later.

packages/turf-boolean-is-valid/index.js
```diff
@@ -0,0 +1,111 @@
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
```
DenisCarriere: We would be adding a `.gitignore` which includes `index.js` to prevent this file from being uploaded

packages/turf-boolean-is-valid/index.ts
```diff
@@ -0,0 +1,99 @@
+import { segmentEach } from '@turf/helpers';
```
DenisCarriere: `segmentEach` does not exist in `@turf/helpers` 🤦‍♂️

packages/turf-boolean-is-valid/index.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @name booleanIsValid
+ * @param {Geometry|Feature<any>} feature GeoJSON Feature or Geometry
+ * @returns {boolean} true/false
+ * @example
+ * var line = turf.lineString([[1, 1], [1, 2], [1, 3], [1, 4]]);
+ *
+ * turf.booleanIsValid(line, point);
+ * //=true
+ */
+export default function booleanIsValid(feature: Feature<any> | Geometry) {
```
DenisCarriere: `Feature` & `Geometry` needs to be imported from `@turf/helpers`

```js
import {Feature, Geometry} from '@turf/helpers';
```

packages/turf-boolean-is-valid/index.ts
```diff
@@ -0,0 +1,99 @@
...
+    const geom = getGeom(feature);
+    const type = getType(feature);
+    const coords = getCoords(feature);
+
+    switch (type) {
+    case 'Point':
+        return coords.length > 1;
+    case 'MultiPoint':
+        for (var i = 0; i < coords.length; i++) {
+            if (coords[i].length < 2) return false;
```
DenisCarriere: These switch cases get tricky... 🤔 Hard to properly handle these scenarios in a "strongly typed" way.

Easy solution would be to convert `coords` to `any`:

```
const coords: any = getCoords(feature);
```

DenisCarriere:
@rowanwins reference the module name, couldn't we call it `boolean-valid` since all boolean module is insinuating the word `boolean-is-<module>`.

We should rename this module as `boolean-valid`?

DenisCarriere:
👍 `boolean-valid-geometry` could also work.

We could expand valid modules to look something like this:

- `boolean-valid` (groups all submodules)
  - `boolean-valid-properties`
  - `boolean-valid-bbox`
  - `boolean-valid-geometry`
  - `boolean-valid-feature`

stebogit:
I like the idea @DenisCarriere :+1:, that makes the development easier and allow faster check if you know what you actually need.
The `boolean-valid` module though should validate the entire GeoJSON. I guess the average user would not really know or appreciate the difference between geometry, feature or geojson object, I bet they would expect to validate the whole object.

We could have a single module, like helpers, exposing different methods and the default one being `valid`(geojson):
```js
import valid from '@turf/boolean-valid';
import {validGeometry} from '@turf/boolean-valid';
```
Is that even possible?

DenisCarriere:
@stebogit Great suggestion, and YES 🚀 that is possible.

The great thing with ES Modules is  you can mix `named exports` & `default exports` (without being considered a "hack").

Here's how the code would look like when exporting your methods:

**setting up exports**
```ts
import validGeometry from "@turf/boolean-valid-geometry";

export default function booleanValid(feature) {
  // Main Boolean Valid method
}
// Export all Boolean Valid sub-modules
export { validGeometry }
```

**importing single valid module**

```ts
import valid from "@turf/boolean-valid"
```

**importing multiple valid module**

```ts
import {default as valid, validGeometry} from "@turf/boolean-valid"
```

CC: @rowanwins

DenisCarriere:
@rowanwins Going to merge this, however I won't publish it just yet until we figure out how we want to implement these methods.

danpe:
@DenisCarriere any news here? is there a plan to publish this feature? what's missing?

PR #1316, opened 2018-03-19
[turf-intersect] MultiPolygon input support

benjamintd:
👋 first time contributing to Turf!

This PR adds support for inputting multi-polygon features to turf-intersect, by running `intersect` on all sub-polygons that compose it.

This fixes the following issues: 
- https://github.com/Turfjs/turf-intersect/issues/13
- https://github.com/Turfjs/turf/issues/702

Additionally, it edits the README to delete the mention of returning Point, MultiPoint, LineString, or MultiLineString features when the polygons share a point or border. This is no longer the case since JSTS was replaced with `martinez-polygon-clipping`.


gajus:
Was this removed?

http://turfjs.org/docs#intersect

gajus:
This wasn't released yet.

gajus:
Whats the reason `@turf/intersect` v6 release exist and `@turf/@turf` v6 does not exist?

rowanwins:
Hi @gajus

Unfortunately v6 was released without a tag and so it should be thought of as an alpha release. It was pushed out as a move to make turf typescript-based but that's not going to eventuate now.

I'm currently working on v7 branch where I'll try to get this merged in.

Cheers

ChristopherTotty:
Would also really like to see this make it to a release soon.

PR #1327, opened 2018-03-23
Convert Quadrat to Typescript

DenisCarriere:
## Convert Quadrat to Typescript

CC: @zhuang-hao-ming

TurfJS is currently in transition from ES Module to Typescript, I've started refactoring some dependant modules of Quadrat, once those modules are refactored it will be a lot easier to use/refactor `Quadrat`.

**Dependant Modules refactored to Typescript:**

- [x] line-segment
- [x] line-intersect
- [x] polygon-to-line
- [x] boolean-disjoint
- [x] boolean-interect
- [x] square-grid
- [x] random
- [x] quadrat


DenisCarriere:
> I am advised to use a typescript style, instead of the old template?

Yes try to use the Typescript style, but this update is fairly new and still ongoing, so things might change later on.

Issue #1342, opened 2018-03-30
A new module directional mean statistic?


zhuang-hao-ming:
I don't seem to find the directional mean statistic module in turf.
If this module exists, please tell me.
The statistic calculate the average angle of the lines, measuring the trend of a set of line features.

If this module proves to be nonexistent, I consider implementing one. 😄

DenisCarriere:
I'm positive that this module does not exist, would be amazing to have one implemented 👍

PR #1343, opened 2018-04-01
New module: directional-mean.

zhuang-hao-ming:
#1342 

As stated in the issue. This module is used to calculate **average direction** of a set of lines.
At the same time, it also calculates another number **circular variance** to represent the variation in
direction between the lines.
The last point, this module supports both projection and geographic coordinates.

## example

The result is a line(the red one) whose length is the average length of all lines and whose center is the 
centroid of  the feature set.

directional_mean

directional_mean1

Please fill in this template.


packages/turf-directional-mean/index.js
```diff
@@ -0,0 +1,239 @@
+"use strict";
```
DenisCarriere: Sounds like we would now need to add `index.js` in the global `.gitignore`

DenisCarriere: I'll merge this PR and delete it afterwards since I can't commit to @zhuang-hao-ming 's `master` branch

zhuang-hao-ming: Haha, I'm sorry, I forgot it. 
Adding it to the global `.gitignore` file seems to affect other modules that have not been refactored yet.

DenisCarriere: I think right now we've got far enough with Typescript that we can't really go back, so every "patch" should also involve refactoring the Typescript before fixing the module.

PR #1382, opened 2018-05-04
Fix for along module to use rhumb calculations.

rowanwins:
Please fill in this template.

Reworking the along module to use the rhumb modules which work on straight lines, rather than great-circle lines, this provides more usable outputs on longer lines.

Also reworked the tests to be a bit more meaningful. Previous tests didn't actually test much at all, not sure how that happened 🤷


rowanwins:
Hey @stebogit 

Would you mind doing a quick review of these changes. 

Travis seems to be failing due to unrelated changes (I'm thinking npm versions or something effectiv decimal precision).

Thanks

packages/turf-along/test.js
```diff
@@ -17,15 +20,31 @@ test('turf-along', t => {
     const pt7 = along(line, 100, options);
     const pt8 = along(line.geometry, 0, options);
     const fc = featureCollection([pt1, pt2, pt3, pt4, pt5, pt6, pt7, pt8]);
-
+    
     fc.features.forEach(f => {
         t.ok(f);
         t.equal(f.type, 'Feature');
         t.equal(f.geometry.type, 'Point');
```
stebogit: why these checks? what else but a point could be returned by `along`?

rowanwins: They were the original tests - they are a bit pointless but I've left them

packages/turf-along/test.js
```diff
@@ -17,15 +20,31 @@ test('turf-along', t => {
...
     });
     t.equal(fc.features.length, 8);
```
stebogit: also this test seems quite meaningless to me

packages/turf-along/test.js
```diff
@@ -17,15 +20,31 @@ test('turf-along', t => {
...
-    t.equal(fc.features[7].geometry.coordinates[0], pt8.geometry.coordinates[0]);
-    t.equal(fc.features[7].geometry.coordinates[1], pt8.geometry.coordinates[1]);
+  
+    t.equal(isPointOnLine(fc.features[5].geometry.coordinates, line), true);
```
stebogit: Wouldn't this be the same, but more clean?
```js
fc.features.forEach(f => {
    t.true(isPointOnLine(f, line));
});
```

rowanwins: This is the more accurate check I'd like to move towards, but for some reason some of the points are falling off the line (by a minutia), I'm not sure if this is a rounding thing or what...

packages/turf-along/test.js
rowanwins: true :) - as above

packages/turf-along/test.js
```diff
@@ -17,15 +20,31 @@ test('turf-along', t => {
...
+    if (process.env.REGEN) write.sync(path.join(__dirname, 'test', 'fixtures', 'dc-points.geojson'), fc);
+    t.deepEqual(fc, load.sync(path.join(__dirname, 'test', 'fixtures', 'dc-points.geojson')));
 
     t.end();
 });
+
+const longLine = load.sync(path.join(__dirname, 'test', 'fixtures', 'long-line.geojson'));
+
+test('turf-along-long', t => {
+  const pt1 = along(longLine.features[0], 200, {units: 'miles'});
```
stebogit: If I'm right you draw a point on the line and then you want the calculated point along the line to match the one you set, correct?
If so I'd write it like this, to make it a little more obvious:
```js

test('turf-along-long', t => {
  const features = load.sync(path.join(__dirname, 'test', 'fixtures', 'long-line.geojson'));
  const longLine = features.features[0];
  const pointOnLongLine = longLine.features[1];

  const pt1 = along(longLine, 200, {units: 'miles'});
  t.deepEqual(pt1, pointOnLongLine);

  t.end();
});

File: packages/turf-along/test/fixtures/dc-points.geojson [hunk 5]
```diff
@@ -1,115 +1,93 @@
...
-        "coordinates": [
-          -77.02342987060545,
-          38.870587377511235
-        ]
-      }
-    }
-  ]
-}
\ No newline at end of file
+	"type": "FeatureCollection",
```
stebogit: not sure what's the style on `JSON` indentation, 2 or 4 spaces?

packages/turf-along/test.js
stebogit: is the test line a rhumb or a great-circle line where it fails? Maybe we are now facing the reverse problem we want to solve with this fix.
I was wondering if moving everything to rhumb functions would actually fail in those other cases where conventional (I don't have a better word for this) functions were actually ok.

rowanwins: Just double checked but the original approach also failed with the same test. I'm thinking it's something to do with floating point math that is causing us to fail on the test.

packages/turf-along/test/fixtures/dc-points.geojson
rowanwins: dont think we've ever cared have we?

rowanwins:
Dropping a note here that travis is failing because of a dev dependency on this module in `@turf/line-slice-along`. The `@turf/line-slice` tests are also failing for a reason less clear to me.

rowanwins:
Closing in favour of v7

tempranova:
I believe this should be re-opened since v7 was dropped!

the-nemz:
@rowanwins Is this still in the cards to be merged? It would be a big help to me!

twelch:
Drive-by comment. I think there is valid use case for great circle method, giving the shortest and most direct path on a sphere.  One thought is this rhumb version should simply become a new `rhumb-along` module?

smallsaucepan:
Drive by reply @twelch, longer term do we look at simply having @turf/along and passing a `mode` option which is either `great-circle` (default) or `rhumb-line`? Cuts down on package numbers and documentation, and fits in with our [2D] [3D] badge in the documentation discussion.

PR #1387, opened 2018-05-14
optimize ring area calculation

pirxpilot:
- unwind special (n - 2 and n - 1) cases to simplify the loop
- keep track lower, middle, and upper value in radians so that we don't need to recalculate them while traversing the ring

bench before:   polygon x 4,280,634 ops/sec ±1.53% (89 runs sampled)
bench after:      polygon x 5,810,796 ops/sec ±1.09% (87 runs sampled)

### DIFF
--- a/packages/turf-area/index.ts
+++ b/packages/turf-area/index.ts
@@ -76,41 +76,50 @@ function polygonArea(coords: any) {
  * @param {Array<Array<number>>} coords Ring Coordinates
  * @returns {number} The approximate signed geodesic area of the polygon in square meters.
  */
+
+const FACTOR = RADIUS * RADIUS / 2;
+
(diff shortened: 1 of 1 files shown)


rowanwins:
Hi @pirxpilot 

Sorry its taken a long time for someone to review this. 

Looking at the Travis test result the change to this module is causing issues in `@turf/isobands`.  I suspect we're seeing a change to decimal precision. Because this module has only 1 test itself and it's rounding the result to the nearest integer it's not exactly a super-thorough test.

I'll need to do a bit more in-depth testing on isobands to make sure it's all working out as expected.

smallsaucepan:
Contents of this PR were incorporated and merged as part of #2520. Thanks for your contribution @pirxpilot!

PR #1401, opened 2018-06-05
1400: Fix @turf/intersect, @turf/difference, and @turf/union import of martinez-polygon-clipping

### LINKED ISSUE #1394: Bug: turf/intersect 6.1.2 fails due to martinez.intersection returning invalid polygon
Depending on the order the polygons are provided to the intersection check, martinez.intersection may return a list of coordinates that is not a valid polygon. (Occurs here https://github.com/Turfjs/turf/blob/master/packages/turf-intersect/index.ts#L52)

Here is a gist demonstrating the behavior: https://gist.github.com/PizzaBrandon/1d897159017fc1654b061611d408ad46

While I believe martinez shouldn't return a bad polygon, turf should filter this before it causes an exception in the intersect method.

### LINKED ISSUE #1400: Error after building @turf/intersect 6.1.2 with webpack
After building @turf/intersect with webpack 3 (using default settings provided by `create-react-app`), intersect does not run due to a problem with turf's import of `martinez-polygon-clipping`:

```
TypeError: martinez.intersection is not a function
    at intersect (http://localhost:3000/static/js/bundle.js:1538:37)
    at Object../src/App.js (http://localhost:3000/static/js/bundle.js:39523:72)
    at __webpack_require__ (http://localhost:3000/static/js/bundle.js:679:30)
    at fn (http://localhost:3000/static/js/bundle.js:89:20)
    at Object../src/index.js (http://localhost:3000/static/js/bundle.js:39638:63)
    at __webpack_require__ (http://localhost:3000/static/js/bundle.js:679:30)
    at fn (http://localhost:3000/static/js/bundle.js:89:20)
    at Object.0 (http://localhost:3000/static/js/bundle.js:39664:18)
    at __webpack_require__ (http://localhost:3000/static/js/bundle.js:679:30)
    at ./node_modules/@turf/helpers/index.js.Object.defineProperty.value (http://localhost:3000/static/js/bundle.js:725:37)
    at http://localhost:3000/static/js/bundle.js:728:10
```

I built a minimal demonstration for this bug: https://github.com/PizzaBrandon/bug-reports/tree/master/turf-intersect-6-1-2

I am pulling down the turf repo now to try to address this, but I wanted to get visibility around this bug now.

Other packages may also be affected by this and I'll update the linked demo as they are discovered.

PizzaBrandon:
Fixes #1400 and #1394 

Starting in 0.4.x, `martinez-polygon-clipping` exports `default`. Replacing `import * as martinez` corrects undesired behavior in built downstream packages where webpack would incorrectly strip away supporting code. This also fixes #1394 by importing 0.4.x of martinez-polygon-clipping.

All tests pass and I have confirmed through `npm link` that the linked issue goes away with this change and importing the newest `martinez-polygon-clipping`.

Related landed PR in martinez: https://github.com/w8r/martinez/pull/82


PizzaBrandon:
Well, that was awkward. I'll figure out how to make yarn pick up the latest version of martinez

PizzaBrandon:
Once this downstream PR is merged (https://github.com/w8r/martinez/pull/83), I will rebuild this PR.

PizzaBrandon:
Sorry for the delay. The PR is now ready and affects `@turf/difference`, `@turf/intersect`, and `@turf/union` as they all import `martinez-polygon-clipping`.

rowanwins:
Hey @PizzaBrandon 

Thanks for taking a look at this and resolving some things upstream in martinez. I'll probably take a full look at this tomorrow night as it'll require a bit of looking at and publishing a few modules.

Cheers

rowanwins:
Hi @PizzaBrandon 

I see some reference in your commit titles that you've updated the Martinez dependency to v0.4.3 - that should be reflected in the `package.json` file for each of the turf modules, but it isn't currently. If you could add that in that would be fab.

PizzaBrandon:
They presently all are set to `*` version, so they should pick up this change.

rowanwins:
Well there you go - didn't know that could even happen! I'll probably put in a minimum version number anyway because I know the more recent releases of martinez are a lot more stable, just to avoid any hiccups.

Thanks again for your work on this one.

PizzaBrandon:
Do you want me to make the minimum version number change? Is there anything else you need to be able to merge and publish?

rmmlopes:
Any idea when this fix will get into a release?

rowanwins:
Hi @rmmlopes 

Soon hopefully, I'm just awaiting a bit of advice regarding the publishing workflow.

Issue #1403, opened 2018-06-07
Support rectangleGrid


stevage:
I'd like to add a new module, rectangleGrid, which is essentially squareGrid without the constraint that height=width. One use case is drone imagery planning, where the rectangular aspect ratio of the camera defines the region to be captured.

Proposed usage:

```
var rectangleGrid = turf.rectangleGrid(bbox, cellWidth, cellHeight, options);
```

Sound good?

(The development time is sponsored by Prairie Apps.)

rowanwins:
Hey @stevage 

@adam3039 previously raised the same issue[here](https://github.com/Turfjs/turf/issues/1251), there was some work done [here](https://github.com/adam3039/rectangle-grid) and [here](https://github.com/adam3039/turf) so might be worth asking how he's going.

Unfortunately things are a bit undocumented at the moment in terms of the typescript transition although hopefully you can follow one of the existing packages (eg [here](https://github.com/Turfjs/turf/tree/master/packages/turf-along)), if not just go with traditional js and one of the contributors can convert it for you.

Thanks to Prairie Apps (whoever you are!)

stevage:
Hmm, one issue I've just encountered (which affects `squareGrid` as well) is where to apply the `units` option. `squareGrid` applies it along the western and southern border. That means that if you ask for rectangles that are, say, 500 miles wide by 100 miles high, for an area that spans from 80 degrees north to 80 degrees south, you'll get rectangles that are much wider than that at the equator.

Obviously, there's something a bit nonsensical about that request, but I feel that converting the units at the centre of the bounding box is more intuitive than at the west and south edges. But then it wouldn't be consistent with squareGrid anymore.

At the very least, this should be documented I think.

Thoughts?

adam3039:
@rowanwins  I'm the owner of Prairie Applications. I couldn't get through the contribution side of things so I reached out to @stevage for a hand. We're in need of this module, so I'm happy to contribute it to turf, it's a fantastic framework that I'd like to support. Thanks for all your work.

Cheers

Issue #1251, opened 2018-02-05
Equivalent to squareGrid for rectangles


adam3039:
I need to split a bounding box into rectangles (specify a length & width of the grid cell). Is there any way to achieve this with existing tools, or potentially a feature that could be added.

rowanwins:
Hi @adam3039 

We dont currently have this functionality although it would probably be relatively simple to implement. 

If you'd like to have a go at implementing it take a look at the source code from the [square-grid](https://github.com/Turfjs/turf/blob/master/packages/turf-square-grid/index.js) module and that should get you on your way, we'd happily accept a pull request for a new module. Otherwise it'll probably be a little while before any of the core team have any time to take a look at it.

Cheers
Rowan

rowanwins:
Closing in favour of #1403 and #1405

PR #1405, opened 2018-06-07
New module: rectangleGrid

stevage:
- [?] Run `npm run lint` to ensure code style at the turf module level.

`npm run lint` doesn't seem to do anything useful at the moment.

See #1403 and #1251.


rowanwins:
Thanks @stevage I'll try and take a look at this over the weekend.

stebogit:
@stevage this package is just an exact copy of `@turf/square-grid`, except the additional parameter for the second dimension of the cell.
Since a square is actually a special case of a rectangle I'd suggest to modify the current `square-grid` package to include the additional param, add a few test for the new features and rename it `@turf/rectangle-grid`.

Then create a new package for `@turf/square-grid` as a simple alias for the specific case:
```js
export default function squareGrid(bbox, cellSize, options) {
    return turf.rectangleGrid(bbox, cellSize, cellSize, options);
}
```

That should save a lot of dev time and avoid redundant code.

@rowanwins @DenisCarriere what do you think?

rowanwins:
You're a smarty pants @stebogit - that does sound like a sensible approach as it would keep the modules in sync if any improvements/fixes need to be made.

PS good to see you :)

stevage:
Yeah, I like that approach - but it definitely doesn't "save a lot of dev time" when the PR is already submitted :)

Anyway I'll have a go.

stevage:
Ok, did I do it right @stebogit ?

packages/turf-rectangle-grid/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/Turfjs/turf.git"
+  },
+  "keywords": [
+    "turf",
+    "grid",
+    "regular",
+    "cartesian",
+    "grid"
```
stebogit: this is duplicated

packages/turf-square-grid/index.d.ts
```diff
@@ -22,9 +21,9 @@ import { BBox, Feature, FeatureCollection, MultiPolygon, Polygon, Properties, Un
  * //addToMap
  * var addToMap = [squareGrid]
  */
-declare function squareGrid<P = Properties>(bbox: BBox, cellSide: number, options?: {
+import { FeatureCollection, Polygon, BBox, Units, Feature, MultiPolygon, Properties } from "@turf/helpers";
```
stebogit: the import should go at the very top of the file

packages/turf-square-grid/index.ts
```diff
@@ -29,56 +21,19 @@ import {getType} from "@turf/invariant";
...
-    const cellWidth = xFraction * (east - west);
-    const yFraction = cellSide / (distance([west, south], [west, north], options));
-    const cellHeight = yFraction * (north - south);
-
-    // rows & columns
-    const bboxWidth = (east - west);
-    const bboxHeight = (north - south);
-    const columns = Math.floor(bboxWidth / cellWidth);
-    const rows = Math.floor(bboxHeight / cellHeight);
+import {
```
stebogit: this should go at the top, same as above, just for consistency within the library

packages/turf-square-grid/package.json
```diff
@@ -43,9 +43,6 @@
     "@types/tape": "*"
   },
   "dependencies": {
-    "@turf/boolean-intersects": "6.x",
-    "@turf/distance": "6.x",
-    "@turf/helpers": "6.x",
```
stebogit: isn't `@turf/helpers` still required here for types?

packages/turf-rectangle-grid/LICENSE
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
+
+Copyright (c) 2017 TurfJS
```
stebogit: I guess this should be 2018

packages/turf-rectangle-grid/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "@turf/truncate": "*",
+    "@std/esm": "*",
+    "benchmark": "*",
+    "rollup": "*",
+    "write-json-file": "*",
+    "load-json-file": "*",
+    "tape": "*"
+  },
+  "dependencies": {
+    "@turf/helpers": "*"
```
stebogit: `@turf/boolean-intersects`, `@turf/distance` and `@turf/invariant` are missing

packages/turf-rectangle-grid/LICENSE
stevage: Done

packages/turf-rectangle-grid/package.json
stevage: Done

rowanwins:
Have just pushed some changes here @stevage & @stebogit to get travis to be happy

- Currently the travis tests fail if the licence date is set to anything other than 2017
- I think the addition to the turf master package was also giving it grief so I've temporarily removed it
  - I need to work out how this part of the process was actually being managed

packages/turf-square-grid/package.json
rowanwins: Yeah from what I can tell `@turf/helpers` now gets brought in pretty much everywhere to help with the typescript side of things.

packages/turf-rectangle-grid/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "main.js",
+    "main.es.js"
+  ],
+  "scripts": {
+    "pretest": "rollup -c ../../rollup.config.js",
+    "test": "node -r @std/esm test.js",
+    "posttest": "node -r @std/esm ../../scripts/validate-es5-dependencies.js",
+    "bench": "node -r @std/esm bench.js",
+    "docs": "node ../../scripts/generate-readmes"
+  },
```
rowanwins: So I think these scripts need changing to match one of the other typescript modules (eg turf-random) just above it.
- Then I think the main.js and the main.es.js can be deleted and the `@std/esm` part of the package.json

packages/turf-square-grid/package.json
rowanwins: Same story in this package.json as above - we need to make it align to the TS world view.

packages/turf-rectangle-grid/package.json
stevage: Ok, done. I'm not quite sure why my generated index.js looks different from the one that was already in this branch. Hopefully not a problem?

stevage:
Just checking in, I *think* I have addressed all the outstanding issues, but want to make sure you're not waiting on me to do something.

rowanwins:
Hi @stevage yep I think it's all good. 

I'll accept the pull request and attempt to publish sometime over the next few nights. The publishing process with [lerna](https://github.com/lerna/lerna) is still a bit mysterious to me so I'm just trying to avoid breaking everything :)

stevage:
Oh awesome 👍

adam3039:
Hey @rowanwins just curious if there had been any headway made on a publish of this latest module. I'm having difficulty bringing in the module from just the github repo alone. It looks like the main turf package in packages/turf doesn't point to the rectangle-grid module at all. Possibly something was missed in the pull request, or I could very well be off my rocker as well. I've tried a few different ways to get it working, but so far only using the official releases from NPM have worked for me.

Thanks for all your help with this.

stevage:
>It looks like the main turf package in packages/turf doesn't point to the rectangle-grid module at all.

That's a problem that has to do with a circular dependency: it seems to be hard to both add the module, and have the main module depend on it, at the same time.

Are you able to make that change locally?

yuqianma:
Currently there's no `rectangleGrid ` in `@turf/turf`. Is it intended?

PR #1472, opened 2018-08-30
Issue-1469: Fixes incorrect implementation for line crosses line

chriserickson:
This fixes Issue-1469, where booleanCrosses did not correctly identify two lines which cross. This was due to some overly aggressive checking for endpoint-on-line, which I re-implemented.

### DIFF
--- a/src/boolean-crosses/index.js
+++ b/src/boolean-crosses/index.js
@@ -1,7 +1,7 @@
 import lineIntersect from '../line-intersect';
 import { polygonToLine } from '../polygon-to-line';
 import booleanPointInPolygon from '../boolean-point-in-polygon';
-import { getGeom, getType } from '../invariant';
+import { getGeom } from '../invariant';
 import { point } from '../helpers';
 
 /**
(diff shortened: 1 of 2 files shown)


src/boolean-crosses/index.js
```diff
@@ -86,21 +86,19 @@ function doMultiPointAndLineStringCross(multiPoint, lineString) {
...
-    return false;
+    const intersections = lineIntersect(lineString1, lineString2);
+
+    const endpoints = [
+        lineString1.coordinates[0], lineString1.coordinates[lineString1.coordinates.length - 1],
+        lineString2.coordinates[0], lineString2.coordinates[lineString2.coordinates.length - 1]
+    ]
+
+    // A line that touches another line should not cross
+    var intersectionsWithoutEndpoints = intersections.features.filter(function (feature) {
```
rowanwins: `filter` functions are generally quite slow @chriserickson (as are many of those es5 array functions like `forEach` and `map`) so generally we try to avoid them if possible.

chriserickson: @rowanwins: In this case the filter is quite overly aggressive too -- we don't need the filtered results.
How do you feel about `array::some()`? According to at least one source (http://www.andygup.net/fastest-way-to-find-an-item-in-a-javascript-array/) it should be plenty fast.

rowanwins:
Thanks for this @chriserickson and for adding the new test case!

jcready:
Can this be back-ported to the current major release version? This is a pretty major bug to put off for two major versions.

mast4461:
Yesterday I too fought this bug for about an hour, before resorting to using `turf.lineSplit` instead:
```
function booleanCrosses(line1, line2) {
  const split = turf.lineSplit(line1, line2);
  return split.features.length > 0;
}
```

I too would be greatful if this fix was released soon. Especially seeing as it's been _quite some time_ since it was merged.

Issue #1587, opened 2019-02-14
The last point of the coordinates returned by lineArc() is not match ‘bearing2‘.


huanglii:
I used the **`lineArc()`** to draw arcs, But the last point of the coordinates returned by **`lineArc()`** is not match **`bearing2`**. If increase the param which named **`steps`**, it will be better. 

With my expectation, the first point and the last point should match the **`bearing1`** and the **`bearing2`**.

[The Sample Code by JSFiddle](https://jsfiddle.net/95j1o2dq/)
![image](https://user-images.githubusercontent.com/20068340/52792774-292d0a00-30a7-11e9-9ad4-ebe6df2e4cbe.png)

huanglii:
the line 52, `360` should be changed to `Math.abs(angle2 - angle1)`
`alfa = arcStartDegree + i * 360 / steps;`

huanglii:
the line 52, `360` should be changed to `(Math.abs(arcEndDegree - arcStartDegree))`
`alfa = arcStartDegree + i * 360 / steps;`

huanglii:
This issue is fired when '(bearing2 - bearing1) % (360 / steps) == 0'. At this time, alfa will eaual to arcEndDegree, so the last point is lost.
from [pull request 1590](https://github.com/Turfjs/turf/pull/1590)

vwong:
I'm just stumbling across this bug now, and independently (before I saw this issue), deduced the same fix. What's stopping this from being merged?

huanglii:
Sorry, Did not reply for a long time, thought the submission was wrong.
A new pr([line-arc fix. Resolves #1587](https://github.com/Turfjs/turf/pull/1778)) was opened.

PR #1588, opened 2019-02-14
fix issue 1587

huanglii:
Ensuring the first point and the last point match the bearing1 and the bearing2. [#1587](https://github.com/Turfjs/turf/issues/1587)

### DIFF
--- a/packages/turf-line-arc/index.js
+++ b/packages/turf-line-arc/index.js
@@ -46,10 +46,10 @@ function lineArc(center, radius, bearing1, bearing2, options) {
     var alfa = arcStartDegree;
     var coordinates = [];
     var i = 0;
-    while (alfa < arcEndDegree) {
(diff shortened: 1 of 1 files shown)


stebogit:
@huanglii would you please add also a test case to confirm this PR does fix #1587? You can use the same example you submitted in the issue.

huanglii:
[test case](https://jsfiddle.net/k5a90vgn/)
image
image

huanglii:
:smile:   **`Math.abs()`** is not necessary.

stebogit:
@huanglii what I meant is, add the test case to the [`test`](https://github.com/Turfjs/turf/tree/master/packages/turf-line-arc/test) folder of the module and make sure all test pass

huanglii:
@stebogit Resubmitted [PR](https://github.com/Turfjs/turf/pull/1590)

PR #1609, opened 2019-03-06
turf-projection: Deactivate longitude wrapping for `toMercator`

### LINKED ISSUE #1602: toMercator changes polygons geometry
When using toMercator on a polygon that overlap the -180 or 180 longitude, without antimeridian cutting.

The current implementation wraps every point one by one which ends up returning a wrong polygon:

Here is a polygon overlapping the -180 longitude:
- https://gist.github.com/Darune/d73952fbe586d1e8116d6e779042ccb3
Here is the result of a `toMercator` transformation reconverted using `toWgs84`:
- https://gist.github.com/Darune/dbd604f5b697cbbd36b918acab2dea94

code to reproduce the issue:

```javascript
import { toWgs84, toMercator } from '@turf/projection';

toWgs84(toMercator({
  "type": "Feature",
  "properties": {}, 
  "geometry": {
    "type": "Polygon",
    "coordinates": [[
        [-181, 60],
        [-181, 61],
        [-179, 61],
        [-179, 60],
        [-181, 60]
    ]]
  }
}));
```

I think that when calling `toMercator` with a GeoJson, turf should not try to wrap longitude as it does.
if this is an acceptable behavior I could submit a PR

Darune:
If a geometry is crossing the 180th meridian we should not wrap its longitude, this would break the geometry of the polygon

this resolves #1602

Please fill in this template.


packages/turf-projection/index.ts
```diff
@@ -82,25 +82,32 @@ function convert(geojson: any, projection: string, options: {mutate?: boolean} =
  *
  * @private
  * @param {Array<number>} lonLat WGS84 point
+ * @param {Object} [options] Optional parameters
+ * @param {boolean} [options.wrapLongitude=true] wrap logitude passing the meridian to a coordinated between -180 and 180
  * @returns {Array<number>} Mercator [x, y] point
  */
-function convertToMercator(lonLat) {
+function convertToMercator(lonLat, options: {wrapLongitude?: boolean} = {wrapLongitude: true}) {
```
stebogit: I'm not sure I understand why you added this optional parameter if the function is only used internally (it's not exported).
besides, I was under the impression wrapping the coordinates was actually the original issue, so why keeping the option... 🤔

packages/turf-projection/test.js
```diff
@@ -50,23 +41,23 @@ const fromMercator = fs.readdirSync(directories.mercator).map(filename => {
...
         t.deepEqual(results, load.sync(directories.out + 'wgs84-' + filename));
     }
     t.end();
 });
 
-
+test('projection -- Round Trips', t => {
+    for (const {geojson}  of fromWgs84) {
+        const source = truncate(geojson);
+        const results = truncate(toWgs84(toMercator(source)));
```
stebogit: 👍

packages/turf-projection/test.js
```diff
@@ -25,17 +23,10 @@ const fromWgs84 = fs.readdirSync(directories.wgs84).map(filename => {
 });
 
 test('to-mercator', t => {
-    for (const {filename, name, geojson}  of fromWgs84) {
-        var expected = clone(geojson);
-        coordEach(expected, function (coord) {
-            var newCoord = proj4('WGS84', 'EPSG:900913', coord);
```
stebogit: are you sure `proj4` returns an incorrect result? if so, would you mind sharing an example?

packages/turf-projection/test/out/mercator-fiji.geojson
```diff
@@ -6,7 +6,7 @@
 		"coordinates": [
 			[
 				[
-					-20031393.340521,
+					20043623.345058,
```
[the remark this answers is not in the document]
Darune: to generate the expected coordinate I used ogr2ogr: 
`ogr2ogr  -f "GeoJSON" /vsistdout/ -t_srs epsg:900913 -s_srs epsg:4326 test/wgs84/polygon.geojson | jq '.features'` for most cases exept `geometry.geojson` where the jq exepression would be `jq '.features[0].geometry'`

I didn't find any online tool to visualize the coordinates so i copy pasted the geojson example from openlayer and put it in a jsfiddle here:

https://jsfiddle.net/z1e5bqLx/1/

packages/turf-projection/index.ts
Darune: In my opinion even it polygons must be allowed to overlap the antimeridian / meridian, for points, I think we would prefer them to always be wrapped. 
So I set the wrapLongitude to true only for geojson, if someone sends a `[lng, lat]` array we keep the longitude wrapping as it would be the behavior i would be expecting.

What do you think ?

packages/turf-projection/test.js
Darune: `proj4.js` returns correct results for point and polygons not crossing the antimeridian.

`proj4` provides a way to disable longitude wrapping with the `+over` flag described here https://proj4.org/usage/projections.html

However i didn't manage to make it work using `proj4.js`

packages/turf-projection/index.ts
```diff
@@ -82,25 +82,32 @@ function convert(geojson: any, projection: string, options: {mutate?: boolean} =
...
+    options = options || {wrapLongitude: true};
```
stebogit: I'm not really familiar with Typescript, but I think this line is unnecessary as `options` has already a default value in the signature

Darune: not familiar with typescript, 
I saw the same code for the options of `toMercator` / `toWgs84`,
here: 
https://github.com/Turfjs/turf/blob/master/packages/turf-projection/index.ts#L57
 
I guessed it was to handle the case where `options` is explicitly set to `null`.

packages/turf-projection/test.js
stebogit: Ah, looks like it's not supported yet: https://github.com/proj4js/proj4js/issues/180.

packages/turf-projection/index.ts
Darune: I'm not sure I can express the same thing with a optional parameter on the `toMercator` / `toWgs84` functions. 

Both method takes either coordinates array or geojson as parameter, the default value we want is different depending on the nature of the parameter type, would how would you name such parameter ?

stebogit: I don't think we want a different default value, and so behavior, depending solely on the type of input, rather just giving to the user the option to decide.

I believe the expected result in Turf would be not having wrapped coords (basically allow exceeding the -180/180 range). That's why the [original issue](#1602) is considered a bug.

stebogit: BTW the parameter `options: {wrapLongitude?: boolean} ` I think is very appropriate also for those functions

packages/turf-projection/test/out/mercator-fiji.geojson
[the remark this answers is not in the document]
Darune: not familliar with the package, but it seems like a wrapper around gdal's ogr2ogr, it's easy to use it like `proj4` was used.
However keep in mind that `gdal` is not a funny dependency to install on different `OS` and adding this dependency to the test would require gdal to be install, maybe i could create another command like : 

`npm generate-toMercator-tests`

`npm generate-toWgs84-tests`

I'm not sure if org2org npm package checks for `gdal` at installation time or at runtime, but if it checks it at runtime that would make `gdal` required only for those command and not at installation time.

I just want you to be aware about the implications of depending on `gdal`

packages/turf-projection/index.ts
Darune: ok then, i'll add the option, and the tests

packages/turf-projection/test/out/mercator-fiji.geojson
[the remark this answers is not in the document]
Darune: So do you want to depend on `gdal` ? 
Should i create additional tests generation npm `scripts` ?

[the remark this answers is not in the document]
stebogit: I don't think it's a good idea, it might become a blocker for possible contributors...
Why not just restoring the `proj4` testing code and simply add a separate test suite for the crossing 180 meridian case using the files you generated with `org2org`? With a proper comment/documentation of why that is necessary.

@morganherlocker @rowanwins what do you think?

smallsaucepan:
Hi @Darune and @stebogit. Did this PR settle on an approach? Is it still worth pursuing?

stebogit:
@smallsaucepan I honestly don't remember the details at this point.
The original idea I think still makes sense, and I think the only concern I had was regarding the test results - generated with an external tool I believe - which were not easy to visualize, thus not obvious to verify. That is, next time one of those tests breaks it would be challenging to understand what exactly happened and then how to fix it generating a new valid output.
But I might be wrong 😅

smallsaucepan:
Thanks @stebogit, and you too @Darune. I think we'll close this PR and issue. If a pressing need arises we can reopen and revisit.

PR #1714, opened 2019-06-17
Polygonize handle isValid and duplicate points/edges

Traksewt:
Polygonize previously had a TODO for isValid (for EdgeRing) and was hardcoded to return true. The original implementation that this was ported from expected the EdgeRings to have the same start and end point, meaning that the minimum number of vertices for an edge ring would be 4 (as point 1 == point 4). However, this implementation does not add the joining last vertex on the edge ring - it adds it as needed when converting to a polygon. This means that the minimum number of points for a edgering is 3 (forming a triangular polygon). 

Polygonize was getting corrupted if the same vertex was getting reused within an edgering. As soon as you encounter a same vertex within an edgering, that edgering can be considered formed and the looping over edges should be aborted. This will allow other joining edges the chance to form other valid edgerings.

Polygonize was not handling when there were multiple connections within the graph between the same nodes. This is fixed by ensuring each edge in the graph is unique. 

Polygonize was too trigger happy culling existing vertices by only considering the inner edges when looking for dangles. This did not consider the case for a single circle of edges, where in this case each vertex only has 1 inner edge - making it a candidate to be removed using the existing algorithm. The original algorithm pre-port was looking for nodes with degree 1. A node (as a part of a linear line) with 1 inner and 1 outer is not degree 1 yet it was getting culled by the existing algorithm. By checking the outer edges as well and only culling if inner edges <= 1 and outer edges == 0, it will avoid the case where this node is pointing to another node. 

Please fill in this template.


mfedderly:
@Traksewt are you still interested in working on this PR since #3045 has been merged recently? The `isValid` stub is still there, and it sounds like there might be a few other changes from this worth bringing forward? Happy to look at new PRs or a rebased version of this one, whatever makes sense.

PR #1768, opened 2019-10-10
Fix undeclared variable i in area

ahopkins:
I was trying to import this into a module and calculate some areas. Because strict mode is on, I am getting:

```
ReferenceError: assignment to undeclared variable i
```

This is because there is no variable declaration inside the `for` loop.

### DIFF
--- a/packages/turf-area/index.ts
+++ b/packages/turf-area/index.ts
@@ -39,7 +39,7 @@ function calculateArea(geom: Geometry): number {
     case "Polygon":
         return polygonArea(geom.coordinates);
     case "MultiPolygon":
-        for (i = 0; i < geom.coordinates.length; i++) {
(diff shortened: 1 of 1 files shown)


mfedderly:
This failed CI because `i` was being redefined. If you look above there's a `let i` declaration above both of the places you changed. What version of @turf/turf-area were you having an issue with?

ahopkins:
`3.0.14`

ahopkins:
Which.... now that I am looking at `package.json` and seeing that it looks incredibly old compared to what is on NPM, I am super confused how `yarn add turf` would have given me this old version. I will close this and upgrade my version.

rowanwins:
@ahopkins - you ought to `yarn install @turf/turf`. 

Unfortunantly `yarn install turf` still works but it points to last version prior to turf as a monorepo.

PR #1771, opened 2019-10-14
fix: lineSlice api: when stopPt equals to line's last position, the api return two same coodinates

zt123123:
fix: when stopPt equals to ends[1], return two same coodinates


mfedderly:
@zt123123 can you explain more about what this PR is supposed to fix? Thanks!

PR #1794, opened 2019-11-21
WIP: turf-cluster-kmeans deduplicate initial centroids

Crisfole:
This is a WIP. I was unable to get the development environment working correctly (master doesn't appear to pass pre-lint). Thus the empty checkboxes below:

Currently if you run kmeans clustering on a series of points with duplicates in the first NUMBER_OF_CLUSTERS coordinates it will fail to produce the right number of clusters; the algorithm ignores half of the duplicates.

There are a two ways I can think of to fix this:

 1. Shift each point by a tiny amount in an arbitrary direction. (.5% of the bbox around the points?) This will reduce the changes of collision dramatically, but I am guessing would be suboptimal for performance).
 1. Keep picking new points until the initial points are unique.

I implemented option 2 which provided a more natural way to determine the actual number of clusters, so I was able to remove some numberOfClusters code and replace it with simply 'Number of clusters found'.

It is possible that there is a better solution to this (from within the algorithm itself). I didn't dig any deeper.

### DIFF
--- a/packages/turf-clusters-kmeans/index.ts
+++ b/packages/turf-clusters-kmeans/index.ts
@@ -1,11 +1,22 @@
 import clone from '@turf/clone';
 import { coordAll, featureEach } from '@turf/meta';
-import { FeatureCollection, Feature, Point, Properties } from '@turf/helpers';
+import { FeatureCollection, Coord, Point, Properties } from '@turf/helpers';
 import * as skmeans from 'skmeans';
 
 export interface KmeansProps extends Properties {
     cluster?: number;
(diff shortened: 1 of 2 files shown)


Crisfole:
I did some more testing and I'm not 100% positive this is correct...I'll need to look into my particular case a bit harder...It worked correctly with random points...but not in my particular case.

Crisfole:
I'm going to close this. I'm solving by simply not using turf's cluster-kmeans module...don't have the time/space to look deeper.

PR #1801, opened 2019-12-03
use npm version of marchingsquares (isobands/isolines)

### LINKED ISSUE #1797: isobands fails on flat data
We are hitting this in a map @superchad and I are developing. When all the points in the grid fall in the same band then `isobands` returns no geometry instead of the bbox of the grid. If one point on the edge has a different band than the hole area is filled in correctly. Instead of a geojson I have a failing test https://github.com/Turfjs/turf/compare/v7...Eh2406:v7

Eh2406:
This closes #1797 by moving to the maintained version of marchingsquares on npm.

The test are updated to match the new observed behavior. Is there a good way to recheck that it is still the correct answer?


Eh2406:
Look like the CI failure is unrelated.

smallsaucepan:
Thanks for submitting this @Eh2406. Have reimplemented the marching squares integration and merged this as PR #2527. Will close this PR in favour of that one.

Though if you're still around are you able to run us through the background of the clean up to orderByArea?

Eh2406:
I am still around, and happy to answer questions. I do not remember the context. Do you have any links that might jar my memory?

smallsaucepan:
Thankyou! It's this commit "small clean up to orderByArea": https://github.com/Turfjs/turf/pull/1801/commits/88b159a69a3ab7678eb0bc98e21a599d18b27db7

Was that only to map-ify the code and make it more readable? Or was it to fix a bug? Looks like the old code might not have actually been comparing the area values?

Eh2406:
I think they are both correct. The old code did compare values, in the sort. The old code is `O(n^2)` and the new code is `O(N*ln(N))`, but I don't know if it matters for the size input.

PR #1813, opened 2020-01-10
union (v7): Change 'fc' to rest argument 'features'

mbullington:
When upgrading from turf v6 to v7 (since union uses the more stable union/difference library), we ran into this issue and it was not immediately clear what the issue was.

I think it makes sense to support both styles, especially since `difference` is hard coded to `a` and `b`.

What the PR makes possible / tries to accomplish:

```js
import { union } from 'turf';

union(feature1, feature2, fc);
```


mbullington:
Closing this up as TURF is going in another direction in regards to v6 / v7.

Cheers!

PR #1844, opened 2020-03-09
Update from @std/esm to esm

mfedderly:
The first in a series of maintenance-related PRs.

- `@std/esm` has been deprecated in favor of `esm`, this fixes building on newer node versions
- Moved the `esm` dependency from every package.json into the root package.json
- Small extra commits for a yarn.lock inconsistency and whitespace change

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "documentation": "*",
     "eslint": "*",
     "eslint-config-mourner": "*",
+    "esm": "^3.2.25",
     "fs-extra": "*",
     "lerna": "2.8.0",
     "load-json-file": "*",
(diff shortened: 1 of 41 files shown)


mfedderly:
@rowanwins for visibility

mfedderly:
This basically comes down to whether test.js and bench.js are written with esm or not (which differs per package). Some packages have it and some don't. I did a project-wide find and replace to find all @std/esm references and then you can also verify that all of the references are removed because it also disappeared from the yarn.lock. There is also an additional layer of safety because the tests still run.

mfedderly:
@rowanwins I see that you +1'd my last comment, but didn't approve the PR. I don't actually have permissions to merge my own commits. Can you either approve the PR or grant me more permissions? Thanks!

PR #1849, opened 2020-03-12
Standardize Typescript builds

### LINKED ISSUE #1390: Missing typescript definitions for center, projection
Some packages such as `@turf/center` or `@turf/projection` at the newest version do not have any typescript typings anymore, while others do (e.g. `@turf/explode`). The package.json even references non-existent file `index.d.ts`

### LINKED ISSUE #1465: Typescript definition is missing with `boolean-within`
Hi,
I'm using node v10.6.0 and npm 6.4.0.

I'm adding `boolean-within` with `$ npm i --save @turf/boolean-within`
and the definition file `index.d.ts` is not downloaded, while in other module it is.

Step to reproduce:

```
$ mkdir /tmp/test
$ cd /tmp/test
$ npm i @turf/boolean-within
$ ls node_modules/@turf/boolean-within
# see that index.d.ts is missing, try another package installed as dep
$ ls node_modules/@turf/boolean-point-in-polygon
# see index.d.ts correctly
```

### LINKED ISSUE #1493: Missing index.d.ts file in distance package
The [@turf/distance](https://github.com/Turfjs/turf/tree/master/packages/turf-distance) package has an `index.ts` file, but not an `index.d.ts`. Therefore, when I use it in my Typescript project and try to compile, I get:

```
Could not find a declaration file for module '@turf/distance'.
```

...because I have `"noImplicitAny": true` set in my project's package.json.

It looks like other packages (such as [@turf/bbox](https://github.com/Turfjs/turf/tree/master/packages/turf-bbox)) **do** have `index.d.ts` files, and I would rather not have to loosen my project configuration to be able to use this package.

**Additional context**
- Some guidelines on publishing types for turf packages: https://github.com/Turfjs/turf/issues/1412#issuecomment-396915653
- The `index.d.ts` file can easily be generated by running the following command:
```
lerna run prepare --scope @turf/distance -- --declaration
```

### LINKED ISSUE #1518: Type checking included turf library in typescript fails for nearest-point-to-line and point-to-line-distance
When type checking a project including turf two `d.ts` files seem to not get registered properly. This can be trivially fixed by adding local declarations (copied from the relevant ones in the repo), but it's not obvious and leads to a rougher onboarding experience.

## To replicate

Install turf as part of a package and try to check using typescript

`tsconfig.json`
```
{
  "compilerOptions": {
    "esModuleInterop": true,
    "isolatedModules": true,
    "jsx": "react",
    "module": "commonjs",
    "noEmit": true,
    "noImplicitAny": true,
    "noUnusedLocals": true,
    "outDir": "./dist/",
    "sourceMap": true,
    "strict": true,
    "target": "es5"
  },
  "include": ["./src/**/*.ts", "./src/**/*.tsx"],
  "exclude": ["./dist/", "./node_modules"]
}
```

```
$ tsc
node_modules/@turf/turf/index.d.ts:34:45 - error TS7016: Could not find a declaration file for module '@turf/nearest-point-to-line'. '/Users/jabrahamson/Development/cibo/ursa-web/node_modules/@turf/nearest-point-to-line/index.js' implicitly has an 'any' type.
  Try `npm install @types/turf__nearest-point-to-line` if it exists or add a new declaration (.d.ts) file containing `declare module '@turf/nearest-point-to-line';`

34 export {default as nearestPointToLine} from '@turf/nearest-point-to-line';
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@turf/turf/index.d.ts:84:46 - error TS7016: Could not find a declaration file for module '@turf/point-to-line-distance'. '/Users/jabrahamson/Development/cibo/ursa-web/node_modules/@turf/point-to-line-distance/index.js' implicitly has an 'any' type.
  Try `npm install @types/turf__point-to-line-distance` if it exists or add a new declaration (.d.ts) file containing `declare module '@turf/point-to-line-distance';`

84 export {default as pointToLineDistance} from '@turf/point-to-line-distance';
                                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

error Command failed with exit code 1.
```

### LINKED ISSUE #1523: Missing Types for TypeScript When Using turf/circle Directly
Using:

```typescript
import * as circle from '@turf/circle';
```

> Could not find declaration file for module '@turf/circle'. node_modules/@turf/circle/index.js implicitly has 'any' type.

Using:

```typescript
import { circle } from '@turf/turf'; // OK
```

It looks like no index.d.ts file is setup here: https://github.com/Turfjs/turf/tree/master/packages/turf-circle Is that needed?

### LINKED ISSUE #1568: Typings for @turf/distance
Hi! It seems that typings for `distance` do not work out-of-the-box.

STR:
1. `npm install --save @turf`
2. `import distance from '@turf/distance';`

Compiler says:
```
Type error: Could not find a declaration file for module '@turf/distance'. 'src/node_modules/@turf/distance/index.js' implicitly has an 'any' type.
  Try `npm install @types/turf__distance` if it exists or add a new declaration (.d.ts) file containing `declare module '@turf/distance';`  TS7016
```
Though if I suppress it, no errors occur while runtime.

When I look into `/distance` module, there is no .ts files there:

screenshot from 2019-01-13 21-19-13

Tried the following steps to fix:
1. `npm install @types/turf__distance` - 404 Not Found (as turf contains types on its own, not in `types` repo
2. `npm install @types/turf` - old version of typings
3. `npm install @turf/turf` - better, I can use typings as well as `distance` itself, but only in way when I import the entire `turf` lib. So that bundle size increased by 300kb, doesn't fit.
4. `dts-gen @turf/distance` - no `.ts` files there to create the `.d.ts` file

**The question is**: how can I use the single `@turf/distance` module along with TypeScript typings?

### LINKED ISSUE #1725: No typings for the standalone @turf/distance package
Hi,

I installed @turf/distance@6.0.1 and it doesn't appear to have TypeScript typings. The package.json says there should be an index.d.ts, but that file isn't present in the package under the node_modules folder.

I think maybe this commit might have broken it https://github.com/Turfjs/turf/commit/364dc81a97d13744b770e1d62802f7b8979d61d6#diff-6d071eff6c43993c40d93a8b2b275048 but I don't really have much context on all this.

To mitigate this for myself I downgraded to 6.0.0 which includes the .ts file and so works fine.

Would appreciate a fix.

### LINKED ISSUE #1733: TypeScript type definitions not included in npm-installed @turf/clusters-dbscan
Example:
```ts
import clustersDbscan from '@turf/clusters-dbscan';

const myFeatureCollection = featureCollection(...);

// return value inferred as `any`
const result = clustersDbscan(myFeatureCollection, 10);

// this should cause a TS compiler error, but doesn't because `result` type is `any`
console.log(result.doesntExist);
```

Expected: the above code should not compile, since `doesntExist` isn't a real field of the return object, which the compiler would know if type definitions for this module were included.

I suspect this is likely also an issue w/ some other turf packages, but haven't checked myself.

Would suggest enabling the [`declaration` option in tsconfig.json](https://www.typescriptlang.org/docs/handbook/compiler-options.html#compiler-options), or otherwise explicitly making sure that a `.d.ts` file is included in the packaged module

### LINKED ISSUE #1742: Typescript type definitions not included in @turf/centroid
The index.d.ts file for the @turf/centroid package is missing.

### LINKED ISSUE #1815: Some turf packages are missing typescript definitions?
Hi i wanted to lower my bundle size, first i tried to use treeshacking using webpack 4, but that did not seem to work at all with Turf.js.

But some packages are missing typescript definitions like center, centroid, distance.
It seems i get the definitions if i use @turf/turf could you please publish new packages that contain the missing index.d.ts file

### LINKED ISSUE #1827: Webpack error when using NPM package @turf/boolean-equal
I have an **Angular 8** application including some Turfjs NPM packages and everything had worked beautifully so far. However, after adding the NPM package **@turf/boolean-equal**, I started getting the following error:

> ERROR in ./node_modules/@turf/clean-coords/index.ts
> Module build failed (from ./node_modules/@ngtools/webpack/src/index.js):
> Error: C:\Projects\PRAI\Code\prai_dmaps\DMapS\ClientSide\dmaps-client\node_modules\@turf\clean-coords\index.ts is missing from the TypeScript compilation. Please make sure it is in your tsconfig via the 'files' or 'include' property.
> The missing file seems to be part of a third party library. TS files in published libraries are often a sign of a badly packaged library. Please open an issue in the library repository to alert its author and ask them to package the library using the Angular Package Format (https://goo.gl/jB3GVv).
>     at AngularCompilerPlugin.getCompiledFile (C:\Projects\PRAI\Code\prai_dmaps\DMapS\ClientSide\dmaps-client\node_modules\@ngtools\webpack\src\angular_compiler_plugin.js:913:23)
>     at plugin.done.then (C:\Projects\PRAI\Code\prai_dmaps\DMapS\ClientSide\dmaps-client\node_modules\@ngtools\webpack\src\loader.js:41:31)
>     at process._tickCallback (internal/process/next_tick.js:68:7)
> 

_@turf/clean-coords_ is a dependency of _@turf/boolean-equal_. I checked and confirmed that the file _./node_modules/@turf/clean-coords/index.ts_ exists, so I guess this file is not being properly referenced.

As suggested in the error, I tried adding _@turf/clean-coords/index.ts_ to the "files" property of my _tsconfig.json_, and the error disappeared but I got two new identical errors, this time for the files _./src/main.ts_ and _./src/polyfills.ts_. These files are fundamental in any Angular application and never had this problem before, so I guess something is still not right. If I add all 3 files to the tsconfig.json then all the errors disappear, but importing these files (main.ts and polyfills.ts) is not something that one should have to do in order to use @turf/boolean-equal.
```
  "files": [
    "node_modules/@turf/clean-coords/index.ts",
    "src/main.ts",
    "src/polyfills.ts"
  ]
```
Just in case it helps, I noticed that _@turf/boolean-equal_ **is missing the file index.d.ts**.

mfedderly:
This is a really big PR that does the same thing over and over in every Typescript package. I made the changes manually using a bunch of package-wide search and replace. I'd like to make this sort of thing automated, but I'm too scared to do that yet without understanding the repo better.

The first step is to have a single tsconfig.json file (tsconfig.shared.json) in the root of the repo that all typescript packages are going to reference. This lets us ensure that the typescript strictness level is the same in every package by just changing it in one place. The tsconfig.json in every typescript package then just references the root one for its settings.

The next step is to have tsc publish the output .js and .d.ts files into a dist directory and use that as the package main. This gives us a single place for everything that winds up in the NPM package, and gives us room to publish a second set of artifacts that use ES modules to support tree shaking (future PR). It also makes the .gitignore situation much cleaner. The standardization also fixes the missing typescript definitions in many packages.

The last step was to re-enable the types.ts test step to make sure the types keep being enforced at build time.

A quick summary of what changed by file:

- `packages/*/.gitignore` these were all removed because the root .gitignore takes care of it now
- `packages/*/bench.js` updated these to point at the built artifacts instead of the ones at the root
- `packages/*/test.js` same as bench.js
- `packages/*/*.d.ts` these are all generated files, they got moved into dist
- `packages/*/package.json` updated main, types, and files to point at the new dist locations, some packages (with types.ts) needed the test script forked to do both tape and types.ts tests.
- `packages/*/tsconfig.json` updated to extend tsconfig.shared.json
- A few packages also had additional .ts files that generated .d.ts and .js files, so the generated files have also been deleted
- Some other files (like packages/turf-boolean-equal/index.ts) had their `import * as foo from "bar"` get rewritten to `import foo from "bar"`, this is due to tsconfig's `esModuleInterop` flag and some import semantics with typescript.

A few packages needed special attention to get them working in the new build setup:

- packages/turf-bbox-clip/lib/lineclip.js was renamed to packages/turf-bbox-clip/lib/lineclip.ts, and only slightly modified to get the build to work again
- packages/turf-boolean-equal/package.json - Had to update the dependency on clean-coords to pick up the one in this repo instead of an older one from npm
- There's a subtle issue with packages/turf/test.js where it needs to be run last because it depends on every package having been built (even the ones not listed as dependencies). I think building in pretest is probably wrong, but this is the most minimal change to get a passing build on this already way-too-big.

Fixes #1827, Fixes #1815, Fixes #1742, Fixes #1733, Fixes #1725, Fixes #1568, Fixes #1523, Fixes #1518, Fixes #1493, Fixes #1465, Fixes #1390


mfedderly:
@rowanwins this is ready for you to take a look at, sorry for it being so gigantic.

rowanwins:
I'm probably going to split up my comments as I review different bits sorry @mfedderly as I probably won't get through reviewing everything in one go :) 

> packages/*/bench.js updated these to point at the built artifacts instead of the ones at the root

Should we be forcing the build process before `npm run bench` is run to ensure that the benchmark is run on the changed files? eg something like
````
    "bench": "npm run prepare && node bench.js",
````

From my perspective if bench is running the `dist` then we should ensure `dist` is up to date. Happy to take thoughts on this.

**Comment by rowanwins** [comment]
> packages/*/.gitignore these were all removed because the root .gitignore takes care of it now

👍

**Comment by rowanwins** [comment]
> packages/*/test.js same as bench.js

The changes to reference the `dist` folder makes sense and doesn't have the same problem as `bench` because of the `pretest` script which ensures things are compiled prior to the tests running.

PS I'm making some of these comments just in case others are watching on and wondering why they are different.

**Comment by mfedderly** [comment]
Good idea about the bench, I used some tooling to apply that change. The scripts entries aren't necessarily in the order I'd prefer, but the tooling I'd like to introduce (https://github.com/monorepolint/monorepolint) can fix that in a later PR.

**Comment by rowanwins** [comment]
Just reviewing this properly and also noticed it's from your fork. Perhaps after this PR it's easier to switch across to using this repo directly rather than your fork?

File: packages/turf-nearest-neighbor-analysis/test.js [hunk 1]
```diff
@@ -7,7 +7,7 @@ const truncate = require('@turf/truncate').default;
 const centroid = require('@turf/centroid').default;
 const { featureEach } = require('@turf/meta');
 const { featureCollection } = require('@turf/helpers');
-const nearestNeighborAnalysis = require('.').default;
+const nearestNeighborAnalysis = require('./dist/js/index.js').default;
```
rowanwins: I'm having troubles getting this test to run locally because of how file paths are handled. SOmething like this should be more robust
````
const fs = require('fs');
const test = require('tape');
const glob = require('glob');
const path = require('path');
const load = require('load-json-file');
const write = require('write-json-file');
const truncate = require('@turf/truncate').default;
const centroid = require('@turf/centroid').default;
const { featureEach } = require('@turf/meta');
const { featureCollection } = require('@turf/helpers');
const nearestNeighborAnalysis = require('./dist/js/index.js').default;

const directories = {
    in: path.join(__dirname, 'test', 'in') + path.sep,
    out: path.join(__dirname, 'test', 'out') + path.sep
};

const fixtures = fs.readdirSync(directories.in).map(filename => {
    return {
        filename,
        name: path.parse(filename).name,
        geojson: load.sync(directories.in + filename)
    };
});

test('turf-nearest-neighbor', t => {
    fixtures.forEach(fixture => {
        const filename = fixture.filename;
        const name = fixture.name;
        const geojson = fixture.geojson;
        const options = geojson.options;
        const results = featureCollection([]);
        featureEach(geojson, feature => results.features.push(truncate(feature)));
        if (geojson.features[0].geometry.type === "Polygon") {
          featureEach(geojson, feature => results.features.push(truncate(centroid(feature, {properties: {"marker-color": "#0a0"}}))));
        }
        results.features.push(truncate(nearestNeighborAnalysis(geojson, options)));
        if (process.env.REGEN) write.sync(directories.out + filename, results);
        t.deepEqual(results, load.sync(directories.out + filename), name);

  });
  t.end();
});
````

rowanwins:
`npm-run-all` hasn't been added within the individuals packages `package.json` in the `devDependencies` which means you can't run `npm run bench` within a package directory. 
Happy to let it slide if it's easier to do once we're using `monorepolint`- would be handy if that could be used to manage core `devDependencies`.
Alternatively I'm pretty sure `&&` is cross-platform friendly so we could swap out
````
    "bench": "npm-run-all prepare bench:run",
````
for 
````
    "bench": "npm run prepare && bench:run",
````

packages/turf-nearest-neighbor-analysis/test.js
mfedderly: I went ahead and made that change, there's a lot of boilerplate in the test.js and bench.js files that maybe can be combined.

mfedderly:
For `npm-run-all` not being available, that's interesting. Are you using NPM instead of yarn? Right now we've got a yarn.lock and CI builds using yarn. I'm on yarn 1.22.1.

npm-run-all has a nice feature where you can do `npm-run-all test:*` that runs test:tarp and test:types which I want to use once we start going to monorepolint.

If its the npm/yarn difference causing this, I'd like to standardize around yarn if we can, just so everyone has the same versions of the dependencies.

Once monorepolint is set up, it is considerably easier to enforce dependencies across the packages, so I'm also happy to just add it everywhere. I'm indifferent on having it in the root package.json only vs all of them.

rowanwins:
Updating my yarn version to the latest hasn't helped the `npm-run-all` issue. 

To confirm my approach
````
// in turf root dir
yarn install
cd packages/turf-along
npm run bench
=> sh: npm-run-all: command not found
````

mfedderly:
The difference in npm-run-all behavior is because you are running the bench script with npm and I've been running it with yarn! Totally on board with fixing this, assigned #1854 to myself.

PR #1856, opened 2020-03-16
Setup monorepolint

mfedderly:
- Monorepolint lets us easily manage all of our subpackages declaratively
- Husky installs commit hooks to prevent invalid commits from even merging (bypass this with `git commit --no-verify`)
- lint-staged gives us rules for custom linting depending on what files changed
- Bump the travis config to use node 12 (current LTS)
- Bump typescript dependency because some package shipped types that are only 3.x compatible.


mfedderly:
Yep that was the minimal change I made first!

PR #1860, opened 2020-03-17
Build Standardization

### LINKED ISSUE #1170: ⚠️ Publishing *.mjs file extension for "module" - TurfJS v6.0
### ⚠️ Breaking Change - Major Release v6.x

To prevent any older Webpack configurations from being accidentally  broken, the `v5.x` release currently uses `*.js` file extension to publish `"module"`. In the next TurfJS major release, the pre-compiled `"module"` file extension will be using `*.mjs` which has already landed in Webpack 4 (https://github.com/webpack/webpack/issues/5686).

Ref: https://github.com/Turfjs/turf/issues/1165 & https://github.com/Turfjs/turf/issues/1133

### LINKED ISSUE #1383: Turf does not work with Webpack 4
I upgraded my project to Angular 6 which migrated to Webpack 4. I have a dependency on `mapbox` which uses turf. After the upgrade I started getting build errors like the following.

```
ERROR in ./node_modules/@turf/meta/index.mjs
997:15-25 Can't import the named export 'lineString' from non EcmaScript module (only default export is available)
 @ ./node_modules/@turf/meta/index.mjs
 @ ./src/index.js
```

It appears to be failing because `@turf/meta` uses `mjs` and `@turf/helpers` is a non `mjs` module. I am not sure if this is just a build configuration in Webpack that needs to be tweaked or if a correction needs to be made to turf. I looked around the Webpack issues and didn't really see anything to help along with a fix.

Here is a repo to easily reproduce this issue.

https://github.com/adamduren/webpack4-turf-issue

### LINKED ISSUE #1833: cannot import @turf/difference
I'm opening this issue #1684 again because it is still not fixed, I can't import @turf/difference with webpack because of the use of a *.mjs file as module.

mfedderly:
Lots more standardization now that we have monorepolint to lean on. Individual commits are marked with [manual] or [automatic] depending on what sort of changes they contain.

- Enforce the existing TS configuration
- Make the JS projects build into the same dist directory
- Testing standardization, adds some missed types.ts testing
- Small fix to turf-dissolve/bench.js
- Small fix to turf-explode/test.js
- Clean up of @turf/turf build (entirely manual)
- turf-meta and turf-difference had .mjs files instead of .js files as their root

Fixes #1833, Fixes #1383, Fixes #1170


.monorepolint.config.ts
```diff
@@ -32,6 +32,78 @@ glob.sync(path.join(__dirname, "packages", "turf-*")).forEach(pk => {
...
+      {
+        options: {
+          entries: {
+            main: "dist/js/index.js",
+            types: "dist/js/index.d.ts",
+            files: ["dist"]
+          }
+        },
+        includePackages: TS_PACKAGES
+      },
```
rowanwins: So this first chunk targets the TS-based packages?

Is there any reason we're doing `dist/js/index.js` rather than `dist/index.js`?

.monorepolint.config.ts
```diff
@@ -32,6 +32,78 @@ glob.sync(path.join(__dirname, "packages", "turf-*")).forEach(pk => {
...
+        options: {
+          entries: {
+            main: "dist/js/index.js",
+            module: "dist/es/index.js",
+            types: "index.d.ts",
+            files: ["dist", "index.d.ts"]
+          }
+        },
+        includePackages: JS_PACKAGES
+      }
```
rowanwins: This chunk targets the plain ol' JS based packages?

And now perhaps I can see the `dist/js` vs `dist/es` difference. I think previously we dumped everything in the dist folder. Think I'm happy enough either way.

.monorepolint.config.ts
```diff
@@ -32,6 +32,78 @@ glob.sync(path.join(__dirname, "packages", "turf-*")).forEach(pk => {
...
+        options: {
+          scripts: {
+            bench: "npm-run-all prepare bench:run",
+            "bench:run": "node bench.js",
+            docs: "node ../../scripts/generate-readmes",
+            prepare: "tsc",
+            pretest: "tsc"
+          }
+        },
+        includePackages: TS_PACKAGES
```
rowanwins: So this will standardise all the scripts part for TS-based packages.

.monorepolint.config.ts
```diff
@@ -32,6 +32,78 @@ glob.sync(path.join(__dirname, "packages", "turf-*")).forEach(pk => {
...
+      },
+      {
+        options: {
+          scripts: {
+            pretest: "rollup -c ../../rollup.config.js",
+            posttest: "node -r esm ../../scripts/validate-es5-dependencies.js"
+          }
+        },
+        includePackages: JS_PACKAGES
+      },
```
rowanwins: This will standardise the script part for the plain JS packages.

Is there any reason we don't include `bench` or `docs`?

.monorepolint.config.ts
```diff
@@ -32,6 +32,78 @@ glob.sync(path.join(__dirname, "packages", "turf-*")).forEach(pk => {
...
+      {
+        options: {
+          scripts: {
+            "test:tape": "node -r esm test.js"
+          }
+        },
+        includePackages: TAPE_PACKAGES
```
rowanwins: This adds the test:tape script to all packages with a test.js file which should be pretty much every package

.monorepolint.config.ts
```diff
@@ -32,6 +32,78 @@ glob.sync(path.join(__dirname, "packages", "turf-*")).forEach(pk => {
...
+      },
+      {
+        options: {
+          scripts: {
+            "test:types": "tsc --noEmit types.ts"
+          }
+        },
+        includePackages: TYPES_PACKAGES
```
rowanwins: Test the types on all packages that have types, but aren't written in TS.

rowanwins:
Any one other thought... does this rollup config generate something that's suitable for us via a cdn?

````
<script src='https://npmcdn.com/@turf/turf/turf.min.js'></script>
<script>
    var bbox = turf.bbox(features);
</script>
````

I'm wasn't sure if `iife` or `umd` we're better options for that use case?

mfedderly:
There's two rollup configs, one that gets used for everything that isn't `@turf/turf`, and one in that package for the main rollup. `@turf/turf` still has `dist/turf.min.js` which is the same UMD export as before.

.monorepolint.config.ts
mfedderly: Yeah it gives us room for dist/es/* for the modules later.

.monorepolint.config.ts
mfedderly: Actually everything used to get dumped in the root of the package which was going to make setting up the linters again really hard.

.monorepolint.config.ts
mfedderly: Half oversight, half work in progress. I'd really want to do `includePackages: [...TS_PACKAGES, JS_PACKAGES]` for all of the standard things, but those blocks are sort of missing at the moment (TS still isn't publishing modules)

.monorepolint.config.ts
mfedderly: Yeah I'd like this to be on all TS and JS packages, not just when a test.js exists, because I'd like to enforce that it exists.

.monorepolint.config.ts
mfedderly: I'm split on whether or not to test TS packages too, it could catch API breaks. test.js and bench.js also sort of already catch API breaks though.

rowanwins:
Ahh I was only looking at the diff which didn't show the `umd` package :joy:

PR #1862, opened 2020-03-18
Fix version inconsistencies

mfedderly:
I ran into a bunch of issues today when moving more fully into ES module territory. In order to fix some of them I decided to synchronize all of the package versions. I was hoping to wait for a release to do this, but there were actually API breaks so we need to fix this before we can release.

Check the individual commits for more details on the issues I ran into.


mfedderly:
Filed #1866 to track updating martinez-polygon-clipping and fixing the tests for real instead of skipping them.

mfedderly:
@rowanwins this is ready for you

PR #1869, opened 2020-03-23
Tree shaking

### LINKED ISSUE #1736: Add `sideEffects` to package.json?
Hi, it would be nice if you could list side-effectful modules somewhere like [here](https://github.com/d3/d3/issues/3131#issuecomment-517316673) or if you could set `sideEffects` in `package.json`s to help tree-shakability.

As a reference, I'd like to tree-shake turf modules [here](https://github.com/nestauk/svizzle/commit/b1e377bebcd8a07d8e526c3a4c1838615e0665de) as `@turf/helpers` and `@turf/meta` seem pretty "big" and I doubt I need to bundle all of their functions but I'm not sure if that will break my builds.

Thanks!

### LINKED ISSUE #1846: How to reduce bundle size?
Hello there,
this is my `webpack bundle analyzer` report, and as you see, the turf size is pretty big (556ko)!

image

I've verified twice, and all my files contains only partial imports, like this:
```js
import {
  getCoords,
  nearestPoint,
  along,
  point
} from "@turf/turf";
```

So I don't understand why this is so big?

Am I doing something wrong?

Thanks!

mfedderly:
Lots of monorepolint-driven changes:
- Typescript publishes ES modules
- All packages have `sideEffects: false` to enable tree shaking
- All projects use prepare instead of duplicating in the pretest command
- Standardize order of package.json files

There's one manual change that will likely need to turn into some kind of productized build step:
- turf-bbox export hack to fix @turf/turf build

I'd like to at least update rollup and see if that fixes the turf-bbox build hack, otherwise I think we might need a post-processing step on the generated Typescript to fix the `.default` exports on packages in es5 code. I'm curious if people run `esModuleInterop` turned on for their Typescript projects, I think it is almost mandatory these days to interoperate with non-TS libraries.

Fixes #1736, Fixes #1846


mfedderly:
@rowanwins This is ready for you

RichardLindhout:
Can we use this from npm?

PR #1870, opened 2020-03-24
Upgrade lerna

mfedderly:
Quick PR to bump the Lerna dependency

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,5 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "npmClient": "yarn"
+  "npmClient": "yarn",
+  "version": "0.0.0"
 }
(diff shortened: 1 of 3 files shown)


lerna.json
```diff
@@ -1,9 +1,7 @@
 {
-  "lerna": "2.8.0",
   "packages": [
     "packages/*"
   ],
-  "version": "independent",
   "npmClient": "yarn",
-  "useWorkspaces": true
+  "version": "0.0.0"
```
mfedderly: This should get bumped to 6.2.0-something when I do the release

lerna.json
```diff
@@ -1,9 +1,7 @@
 {
-  "lerna": "2.8.0",
   "packages": [
     "packages/*"
   ],
-  "version": "independent",
   "npmClient": "yarn",
-  "useWorkspaces": true
```
mfedderly: Seems like this is no longer required always, only with the bootstrap command.

PR #1874, opened 2020-03-26
Release 6.2.0.alpha.1

mfedderly:
With changelog updates


lerna.json
```diff
@@ -1,10 +1,12 @@
 {
-  "packages": ["packages/*"],
+  "packages": [
```
mfedderly: The bots are fighting about whitespace

rowanwins: solid

Issue #1875, opened 2020-03-27
Latest release not available at posted endpoint


kuanb:
In Turf docs, under https://turfjs.org/getting-started, it indicates that one can require Turf from the following endpoint: https://npmcdn.com/@turf/turf@6.2.0-alpha.1/turf.min.js

![image](https://user-images.githubusercontent.com/6053396/77714709-ef5f5000-6f96-11ea-8a4a-1ed85b11d8a0.png)

At present, that endpoint returns: ```
Cannot find "/turf.min.js" in @turf/turf@6.2.0-alpha.1```

vincentsarago:
Same 😢. Any idea what's going on? I've check Travis log and couldn't find anything

rowanwins:
Due to some changes in packaging it's now at 
https://npmcdn.com/@turf/turf@6.2.0-alpha.1/dist/turf.min.js

rowanwins:
Turns out this broke all the docs :(

vincentsarago:
@rowanwins I don't mean to be a pain in the ... but this change is kind of sad. It breaks every web page I've done where I use turf 🙁. To be honest this is a kinda unexpected change especially for an alpha release !

rowanwins:
Hey @vincentsarago 

Sorry about that, we've just gotten a new contributor on board and this was our first release in years!

Unfortunately with those cdn links if you don't use a named release just picks out the most recent release, and because we changed the location of `turf.min.js` it has freaked.

The old version you've would've been hitting is still available at 
https://unpkg.com/@turf/turf@5.1.6/turf.min.js

@mfedderly - we'll need to have a think about this because this is breaking change. Perhaps even as a temporary measure we can put the compiled output back in the root dir as well.

Cheers
Rowan

rowanwins:
The other thing to mention is that using a CDN hosted version of a library is a somewhat risky option for a production app. Bundling in your dependencies into your application is a more standard low risk approach as you're less likely to get hit by stuff like this.

rowanwins:
Hi @kuanb & @vincentsarago 

We've amended the npm package so that the latest points back to the older version but doesn't change the dist directory so all old CDN links should now work again.

vincentsarago:
thanks a lot @rowanwins this is really appreciated 🙏 
Awesome work

mfedderly:
[...]

PR #1876, opened 2020-03-27
Revert @turf/turf dist file location change

mfedderly:
Moving the dist file for @turf/turf is pretty problematic for CDN users. (#1875)
This moves it back to its prior location and enforces it with monorepolint.


mfedderly:
It already works with `browser` I think. But people will have direct links to that .js file so lets just leave it where it is.

PR #1879, opened 2020-03-30
lineToPolygon fix. Resolves #1812

### LINKED ISSUE #1812: lineToPolygon(line) mutates the underlying LineString
[Here's a JSFiddle](https://jsfiddle.net/hesfyndk/) and [a sample geojson](http://geojson.io/#map=8/49.440/11.448).

Basically, if I have a LineString of length `k`, calling `lineToPolygon(line)` (which returns a "new" polygon) also mutates `line` to have `k + 1` points. Of course, you can set the property `autoComplete`, but I would never expect `lineToPolygon(line)` to both return a `Polygon` and also mutate the input. In this case, I *want* the resulting Polygon to autoComplete, but I don't want to change the input string in place

My current workaround is `lineToPolygon(clone(line))`

```
var linestring1 = turf.lineString([[10, 50], [10.3, 50.3], [11.1, 49.5]]);
var line1Len = turf.getCoords(linestring1).length

var poly1 = turf.transformTranslate(turf.lineToPolygon(linestring1), 20, 35); 
var poly1Len = turf.getCoords(linestring1).length

console.log(line1Len, poly1Len, turf.getCoords(linestring1).length); // 3, 4, 4
console.log(line1Len === turf.getCoords(linestring1).length); // false
```

MortenBirk:
Ensure that the input is by default not modified when lineToPolygon is called. Resolves #1812

An optional mutate prop has been added to options. If true the input will be modified, if false (default) the input will not be modified.

### DIFF
--- a/packages/turf-line-to-polygon/index.ts
+++ b/packages/turf-line-to-polygon/index.ts
@@ -4,6 +4,7 @@ import {
     polygon, multiPolygon, lineString, isObject,
     Feature, FeatureCollection, MultiLineString, LineString, Polygon, MultiPolygon, Properties
 } from '@turf/helpers';
+import clone from '@turf/clone';
 
 /**
  * Converts (Multi)LineString(s) to Polygon(s).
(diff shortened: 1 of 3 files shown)


packages/turf-line-to-polygon/index.ts
```diff
@@ -29,16 +31,23 @@ function lineToPolygon<G extends LineString|MultiLineString>(
         properties?: Properties,
         autoComplete?: boolean,
         orderCoords?: boolean,
+        mutate?: false,
```
mfedderly, line 34: Shouldn't this have a type of `boolean`? Or you could get rid of line 46 and do `mutate = false`.

MortenBirk: Yes it should of course be boolean, it should have been fixed now. For consistency with the rest of the modules code base I think line 46 should stay, even though I do find `mutate = false` to be more readable.

rowanwins:
I think this is fine as a bugfix even though the default behaviour is changing.

mfedderly:
Thanks!

PR #1887, opened 2020-04-23
point-grid - Use center latitude to compute horizontal distance

### LINKED ISSUE #1886: point-grid bag
Hi, I found a problem when building a grid for a big box.

Sample this prolem:
GeoJSON: https://gist.github.com/entagir/8a2f8e003731ce52ab87fe230dc2ab7f
             or: https://jsonkeeper.com/b/GHL4
Code: https://jsfiddle.net/entagir/dsahcfej/11/

Thank!

entagir:
Correction of xFraction variable calculation.

I propose to calculate the distance in latitude from the edge of the box to the center and multiply by 2, because with a large box (more than half the length of the latitude), the distance is not calculated correctly (the shortest distance from edge to edge will envelope the Earth from the other side and will not be the length of the box).

I also propose to calculate the distance in latitude, located in the center of the height of the box, because Now it is calculated on the lower edge (south), which gives a very small number of points for large boxes (at a height from the equator to the pole), because at the northern and southern latitudes their length decreases with distance from the equator. Also, with the same small boxes (one in the northern hemisphere, the other in the southern), the difference in result will be visible. In this regard, I think it will be fair to consider the latitude located in the center of the box.

And you also need to revise the tests for this module.

Thank.

PS: I attach the results and testing code to the tests of the module turf-point-grid. For comparison.
[turf-point-grid-test.zip](https://github.com/Turfjs/turf/files/4521022/turf-point-grid-test.zip)

Resolves #1886

### DIFF
--- a/packages/turf-point-grid/index.ts
+++ b/packages/turf-point-grid/index.ts
@@ -52,20 +52,20 @@ function pointGrid<P = Properties>(bbox: BBox, cellSide: number, options: {
     var north = bbox[3];
 
     var xCenter = 0;
-	var yCenter = 0;
+    var yCenter = 0;
 	
-	if(east > west)
-	{
-		xCenter = west + (east - west) / 2;
-	}
-	else if(east < west)
-	{
-		xCenter = east - (west + east) / 2;
-	}
-	
-	yCenter = north - (north - south) / 2;
+    if(east > west)
+    {
+        xCenter = west + (east - west) / 2;
+    }
+    else if(east < west)
+    {
+        xCenter = east - (west + east) / 2;
+    }
+    
+    yCenter = north - (north - south) / 2;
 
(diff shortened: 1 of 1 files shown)


JamesLMilner:
@rowanwins is this PR still relevant after your changes here https://github.com/Turfjs/turf/pull/2106 ? If not we can close perhaps?

mfedderly:
Currently, point-grid calculates the x distance using the south side of the bounding box. This can result in a very low number of points being used when that side of the bounding box is close to the pole. By using the center latitude, it computes a more representative number of x steps.

Given that this can result in a large increase in total points, I'd like to add this to the v8 milestone just in case.

If you get a chance can you rebase this? Otherwise I will revisit in the 8.0 work eventually.

It'll need at least this to update the formatting and test results:
```
REGEN=1 pnpm test
pnpm prettier --write index.ts test/*
```

Issue #1826, opened 2020-01-30
In some cases `pointsWithinPolygon` is not returning all features


AScripnic:
https://jsfiddle.net/8w9dbpvm/2/

as you can see in the example above I have 2 different points with exactly same behavior.
In one scenario the returned value is 0 and in another 1 (which is correct)

smallsaucepan:
Suspect this is because point `a` is above 90 north latitude, so past the pole and literally off the map. I don't think Turf functions are expected to work with values that stray outside -180 / 180 longitude or -90 / 90 latitude.

```javascript
const a = [11.6711, 104.063];
const b = [31.7475, 35.4756];
```

Closing, though please reopen if this doesn't explain what OP was seeing.

Issue #1755, opened 2019-09-04
`booleanPointInPolygon` method may not works as designed


Coosec:
Hi. I've prepared example geojson with quite complicated polygon and one marker within it. When I use `booleanPointInPolygon` to check if marker is in inside polygon, this method returns `false`. Even though I am 100% sure it should returns `true`. My geojson and simple jsfiddle checker are below:

- [X] GeoJSON - https://gist.github.com/Coosec/f92b7135a2e1a8110ed02777a6e29002
- [X] jsfiddle - https://jsfiddle.net/4u2hoztv/1/

rowanwins:
Hi @Coosec 

The current method doesn't work with degenerate polygons (eg with many self-intersections).

I suggest checking out the [point-in-polygon-hao library](https://github.com/rowanwins/point-in-polygon-hao) which I authored which should handle that case. I'm eventually hoping to replacing the existing turf function with it.

Hope that helps

Coosec:
Hi @rowanwins 

What about simple polygons? I've prepared example with not degenerated polygon which does not work neither.

- [X] GeoJSON - https://gist.github.com/Coosec/446761236d18d3eb6fa3f6597227458b
- [X] jsfiddle - https://jsfiddle.net/tbpzcwr6/

rowanwins:
Hi @Coosec 

That is strange that it didn't work for that simple polygon... that said my replacement module works fine so I'd suggest using it.

kshetline:
I'd like to add that I'm having problems with geojson data as well. I'm trying to identify the timezones to which latitude/longitude pairs belong, using shapes defined for timezones in https://github.com/evansiroky/timezone-boundary-builder/.

At first I thought I was just seeing a lot of errors in this set of timezone boundaries, but I've since mapped several sample timezones with https://geojson.tools/ and have discovered the errors appear to be from `booleanPointInPolygon` behaving badly. Many of these timezones are multipolygons, so simply using `point-in-polygon-hao` instead wouldn't help (or at least wouldn't be straight-forward). I don't see any evidence that these timezones use degenerate polygons anyway.

I don't believe there are any errors in my logic for checking points against timezone shapes, but here's a snippet of the code anyway so you can see my basic approach:

```
function weedZones(x: number, y: number): Feature[] {
  let zones = zoneGrid[x][y];

  if (zones?.length > 1) {
    const weededZones: Feature<any>[] = [];
    const cell = polygon([[[x - 0.1, y - 90.1],
[...]

PR #1893, opened 2020-05-03
Update boolean-point-in-polygon to use point-in-polygon-hao library

rowanwins:

This PR updates boolean-point-in-polygon module to use the `point-in-polygon-hao` library which is more robust and faster than the current implementation.

I'm also wondering if we want to tweak the API a little  - we currently have an option called `ignoreBoundary` which modifies how we handle the result. However I'm wondering if we should actually provide an option that allows us to report a point lying on the boundary because there may be scenarios where that is helpful to know, however it seems. a bit clunky to have another option thats called `reportOnBoundary`... so I'm open to ideas on if and how we might tackle that. Any changes here might also be potentially breaking so we'd need to consider that...

### Benchmarks
````
// This PR
simple x 16,449,144 ops/sec ±1.45% (88 runs sampled)
multiPolyHole - inside x 7,624,379 ops/sec ±2.08% (89 runs sampled)
multiPolyHole - outside x 88,233,389 ops/sec ±1.70% (87 runs sampled)

// OLD
simple x 9,041,375 ops/sec ±2.02% (90 runs sampled)
multiPolyHole - inside x 3,853,073 ops/sec ±0.61% (88 runs sampled)
multiPolyHole - outside x 72,504,154 ops/sec ±0.47% (92 runs sampled)
````

### CODE REVIEW

**Comment by rowanwins** [comment]
I reckon let's leave the API change till later until we've got a bit more of a rhythm and we might identify other breaking changes to tackle at the same time.

Re a seperate point-on-ring module we already have `booleanPointOnLine` which is pretty much the same.

**Comment by kaligrafy** [comment]
Any news on this? I don't see why it was not merged already since the current implementation does not detect boundary nodes correctly and is slower...

**Comment by mfedderly** [comment]
This just needs a new release from the point-in-polygon-hao module for https://github.com/rowanwins/point-in-polygon-hao/pull/14 and then it should be ready to merge.

**Comment by rowanwins** [comment]
Just released v1.1.0 for point-in-polygon-hao

**Comment by mfedderly** [comment]
@rowanwins any idea why this does not address #1755 and #2048? They both have points that appear to not be in their polygons, but they return true on the current version of this branch.

Here's the two tickets as tests:

```

test("boolean-point-in-polygon -- issue #1755", (t) => {
  const pt = point([18.957922, 61.5439]);
  const poly = polygon([
    [
      [42.14295462, 71.62860768],
      [37.39686087, 54.86312041],
      [-10.59142038, 71.40570518],
      [42.14295462, 71.62860768],
    ],
  ]);
  t.false(booleanPointInPolygon(pt, poly));
  t.end();
});

test("boolean-point-in-polygon -- issue #2048", (t) => {
  const pt = point([-84.44876, 42.79741]);
  const poly = polygon([
    [
      [-86.3559, 41.2203],
      [-69.9985, 54.7459],
      [-69.9782, 54.7441],
      [-69.9725, 54.7484],
      [-69.9672, 54.7496],
      [-69.9514, 54.758],
      [-69.9469, 54.7625],
      [-69.9457, 54.7669],
      [-69.9458, 54.7736],
      [-69.9479, 54.7769],
      [-69.9546, 54.7777],
      [-69.9612, 54.775],
      [-69.9644, 54.7769],
      [-69.9644, 54.7803],
      [-69.9553, 54.7926],
      [-69.9587, 54.793],
      [-86.3559, 41.2203],
    ],
  ]);
  t.false(booleanPointInPolygon(pt, poly));
  t.end();
});
```

rowanwins:
@mfedderly this is one of the quirks of projecting coords onto a map using WebMercator. Attached is a pic showing the first case from a planar geometry viewer
Geometry - GeoGebra 2021-08-06 10-46-49

QGIS also shows similar when rendering the points using WGS84.
*Untitled Project — QGIS 2021-08-06 10-52-07

The existing implementation in master returns the same thing so I don't think we've got any regression.

I must admit you had me very concerned for 20 mins. 🎢  of emotions like 😭  then 😌  
😆 

I wonder if this is responsible for some issues like #1826 ?

mfedderly:
@rowanwins I'm going to get this merged. If you want to add more behavior changes to ignoreBoundary/reportOnBoundary before 7.0.0 releases, either file a PR, or an issue in the 7.0 milestone.

PR #1896, opened 2020-05-03
Update kinks to use sweepline-intersections library

rowanwins:

This replaces the naive algorithm in `kinks` to use the `sweepline-intersections` [library](https://github.com/rowanwins/sweepline-intersections).

I've added a larger geometry (switzerlandKinked - which has 700 vertices) which really highlights how slow the existing naive implementation is. The existing implementation works ok if dealing with <20 vertices but will quickly grind to a halt as it checks every segment against every segment.

````
// NEW
---> switzerlandKinked x 1,907 ops/sec ±1.28% (91 runs sampled)
hourglass x 898,096 ops/sec ±6.84% (79 runs sampled)
multi-linestring x 392,678 ops/sec ±1.52% (90 runs sampled)
multi-polygon x 403,615 ops/sec ±2.56% (90 runs sampled)
open-hourglass x 1,197,766 ops/sec ±3.54% (75 runs sampled)
triple x 364,118 ops/sec ±1.31% (88 runs sampled)

// OLD
---> switzerlandKinked x 88.20 ops/sec ±0.78% (75 runs sampled)
hourglass x 3,919,439 ops/sec ±1.87% (90 runs sampled)
multi-linestring x 620,491 ops/sec ±2.10% (73 runs sampled)
multi-polygon x 536,627 ops/sec ±2.36% (67 runs sampled)
open-hourglass x 3,077,403 ops/sec ±3.21% (78 runs sampled)
triple x 804,901 ops/sec ±1.29% (87 runs sampled)
````

### CODE REVIEW

File: packages/turf-kinks/index.ts [hunk 1]
```diff
@@ -40,114 +41,16 @@ export default function kinks<T extends LineString | MultiLineString | Polygon |
...
-        coordinates = [feature.coordinates];
-    } else if (feature.type === "MultiLineString") {
-        coordinates = feature.coordinates;
-    } else if (feature.type === "MultiPolygon") {
-        coordinates = [].concat.apply([], feature.coordinates);
-    } else if (feature.type === "Polygon") {
-        coordinates = feature.coordinates;
-    } else {
+
+    // @ts-ignore
```
mfedderly, line 45: Why is there a @ts-ignore here?

packages/turf-kinks/index.ts
```diff
@@ -40,114 +41,16 @@ export default function kinks<T extends LineString | MultiLineString | Polygon |
...
+    if (featureIn.type === "Point" || featureIn.type === 'MultiPoint') {
```
mfedderly, line 46: Should this also check for `Feature<Point>` and `Feature<MultiPoint>`?

packages/turf-kinks/index.ts
```diff
@@ -40,114 +41,16 @@ export default function kinks<T extends LineString | MultiLineString | Polygon |
...
-    // if line2 is a segment and line1 is infinite, they intersect if:
-    if (b >= 0 && b <= 1) {
-        result.onLine2 = true;
-    }
-    // if line1 and line2 are segments, they intersect if both of the above are true
-    if (result.onLine1 && result.onLine2) {
-        return [result.x, result.y];
-    } else {
-        return false;
+    const intersections = findIntersections(featureIn)
```
mfedderly, line 50: This seems a lot cleaner!

packages/turf-kinks/index.ts
rowanwins: Probably because I don't understand ts :)

packages/turf-kinks/index.ts
rowanwins: This might help resolve the above :)

packages/turf-kinks/index.ts
rowanwins: There is similiar logic in the sweepline-intersections library - but it does make our package look neater :)

rowanwins:
In regards to performance - those old benchmarks are done on trivial geometries where the number of edges are between 5 & 10. The new algorithm sets up a few data structures which allow it to scale on larger geometries (eg the switzerland example I added with 700 edges), but that adds some initial overhead. The more edges you have the worse the old brute force algorithm will perform.

Conceivably we could set some arbitrary limit where we use the brute force algorithm in number of edges is < x. What do you think @mfedderly ?

packages/turf-kinks/index.ts
mfedderly: `This condition will always return 'false' since the types '"LineString" | "MultiLineString" | "Polygon" | "MultiPolygon" | "Feature"' and '"Point"' have no overlap.` This is saying that because of the typings for `featureIn` it wouldn't actually expect `Point` to ever come in. This makes sense if you're a typescript consumer but isn't necessarily safe if you're a javascript consumer. You can cast featureIn to allow the checks like so:

```
// at the top
import {GeometryObject} from "@turf/helpers";
// down below
(featureIn as GeometryObject | Feature<GeometryObject>).type === "Point"
```

mfedderly:
Yeah I think splitting based on number of edges is a good solution, some amount of testing would have to be required to figure out what the inflection point is. You could maybe use @turf/circle to generate arbitrarily complex circle polygons to test with

rowanwins:
@mfedderly - took your advice re the benchmarks with testing for the inflection point using @turf/circle to generate edges and got the following results. So it looks like the infection point is probably around 10 edges.

````
// New version
steps - 5 x 470,655 ops/sec ±0.72% (93 runs sampled)
steps - 10 x 254,759 ops/sec ±2.27% (93 runs sampled)
steps - 15 x 163,340 ops/sec ±0.30% (94 runs sampled)
steps - 20 x 115,420 ops/sec ±2.33% (91 runs sampled)

// Old version
steps - 5 x 511,895 ops/sec ±2.93% (84 runs sampled)
steps - 10 x 249,804 ops/sec ±1.94% (88 runs sampled)
steps - 15 x 135,428 ops/sec ±1.98% (87 runs sampled)
steps - 20 x 85,685 ops/sec ±1.59% (89 runs sampled)
````

File: packages/turf-kinks/index.ts [hunk 4]
```diff
@@ -42,8 +43,7 @@ export default function kinks<T extends LineString | MultiLineString | Polygon |
         features: [],
     };
 
-    // @ts-ignore
-    if (featureIn.type === "Point" || featureIn.type === 'MultiPoint') {
+    if ((featureIn as GeometryObject | Feature<GeometryObject>).type === "Point" || (featureIn as GeometryObject | Feature<GeometryObject>).type === 'MultiPoint') {
```
mfedderly, line 46: This still doesn't handle Features correctly, I think you want something like:

```
if (featureIn as GeometryObject).type === "Point" ||
  featureIn as GeometryObject).type === 'MultiPoint' ||
  (featureIn as Feature<GeometryObject>.type === 'Feature' && (
    featureIn.geometry.type === 'Point' ||
    featureIn.geometry.type === 'MultiPoint'))) {
```

mfedderly:
Similar objection here with the compiled code that wound up vendored https://github.com/Turfjs/turf/pull/2033#issuecomment-893682955

mfedderly:
Same thoughts as https://github.com/turfjs/turf/pull/2033#issuecomment-1014898084 to go back to using the proper npm dependency.

packages/turf-kinks/index.ts
rowanwins: Think I've sorted this by limiting support to Features rather than Feature | Geometry

mfedderly (approved):
🎉 woohoo

Issue #261, opened 2015-06-01
error loading turf with browser policy (CSP) and disableEval()


tcastelli:
Hello, not sure if this is something easy/possible, but when using CSP on a site and disabling the use of eval() or similar to avoid XSS on websites, i get the following error while loading turjs
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'unsafe-inline'
 turf.min.js:16:16187

Is it possible to refactor the library to avoid using eval strings to code?

Thanks! :)

tmcw:
@tcastelli as far as I can see, we don't have any calls to eval() in any of our packages: can you use an uncompressed build of Turf to find a more precise error report?

tmcw:
Closing: please reopen if you can give a more specific report.

Lknechtli:
I'm getting a similar error when the code is evaluated: 
```Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' 'unsafe-inline'```

The referenced lines are lines 5625-5628 in the un-minified turf.js I got off github for v2.0.0
Is there any way around this without modifying the CSP?

tcql:
@Lknechtli Looks like this is the result of [code in `robust-orientation`](https://github.com/mikolalysenko/robust-orientation/blob/1ae9e492e903f8ef50337678adcb4892a50237da/orientation.js#L87), one of the nested dependencies of turf-convex, using `new Function` to generate functions. 

Without entirely replacing / reimplementing turf-convex to avoid the evals, it looks like you're stuck modifying CSP, or using a custom turf build that doesn't include turf-convex

rimig:
Does that mean this is not being considered? 

I'm kind of blocked on using turf, I'd really prefer to not fork.

DenisCarriere:
I doubt this will be considered for now until the dependency is fixed.

Have you considered simply importing the modules you want via npm and building your own bundle with Webpack or Browserify?

@rimig Which Turf modules are you interested in using? This issue is relevant for `@turf/convex` and the entire browser Turf build.

DenisCarriere:
👍 Awesome stuff, it's definitely easier to load individual modules instead of the whole thing.

toolchild:
I'm trying to use @turf/center-of-mass, but I get the same error. 
[...]

PR #1907, opened 2020-05-18
update concaveman

### LINKED ISSUE #1903: turf Content Security Policy error
I'm getting the following error when loading turf from the CDN:
"https://cdn.jsdelivr.net/npm/@turf/turf@5/turf.min.js"

```
turf.min.js:1 Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self' https:".

    at new Function (<anonymous>)
    at o (turf.min.js:1)
    at turf.min.js:1
    at mt (turf.min.js:1)
    at turf.min.js:1
    at turf.min.js:1
    at turf.min.js:1
```

Is there a way to get around this issue without having to add 'unsafe-eval' to the Content Security Policy. Everything I've read indicates adding "unsafe-eval" exposed the application to security risks.

rowanwins:

This PR updates the concaveman dependency in a number of packages after a new release of that package addressing a security issue.

Resolves #1903
Also related to #261

### DIFF
--- a/packages/turf-clusters-dbscan/package.json
+++ b/packages/turf-clusters-dbscan/package.json
@@ -53,7 +53,7 @@
     "@types/tape": "*",
     "benchmark": "*",
     "chromatism": "*",
-    "concaveman": "*",
+    "concaveman": "^1.2.0",
     "load-json-file": "*",
     "npm-run-all": "*",
     "tape": "*",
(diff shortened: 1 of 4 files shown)


donmccurdy:
Is this change needed for downstream users to get the latest version of `concaveman`? From https://unpkg.com/@turf/convex@6.3.0/package.json it seems to be already published with version `*`, and I'm not sure what effect the `*` -> `^1.2.0` change has except to prevent users from getting a version more recent than `v1.2.0`.

rowanwins:
You're correct thanks @donmccurdy

PR #1927, opened 2020-07-13
Add some quick notes to document the release process

mfedderly:
A brief write up of the release process in its current state so we have something to work from in the future.


RELEASING.md
```diff
@@ -0,0 +1,18 @@
+# Release process notes
+
+## Prerelease
+
+- create a new branch with some name
+- write up changelog changes (manually), commit that
+- yarn lerna version --no-push 6.2.0-alpha.2
```
rowanwins, line 7: Perhaps we can drop some notes as subpoints where required which explains what the cmd does and what we'd expect to see, eg 
- yarn lerna version --no-push 6.2.0-alpha.2
  - `yarn lerna version` bumps the version of all packages changed since the last release. 
  - Adding `--no-push` means the changes aren't pushed to git, we do this because.... 
  - We manually specify the tag at the moment because we're doing alpha releases but down the track we can use keywords like `minor` or `patch`.

RELEASING.md
```diff
@@ -0,0 +1,18 @@
...
+- git push origin --follow-tags \$branch
+- make PR
+- merge PR
+- re-fetch the new master and check it out locally
+- yarn install
+- yarn lerna run prepare
```
rowanwins, line 13: Similar to above
- yarn lerna run prepare
  - Runs the `prepare` script in each package directory (which does the compilation from ts to js)

rowanwins (approved):
Awesome

Issue #1671, opened 2019-05-02
greatCircle returns MultiLineString


csimpi:
Please provide the following when reporting an issue:

- [ ] GeoJSON data as a [gist file](https://gist.github.com/) or [geojson.io](http://geojson.io/) (filename extension must be `.geojson`).
- [x] Snippet of source code or for complex examples use [jsfiddle](https://jsfiddle.net/).

Dear Maintainers,
thank you so much for this awesome library!

I have some troubles with `greatCircle`, sometimes it returns with `MultiLineString` instead of `LineString`. I can't find any clue about this in the documentation, it says the method returns as a `LineString`.

For example:
Origin:
```
{lng:-122.389881,lat:37.615215}
```
Destination:
```
{lng:126.451234,lat:37.448526}
```

Reproduce and explanation:
https://jsfiddle.net/Lx2gtyzp/1/

I suppose this is happening when the link crosses the 180th longitude.
Is there any proper/better way to show the arc as a LineString? Should I concatenate the coordinates? 

**My real problem with this is that** I have to use `turf.along()` with the generated arc, and unfortunately  `turf.along()` can't handle MultiLinestring. It would be great if `turf.along()` would be able to work with MultiLineString the same way is with LineString.

rowanwins:
Hi @csimpi 

It looks like your output line is being sliced in 2 by the anti-meridian. Unfortunately the anti-meridian is one of the complicated beasts of the spatial world ([see this blog post](https://macwright.org/2016/09/26/the-180th-meridian.html)).

I'll have a think about how you could possibly tie in `along` with that output but I'm not sure I'm going to be able to come up with anything....

csimpi:
@rowanwins Thank you, the blog post is very interesting, I'll dive into deeper. I'd use a combination of angle guessing+closer way between two points (that crosses 180th). 
Btw, Google handles the situation pretty well:
https://goo.gl/maps/73L5LRi4NEhqB7x37 

Let me know if you have any ideas with `along`. + MultiLine. I have no more ideas today :)

csimpi:
After a lot of studying it came up the GeoJSON standard suggests to use `MultiLineString` in antimeridian cases so this part is OK.
The real problem is that[ `turf.along()` doesn't support `MultiLineString`.](https://github.com/Turfjs/turf/issues/1699)
https://github.com/Turfjs/turf/issues/1699

My workaround:
https://jsfiddle.net/bw1rxs5h/3/

PR #1928, opened 2020-07-14
Include MultiLineString output type for `greatCircle` module

kronick:
This PR updates the Typescript type definitions and corresponding docs for the `greatCircle` function to include the possibility of returning a `MultiLineString` when the input spans the antimeridian. This is described as [expected behavior in this issue](https://github.com/Turfjs/turf/issues/1671). Without this fix, Typescript will return an error if you try to test, for example:
```typescript
const greatCircleGeojson = greatCircle(start, end);
if(greatCircleGeojson?.geometry.type === "MultiLineString") {
 // ... handle this case
}
```

Typescript complains: `This condition will always return 'false' since the types '"LineString" | undefined' and '"MultiLineString"' have no overlap.`


kronick:
@rowanwins Thanks for the ✅ — will you handle merging this for the next release? Is there anything else I need to do?

regileeso:
I'd love to see these changes in a new release as soon as possible 👍

Issue #1940, opened 2020-09-05
Node fails to import ESM


diachedelic:
### Versions
- @turf/helpers v6.2.0-alpha.2
- Node.js v14.9.0

```js
// ./foo.js
import { point } from "@turf/helpers";
console.log(point);

// ./package.json
{ "type": "module" }
```

```sh
$ node foo.js
import { point } from "@turf/helpers";
         ^^^^^
SyntaxError: The requested module '@turf/helpers' is expected to be of type CommonJS, which does not support named exports. CommonJS modules can be imported by importing the default export.
For example:
import pkg from '@turf/helpers';
const { point } = pkg;
```

Replacing `"@turf/helpers"` with `"@turf/helpers/dist/es/index.js"` the same error message. Now, if I add `"type": "module"` to `node_modules/@turf/helpers/package.json`:

```sh
$ node foo.js 
[Function: point]
```

However, if I change the import back to `"@turf/helpers"`:
```sh
$ node foo.js 
file:///Users/jamesdiacono/Code/monorepo/pkg/app/foo.js:1
import { point } from "@turf/helpers";
         ^^^^^
SyntaxError: The requested module '@turf/helpers' does not provide an export named 'point'
    at ModuleJob._instantiate (internal/modules/esm/module_job.js:98:21)
    at async ModuleJob.run (internal/modules/esm/module_job.js:137:5)
    at async Loader.import (internal/modules/esm/loader.js:165:24)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
```

It appears that if package.json specifies `"type": "module"`, then the `"module"` property is ignored! More discussion at https://github.com/nodejs/modules/issues/432.

diachedelic:
It may be a solution to use the `.mjs` file extension for the es module:

```diff
// package.json
-"module": "dist/es/index.js",
+"module": "dist/es/index.mjs",
```

PR #1942, opened 2020-09-09
Node ESM compatibility

diachedelic:
Does two things to ensure turf ESM packages can be imported in Node:

1. Configures the `"exports"` in package.json, a successor to the `"module"` property.
2. Create a package.json containing `{"type":"module"}` in `dist/es/`. This ensures that `dist/es/index.js` is loaded as an ESM module, even if the package's package.json is *not* `"type": "module"`


diachedelic:
Fixes #1940

mfedderly:
Thanks for catching this! There's one package that isn't quite the same as everything else, can you make the same update in `packages/turf/package.json` as well?

diachedelic:
@mfedderly I thought I caught that one?

mfedderly:
Oh I see - not all of the changes in this PR came from the monorepolint changes. You can actually do all of this from the .monorepolint.config.js directly. I pushed a branch that I think does what you want.

https://github.com/Turfjs/turf/commit/1d360cc3e7ef9bac753a05446fe3e31b2822dec9

If you can verify that the change I made was correct and update your PR I can get this merged tomorrow. I'm pretty sure if you pull my own commit in it will block me from being an approver on a branch I contributed to, so its easiest if you make your own commit with those changes if possible.

diachedelic:
Are you saying that you make that change to `.monorepolint.config.js` and then run `mrl check --fix`, rather than modifying all of the individual package.json's manually?

diachedelic:
Okay I've incorporated your change.

mfedderly (approved):
Thanks! I mostly consume Turf through webpack, so I would've missed this entirely.

lcrosetto:
It is breaking on webpack v5 because of missing file extensions on imports, as described here:

https://nodejs.org/api/esm.html#esm_mandatory_file_extensions

Sample error:

```
ERROR in ../node_modules/@turf/polygonize/dist/es/index.js 3:0-38
Module not found: Error: Can't resolve './lib/EdgeRing' in './node_modules/@turf/polygonize/dist/es'
Did you mean 'EdgeRing.js'?
BREAKING CHANGE: The request './lib/EdgeRing' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ../node_modules/@turf/turf/dist/es/index.js 64:0-57 64:0-57
```

mfedderly:
@lcrosetto what version? 6.2.0 released today or another version?

lcrosetto:
> @lcrosetto what version? 6.2.0 released today or another version?

Yes, 6.2.0. Some of the imports from ./lib would need the file extension added, e.g.

`import EdgeRing from "./lib/EdgeRing.js";`

Issue #1547, opened 2018-12-08
Buffer distance


flightsurvey:
I have created a buffer around a geojson object in Leaflet in the following way:

```
var geoJSON ={
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              -4.491606,
              54.143007
            ],
            [
              -4.499331,
              54.139085
            ],
            [
              -4.497614,
              54.13828
            ],
            [
              -4.495983,
              54.139135
            ],
            [
              -4.495296,
              54.138783
            ],
            [
              -4.490061,
              54.141146
            ],
            [
              -4.491606,
              54.143007
            ]
          ]
        ]
      }
    }
  ]
};

var area = L.geoJSON(JSON.parse(geoJSON)), {
	color: 'green',
	weight: 2,
	onEachFeature: function (feature, layer) {
		layer.on("click", function (e) {
			map.fitBounds(layer.getBounds());
		});
	}
}).addTo(map);

var distance = 50;

var buffer = L.geoJSON(turf.buffer(JSON.parse(area), distance / 1000, {units:'kilometers'}), {
	color: 'orange',
	weight: 2,
	onEachFeature: function (feature, layer) {
		layer.on("click", function (e) {
			map.fitBounds(layer.getBounds());
		});
	}
}).addTo(map);
```

I then measured a perpendicular line between the area and the buffer - albeit freehand but it seems to be pretty true - and it appears to be around 44m.

To view the output paste the following into geojson.io:

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              -4.491008208047599,
              54.14317729554968
            ],
            [
              -4.4910793289379,
              54.14324457294276
            ],
            [
              -4.491172202002483,
              54.14330203846458
            ],
            [
              -4.491282991531243,
              54.14334731869702
            ],
            [
              -4.4914071217792,
              54.14337854349127
            ],
            [
              -4.4915394659636,
              54.14339442321583
            ],
            [
              -4.491674558029246,
              54.14339430202498
            ],
            [
              -4.4918068184314475,
              54.14337818494692
            ],
            [
              -4.491930784606396,
              54.14334673767213
            ],
            [
              -4.492041336604907,
              54.143301259052194
            ],
            [
[...]

Issue #1484, opened 2018-09-14
Error in the radius of turf.buffer


Fededgs:
Working with turf.buffer I notice that the buffering figure of a Polygon was not correct.
In particular, the radius (expressed in meter) wasn't respected during the dilatation.

I report below a test done for buffer a point.
(I know that there is turf.circle but i want to demonstrate the error for the radius).
Code:
```
   var point={
                    "type": "Feature",
                    "properties": {},
                    "geometry": {
                        "type": "Point",
                        "coordinates": [
                        7.237860560417174,
                        45.12233909279541
                        ]
                    }
                   };
   var p = turf.buffer(point, 50, {units:'meters'});
   console.log(turf.area(p));
```
Console:
```
  3885.60
```
Check resultant area:
3885.60=![image](https://user-images.githubusercontent.com/32876183/45543315-909f0f00-b814-11e8-9863-2a7ea1f632c1.png)* R² --> R=35.17m

So,the radius doesn't match the radius of the function.
(It was acceptable if the area was higher for the terrestrial curvature)

With other test, i find out that using radius=71.1 with 'meters', the area after buffer is 7857.98mq and R=50.01m with the formula used before.

Using turf@5.1.6

thanks

rowanwins:
Hi @Fededgs 

Yeah there are some known issues with `@turf/buffer` at the moment, perhaps for the timebeing try rolling back to v4.7.3, and it'll be addressed in v7.

PS Thanks for the handy test case, I'll try and add something like that into the test suite as it should hopefully help avoid mistakes like the one that's currently causing grief.

morganherlocker:
This is likely caused by geodesic and non-geodesic calculations being mixed, but would need to do more digging to confirm.

jonnedeprez:
I just calculated a 13km buffer around a point, and measured 17.6km for the diameter of the resulting circle, which is obviously too small. Interestingly, when I take two opposite points on that same circle, and then for both swap their longitude and latitude, the distance between those two new points is... 26km :-)

Seems like this is a pretty simple bug to me!

JamesLMilner:
@chrispahm thanks for confirming! I will close. Folks can follow up with a new issue if they are still seeing any problems.

Issue #1246, opened 2018-02-02
Testing turf.buffer accuracy


SteveCline:
I am getting a rather large difference comparing turf.buffer to the result of a buffer generated in GIS software. The GIS buffer was projected in MD state plane, then queried to a geoJson feature using SR 4326 via esri's REST api. Using a leaflet measure tool to ground truth, the GIS buffer was accurate. Does anyone have a good explanation?

```
var point = turf.point([-76.675801, 38.935749]);
var buffered = turf.buffer(point, 10, {units: 'miles'});
```
![turf](https://user-images.githubusercontent.com/32917766/35756134-24fb5146-0838-11e8-9638-30fa227b12f0.jpg)

rowanwins:
Hi @SteveCline 

I've just run the operation and opened the result in QGIS, the width measured in QGIS comes in at 20 miles as expected however the height comes in around 15miles. 

Under the hood turf/buffer projects input geometries to a [transverse mercator projection](https://github.com/d3/d3-geo/blob/master/README.md#geoTransverseMercator), I suspect this is the cause of the discrepancy between the results you're getting. It's not documented that it's happening and I wonder if we'd be better off providing an option allowing people to pass in a chosen projection?

Projections are one of those complicated things and Turf generally tries to avoid providing options around them to users, we generally go for a 'just works as expected' philosophy, although buffering is one are in particular where I think we could perhaps provide some more options for educated users.

Hope that helps explain things
Cheers
Rowan

SteveCline:
Thanks @rowanwins 

I think passing a spatial reference code as an option would be very helpful for the buffer and probably some other methods too. I would argue, if turf's philosophy is to 'just work as expected', using proper spatial reference is vitally important to that expectation of working.

Mid latitude buffers will be off by 25% and running turf.circle will give a more accurate result in these parts of the globe.

If turf has the goal of being the geoprocessing engine of the future, which I hope it does, it can't have inaccuracies this large and be used in serious GIS applications. Thanks again for your help and good work with Turf!

rowanwins:
No worries @SteveCline - we'll have a think about the best way we can expose that functionality.

[...]

PR #1956, opened 2020-10-12
[@turf/buffer] Fix projection issues that produce undersized buffers at non-equatorial latitudes

### LINKED ISSUE #1470: Breaking change in buffer between 4.7.3 and 4.7.4
> Please provide the following when reporting an issue:

- [x] GeoJSON data: https://gist.github.com/garthk/888c7a0b8a464ec02a3bd1c4d08f8882
- [x] Snippet of source code: https://github.com/garthk/turf-buffer-change

There was a breaking change to `@turf/buffer` in a patch release:

> Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API.

The schema is the same, but I strongly disagree with the proposition that a result change this big is backwards compatible.

dyakovlev:
Hi folks, my company recently had an issue of undersized buffers reported that prompted me to spend some quality time in Turf and JSTS source. It seems that this has been an issue reported a number of times (#1246, #1470, #1484, #1547). I think I've found the underlying issue to be a misconfiguration in the projection used in the `buffer` implementation. I hope this can help!

---

---

The existing buffer code has three problems:
- it uses two different projections depending on the latitude of the input geometry. This results in an error discontinuity at 50N.
- it switches projections only for the northern hemisphere.
- it reverses the (lnglat) center, which has the eventual result of rotating the projection to a skewed center.

The third problem is one of the causes of `buffer`'s size issues. At latitudes above 50N, the projection distorts input distances due to not being rotated to the correct input geometry center. At latitudes below 50N, the Mercator projection does not receive a correction for the input distance. If using the Mercator projection, the actual distance used as an input to JSTS BufferOp should be distance / cos((lat * π) / 180). [1]

As recommended in the original[2] exploration that prompted the fix introduced in #938, this patch switches the buffer prep code to consistently using the Azimuthal Equidistant projection rather than switching between Mercator and Transverse Mercator. When centered on the input geometry, this projection should produce less distortion in all directions for input geometries of reasonable size compared to a cylindrical one.

[1] https://gis.stackexchange.com/a/347573
[2] https://github.com/w8r/moscow-rings#solution

---

Question - as mentioned in the contributing guidelines, I regenerated READMEs using the provided script. This produced a lot of diffs to links and looked like either a tool misconfiguration or drift in some underlying system. I've chosen to omit this for the time being; was that the correct choice?

Validation - I've visually checked the generated output geojson. Wondering if there's an automated test that would be reasonable to implement - maybe construct a point grid, buffer all of the points, verify their radii to be within some ε of the input distance?


dyakovlev:
tagging @rowanwins @stebogit as people with some context

mattzucker:
Thank you for this fix!  Tried a couple of tests with my project and the distance projections are so much better.
The only thing I'm having trouble with is getting a minified release build.  I ran `yarn lerna run prepare` which generated packages/turf/turf.min.js and packages/turf/dist/*

If I use the index.js file from dist/ then everything works great, but turf.min.js throws this error when trying to use buffer:
`Uncaught TypeError: l is not a function
http://localhost:8080/scripts/turf.min.js:1`

It's difficult for me to debug it because of the minified file, and the non-minified file doesn't error.  I don't even know if this is a bug with your pull request or my environment, but could you please test that for me?

dyakovlev:
Thanks for checking it out, @mattzucker. I'm seeing the same issue and can confirm that it's something to do with my change - works fine in `master`.  I'll see if I can get it working..

rowanwins (approved):
This looks really great thanks @dyakovlev - sorry it's taken so long to review!

The code is much simpler to follow AND produces more consistent and more accurate results!

mfedderly:
For the `l is not a function` error, this likely has to do with how the (out of date) rollup build is handling d3-geo's aziumuthalEquidistant.js file, specifically the projection import.

https://github.com/d3/d3-geo/blob/master/src/projection/azimuthalEquidistant.js#L14

dyakovlev:
Do you think there's a tree shaking issue of some sort. Is there any followup I can help with here? Happy to spend some cycles in turf.

mfedderly:
I'm going to go ahead and just upgrade our whole rollup setup from 0.55.5 to 2.34.2 and hope that fixes the issue. That's nearly 3 years of rollup development so I'm hoping its just magically fixed once I'm done. Its been on my list of improvements to make but I haven't had a compelling reason to do it immediately until now. If that doesn't fix it I'll re-ping you though.

I really appreciate the help, when I joined the Turf project my goal was to do a lot of build and infrastructure work, so some of the packages are still a mystery to me. I'm glad you were able to make such a great fix.

mfedderly:
@dyakovlev the upgrade seems to have done the trick, you're off the hook this time :)

Issue #1941, opened 2020-09-09
point() no longer throws if given invalid args


diachedelic:
As of v6.2.0-alpha.2, you can pass `@turf/helper`'s `point` function literally anything and it will not complain. This a breaking change for me - if it's intentional could it be mentioned in the changelog?

```js
import { point } from "@turf/helpers";
point("hello");
// {
//  type: 'Feature',
//  properties: {},
//  geometry: { type: 'Point', coordinates: 'hello' }
// }
```

mfedderly:
Looks like this got dropped when helpers was converted to Typescript. Typescript would catch errors like this at compile time, but obviously not all of our consumers are using Typescript.

diachedelic:
Thanks!

PR #1964, opened 2020-11-16
Add point() runtime checks to fix API break

mfedderly:
Fixes https://github.com/Turfjs/turf/issues/1941

All I did was go back in the git history and grab these checks when they were removed in the conversion to Typescript and then add a test to prevent this from happening again.


morganherlocker (approved):
Looks great! :+1:

PR #1965, opened 2020-11-17
Add Prettier and ESLint to the builds

mfedderly:
Add ESLint and Prettier to the builds. ESLint is enabled and working, but all of the rules that caused errors have been disabled in order to make it easier to see the ESLint cleanup in future PRs. I also removed all references to TSLint as ESLint now can be used on Typescript and is the preferred path forward for linting.

The first has manual actions and the second is just applying prettier to bring everything into expected format going forward.


rheh (approved):
LGMT

rowanwins (approved):
👍

PR #1968, opened 2020-11-19
Bench + Test cleanup for faster builds and easier publish

mfedderly:
- No longer have bench and test call prepare explicitly
- bench.js, test.js, types.ts all use the local copy of the code, not the built copy
- Fixed quite a few bench.js files that were just broken
- Rework root package.json for build clarity
- Change prepare scripts to build so that they don't run on publish


rowanwins (approved):
Looks good!

PR #1975, opened 2020-12-09
Upgrade rollup to fix @turf/turf build issue

mfedderly:
Fixes a rollup build error on @turf/turf that we hit while fixing https://github.com/Turfjs/turf/pull/1956

This was actually surprisingly easy, I expected more breaks than what we wound up with.
I also removed a few of the older standardization scripts that are no longer necessary now that we have monorepolint doing its thing.


mfedderly:
@dyakovlev fysa

packages/turf/package.json
```diff
@@ -64,14 +64,14 @@
     "test": "echo '@turf/turf tests run in the last-checks step'"
   },
   "devDependencies": {
+    "@rollup/plugin-buble": "^0.21.3",
```
rowanwins, line 67: Are we actually using `@rollup/plugin-buble` anywhere... I couldn't find any use of it?

mfedderly: On interesting, I thought we were using it in @turf/turf's special rollup but I guess not. Its on my todo list to swap to babel anyhow, since babel is roughly 60x more popular. That would help fix other issues related to es6 syntax creeping in.

rowanwins: I'm happy for it to sit there unused and it'll eventually go with the switch to babel.

rowanwins (approved):
👍

PR #1978, opened 2020-12-10
Don't override Number.prototype.modulo

mfedderly:
It is weirdly side-effectful to override builtin prototypes and supposedly this way is more performant.


rowanwins (approved):
Approved!

PR #1979, opened 2020-12-10
[@turf/along] Implement percentage units

tuukka:
Would you be interested in implementing percentage units for `along` like this draft PR proposes?

### DIFF
--- a/packages/turf-along/index.ts
+++ b/packages/turf-along/index.ts
@@ -1,6 +1,7 @@
 import bearing from "@turf/bearing";
 import destination from "@turf/destination";
 import measureDistance from "@turf/distance";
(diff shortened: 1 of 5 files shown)


twelch:
Thanks for opening @tuukka, I like the spirit of this.  That said, I think a percentage is inherently unit-less, and isn't a good fit to add as a supported `Unit` type for the whole library.

With that I'm going to close this PR.  I think if you were to rework as an optional parameter such as `percent: boolean` that defaults to false and, if present, does the percent calculation and skips the units, I personally would accept it.  I think `lineSliceAlong` could get the same treatment.

tuukka:
@twelch Thank you for the feedback, makes perfect sense!

PR #1984, opened 2020-12-16
Only include ES5 code, remove some ES6 that had crept in

mfedderly:
This PR includes a bunch of small efforts to get Turf back to shipping es5-only code.

- Move @turf/polygonize to typescript instead of its own custom rollup. This had a custom rollup step that was only manually run, and the rollup output was being included in the module code. I moved this to typescript because it was making use of `class`.
- validate-es5-dependencies -> es-check. validate-es5-dependencies stopped actually failing builds, and is prone to some false positives. es-check seems to work well and we don't need to maintain it.
- @turf/buffer const -> var. Just a quick syntax change here.
- polygon-clipping dependency from 0.14.3 -> 0.15.2 to swap es6 code for es5 code. This cleans up some es6 code that was sneaking in via the dependency on splaytree.
- @turf/turf - Change rollup config to better support tree shaking, remove buble reference. This brings us more in line with the other packages. The browser target is still the combined minified version as expected.


rowanwins (approved):
Nice!

PR #1985, opened 2020-12-16
Remove @turf/isolines and @turf/isobands due to potential license conflicts

mfedderly:
These modules use some code from an older version of the marchingsquares package. That package is licensed with AGPL with a special additional wording that may make it compatible with MIT. Out of an abundance of caution we should probably remove this package and revisit adding it again in the future if there is demand.


stebogit:
@RaumZeit maybe you have some input on this? After all I know you were very specific when defining the licence for your library, so I trust you know this stuff. Do you think there is any actual limitation/incompatibility with (Turf's) MIT license here?
I would be available to import the latest MarchingSquares.js version into these two Turf modules if that license was more appropriate, if at all different.

I have been wanting to do so, which would likely fix several related issues, for a while, but I have been discouraged by the current condition of the project at the moment; I was waiting for @mfedderly to complete his attempt to fix the mess (thank you @mfedderly for your effort BTW!)

mfedderly:
@stebogit I did try an upgrade on isolines and there were some weird bugs still. You can see the geojson diffs here: 
https://github.com/Turfjs/turf/compare/mf/clarify-marchingsquares-licensing?expand=1&short_path=24ee86c#diff-24ee86cf81a547d1436258deb8a72b750fe45a81e4beeb4b10ffbb5031cfd9a5

At the end of the day I think having the GPL header in a file (current version) or linking to a library who's package.json contains the license entry for the same license is making people uncomfortable. I have seen the license file in that repo which does potentially remove the viral part of the GPL, but that nuance is hard to convey to tools that just look at the license entries in the package.json files.

mfedderly:
See https://github.com/Turfjs/turf/pull/1993

mfedderly:
Closing in favor of https://github.com/Turfjs/turf/pull/1993

PR #1993, opened 2021-01-05
Clarify MarchingSquares.js license

mfedderly:
There was a bit of confusion about the license on the MarchingSquares code. In order to make it clearer that this particular code is compatible with Turf's chosen MIT license, I went ahead and copied the entire license block from the source repo and put it inline to make things clearer for the next person that comes along.

Related original discussion: https://github.com/RaumZeit/MarchingSquares.js/issues/3

@stebogit @RaumZeit @rowanwins FYSA

There's also an unrelated small yarn.lock update that somehow wasn't caught by the build system.

### DIFF
--- a/packages/turf-isobands/lib/marchingsquares-isobands.js
+++ b/packages/turf-isobands/lib/marchingsquares-isobands.js
@@ -3,8 +3,38 @@
  * Copyright (c) 2015, 2015 Ronny Lorenz <ronny@tbi.univie.ac.at>
  * v. 1.2.0
  * https://github.com/RaumZeit/MarchingSquares.js
+ *
+ * MarchingSquaresJS is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


packages/turf-isobands/lib/marchingsquares-isobands.js
```diff
@@ -3,8 +3,38 @@
...
+ * Any modifications to MarchingSquaresJS, however, must be shared with the public
+ * and made available.
+ *
+ * In summary this:
+ * - allows you to use MarchingSquaresJS at no cost
+ * - allows you to use MarchingSquaresJS for both personal and commercial purposes
+ * - allows you to distribute UNMODIFIED VERSIONS of MarchingSquaresJS under any
+ *   license as long as this license notice is included
+ * - enables you to keep the source code of your program that uses MarchingSquaresJS
+ *   disclosed
```
RaumZeit, line 31: should be 'undisclosed' (see https://github.com/RaumZeit/MarchingSquares.js/pull/25)

rowanwins:
Looks good to me @mfedderly - I think there is enough material that should come up on keyword searches in this repo which point to the history of how this came about so I think it's good to proceed.

Issue #1499, opened 2018-10-02
v7 alpha release


rowanwins:
Inline with the suggestions contained within #1428 I've been working on getting a v7 release ready (see #1432)

## API Changes
- There haven't been any changes to module API's

## Higher level changes
- Modules are now [ES6 modules](http://2ality.com/2014/09/es6-modules-final.html), written in plain ol' js, and built into a single package using `rollup`
- Lerna is now no longer used for publishing as only a single package is published

## Module enhancements
- JSTS has been replaced in `union`, `difference`, `intersect` by [polygon-clipping](https://github.com/mfogel/polygon-clipping)
 -`dissolve` has been overhauled
- a number of other minor enhancements and fixes to modules...

## Testing
- My initial testing of the alpha bundle has been going ok
- @shongololo has raised an [issue](https://github.com/Turfjs/turf/pull/1432#issuecomment-425570132) around `ReferenceError: m is not defined`  when buffering linestrings.
  - need to do a bit of digging here, may be a minification issue or something else... although see note below, ideally I'm hoping to do away with the jsts dependency in the buffer module

## Outstanding issues 
  - The only module where `jsts` remains in use is buffer. I'm currently working on a replacement [here](https://github.com/rowanwins/geojson-buffer)
- In the buffer replacement above I'm trying to build my library pulling in a couple of turf modules (eg `coordEach`). For some reason the build is pulling in all of turf rather than just the couple of modules I want (aka tree-shaking doesn't appear to be working...)
  - In my case I'm again using `rollup` as my build tool. I could perhaps investigate using webpack although it would be nice to have turf 'just working' irrespective of build tool choice.

## Where to find it
- v7.0.0-alpha.1 has been [published to npm](https://www.npmjs.com/package/turf/v/7.0.0-alpha.1)
- Or it can be found in the v7 [branch](https://github.com/Turfjs/turf/pull/1432)

## What next
- Try to resolve tree-shaking issue noted above
- Publicise changes more widely

cc @stebogit @tmcw

shongololo:
@rowanwins this is a huge step forward, thank you.

Regarding tree shaking, I'm going by memory, but I believe that it didn't work with webpack either. I'll update this with further comments if I figure anything out.

rowanwins:
Yep confirmed webpack also isn't succesful in tree-shaking, need to do some more digging on the turf config end...

rowanwins:
Hmm so I've done some more digging and it found a few things
[...]

Issue #1257, opened 2018-02-12
How to create a buffer with flat linecap?


mickeyjohn:
Hi everyone,

Turf is a very nice geo lib. When using the .buffer function. It can create a nicely buffer polygon with a round linecap around the given polyline. My question is how to create a polyline with flat linecap with turf? Is there any way to achieve that?

Best Regards,
John

rowanwins:
Hi @mickeyjohn 

At the moment it's not possible with Turf. Using the [jsts](https://github.com/bjornharrtell/jsts) lib (on which Turf currently relies) I believe it's doable.

We do already have an [item](#639) to support endcap strategies so Im going to close this in favour of that issue. We're currently trying to remove the dependency on jsts as it is a very large library but when we do we'll try and make sure we have those endcap options in place.

Cheers
Rowan

Issue #639, opened 2017-04-02
Requesting for more strategies in turf buffer


dhivehi:
Currently turf buffer seems to support "end_round" as end strategy and "join_round" as join strategy. It would be nice if your can provide the strategies as follows:

- For join strategies, 'join_round' and 'join_miter'.
- For end strategies, 'end_round' and 'end_flat'. 

More ref: https://dev.mysql.com/doc/refman/5.7/en/spatial-operator-functions.html#function_st-buffer-strategy

rowanwins:
Thanks for the suggestions @dhivehi 

Currently the buffer module is a wrapper around jsts and we're trying to remove this dependency. Once we've done that hopefully we'll be in a better position to implement those additional suggestions.

Cheers

nboisteault:
:+1:

bartvde:
Are there any plans for incorporating this into turf? I recently did this by using mainstream jsts instead of the turf ES6 fork of JSTS, and passing in the endCap here:

```
  // 2 is endCap flat
  let buffered = BufferOp.bufferOp(geom, distance, undefined, 2);
```
but it required copying some of the turf buffer code to achieve it.

Any guidance appreciated. Are you open for PRs to change to mainstream jsts and add this to the buffer operation? Or what would be a good strategy?

oli-clive-griffin:
@bartvde long shot but are you able to remember how using jsts went? I'm really needing a mitre join buffer in js and this is looking like one of the best options

axekan:
For those wanting to use this right now using jsts, you can use [patch-package](https://github.com/ds300/patch-package) to add the required changes to 6.5.0 until support for this gets added. 

Below is the patch file I used. With these changes you can run :

```typescript
buffer(feature, 100, { units: 'meters', joinStyle: 'mitre' })
```
The documentation for the parameters can be found in the docs for the java library https://locationtech.github.io/jts/javadoc/.
I'm unsure what the purpose of the `steps` parameter was in the original version, but I used it as the 'Quadrant segments' parameter in `BufferParameters`.

```diff
diff --git a/node_modules/@turf/buffer/dist/es/index.js b/node_modules/@turf/buffer/dist/es/index.js
index 0ba5731..efe8a35 100644
--- a/node_modules/@turf/buffer/dist/es/index.js
+++ b/node_modules/@turf/buffer/dist/es/index.js
@@ -1,9 +1,22 @@
 import center from '@turf/center';
[...]

PR #1994, opened 2021-01-06
Add available end cap style options from turf-jsts to @turf/buffer

chrispahm:
This PR adds an option to @turf/buffer to specify different end cap styles (round = default, flat, square) for (line) buffers (solving #1257 and #639). 

The change to the codebase is minimal, as the [turf-jsts](https://github.com/DenisCarriere/turf-jsts/blob/da0872e55b4ec170d60ea13f464fec696ee127fa/src/org/locationtech/jts/operation/buffer/BufferOp.js#L119) dependency already incorporates this feature and we would just be passing the option along.

I am fully aware that the issues #1257 and #639 have been originally closed because of ongoing efforts to remove the JSTS dependency entirely (e.g. #88 and the V7 branch). However, following the discussion around #1499, it seems like there is no viable alternative to JSTS regarding the buffer operation yet. In the meantime, I would argue it is fair to users to expose the full functionality of turf-jsts and its buffer operation since the relevant code is included anyway.

### Demo
https://observablehq.com/@chrispahm/turf-js-buffer-end-cap-styles

Please fill in this template.


aaronkyle:
Nice.  These endcaps really make things look prettier!

throberto:
Good! I need it!

shenzhuxi:
It's a great feature to have. 
But with this fork, if I use flat end and 1 step on a 2-points line, it will return 5 points. 
```
line = turf.lineString([start, end]);
buffered = turf.buffer(line, 2, {units: 'meters', endCapStyle: 'flat', steps: 1});
console.log(buffered.geometry.coordinates); //[Array(5)]
```
Shouldn't it return a rectangle?

chrispahm:
Yup, that's due to the way LinearRings are defined in the [GeoJSON spec](https://geojson.org/geojson-spec.html):

> A LinearRing is closed LineString with 4 or more positions. The first and last positions are equivalent (they represent equivalent points).

Here's a great blog post explaining the GeoJSON spec with great examples: https://macwright.com/2015/03/23/geojson-second-bite.html

shenzhuxi:
Thanks @chrispahm !
The code seems very straight forward. Why hasn't it been merged yet?

shenzhuxi (approved):

shenzhuxi (approved):

chrispahm:
Hey @shenzhuxi,

Thanks for your approval, however the PR requires at least 1 approving review by one of the maintainers (@mfedderly and @rowanwins) or anyone with write access to the repo in order to be considered for merging.

Anyways, thanks for having a look!
Best
Chris

JamesLMilner:
Hi @chrispahm, I can see that it is a simple and useful change so just want to update here; I think the reason this hasn't been merged yet is because we were looking at removing JSTS completely from the turfjs - I think @rowanwins has been rewriting the buffer library which would mean if we merge this we would have to maintain compatibility which might slow us getting the new buffer implementation in. JSTS is a huge pain for us because it creates very big bundle sizes because of it's size, hence wanting to remove it.

chrispahm:
Hi @JamesLMilner,

Absolutely understanding your point and the issues with JSTS:

> I am fully aware that the issues https://github.com/Turfjs/turf/issues/1257 and https://github.com/Turfjs/turf/issues/639 have been originally closed because of ongoing efforts to remove the JSTS dependency entirely (e.g. https://github.com/Turfjs/turf/issues/88 and the V7 branch). However, following the discussion around https://github.com/Turfjs/turf/issues/1499, it seems like there is no viable alternative to JSTS regarding the buffer operation yet. In the meantime, I would argue it is fair to users to expose the full functionality of turf-jsts and its buffer operation since the relevant code is included anyway.

I'm just wondering if the switch to [geojson-buffer](https://github.com/rowanwins/geojson-buffer) (or any other library) will happen anytime soon, especially considering the discussion in https://github.com/rowanwins/geojson-buffer/pull/4#issuecomment-486229393. 
Therefore, might it not just be (more) sensible to just switch to upstream jsts as a start or even use @rycgar's [BufferOp port](https://github.com/Turfjs/turf/issues/88#issuecomment-850333577) instead? Thought this was a similar to what you were proposing https://github.com/Turfjs/turf/issues/88#issuecomment-868422036 😊

GetecHarry:
Hi just wanted to drop in and say that this PR should definitely be approved in my opinion, this is exactly what we needed and Chris was super helpful in providing the minified JS file when emailing him. Thanks again!

PR #1996, opened 2021-01-11
[@turf/tag] Confirm support for multipolygons and add to TypeScript definitions

bryceroney:
Hi there

This is my first time contributing to a large OSS project, so I hope I have met the requirements in the contributing guide.

I've started using TypeScript more for my JavaScript projects, and when working on a project using turf noticed that the definitions for `@turf/tag` do not include MultiPolygon as well as Polygon.

This PR adds that definition to the TypeScript file, but also adds tests to ensure it does work if changes are made to `@turf/boolean-point-in-polygon`

I am not sure why yarn.lock changed.


mfedderly (approved):
Looks good. Thanks for adding the tests as well!

The yarn.lock change was leftover from a different PR but has already been fixed on the master branch. After updating your PR with the latest it went away.

PR #1997, opened 2021-01-11
Remove '.prettierrc.json'

mondeja:
An empty Prettier configuration file was added in #1965, but is useless since does not define any configuration.


mfedderly:
This opts us into the prettier defaults, and lets editors that are set to format on save when there is a prettier config work correctly.

PR #2008, opened 2021-01-18
Overhaul approach for dissolve module

rowanwins:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Have read [How To Contribute](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#how-to-contribute).
- [x] Run `npm test` at the sub modules where changes have occurred.
- [x] Run `npm run lint` to ensure code style at the turf module level.

This PR completely rewrites the dissolve module, the logic is now a lot simpler, and performance is improved.

Resolves #1806
Resolves #1912 
Resolves #1237
Resolves #1208
Resolves #742 
Resolves #2027
Resolves #2074

PR #2010, opened 2021-01-18
Update changelog for 6.3.0

mfedderly:
I'd like to make a quick follow-up release to mostly fix the package.json import issues, but we might as well pull in the updates you made as well while we're at it.

I'll rebase this after https://github.com/Turfjs/turf/pull/2007 and https://github.com/Turfjs/turf/pull/2008 are merged and then do the release.


rowanwins (approved):
👍

PR #2011, opened 2021-01-18
Add .js to relative import statements in es output at build time

mfedderly:
In https://github.com/Turfjs/turf/pull/1942#issuecomment-759070517 it was pointed out that Turf does not quite work in Webpack 5 without a [workaround](https://github.com/webpack/webpack/issues/11467#issuecomment-691873586).

This change adds a build-time transform onto the dist/es output files that adds the .js extensions to all of the outputs.

I would have preferred to specify the .js extension directly in the imports in the source, but that was not working well with the way we use ts-node to run test and bench. I'm a little worried that this is a little fragile because it is a regular expression, but I'm hoping that this ages gracefully and prevents us from needing an entire codemod framework to do the same thing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lint:escheck-js": "es-check es5 packages/*/dist/js/index.js packages/turf/turf.min.js",
     "lint:escheck-es": "es-check --module es5 packages/*/dist/es/index.js",
     "postlint": "documentation lint packages/turf-*/index.js",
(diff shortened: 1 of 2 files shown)


mfedderly:
@lcrosetto FYSA

rowanwins (approved):
I set this up locally and it seems straight forward enough, a bit finickity that we have to do it but 🤷‍♂️ what can you do, maybe one day we'll be able to dispense with it.

PR #2016, opened 2021-01-22
turf-transform-rotate/turf-ellipse: Fix documentation for angle parameter 

### LINKED ISSUE #2015: Direction of rotation in transformRotate?
The documentation for transformRotate says `angle: of rotation (along the vertical axis), from North in decimal degrees, negative clockwise` and `all rotations follow the right-hand rule` (which here means positive = counterclockwise, right?).

But the example in http://turfjs.org/docs/#transformRotate shows that a positive angle (10 degrees in the example) rotates the triangle **clockwise**.

(And this would be consistent with the "bearing" parameter for some other Turf.js functions, which explicitly say `positive clockwise`).

Which is the intended behavior here? 

(btw, turf/ellipse has the same description for its "angle" parameter, and it uses transformRotate internally).

pasieronen:
Fixes #2015 

Please fill in this template.


mfedderly (approved):

mfedderly:
Thanks!

PR #2030, opened 2021-02-10
[@turf/line-slice-along] Fix offset distance equal to line length

EricPKerr:
`@turf/line-slice-along` currently returns an invalid GeoJSON `LineString` if you pass in an offset that is equal to the line length.

The end of the module returns:
```
return lineString(coords[coords.length - 1]);
```

Which results in:
```
{
  type: 'Feature',
  properties: {},
  geometry: { type: 'LineString', coordinates: [ lon, lat ] }
}
```

This is a Point and not a valid LineString geometry (list of coordinates).  The modification here returns the last point as a valid LineString geometry.


mfedderly (approved):
Thanks!

PR #2033, opened 2021-02-13
Rewrite line-intersect module

### LINKED ISSUE #2024: useless comparison inside line intersection algorithm
https://github.com/Turfjs/turf/blob/c952c44a25cb9ceba82f140b2b0407d8484afac4/packages/turf-line-intersect/index.ts#L116-L121

the function will always return null after denom === 0, the if `(numeA === 0 && numeB === 0) {` comparison is either useless or should be outside the if

### LINKED ISSUE #2029: lineStringsIntersect in geojson-utils a lot faster than turf.lineIntersect
I have a usecase where I need to check for a lot of lines where they intersect which a lot of Polygons. I really like turf and I use it a lot in my project but I noticed that turf.lineIntersect is relatively slow.

I then stumbled upon https://github.com/maxogden/geojson-js-utils (which does not seem to be actively maintained anymore) and their lineStringsIntersect method.

This seems to be (in my usecase) significantly faster and I wonder why. In case the algorithm is just way quicker, maybe it would be cool to integrate the faster algorithm into turf?

rowanwins:
Please fill in this template.

This PR provides an overhaul of the `line-intersect` module primarily to make it more performant for larger geometries. The previous implementation used a fairly brute force approach while this PR make uses a sweepline.

I did have to comment out one test for handling `GeometryCollection` as I couldn't make typescript happy.

The benchmark shows the improvement, the only exception is the first case which i only a single segment.
````
// CURRENT
2-vertex-segment x 4,123,821 ops/sec ±12.92% (74 runs sampled)
double-intersect x 53,118 ops/sec ±4.22% (72 runs sampled)
multi-linestring x 16,417 ops/sec ±2.31% (77 runs sampled)
polygons-with-holes x 9,739 ops/sec ±2.55% (85 runs sampled)
same-coordinates x 51,303 ops/sec ±4.23% (71 runs sampled)

// PROPOSED
2-vertex-segment x 1,467,485 ops/sec ±1.74% (89 runs sampled)
double-intersect x 307,665 ops/sec ±1.70% (91 runs sampled)
multi-linestring x 81,337 ops/sec ±0.67% (94 runs sampled)
polygons-with-holes x 27,711 ops/sec ±0.70% (92 runs sampled)
same-coordinates x 521,277 ops/sec ±0.75% (92 runs sampled)
````

Resolves #2024
Resolves #2029

### CODE REVIEW

File: packages/turf-line-intersect/index.ts [hunk 1]
```diff
@@ -35,100 +35,56 @@ function lineIntersect<
...
-    featureEach(tree.search(segment), (match) => {
-      const intersect = intersects(segment, match);
-      if (intersect) {
-        // prevent duplicate points https://github.com/Turfjs/turf/issues/688
-        const key = getCoords(intersect).join(",");
-        if (!unique[key]) {
-          unique[key] = true;
-          results.push(intersect);
-        }
+  const intersections = findIntersections(featureCollection(features));
```
mfedderly, line 73: This seems like it works differently than the previous version. Before it would compare line1 to line2, but now if you have a FeatureCollection for line1 it will be intersected with its own geometries?

mfedderly (changes requested):

packages/turf-line-intersect/index.ts
rowanwins: Hmm yes good pickup on a subtle difference I hadn't thought of - this new implementation will find self-intersections in a geometry, as well as intersections between two geometries. I could possibly add something to the `sweepline-intersections` lib to handle that if we thought that was problematic.

packages/turf-line-intersect/test.js
```diff
@@ -78,13 +76,13 @@ test("turf-line-intersect - Geometry Objects", (t) => {
...
     "support Feature Collection"
   );
-  t.ok(
-    lineIntersect(
-      geometryCollection([line1.geometry]),
-      geometryCollection([line2.geometry])
-    ).features.length,
-    "support Geometry Collection"
-  );
+  // t.ok(
```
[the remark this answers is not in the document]
rowanwins: TS 🤷‍♂️  . Use of `GeometryCollection` is such a rarity I'd be happy to delete the test. It's also not specified in the docs as a supported input so whether anyone would notice or not who knows.

packages/turf-line-intersect/index.ts
```diff
@@ -10,17 +10,17 @@ import {
...
+import findIntersections from "sweepline-intersections";
 
 /**
  * Takes any LineString or Polygon GeoJSON and returns the intersecting point(s).
  *
  * @name lineIntersect
  * @param {GeoJSON} line1 any LineString or Polygon
  * @param {GeoJSON} line2 any LineString or Polygon
+ * @param {Object} [options={}] Optional parameters
+ * @param {number} [options.removeDuplicates=true] remove duplicate intersections
```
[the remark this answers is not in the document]
rowanwins: 👍 
Slowly getting used to writing jsdoc again!

rowanwins:
Also resolves #2024

matthiasfeist (approved):

packages/turf-line-intersect/index.ts
mfedderly: Its a breaking change in a way that isn't really possible to even work around as a library consumer. That seems pretty bad.

rowanwins: I've updated my sweepline-intersections library to allow libraries to specify whether they want to detect self-intersections or not, and made the subsequent change in how we use it here in Turf. So I think this is good to go now.

mfedderly:
Any chance you can change this back to your npm sweeplines-intersections package and get it to pass the build that way like with point-in-polygon-hao? The lack of licensing header and babeled code breaking eslint makes me feel bad.

mfedderly:
@rowanwins I think the build will pass if you go back to depending on your npm package now that we started targeting es2017? We should also add that package name to https://github.com/Turfjs/turf/blob/master/README.md#browser-support

twelch (approved):
I have one comment to remove geojson-rbush dep if no longer needed.  approving otherwise.

mfedderly (commented):
This looks awesome. Can you publish the typescript defs in the package itself? I made two PRs to accomplish this:
https://github.com/rowanwins/sweepline-intersections/pull/12
https://github.com/Turfjs/turf/pull/2274

packages/turf-line-intersect/package.json
```diff
@@ -68,6 +68,7 @@
     "@turf/invariant": "^6.5.0",
     "@turf/line-segment": "^6.5.0",
     "@turf/meta": "^6.5.0",
+    "sweepline-intersections": "^1.3.1",
```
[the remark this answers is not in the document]
rowanwins: Done with #2274

packages/turf-line-intersect/package.json
```diff
@@ -63,11 +63,8 @@
     "write-json-file": "*"
   },
   "dependencies": {
-    "@turf/geojson-rbush": "^3.2.0",
     "@turf/helpers": "^6.5.0",
-    "@turf/invariant": "^6.5.0",
-    "@turf/line-segment": "^6.5.0",
-    "@turf/meta": "^6.5.0",
+    "sweepline-intersections": "^1.3.1",
```
mfedderly, line 67: Gotta bump this to 1.4.0 and yarn install again. Hopefully I got the change right :)

twelch: See my comment for further improving the type signature - https://github.com/rowanwins/sweepline-intersections/pull/12/files#r831527331

mfedderly:
I'm going to go ahead and merge this

matthiasfeist:
Hi quick question as this PR is merged but it's not released yet to npm. When and how is that gonna happen?

diogormsf:
> Hi quick question as this PR is merged but it's not released yet to npm. When and how is that gonna happen?

+1 on the latest comment. I would love to see this being merged to npm, it would allow me to remove an algorithm created as a workaround.

pm0u:
I believe this PR introduced a breaking change that means that self intersections are checked by default where previously they were not.

https://github.com/Turfjs/turf/issues/2707

Issue #2017, opened 2021-01-22
TypeScript errors in turf 6 (TS2304 and TS2312)


everhardt:
I've updated @turf/turf from 5.1.6 to 6.3.0 and now TypeScript is complaining about the following when I transpile my project:

```
node_modules/@turf/center-median/dist/js/index.d.ts:59:29 - error TS2304: Cannot find name 'Position'.

59     medianCandidates: Array<Position>;
                               ~~~~~~~~

node_modules/@turf/clusters-dbscan/dist/js/index.d.ts:3:38 - error TS2312: An interface can only extend an object type or intersection of object types with statically known members.

3 export interface DbscanProps extends Properties {
                                       ~~~~~~~~~~

node_modules/@turf/clusters-kmeans/dist/js/index.d.ts:2:38 - error TS2312: An interface can only extend an object type or intersection of object types with statically known members.

2 export interface KmeansProps extends Properties {
                                       ~~~~~~~~~~
```

My tsconfig.json:
```
{
  "compilerOptions": {
    "module": "commonjs",
    "esModuleInterop": true,
    "allowJs": false,
    "allowSyntheticDefaultImports": true,
    "target": "es2019",
    "noImplicitAny": true,
    "strictNullChecks": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "moduleResolution": "node",
    "preserveSymlinks": true,
    "sourceMap": true,
    "outDir": "dist",
    "baseUrl": ".",
    "paths": {
      "*": [
        "node_modules/*",
        "src/types/*",
      ]
    },
    "typeRoots": [
      "./node_modules/@types",
      "./src/types"
    ]
  },
  "include": [
    "src/**/*",
    "test/**/*"
  ]
}
```

I just tried with 6.2.0 and that gives the same issue. For now, I've reverted back to 5.1.6

mfedderly:
Your typescript settings are likely more strict than Turf's. I'm hoping to enable all the strictness but in the meantime you might want to look into the `skipLibCheck` option.

https://www.typescriptlang.org/docs/handbook/compiler-options.html

everhardt:
Thanks for the quick response. I looked into that but I have some custom declaration files in the project as well, that I do like to get checked. I'm not in a hurry, so I'll just hold off on updating for now.

antonazaides:
Hi @mfedderly, I have a similar issue.

[...]

PR #2053, opened 2021-03-13
Work towards enabling typescript's strict mode

mfedderly:
This isn't *everything* for typescript's strict mode, but its most of it.

For the most part this is just using what the jsdoc comments were, or inferring the types and making sure the compiler was happy. For the most part no behavior changed, except for `@turf/boolean-crosses` where the MultiPoint+Polygon case was just wrong.

Still to do after this PR, most of them require types getting added to DefinitelyTyped. I'll work on that later.
- clusters-kmeans
- clusters-dbscan
- boolean-equal
- boolean-overlap
- polygonize


JamesLMilner (approved):
This looks good to me - should resolve https://github.com/Turfjs/turf/issues/2017

rowanwins (approved):
👍

PR #2090, opened 2021-05-31
Fix turf/bezier-spline to ensure spline reaches last coordinate of line string

the-nemz:
I discovered a bug where the array of coordinates produced by bezier-spline did not reach the last coordinate of the source line-string. This was because the for loop divided up the duration into segments of 10, and the conditional was `i < spline.duration` where `i` was the time counter. This meant that what should have been the end of the spline – where `i === time.duration` (and up to the previous 9 time units) – was excluded from the loop and therefore the resulting spline.

I fixed this by leaving the conditional as `<` and simply adding the corresponding coordinate at the final time after the loop. This accounts for the case where `spline.duration` both is and is not divisible by 10.


rowanwins (approved):
This looks really good thanks @the-nemz

PR #2105, opened 2021-06-18
Bug fix for rectangle and square grid

### LINKED ISSUE #2079: Global squaregrid
Hi,
Im trying to create a square grid that covers the entire map by using https://turfjs.org/docs/#squareGrid

The default example works great, but if i expand the bbox to `[-180, -90, 180, 90]` it returns an empty feature set.
This is my code:  

```javascript
var bbox = [-180, -90, 180, 90];  
var cellSide = 1;  
var options = { units: "degrees" };  
var squareGrid = turf.squareGrid(bbox, cellSide, options);  
```

Is there a bbox limit or am i doing something wrong?

rowanwins:
Fixes a bug in calculation of x width of the `rectangle-grid` (which is used upstream in `square-grid`)
The most obvious improvement is in the [big-bbox-500x100-miles test fixture]( https://github.com/Turfjs/turf/blob/master/packages/turf-rectangle-grid/test/out/big-bbox-500x100-miles.geojson
).

Resolves #2079


rowanwins:
Actually nope, this isn't the right fix

PR #2106, opened 2021-06-18
Fix for rectangle and square grid dimensions

### LINKED ISSUE #1598: Grids generation problems on the whole sphere
First, I would like to thank you for your work. It is a great package for spatial analysis.

I try to get a grid of points on the whole sphere using http://turfjs.org/docs/#pointGrid.

This geoJSON (https://gist.github.com/chasset/2189b121941234acf3dbb90b1b0f111b
) is produced by :

```
turf.pointGrid([-179,-89,180,90], 100)
```

I also tried the following example. It doesn't produce any feature :

```
turf.pointGrid([-180,-90,180,90], 100)
turf.hexGrid([-179,-89,180,90], 100)
turf.squareGrid([-179,-89,180,90], 100)
```

### LINKED ISSUE #1948: squareGrid returns rectangles?
I generated a bbox from these polygons:
image

this is my code:

```
const tbox = require('@turf/bbox');
const arctic = require('./arctic.json');
const bbox = tbox.default(arctic);
const sq = require('@turf/square-grid');

const cellSide = 40;
console.log(JSON.stringify(sq.default(bbox, cellSide, {units: 'kilometers'})));
```
turf-bbox gives me this:

image

### LINKED ISSUE #2079: Global squaregrid

rowanwins:
So in digging into #2079 I found that we using the haversine distance formula for calculating how big to make the grid cells. 

Where this falls down is where you bbox if > 180 degrees because it then finds the shortest distance, and the wider your bbox of interest the worse the effects would be.
````
// A straight line 
turf.distance([-180, -90], [180, -90], {units: 'degrees'})
==> 0
````

Swapping to a flat earth earth calc we do something like
````
180 - -180
==> 360
````

There doesn't seem to be an easy way to hack our distance calc to produce something similar (at least in my initial play) so adjusting the approach seems reasonable and gives more expected results when generating large grids (eg see https://github.com/Turfjs/turf/blob/master/packages/turf-rectangle-grid/test/out/big-bbox-500x100-miles.geojson) 

Resolves #2079
Resolves #1948
Resolves #1598

The same lines of code possibly need to be ported to point-grid and triangle-grid. See #1886

### CODE REVIEW

**Comment by rowanwins** [comment]
Alternatively I need to look into how calculate a harversine distance going the long way around (when the bbox is >180), that seems achievable, although a bit fiddlier

File: packages/turf-rectangle-grid/test/in/victoria-20x100-km.json [hunk 1]
```diff
@@ -1,5 +1,5 @@
 {
-  "bbox": [141, -34, 150, -39],
+  "bbox": [141, -39, 150, -34],
```
mfedderly, line 2: Did you flip this because the behavior is changing? Or to alter the test?

mfedderly (commented):
It looks like this broke the tests in @turf/square-grid, they probably just need to be re-updated. Github also isn't letting me do the rich diff view for these geojsons, but I'm assuming you looked at the output and think its better?

packages/turf-rectangle-grid/test/in/victoria-20x100-km.json
rowanwins: Technically the previous bbox was incorrect because the miny is supposed to come first

rowanwins:
I've checked the results of the proposed PR against QGIS and we now generate almost the exact same results.

mfedderly (approved):
This looks great to me

PR #2108, opened 2021-06-18
Enable more lint rules

mfedderly:
Continuing along the path to making the robots more helpful, I've enabled several more of the eslint rules. The remaining disabled lint rules are going to be more involved, so I'll do them later.

As always I'm happy to have a discussion on eslint/prettier configurations we do and do not like for Turf as a project, I'm mostly just doing the minimal config that gets us in a good place to have those conversations and fix a few bugs along the way.

### DIFF
--- a/packages/turf-clean-coords/index.ts
+++ b/packages/turf-clean-coords/index.ts
@@ -60,7 +60,7 @@ function cleanCoords(
       var existing: Record<string, true> = {};
       getCoords(geojson).forEach(function (coord: any) {
         var key = coord.join("-");
-        if (!existing.hasOwnProperty(key)) {
+        if (!Object.prototype.hasOwnProperty.call(existing, key)) {
           newCoords.push(coord);
           existing[key] = true;
         }
--- a/packages/turf-invariant/test.js
+++ b/packages/turf-invariant/test.js
@@ -300,8 +300,12 @@ test("invariant -- getCoords", (t) => {
 
   t.throws(() => invariant.getCoords(false));
   t.throws(() => invariant.getCoords(null));
-  t.throws(() => containsNumber(invariant.getCoords(["A", "B", "C"])));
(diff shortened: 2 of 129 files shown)


packages/turf-invariant/test.js
```diff
@@ -300,8 +300,12 @@ test("invariant -- getCoords", (t) => {
 
   t.throws(() => invariant.getCoords(false));
   t.throws(() => invariant.getCoords(null));
-  t.throws(() => containsNumber(invariant.getCoords(["A", "B", "C"])));
-  t.throws(() => containsNumber(invariant.getCoords([1, "foo", "bar"])));
+  t.throws(() =>
```
mfedderly, line 303: These were actual bugs in the tests, it would throw an error before because containsNumber was undefined

packages/turf-clean-coords/index.ts
```diff
@@ -60,7 +60,7 @@ function cleanCoords(
       var existing: Record<string, true> = {};
       getCoords(geojson).forEach(function (coord: any) {
         var key = coord.join("-");
-        if (!existing.hasOwnProperty(key)) {
+        if (!Object.prototype.hasOwnProperty.call(existing, key)) {
```
twelch, line 63: I'm surprised it's requiring you to use hasOwnProperty in this way?

mfedderly: Its a defensive change against someone modifying .hasOwnProperty on your object. See: https://eslint.org/docs/rules/no-prototype-builtins

rowanwins (approved):
All looks reasonable to me 👍 

I am somewhat surprised that the spread operator is preferred and passes the es5 check, I thought that was an es6 thing?

mfedderly:
> All looks reasonable to me 👍
> 
> I am somewhat surprised that the spread operator is preferred and passes the es5 check, I thought that was an es6 thing?

The typescript compiler compiles down to ES5 so I get to use it :).

PR #2113, opened 2021-06-22
Update github actions to cache yarn

mfedderly:
Right now the yarn install time takes up over half of our build time, this will probably cut that down dramatically. I copy pasted it directly from their examples: https://github.com/actions/cache/blob/main/examples.md#node---yarn


rowanwins (approved):
👍

PR #2131, opened 2021-06-26
Proof of Concept - Remove turf-jsts replace with custom JSTS build in turf-buffer

JamesLMilner:
Please fill in this template.

At the moment JSTS is only used in one module; `turf-buffer`. As it stands we rely on [`turf-jsts`](https://github.com/DenisCarriere/turf-jsts#readme) which although serves its purpose, hasn't been touched in 4 years. 

@rycgar did the work of creating a custom build for JSTS which strips back to the necessary JSTS code that is required to operate `turf-buffer`. I stripped this back very minorly by removing small amounts of dead code + IE polyfills that aren't required.

As it stands building with `terser` and gzipping gives the following file sizes for a minimal use case of `turf-buffer`:

image

This PR, using `turf-buffer` would come in at under half the size for the regular case once minified/gzipped:

image

### DIFF
--- a/packages/turf-buffer/index.js
+++ b/packages/turf-buffer/index.js
@@ -1,5 +1,5 @@
 import center from "@turf/center";
-import { BufferOp, GeoJSONReader, GeoJSONWriter } from "turf-jsts";
+import { bufferOp } from "./lib/bufferOp";
 import { geomEach, featureEach } from "@turf/meta";
 import { geoAzimuthalEquidistant } from "d3-geo";
 import {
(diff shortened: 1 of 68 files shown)


mfedderly:
I like the savings from doing this. These files all probably need license headers because they've been vendored which would get quite large unless they get combined into a single file. I wish we could just get a fresh implementation because I think as things stand we still have a bunch of bugs from using this buffer implementation anyhow. https://github.com/Turfjs/turf/issues?q=is%3Aopen+is%3Aissue+label%3A%40turf%2Fbuffer

PR #2132, opened 2021-06-26
Update prettier on a bunch of files

rowanwins:
Looks like a new version of prettier dropped a few hours ago. This PR fixes a bunch of files that are failing the CI.


mfedderly:
Not sure what happened with this PR. If you want to upgrade the prettier, you would upgrade the package.json, run yarn install to update the yarn.lock and then rewrite all the files using prettier and commit all the changes. I don't see the package.json changes that I would expect. If you aren't using yarn to install I suspect that npm picked a different version of prettier that the repo does not intend to use.

rowanwins:
> Not sure what happened with this PR. If you want to upgrade the prettier, you would upgrade the package.json, run yarn install to update the yarn.lock and then rewrite all the files using prettier and commit all the changes. I don't see the package.json changes that I would expect. If you aren't using yarn to install I suspect that npm picked a different version of prettier that the repo does not intend to use.


Ahh thats likely it. Although I'm not sure why I was seeing some CI builds fail?

PR #2134, opened 2021-06-27
Bugfix: identify multiple overlaps on one segment

### LINKED ISSUE #2068: lineOverlap fails to find multiple overlaps on one segment
if line1 overlaps with line2 multiple times on only one line2's segment, lineOverlap detects only the first overlap, but lineIntersect detects the intersects properly 
```
const l1 = lineString([[0, 1], [1, 1], [1, 0], [2, 0], [2, 1], [3, 1], [3, 0], [4, 0], [4, 1], [4,0]]);
const l2 = lineString([[0, 0], [6, 0]]);
const overlap = lineOverlap(l1,l2);
L.geoJson(overlap).addTo(map);
L.geoJson(lineIntersect(l1,l2)).addTo(map);
```
Screenshot 2021-04-03 at 19 24 04
when calling lineIntersect with reverted order of lines, it works properply too:

```
const overlap = lineOverlap(l2,l1);
L.geoJson(overlap).addTo(map);
```
Screenshot 2021-04-03 at 19 25 48

```
"@turf/line-overlap": "^6.3.0",
```
[jsfiddle](http://jsfiddle.net/xuszkpqb/)

patrickbrett:
Resolves the issue identified by @nasmuris in #2068 and adds their example as a test case. Appears to be working though I'm not extremely familiar with turf internals so a thorough review before merging would be appreciated.

Would also appreciate any help in coming up with additional test cases to ensure this doesn't break anything. We currently have hardly any test cases for this module 😢 

Resolves #2068 


JamesLMilner:
Thanks for raising this @patrickbrett 

Probably one for @rowanwins @mfedderly to take a look at?

patrickbrett:
Sounds good, I'll add them as reviewers

packages/turf-line-overlap/index.ts
```diff
@@ -148,6 +160,9 @@ function concatSegment(
   else if (equal(coords[0], end)) geom.push(coords[1]);
   else if (equal(coords[1], start)) geom.unshift(coords[0]);
   else if (equal(coords[1], end)) geom.push(coords[0]);
+  else return; // If the overlap leaves the segment unchanged, return undefined so that this can be identified.
+
+  // Otherwise return the new line.
```
mfedderly, line 165: This comment is kind of misleading, it isn't a new line its the line that has been mutated

mfedderly (commented):
This is the first time I'm looking at this code and I'm finding it really hard to follow.
I'm nervous about the getCoords().sort() lines as well, I'm not sure how that works and whether it is correct.

Is there a way to simplify this by potentially removing doesOverlaps boolean (just null check against overlapSegment instead?) and combine overlapSegment and additionalSegments?

It seems like this code *probably* works especially because the tests pass, but I'm definitely nervous to merge this.

patrickbrett:
> This is the first time I'm looking at this code and I'm finding it really hard to follow.
> I'm nervous about the getCoords().sort() lines as well, I'm not sure how that works and whether it is correct.
> 
> Is there a way to simplify this by potentially removing doesOverlaps boolean (just null check against overlapSegment instead?) and combine overlapSegment and additionalSegments?
> 
> It seems like this code _probably_ works especially because the tests pass, but I'm definitely nervous to merge this.

I agree the code is a bit of a mess, since this is my first PR I tried to modify the existing logic as little as possible, but more than happy to overhaul it to make it easier to follow and write more extensive tests.

mfedderly:
I also ran a code coverage check and this is actually at like 90% coverage already. It would be cool to add the code coverage numbers to the build at some point.

```
----------|---------|----------|---------|---------|-------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------|---------|----------|---------|---------|-------------------
All files |      90 |       86 |     100 |   90.74 |                   
 index.ts |      90 |       86 |     100 |   90.74 | 69,95-98          
----------|---------|----------|---------|---------|-------------------
```

rowanwins:
@patrickbrett I wouldn't be afraid of giving it an entire overhaul  :)
I'd perhaps look at incorporating [rbush-knn](https://github.com/mourner/rbush-knn) for the tolerance checks

mfedderly (approved):
I don't want to let perfect be the enemy of good here, we can rework this implementation in the future but this at least improves the quality and the test coverage has me feeling OK with this change.

Issue #2099, opened 2021-06-16
Feature Request: Support for FeatureCollection<MultiPoint> to first parameter of pointsWithinPolygon


justinbangerter:
- [version 5.1.6 and 6.3.0] 
```
const geojson = turf.featureCollection([
  turf.multiPoint([
    [10, 10],
    [-10, -10]
  ])
]);

const polygon = turf.polygon([[
  [0, 0],
  [15, 0],
  [15, 15],
  [0, 15],
  [0,0]
]]);

turf.pointsWithinPolygon(geojson, polygon);

> "Error: coord must be GeoJSON Point or an Array of numbers"
```
Thanks for the awesome library.

rowanwins:
Good suggestion thanks @justinbangerter !

rowanwins:
@micaminoff or @twelch - you've both expressed interest in helping out with TurfJS recently, this might good be a good first PR to tackle :)

twelch:
I can take this on.  I'd like some feedback on the expected result. @justinbangerter what were you hoping for?

Here's the current docs:
```
@param {Feature|FeatureCollection<Point|MultiPoint>} points Points as input search
@param {FeatureCollection|Geometry|Feature<Polygon|MultiPolygon>} polygons Points must be within these (Multi)Polygon(s)
@returns {FeatureCollection<Point>} points that land within at least one polygon
```

So the `points` argument would be expanded to include MultiPoint.
```
@param {Feature|FeatureCollection<Point|MultiPoint>} points Points as input search
```

The return value is more interesting depending on what you think the spirit of the function is, and whether the MultiPoint structure should be maintained or altered.    For each MultiPoint input the function could:
1. break the MultiPoint into individual Point features and return only the ones that fall within a Polygon.
2. return the whole MultiPoint if any of its point coordinates fall within a Polygon
3. return a MultiPoint with just the point coordinates that fall within a Polygon

I think the user can/should achieve 1 with the existing function by doing a `flatten` first to get a collection of Points.  By that logic I'm thinking there's only value in supporting MultiPoint for this function if the MultiPoint is maintained (2 or 3).

justinbangerter:
I will trust your judgement here. My opinion is that 3 is the best.

[...]

PR #2137, opened 2021-06-29
pointsWithinPolygon: add MultiPoint support

twelch:
This PR adds support for MultiPoints to pointsWithinPolygon per issue #2099.

This implements Option 3 discussed, for each MultiPoint passed, returns a MultiPoint with just the point coordinates that fall within a Polygon.

### DIFF
--- a/packages/turf-points-within-polygon/index.d.ts
+++ b/packages/turf-points-within-polygon/index.d.ts
@@ -3,6 +3,7 @@ import {
   FeatureCollection,
   Polygon,
   MultiPolygon,
+  MultiPoint,
   Point,
   Properties,
 } from "@turf/helpers";
@@ -14,6 +15,8 @@ export default function pointsWithinPolygon<
   G extends Polygon | MultiPolygon,
   P = Properties
 >(
-  points: Feature<Point, P> | FeatureCollection<Point, P>,
+  points:
+    | Feature<Point | MultiPoint, P>
+    | FeatureCollection<Point | MultiPoint, P>,
   polygons: Feature<G> | FeatureCollection<G> | G
 ): FeatureCollection<Point, P>;
--- a/packages/turf-points-within-polygon/index.d.ts
+++ b/packages/turf-points-within-polygon/index.d.ts
@@ -12,11 +12,10 @@ import {
  * http://turfjs.org/docs/#pointswithinpolygon
  */
 export default function pointsWithinPolygon<
+  F extends Point | MultiPoint,
   G extends Polygon | MultiPolygon,
   P = Properties
 >(
-  points:
-    | Feature<Point | MultiPoint, P>
-    | FeatureCollection<Point | MultiPoint, P>,
+  points: Feature<F, P> | FeatureCollection<F, P>,
   polygons: Feature<G> | FeatureCollection<G> | G
-): FeatureCollection<Point | MultiPoint, P>;
+): FeatureCollection<F, P>;
--- a/packages/turf-points-within-polygon/index.d.ts
+++ b/packages/turf-points-within-polygon/index.d.ts
@@ -19,4 +19,4 @@ export default function pointsWithinPolygon<
     | Feature<Point | MultiPoint, P>
     | FeatureCollection<Point | MultiPoint, P>,
   polygons: Feature<G> | FeatureCollection<G> | G
-): FeatureCollection<Point, P>;
+): FeatureCollection<Point | MultiPoint, P>;
--- a/packages/turf-points-within-polygon/index.js
+++ b/packages/turf-points-within-polygon/index.js
@@ -1,5 +1,5 @@
 import pointInPolygon from "@turf/boolean-point-in-polygon";
-import { featureCollection } from "@turf/helpers";
+import { featureCollection, multiPoint } from "@turf/helpers";
 import { geomEach, featureEach, coordEach } from "@turf/meta";
 
 /**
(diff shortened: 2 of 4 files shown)


packages/turf-points-within-polygon/index.d.ts
```diff
@@ -14,6 +15,8 @@ export default function pointsWithinPolygon<
   G extends Polygon | MultiPolygon,
   P = Properties
 >(
-  points: Feature<Point, P> | FeatureCollection<Point, P>,
+  points:
+    | Feature<Point | MultiPoint, P>
+    | FeatureCollection<Point | MultiPoint, P>,
   polygons: Feature<G> | FeatureCollection<G> | G
-): FeatureCollection<Point, P>;
+): FeatureCollection<Point | MultiPoint, P>;
```
twelch, line 22: The downside to this return signature (in Typescript anyway) is that it narrowly implies the FC can contain Points or MultiPoints but in fact it is one or the other, not both.  So the user can run into some friction that requires type narrowing.  This may be acceptable because the two types are so similar, varying only in number of coordinate Positions.  Otherwise, a solution to overcome is Typescript function overloading

packages/turf-points-within-polygon/index.d.ts
```diff
@@ -11,9 +12,10 @@ import {
...
 export default function pointsWithinPolygon<
+  F extends Point | MultiPoint,
   G extends Polygon | MultiPolygon,
   P = Properties
 >(
-  points: Feature<Point, P> | FeatureCollection<Point, P>,
+  points: Feature<F, P> | FeatureCollection<F, P>,
   polygons: Feature<G> | FeatureCollection<G> | G
-): FeatureCollection<Point, P>;
+): FeatureCollection<F, P>;
```
twelch, line 21: This makes the point input geometry generic as well, so that the return geometry will match what the user passes in.

packages/turf-points-within-polygon/index.js
```diff
@@ -41,11 +41,28 @@ function pointsWithinPolygon(points, polygons) {
...
-      results.push(point);
+    if (point.geometry.type === "Point") {
+      geomEach(polygons, function (polygon) {
+        if (pointInPolygon(point, polygon)) contained = true;
+      });
+      if (contained) {
+        results.push(point);
+      }
+    } else if (point.geometry.type === "MultiPoint") {
+      var pointsWithin = [];
```
twelch, line 52: Just an fyi, I tried to use `let` over `var` here out of habit to narrow the scope to the code block, but the CI linter complained that `let` is reserved.   I was surprised VSCode didn't warn me, I have the eslint module turned on.  @mfedderly I thought I saw you were taking suggestions for relaxing linter rules, perhaps this is one.

https://github.com/Turfjs/turf/runs/2938648986?check_suite_focus=true
```
          ES-Check Error:
          ----
          · erroring file: packages/turf-points-within-polygon/dist/js/index.js
          · error: SyntaxError: The keyword 'let' is reserved (58:6)
```

mfedderly: In this case you aren't hitting eslint rules, but rather the es-check rule. My understanding is that you should not ship non-es5 code as part of your packages. Although the typescript packages in the repo allow us to use the newer es6 features, we do not currently have a transpilation pipeline set up for the javascript packages. Ideally this check is still left in to make sure none of our dependencies accidentally ship es6 as well.

If my knowledge is outdated (caniuse.com says almost 95% of all clients support the let keyword at this point), then we should figure out what we want our officially supported targets to be and go from there.

twelch: Makes sense.  I guess the presence of `const` in the tests that made me think it was okay.  I can see es-check is only run on the dist

rowanwins:
From my initial pass through this all looks good I think @twelch ! 
I'll just leave @mfedderly to do a final check on some of the ts stuff as that's not really my expertise, but it's got my approval.

packages/turf-points-within-polygon/index.d.ts
mfedderly: Yeah this is great.

packages/turf-points-within-polygon/index.js
```diff
@@ -1,14 +1,14 @@
...
 
 /**
- * Finds {@link Points} that fall within {@link (Multi)Polygon(s)}.
+ * Finds {@link Points} or {@link MultiPoint} coordinate positions that fall within {@link (Multi)Polygon(s)}.
  *
  * @name pointsWithinPolygon
- * @param {Feature|FeatureCollection<Point>} points Points as input search
- * @param {FeatureCollection|Geometry|Feature<Polygon|MultiPolygon>} polygons Points must be within these (Multi)Polygon(s)
- * @returns {FeatureCollection<Point>} points that land within at least one polygon
+ * @param {Feature|FeatureCollection<Point|MultiPoint>} points (Multi)Point(s) as input search
```
mfedderly, line 9: The typescript version of this would be `Feature<Point|MultiPoint> | FeatureCollection<Point|Multipoint>` but maybe this is fine for JSDoc?

packages/turf-points-within-polygon/index.js
```diff
@@ -1,14 +1,14 @@
...
+ * @param {FeatureCollection|Geometry|Feature<Polygon|MultiPolygon>} polygons (Multi)Polygon(s) to check if points are within
+ * @returns {FeatureCollection<Point|MultiPoint>} (Multi)Points that land within at least one polygon
```
mfedderly, line 11: It might be worth calling out that the Point | MultiPoint depends on what the input was and is not arbitrary.

mfedderly (commented):
This seems great!

Can you update the @returns annotation to just clarify for JS consumers looking at the docs that the return type mirrors what they passed in?

Also what happens to the typescript types if you pass in a `FeatureCollection<Point | Multipoint>` for points? Does it allow it and correctly infer `FeatureCollection<Point | Multipoint>` for the output? Maybe add a types.ts test for that.

packages/turf-points-within-polygon/index.js
twelch: I agree, as a TS user myself, your suggestion is more correct but looking at the docs for different functions I sensed there was an established convention already.  I don't mind changing them, I would just change the params too to match.  Since each type becomes a link in the docs, maybe it was done for simplicity.

image

packages/turf-points-within-polygon/index.js
twelch: done

twelch:
@mfedderly good suggestions.  I have updated the docs and added a mixed Point|Multipoint use case to confirm the returned type is as expected.  I hadn't really thought of the user being able to pass in a feature collection containing both types but it is supported so I added a unit test for that as well.

mfedderly (approved):
Thanks! Good point about the jsdoc convention being different than the typescript version. Happy to stick with what we've got already.

PR #2146, opened 2021-07-15
README update: Remove reference to bower, update CDN version to latest version

JamesLMilner:
- Removes reference to Bower in the README. Realistically bower has been deprecated for a few years now and I've hardly heard anyone mention it since it's deprecation - I think we should save the README real estate for more meaningful information
- Update the CDN version to the latest version (we should probably make a habit of this)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Download the [minified file](https://npmcdn.com/@turf/turf/turf.min.js), and inc
 You can also include it directly from a CDN:
 
 ```html
-<script src="https://cdn.jsdelivr.net/npm/@turf/turf@6.5.0/turf.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/@turf/turf@6/turf.min.js"></script>
 ```
 
 You can create light-weight turf builds with only the functions you need using the [turfjs-builder UI](https://turfjs-builder.herokuapp.com/) or using browserify as described below.
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Download the [minified file](https://npmcdn.com/@turf/turf/turf.min.js), and inc
 You can also include it directly from a CDN:
 
 ```html
-<script src="https://cdn.jsdelivr.net/npm/@turf/turf@5/turf.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/@turf/turf@6.5.0/turf.min.js"></script>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,7 +40,7 @@ Download the [minified file](https://npmcdn.com/@turf/turf/turf.min.js), and inc
 You can also include it directly from a CDN:
 
 ```html
-<script src="https://cdn.jsdelivr.net/npm/@turf/turf@5/turf.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/@turf/turf@6.5.0/turf.min.js"></script>
```
mfedderly, line 43: Why not @6 so we don't have to remember to bump it every release?

README.md
```diff
@@ -53,10 +53,6 @@ All of Turf's functions can also be installed as separate modules. This works we
 npm install @turf/helpers @turf/buffer
 ```
 
-### Bower [NOTE: Bower support has been dropped Jan 1st 2017]
-
-_Not recommended. Please don't use Bower. Use Browserify, Webpack, or the CDN instead._
```
mfedderly, line 58: lol

README.md
rowanwins: Or do we drop the version reference all together then we never have to update it?

mfedderly: I think that would encourage people to omit the major version specifier and then get upset when we release 7 with breaking changes. We already got a small taste of that when I accidentally published the beta version with the `latest` tag.

twelch: I noticed that the Getting Started page on the docs website has a link to jsdelivr as well, and it's stale at 6.3.0 - http://turfjs.org/getting-started

One idea, is to drop this section from the readme altogether and just link to the Getting Started page.  Then, potentially automate it in the docs by parsing the package version out of package.json at doc build time and pass it into the getting started page.  Though, I can't seem to find where the getting started page in the organization or repo to see if it's template-driven.  This technique is what the openlayers repo does for example - https://github.com/openlayers/openlayers/blob/7a2f87caca9ddc1912d910f56eb5637445fc11f6/doc/quickstart.hbs#L28

rowanwins: I've just updated the getting-started page to just have the major version number referenced, and I can look into the pulling a version number from the package.json...

rowanwins (approved):
👍

PR #2149, opened 2021-07-17
polygonSmooth: make options optional

geirsagberg:
Noticed I got an error when calling `polygonSmooth` without a second options object, even though it was marked as optional. Fixed it by adding an empty object as default.

### DIFF
--- a/packages/turf-polygon-smooth/index.js
+++ b/packages/turf-polygon-smooth/index.js
@@ -1,5 +1,5 @@
+import { featureCollection, multiPolygon, polygon } from "@turf/helpers";
 import { coordEach, geomEach } from "@turf/meta";
-import { featureCollection, polygon, multiPolygon } from "@turf/helpers";
 
 /**
  * Smooths a {@link Polygon} or {@link MultiPolygon}. Based on [Chaikin's algorithm](http://graphics.cs.ucdavis.edu/education/CAGDNotes/Chaikins-Algorithm/Chaikins-Algorithm.html).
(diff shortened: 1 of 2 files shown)


twelch (approved):
Error confirmed.  Looks good to me, thanks @geirsagberg.

JamesLMilner:
Seems sensible to me too, appreciate you adding a test.

packages/turf-polygon-smooth/index.js
```diff
@@ -18,7 +18,7 @@ import { featureCollection, polygon, multiPolygon } from "@turf/helpers";
  * //addToMap
  * var addToMap = [smoothed, polygon];
  */
-function polygonSmooth(inputPolys, options) {
+function polygonSmooth(inputPolys, options = {}) {
```
rowanwins, line 21: @geirsagberg this is using an ES6 syntax and we currently only support es5 for `.js` files.

Check out [this handling](https://github.com/Turfjs/turf/blob/master/packages/turf-sector/index.js#L33-L34
) for how we've typically done this.

rowanwins (changes requested):
This PR needs to be changed to make it ES5 compatible.

packages/turf-polygon-smooth/index.js
geirsagberg: Gotcha, fixed.

geirsagberg:
> This PR needs to be changed to make it ES5 compatible.

Should be fixed now, checks are green.

rowanwins (approved):
Perfecto - thanks @geirsagberg !

JamesLMilner:
🎉

PR #2156, opened 2021-07-25
[Proof of Concept Get coverage by using nyc in turf-along package

JamesLMilner:
Please fill in this template.

- [ ] Use a meaningful title for the pull request. Include the name of the package modified.


rowanwins:
What is the intent with how we'd apply this @JamesLMilner ? Is there an automated test that checks if coverage is being improved or if it declines? Or is there are report that we'll be able to see via the checks above?

JamesLMilner:
@rowanwins the aim here is that we firstly want to be able to see code coverage on push. Ideally we also want to be able to determine how it's changed between master. I'm working on adding another GitHub Action that will run coverage on the current branch and on master so you can compare the difference. I am looking to see if there's a way to write a script that will do a diff on the outputs somehow.

.github/workflows/turf.yml
```diff
@@ -38,3 +38,53 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: git diff --exit-code
       - run: yarn test
+
+  coverage:
```
mfedderly, line 42: Coverage winds up taking a long time because it has to do all of the builds in the `yarn install` step. Circle has the ability to save a workspace and then fork into multiple other tasks. Maybe we can do that here and spend a lot less time on the coverage step? Otherwise we could just calculate it as part of the regular builds since they run tests anyhow.

Also having to re-run coverage on master for every PR seems like its just burning CPU. Maybe we can store the results somewhere and use that for the diff instead of needing to recalculate them every time. I'm not actually sure how people do this in other repos.

package.json
```diff
@@ -16,7 +16,8 @@
     "pretest": "npm run lint",
     "test": "lerna run test",
     "posttest": "lerna run --scope @turf/turf last-checks",
-    "docs": "node ./scripts/generate-readmes"
+    "docs": "node ./scripts/generate-readmes",
+    "coverage": "node ./scripts/coverage"
```
mfedderly, line 20: It looks like you added a lot of the package.json changes manually. You could've used `.monorepolint.config.js` to both apply these and enforce it going forward.

scripts/coverage.js
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env node
```
mfedderly, line 1: this entire script is basically just `yarn lerna run coverage`, right?

package.json
JamesLMilner: Ah! I didn't realise this was how this works - still getting my head around all the monorepo scaffolding...

scripts/coverage.js
JamesLMilner: It appears so! Thanks - again wasn't aware that was possible so good to know

package.json
mfedderly: Yeah monorepolint is something one of my coworkers wrote. Its great but also nobody really knows about it.

.github/workflows/turf.yml
JamesLMilner: I'm not super concerned here because the job processes run in parallel - they shouldn't impact each others performance. 

I've removed `master` for now, better to just get coverage working and then figure out how to do a diff against master.

.monorepolint.config.ts
```diff
@@ -171,6 +171,22 @@ module.exports = {
...
+          scripts: {
+            coverage: "nyc ts-node -r esm test.js",
+          },
+        },
+        includePackages: [...TS_PACKAGES],
+      },
+      {
+        options: {
+          scripts: {
+            coverage: "nyc  node -r esm test.js",
```
mfedderly, line 185: extra space between nyc and node here ^

.monorepolint.config.ts
```diff
@@ -249,6 +265,14 @@ module.exports = {
           dependencies: {
             tslib: "^2.3.0",
           },
+          devDependencies: {
+            nyc: "*",
+          },
+        },
+        includePackages: [...TS_PACKAGES, ...JS_PACKAGES],
```
mfedderly, line 272: This got included in a weird way. tslib dependency should only apply to TS_PACKAGES, I guess just move the above dependencies block to the next entry? It also wound up getting added to all the package.json files so that'll need to get pulled back out. I think you can do that with monorepolint by setting an explicit removal for the JS_PACKAGES on tslib dependency and then running monorepolint and then deleting the rule before committing.

JamesLMilner: Good spot, sorry about that. I figured out how to remove it; you can do:

```
    ":banned-dependencies": {
      options: {
        bannedDependencies: ["tslib"],
      },
      includePackages: [...JS_PACKAGES],
    },
 ```
 
 For future reference

JamesLMilner:
I think this should be fixed up now

mfedderly (approved):
LGTM I'd like to merge this but the branch is in your fork. Want to update yarn.lock and get this merged?

mfedderly:
@JamesLMilner are you still interested in adding coverage data via github actions?

PR #2157, opened 2021-07-28
Add booleanIntersects to @turf/turf typescript types

### LINKED ISSUE #2034: booleanIntersects is not exported from @turf/turf
- [x] **Turf version:** @turf/turf@6.3.0
- [x] **GeoJSON data:** N/A
- [x] **Snippet of source:**

```ts
import * as turf from "@turf/turf";
import { a, b } from "./myGeoFeatures";

const check = turf.booleanIntersects(a, b) // ❌ Property 'booleanIntersects' does not exist on type 'typeof import("/path/to/node_modules/@turf/turf/index")'.ts(2339)
const workaround = !turf.booleanDisjoint(a, b) // ✅ OK
```

Looks like `booleanIntersects` is simply missing from module re-exports.

mfedderly:
Fixes #2034
cc @kachkaev


twelch (approved):

kachkaev:
What’s your expected release timeline folks? Not rushing it, just curious about the repo’s release cycle 🙂

mfedderly:
We're working on making a bunch of breaking changes for 7.0 and then we also got a bit derailed as everyone's lives changed due to covid. You can follow along with the 7.0 milestone here https://github.com/Turfjs/turf/milestone/19. I feel like we're close to having most of the big items knocked out, but progress is also admittedly slow.

mschilde:
Ah, i was just about to report this as a bug. Is there a chance to get this released as a bugfix on the 6.x version?

just-Bri:
> Ah, i was just about to report this as a bug. Is there a chance to get this released as a bugfix on the 6.x version?

I'm running in to this issue on version 6.5.0.
package.json: `"@turf/turf": "6.5.0",`
image

markusand:
Not available in 6.5.0, neither is booleanTouches

chrisbitmead:
Is there a workaround on this lack of booleanIntersects folks? Sorry if this is a stupid question, I'm not a typescript guru.

kachkaev:
@chrisbitmead you can use `!` + `booleanDisjoint` instead of `booleanIntersects` for now (https://github.com/Turfjs/turf/issues/2034)

mfedderly:
If you're consuming with javascript alone, then you can just use 6.5.0 and the javascript code is there. If you're using typescript you can use `7.0.0-alpha.0`

kachkaev:
Ah right: https://www.npmjs.com/package/@turf/turf/v/7.0.0-alpha.0 was released in Nov 2022. This pre-release does not show up in https://github.com/Turfjs/turf/releases, so I did not suspect about its existence.

mfedderly:
That 7.0.0 has a bunch of nice stuff in it, but there's still more breaking changes that we mean to make before releasing a 7.0 proper, but progress is rather slow.

berci-i:
Hi, 

Thank you for the updates.
I tried to add the alpha version but I ran into the following error (while using yarn):

```
SyntaxError: Named export 'BufferOp' not found. The requested module '@turf/jsts' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from '@turf/jsts';
const { GeoJSONReader, BufferOp, GeoJSONWriter } = pkg;
```

I tried to generate `yarn.lock` again from scratch but the error is still there.

mfedderly:
@JamesLMilner does the @turf/jsts error above look familiar to you?

PR #2165, opened 2021-08-02
Use tslib for Typescript transpile support

### LINKED ISSUE #2151: @turf/random's module is "illegal", causing rollup to report an error

- [x ] The version of Turf you are using, and any other relevant versions.

        "@turf/turf": "^6.5.0",

- [x ] GeoJSON data as a [gist file](https://gist.github.com/) or [geojson.io](http://geojson.io/) (filename extension must be `.geojson`).

Not applicable

I'm using rollup to create a large single Turf es6 module. The presence of "this" & code before it's imports in 

    node_modules/@turf/random/dist/es/index.js

produces errors in rollup. This fixes the problem:

```
// var __spreadArrays = (this && this.__spreadArrays) || function () {
//     for (var s = 0, i = 0, il = arguments.length; i < il; i++) s += arguments[i].length;
//     for (var r = Array(s), k = 0, i = 0; i < il; i++)
//         for (var a = arguments[i], j = 0, jl = a.length; j < jl; j++, k++)
//             r[k] = a[j];
//     return r;
// };
import { featureCollection, isNumber, isObject, lineString, point, polygon, } from "@turf/helpers";
var __spreadArrays = function () {
    for (var s = 0, i = 0, il = arguments.length; i < il; i++) s += arguments[i].length;
    for (var r = Array(s), k = 0, i = 0; i < il; i++)
        for (var a = arguments[i], j = 0, jl = a.length; j < jl; j++, k++)
            r[k] = a[j];
    return r;
};
```

With this fix, I still get circular dependencies reported by rollup:
```
(!) Circular dependencies
node_modules/d3-voronoi/src/Diagram.js -> node_modules/d3-voronoi/src/Beach.js -> node_modules/d3-voronoi/src/Cell.js -> node_modules/d3-voronoi/src/Edge.js -> node_modules/d3-voronoi/src/Diagram.js
node_modules/d3-voronoi/src/Diagram.js -> node_modules/d3-voronoi/src/Beach.js -> node_modules/d3-voronoi/src/Cell.js -> node_modules/d3-voronoi/src/Diagram.js
node_modules/d3-voronoi/src/Diagram.js -> node_modules/d3-voronoi/src/Beach.js -> node_modules/d3-voronoi/src/Circle.js -> node_modules/d3-voronoi/src/Diagram.js
node_modules/d3-voronoi/src/Diagram.js -> node_modules/d3-voronoi/src/Beach.js -> node_modules/d3-voronoi/src/Diagram.js
```

mfedderly:
Without importHelpers, tsc would put the transpile support methods at the top of the output file. This apparently breaks rollup and potentially other bundlers. By using tslib, all of these become import statements instead. It also should have the effect of shrinking everyone's output bundles by not duplicating these methods everywhere.

Fixes #2151


twelch:
@mfedderly neat, should this just get added at the top-level package.json alongside typescript? 
 To make it available to all packages - https://github.com/Turfjs/turf/blob/master/package.json#L62

mfedderly:
Sadly it has to be in the individual package.json files because it is a dependency not a devDependency.

twelch (approved):

PR #2166, opened 2021-08-02
Change @turf/area earth radius from the semi-major access to the WGS84 median earth radius

### LINKED ISSUE #2045: Function turf-area should use turf-helpers/earthRadius
Hi,

It called my attention that calculating the area of geographical polygons with Turf JS was giving slightly different results than in PostGIS. I found that these differences came from the value of the Earth radius.

After discussion in previous issues in this repository, the Earth radius was eventually defined as **6371008.8** (earthRadius in turf-helpers), which is the same used in PostGIS:
https://github.com/Turfjs/turf/blob/41a123a0e151be6a7e312dfecb91b69c4ff3f3f2/packages/turf-helpers/index.ts#L84

However, the code that calculates the area (in turf-area) it is still defined as **6378137**, and there is a comment to change it to earthRadius from turf-helpers:
https://github.com/Turfjs/turf/blob/41a123a0e151be6a7e312dfecb91b69c4ff3f3f2/packages/turf-area/index.ts#L4-L5

Do you agree that it should indeed be changed?
If so, is this something I can do myself via a Pull request, or what is the best procedure?

Thank you

mfedderly:
Change from using the semi-major earth radius to the median earth radius.

Fixes #2045


rowanwins (approved):
👍

mfedderly:
Yeah it seems like changing the output of area is a pretty meaningful change even if it doesn't break the function's signature.

For what its worth there is a `7.0.0-alpha.0` you can try, that's got most of our work recently but we don't promise not to break more things before 7.0.0 proper.

PR #2167, opened 2021-08-02
@turf/invariant's getCoord now shallow clones its result

### LINKED ISSUE #2043: @turf/rhumb-distance mutates its argument
I would expect it to behave like the other pure functions, but it [modifies the coordinate of the point passed to it](https://github.com/Turfjs/turf/blob/41a123a0e151be6a7e312dfecb91b69c4ff3f3f2/packages/turf-rhumb-distance/index.ts#L39). This seems like an accident!

mfedderly:
Previously this led to an issue in @turf/rhumb-distance mutating its argument.
I didn't see any other issues in the typescript codebase, so I'm hoping this was an isolated issue.

Fixes #2043


rowanwins (approved):
Looks reasonable

PR #2169, opened 2021-08-02
transform-scale factor should be greater than zero

### LINKED ISSUE #1831: Documentation: Invalid Factor Definition on npm
In your NPM documentation it has a logical fallacy:

`factor number of scaling, positive or negative values greater than 0`

https://www.npmjs.com/package/@turf/transform-scale

### LINKED ISSUE #1959: Transform Scale Geometry Smaller
Hi,
i am trying to scale  geometry smaller by passing negative value using transform-scale, but it cannot, factor parameter should be greater than zero

but in the  @turf/transform-scale documentation , it says  "_factor number of scaling, positive or **negative** values greater than 0_"

why mentioning negative number when it should be greater than zero  and how to scale geometry to make it smaller than the original.

Thank you.

mfedderly:
Fix jsdoc and input check, add test to make sure we don't regress it

Fixes #1959
Fixes #1831


packages/turf-transform-scale/index.js
```diff
@@ -15,7 +15,7 @@ import { getCoord, getCoords, getType } from "@turf/invariant";
  *
  * @name transformScale
  * @param {GeoJSON} geojson GeoJSON to be scaled
- * @param {number} factor of scaling, positive or negative values greater than 0
+ * @param {number} factor of scaling, positive values greater than 0
```
rowanwins, line 18: Perhaps we can add a note that setting the factor to 1 will return the current size, so setting to between 0 & 1 will make the geometry smaller, and setting to above 1 will return a larger geometry.

rowanwins (approved):
Perfecto

PR #2171, opened 2021-08-02
@turf/bbox uses input geojson.bbox where possible

### LINKED ISSUE #1921: bbox doesn't check to see if a bbox exists before calculating
bbox test is used extensively in many library functions to optimize calculations. However in most of the ones I have found (the booleanContains lib is a great example) each library function recalculates the bbox without checking to see if one already exists.  Given how long this can take, this is drastically slowing down calculations instead of speeding them up. 

Suggestion is to make things consistent across the board, have the `bbox` function itself first check to see if a bbox exists on the feature it has been passed and if so, returns that.

mfedderly:
Fixes #1921

There was an issue in @turf/transform-scale where the feature returned by that module would have an invalid bounding box. This also fixes that bug. I caught this because the @turf/shortest-path tests broke.


mfedderly:
This definitely broke @turf/shortest-path but I have no explanation for why

rowanwins (approved):
👍

PR #2173, opened 2021-08-02
@turf/helpers polygon() will throw errors if the first and last point in a ring have different dimensions

### LINKED ISSUE #1820: [@turf/helpers] Polygon: First and last position check is not according to the spec
According to [section 3.1.6](https://tools.ietf.org/html/rfc7946#section-3.1.6):
>  The first and last positions are equivalent, and they MUST contain identical values; their representation SHOULD also be identical.

In the example below, both are invalid, but we throw an error only for the second one

```javascript
const { polygon } = require('@turf/helpers');
console.log('valid', polygon([[
    [ 1,2,3],
    [ 4,5 ],
    [ 6,7 ],
    [ 1,2 ]
]]).geometry)

console.log('invalid', polygon([[
    [ 1,2 ],
    [ 4,5 ],
    [ 6,7 ],
    [ 1,2,3 ]
]]).geometry)
```

- [X] Snippet of source code or for complex examples use [jsfiddle](https://jsfiddle.net/).

mfedderly:
Fixes #1820


rowanwins (approved):
🎉

PR #2177, opened 2021-08-03
@turf/helpers update factors value for degrees, remove unitsFactors entirely as it was unused

### LINKED ISSUE #1406: Degree-Radian conversion scale factor error.
Converting radians to degrees results in what look like mathematical errors. PI radians should be exactly 180 degrees:
```
t.convertLength(Math.PI,"radians","degrees")
179.78993435469053
```

the conversion is done by one of the `factors` in turf-helpers (extract):

```
export let earthRadius = 6371008.8;
...
degrees: earthRadius / 111325,
```

this results in a degree-radian conversion factor of 57.22891, where as 180/pi is 57.29578

Either there's some odd reason why this is set this way, or possibly the `earthRadius` was changed and the corresponding denominator in the degrees calculation wasn't changed. If the intention is to have the mathematical definition of degrees to radians, then the parameter in the factors table should be 180.0/Math.PI and then not depend on the earth radius.

If the current behaviour *is* intentional, then the magic number (111325) should be documented so the next person who sees this doesn't submit an issue (there's quite a few magic numbers in that table...).

mfedderly:
Fixes #1406 

Change the output of:
- @turf/distance
- @turf/rhumb-distance
- @turf/ellipse
- @turf/rectangle-grid


mfedderly:
Interestingly the packages/turf-shortest-path/test/out/simple.json seems to be worse, but I'm beginning to suspect that is an issue with shortest-path and not this change.

rowanwins (approved):
This change was always bound to have reverberations.

PR #2182, opened 2021-08-04
Fix @turf/quadrat-analysis test flake

### LINKED ISSUE #2181: @turf/quadrat-analysis test flaking
I had a post-merge build fail last night that was kind of weird.
https://github.com/Turfjs/turf/runs/3236418556?check_suite_focus=true

Turns out those tests generate a random input that sometimes quadrat-analysis declares to not be random. One such input is here: [randomPointSet.bad.txt](https://github.com/Turfjs/turf/files/6932124/randomPointSet.bad.txt)

In order to keep the tests from flaking, we can just retry with a set of known-good points that I generated.

mfedderly:
Just retry with a set of known good points if the first set of random points doesn't pass

Fixes #2181


twelch:
Why is it important to test using random input generated at runtime?  Is it adding value beyond the diversity of input generated over a long period of time and test runs?

mfedderly:
Yeah basically just slowly fuzzing us for errors with random input over several runs of the test, sqlite does something similar. We could just use the pre-selected one if we wanted but it didn't seem necessary to always use it given how infrequently the random stuff broke (1% of the time? maybe less?).

rowanwins (approved):
Seems quirky but 🤷‍♂️

PR #2183, opened 2021-08-04
Remove @turf/turf re-exports for 4.x compatability

### LINKED ISSUE #2180: Consider removing @turf/turf backwards compatible exports
I'd like to remove the following exported items from @turf/turf, which have been re-exported for backwards compatability.
https://github.com/Turfjs/turf/blob/master/packages/turf/index.mjs#L125-L147

If someone would like to make the argument that we should keep it, that's fine too, but if we're considering doing a bunch of breaks now would be a good time.

mfedderly:
Fixes #2180 

Reasoning here is that these have deprecated for a long while now, and people have had time to move off of them.


twelch (approved):
👍 I think there's agreement here, these should be removed.

PR #2185, opened 2021-08-04
@turf/union - Support arbitrary numbers of geojson inputs

### LINKED ISSUE #2162: turf.union not accepting unpacking: turf.union(...array)

- [6.5.0 / 5.1.6 ] The version of Turf you are using, and any other relevant versions.

I built an app last year to take a long list of 655 parcel coordinates and plot them on a map. Rather than load each parcel as a separate layer, I used turf.union to merge the layers into one. This reduced overhead in the browser and solved the problem of overlapping layers changing colors. I used unpacking like buffers_merged = turf.union(...buffers), where buffers is just an array of polygon features. This worked just fine when I build it but my app stopped working a while back. After spending some time troubleshooting, discovering the 6.5.0 version on github seemed to only accept two inputs, I reverted to the version of turf that was older than when I built the app, which happened to be 5.1.6 and this solved my problem. Not sure why the newer versions stopped looping through args...
image
image

mfedderly:
Related to #2162

I think I see why this method lost the ability to take arbitrary numbers of inputs at some point in history. The typescript definitions are a little wonky because you want both a `...rest` arg and an optional options object at the end. I made the choice to accept multiple options objects in the ...rest argument, but just use the one at the end of the file by just checking the arguments to see whether it was geojson or not.

Open to discussion on whether there's a better way to do the arguments.


rowanwins:
@mfedderly another option I thought about was whether for union we swap to a single argument of a `FeatureCollection`? 

Sometimes with these things I find it helpful to refer to doco from other spatial analysis libs,

[ArcGIS](https://pro.arcgis.com/en/pro-app/latest/tool-reference/analysis/how-union-analysis-works.htm) for example just has one argument which accepts a bunch of features (equivalent of a FC).

[QGIS accepts two arguments](https://docs.qgis.org/3.16/en/docs/user_manual/processing_algs/qgis/vectoroverlay.html#qgisunion), the first being an FeatureCollection, and the second of which is optional FeatureCollection, although this second argument doesn't do anything similar to what Turf currently provides through `polygon-clipping`.

mfedderly:
> single argument

Yeah that's totally better. I used a slightly more permissive array structure so that people don't need to wrap their geometries with features if they didn't already have them as features.

rowanwins:
Ideally we'd also do something similar to `@turf/intersect` and `@turf/difference` in terms of argument structure.

rowanwins (approved):

tremby:
This caught me out during a recent dependency upgrade. The change from n arguments to 2 arguments is not documented in the changelog but was definitely a breaking change! Adding a note about it there might help the next person in my position.

It looks like this commit caused the change: https://github.com/Turfjs/turf/commit/a19953bba7e21a92f77b12fe3ab49ba48978b483#diff-8f96011bd4ca99e0ab3677f6c968e9ebc320c46a7202239fcc7df63560a61540

packages/turf-union/index.ts
```diff
@@ -7,8 +7,7 @@ import { Feature, Polygon, MultiPolygon, Properties } from "@turf/helpers";
  * Takes two {@link (Multi)Polygon(s)} and returns a combined polygon. If the input polygons are not contiguous, this function returns a {@link MultiPolygon} feature.
  *
  * @name union
- * @param {Feature<Polygon|MultiPolygon>} polygon1 input Polygon feature
- * @param {Feature<Polygon|MultiPolygon>} polygon2 Polygon feature to difference from polygon1
+ * @param {Array<Feature<Polygon|MultiPolygon> | Polygon | MultiPolygon>} features Input features to be unioned
```
rowanwins, line 10: Is there any reason we don't make it a `FeatureCollection` rather than array? I think more traditionally we've used standard geojson arguments. 
Not going to die in a ditch about it though :)

mfedderly: Yeah I'll rework this to take a FeatureCollection to be consistent.

mfedderly:
Closed in favor of https://github.com/Turfjs/turf/pull/2247

Issue #2163, opened 2021-07-31
Proposal: Module renaming for v7


rowanwins:
If we're working towards a release of v7 then now would be a good time to rename any modules as a breaking change.

- [ ] `lineIntersect` to become `intersectionPoints`. Rationale being is that this module supports both polygons and lines, and I think the proposed name is a bit clearer as to what it does and returns.

Any other thoughts for candidates @mfedderly @JamesLMilner @twelch ?

JamesLMilner:
I agree that lineIntersect is a bit confusing considering it works on LineStrings and Polygons. I would be happy to rename it to intersectionPoints if that helps provide clarity to users

mfedderly:
I don't know of any other renames offhand, but I made a PR for lineIntersect https://github.com/Turfjs/turf/pull/2187

PR #2187, opened 2021-08-04
Rename @turf/line-intersect -> @turf/intersection-points

mfedderly:
Per #2163

Maybe worth holding off merging this until https://github.com/Turfjs/turf/pull/2033 lands.


twelch:
> Per #2163
> 
> Maybe worth holding off merging this until #2033 lands.

@mfedderly just noting that #2033 has landed, this might be ready to clean up and merge

PR #1749, opened 2019-08-22
Buffer TypeScript definition for negative radius

GabiAxel:
When a negative `radius` is used with `buffer` for a polygon, it is possible the result will be a multi-polygon when the original polygon is shaped like several "blobs" connected with narrow "tubes" - the tubes disappear entirely if they are thinner than radius*2 and the blobs shrink into separated polygons in a multi-polygon.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Have read [How To Contribute](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#how-to-contribute).
- [x] Run `npm test` at the sub modules where changes have occurred.
- [x] Run `npm run lint` to ensure code style at the turf module level.

Submitting a new TurfJS Module.

- [ ] Overview description of proposed module.
- [ ] Include JSDocs with a basic example.
- [ ] Execute `./scripts/generate-readmes` to create `README.md`.
- [ ] Add yourself to **contributors** in `package.json` using "Full Name <@GitHub Username>".

PR #2188, opened 2021-08-05
@turf/buffer - Update typescript definitions

mfedderly:
- The existing definitions seemed unnecessarily convoluted
- The existing definitions did not correctly include that a negative radius could result in a MultiPolygon for certain inputs (#1749)

I think I got this right but would appreciate a double check to make sure I didn't screw anything up.


twelch (approved):
The simplification looks right, based on a negative buffer radius producing a multi-geometry from a single.

PR #2218, opened 2021-10-31
Turf/Buffer - More precise documentation of supported units for buffering

marcjansen:
This changes the documentation of the buffer module to list all supported units. 

[`lengthToUnits`](https://github.com/Turfjs/turf/blob/cd719cde909db79340d390de39d2c6afe3173062/packages/turf-helpers/index.ts#L624-L633) (which is used when buffering) supports more than _just_ `miles, kilometers, and degrees`. I am not sure if this was not documented on purpose. 

I also wasn't sure whether I was supposed to commit the changed `README.md` as well.

Please review.

-------------------

Please fill in this template.

~~Submitting a new TurfJS Module.~~

Not applicable

### DIFF
--- a/packages/turf-buffer/index.js
+++ b/packages/turf-buffer/index.js
@@ -11,7 +11,18 @@ import {
 } from "@turf/helpers";
 
 /**
- * Calculates a buffer for input features for a given radius. Units supported are miles, kilometers, and degrees.
+ * Calculates a buffer for input features for a given radius. Units supported are:
+ *   * centimeters or centimetres
+ *   * degrees
+ *   * feet
+ *   * inches
+ *   * kilometers or kilometres
+ *   * meters, metres
+ *   * miles
+ *   * millimeters, millimetres
+ *   * nauticalmiles
+ *   * radians
+ *   * yards
  *
  * When using a negative radius, the resulting geometry may be invalid if
  * it's too small compared to the radius magnitude. If the input is a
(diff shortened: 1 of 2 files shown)


packages/turf-buffer/index.js
```diff
@@ -11,7 +11,18 @@ import {
 } from "@turf/helpers";
 
 /**
- * Calculates a buffer for input features for a given radius. Units supported are miles, kilometers, and degrees.
+ * Calculates a buffer for input features for a given radius. Units supported are:
```
mfedderly, line 14: Do you think that this is still a useful thing to add to the docs? I'd be inclined to just point users towards the typescript autocomplete for this field. Otherwise I'm mostly just concerned about keeping this list up to date with the actual definition in @turf/helpers. Maybe that's a silly concern since we probably don't often get new units of length.

smallsaucepan: We should definitely try to avoid repeatedly listing all the options in the individual packages.

There might have been autocomplete issues in the past due to .js/.dts/JSDoc inconsistencies. For example, some places have Units in the function definition, but then just string in the JSDoc. Hopefully as we weed out the last few JS packages that will self resolve.

For this case maybe we say, leave any mention of units off completely. What do you think @marcjansen  and @mfedderly?

mfedderly:
This appears to have been fixed https://github.com/Turfjs/turf/pull/2838

PR #2225, opened 2021-11-19
Upgrade rbush versions to avoid bundling issues with @turf/turf-unkink-polygon and @turf/collect

### LINKED ISSUE #2200: RBush is not a constructor
## Sample Code
https://codesandbox.io/s/turf-convex-error-h9pe5

```js
import { point, featureCollection } from "@turf/helpers";
import convex from "@turf/convex";

// data from https://turfjs.org/docs/#convex
const points = featureCollection([
  point([10.195312, 43.755225]),
  point([10.404052, 43.8424511]),
  point([10.579833, 43.659924]),
  point([10.360107, 43.516688]),
  point([10.14038, 43.588348]),
  point([10.195312, 43.755225])
]);

const hull = convex(points); // This line generates error
```
image

## Details
Turf Version: 6.5.0

- [x] The version of Turf you are using, and any other relevant versions.

JamesLMilner:
Please fill in this template.

 Resolves #2200 by:
 
 * Bumping `rbush` in `turf-unkink-polygon` to `3.0.1` and use `@turf/geojson-rbush` in `@turf/collect`  
 * Use new with `rbush` in respective places as it is now an ES6 class

### DIFF
--- a/packages/turf-collect/index.ts
+++ b/packages/turf-collect/index.ts
@@ -47,7 +47,7 @@ function collect(
   inProperty: string,
   outProperty: string
 ): FeatureCollection<Polygon> {
-  var rtree = rbush<Entry>(6);
+  var rtree = new rbush<Entry>(6);
 
   var treeItems = points.features.map(function (item) {
     return {
--- a/packages/turf-collect/index.ts
+++ b/packages/turf-collect/index.ts
@@ -1,15 +1,6 @@
-import { FeatureCollection, Polygon, Point } from "geojson";
-import turfbbox from "@turf/bbox";
+import { FeatureCollection, Polygon, Point, Feature } from "geojson";
 import booleanPointInPolygon from "@turf/boolean-point-in-polygon";
-import rbush from "rbush";
-
-interface Entry {
-  minX: number;
-  minY: number;
-  maxX: number;
-  maxY: number;
-  property: any;
-}
+import rbush from "@turf/geojson-rbush";
 
 /**
  * Merges a specified property from a FeatureCollection of points into a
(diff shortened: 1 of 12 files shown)


JamesLMilner:
This turned out to be more complicated then I initially thought, see https://github.com/Turfjs/turf/issues/2200#issuecomment-974478928

packages/turf-collect/index.ts
```diff
@@ -47,34 +38,24 @@ function collect(
...
     }
-    var bbox = turfbbox(poly);
-    var potentialPoints = rtree.search({
-      minX: bbox[0],
-      minY: bbox[1],
-      maxX: bbox[2],
-      maxY: bbox[3],
-    });
+
+    var potentialPoints = rtree.search(poly) as FeatureCollection<Point>;
```
twelch, line 49: I think the rbush constructor has a generic that lets you specify what type of thing will be loaded into the tree, then the search method (and other methods) will automatically return features of that type without needing to cast.

Notice line 52 on the left above previously specified the tree would contain type `Entry`
```
var rtree = rbush<Entry>(6)
```

Here's another example: https://github.com/Turfjs/turf/blob/d72985ce1a577b42340fed5fc70efe8e4bc8b062/packages/turf-line-overlap/index.ts#L54

You can do the same like

twelch (changes requested):
This PR looks good other than the comment I made to clean up the types.  Nice work.

packages/turf-collect/index.ts
JamesLMilner: I'm not sure we can do that here because the search type and initiated rbush type are different, i.e. we ingest points by search with a polygon. It ends up with a type error:

```
 Type '"Polygon"' is not assignable to type '"Point"'.
```

rowanwins:
@JamesLMilner is this PR now superceded by #2237 ?

JamesLMilner:
@rowanwins I believe so! I can close this out now

PR #2227, opened 2021-11-25
[@turf/points-within-polygon] Fix multipoint dropping properties.

### LINKED ISSUE #2222: PointsWithinPolygon removes properties from MultiPoint feature

- [x] The version of Turf you are using, and any other relevant versions.

Version used:  `"@turf/points-within-polygon": "^6.5.0"`
data that I'm trying to check: https://gist.github.com/KristofVDB1/b34dd4a3a342f39485ec4dd2a81b2aba 
Code: 
```js
const pwp = require('@turf/points-within-polygon').default;

const clipLayer = (geojson, boundingGeojson) => {
  let clippedFeatures = [];
  if (geojson.features[0].geometry.type === 'Point' || geojson.features[0].geometry.type === 'MultiPoint' ) {
    clippedFeatures = pwp(geojson, boundingGeojson).features;
  }
})
```

I currently am trying to use pwp to clip certain points when they're within a certain bbox. It works lovely for regular points, but for some reason, when I use this on a `MultiPoint`, it removes the properties from the feature. The clipping process still functions as intended. The gist contains two features. One from before being clipped, and one right after.

dpmcmlxxvi:
Please fill in this template.

Fixes #2222:
 - Adds the `properties` attribute from the input feature to the output feature
 - Does not add the optional `bbox` attribute. Not clear what to do. The bounding box may have changed and the overhead decision to compute a new one should probably be left to the user.
 - Does not add the optional `id` attribute. Since this is a new feature the `id` probably should not be carried over.

### DIFF
--- a/packages/turf-points-within-polygon/index.js
+++ b/packages/turf-points-within-polygon/index.js
@@ -59,7 +59,7 @@ function pointsWithinPolygon(points, polygons) {
         });
       });
       if (contained) {
-        results.push(multiPoint(pointsWithin));
+        results.push(multiPoint(pointsWithin, point.properties || {}));
       }
     } else {
       throw new Error("Input geometry must be a Point or MultiPoint");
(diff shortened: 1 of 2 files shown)


rowanwins (approved):
Looks good thanks @dpmcmlxxvi !

PR #2231, opened 2021-12-01
@turf/boolean-equal Add precision option

### LINKED ISSUE #1929: boolean-equal only works up to precision 6

boolean-equal only gives correct results for coordinates having 6 or less decimals.

boolean-equal incorrectly returns true for coordinates that are identical up to the 6th decimal, but have different 7th decimals.

Perhaps there should be an option to select the precision where features are practically equal

tunnelpuzzle:
Resolves #1929 by adding a precision option to booleanEqual.

I didn't include any change to README.md in this commit. It looks like the generate-readmes script doesn't look for typescript files so I'm assuming the docs are generated in another step?


rowanwins (approved):
Great work thanks @tunnelpuzzle !

alexcroox:
Did this make it live, seems undocumented currently

Issue #2191, opened 2021-08-08
Bundling @turf/convex with Webpack not working due to concaveman dependency


braco:
@turf/concave is erroring with `Queue is not a constructor` due to a longstanding issue with one of the dependencies:

https://github.com/mapbox/concaveman/issues/18

https://github.com/Turfjs/turf/blob/master/packages/turf-convex/package.json#L62

```
"@turf/convex": "^6.5.0",
```
```javascript
  const points = featureCollection([
    point([10.195312, 43.755225]),
    point([10.404052, 43.8424511]),
    point([10.579833, 43.659924]),
    point([10.360107, 43.516688]),
    point([10.14038, 43.588348]),
    point([10.195312, 43.755225]),
  ]);
  convex(points);
```

JamesLMilner:
Hi @braco, yes unfortunately this is a known issue on the concaveman project which effects us downstream. I believe the actual underlying issue is actually a problem with Webpack itself (https://github.com/mapbox/concaveman/issues/18#issuecomment-637603023). I wrote a response in that thread that explains [how to circumvent it short term if you're using Webpack](https://github.com/mapbox/concaveman/issues/18#issuecomment-873012722).

braco:
Thanks @JamesLMilner. I think that's a yarn only fix , without [npm-force-resolutions](https://github.com/rogeriochaves/npm-force-resolutions) or [overrides](https://github.com/npm/statusboard/issues/343)? I switched to [hull.js](https://www.npmjs.com/package/hull.js), which has been a much simpler fix.

mourner:
Released `concaveman` v1.2.1 with the fix, hope that helps.

stebogit:
Hey guys, just to make clear, would this be an issue on `@turf/concave` or `@turf/convex`? The title of the issue and the example say two different things.

Regardless, I don't see any issue with either modules, at least not with the doc's example reported by @braco:

https://turf-sandbox.netlify.app/?gist=85c24aba86193f3542ae1493d90e986d
![Screen Shot 2021-08-09 at 9 28 15 AM](https://user-images.githubusercontent.com/12717225/128740599-a4e18dc9-a63f-4abe-b504-add538ccc825.png)

JamesLMilner:
I think title/text should be `convex`. concave man is not used in concave : https://github.com/Turfjs/turf/blob/84110709afda447a686ccdf55724af6ca755c1f8/packages/turf-boolean-concave/package.json

[...]

PR #2236, opened 2021-12-21
Use buble to prohibit ES6 code ending up in turf.min.js

JamesLMilner:
So I've been thinking about  #2191  #2200 more extensively and realised the underlying problem is that as it stands we can't upgrade `rbush` to the latest version (and in turn `concaveman` which relies on it) directly because it is now an ES6 module. As more of the Node ecosystem moves to using ES6 this will become more and more common place and we can't keep relying on external deps (and there deps) to be written in ES5.

I was attempting in #2225 to skirt around this by trying to use the ES5 export in `rbush`, but ultimately we can't do the same for concaveman because it uses the latest version of `rbush`, which eventually ends up in `turf.min.js`.

Here I was thinking we could run [buble](https://github.com/rollup/plugins/tree/master/packages/buble) a minimal ES2015+ transpiler over the code that goes into `turf.min.js` to allow us to use any dependency we want inside our packages. Here I've bumped concaveman and rbush and you can see no ES6 code ends up in `turf.min.js` because it gets transpiled (this wouldn't be possible without buble).

What do you folks think?

### DIFF
--- a/packages/turf-collect/package.json
+++ b/packages/turf-collect/package.json
@@ -64,7 +64,7 @@
     "@turf/bbox": "^6.5.0",
     "@turf/boolean-point-in-polygon": "^6.5.0",
     "@turf/helpers": "^6.5.0",
-    "rbush": "2.x",
+    "rbush": "3.0.1",
     "tslib": "^2.3.0"
   }
 }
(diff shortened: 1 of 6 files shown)


JamesLMilner:
Closing in favour of #2237

PR #2237, opened 2021-12-22
Target ES2017

mfedderly:
Target ES2017 instead of ES5 for *most* of our packages.

Turf is a pretty small team, and we're spending a lot of time chasing down packages upstream of us that are not shipping ES5. At this point IE11 has fairly low marketshare and is due to be end-of-life'd in June of 2022. We're also actively making our packages larger and slower than they need to be for almost all users just to support IE11. To ease the transition, we can provide guidance to people consuming Turf packages to help them configure their own transpilation pipelines to keep supporting IE11 for their use cases.

In this PR:
- Target ES2017 for our typescript output
- Enforce ES2017 in the builds for our packages and our dependencies
- Note this in README.md and even note which packages will require transpilation for IE11 support
- Include a script that lets us find the packages needed for the README documentation
- Transpile @turf/turf's browser target with Babel for anyone using the main turf package from a CDN


rowanwins (approved):
LGTM

markstos:
I'm looking at possibly creating  a PR to replace `polygon-clipping` with `polyclip-ts` as a dep, but I'm running into problems because `polyclip-ts` doesn't support a `require()` interface. This PR seems to give permission to drop support for that, but I don't see where to drop support for Common JS for the modules that need to use this dep. Or perhaps I misunderstand what's happening in this PR. Thanks.

mfedderly:
@markstos This PR lets us ship stuff like `class` and `let` keywords that were otherwise prohibited before. Now that the browser ecosystem has mostly moved along (and IE is finally gone). Require vs import/export is actually sort of a separate concern, but from a quick look at the polyclip-ts package.json they ship UMD packages for CDN style consumption, and then code with import/exports to people who consume the main or module endpoint. Anyone consuming those should either reasonably support import/export (node) or be bundling their code anyhow (webpack, etc) as part of a build step.

As for swapping polygon-clipping for polyclip-ts, all of the consumers in the codebase are already using `import` statements anyhow. If polyclip-ts is shipping newer JS syntax than ES2017 in their npm package, the build should catch it and throw an error. Maybe swap out a single turf module, then run these commands in the root:

```
yarn build
yarn lint:escheck-require
yarn lint:escheck-esm
yarn lint:escheck-es5
```

That should make sure that everything meets the maximum ecma rules for Turf, which seems likely from a quick check.

I see that polyclip-ts is newer and shares some of the same authors. Is there any particular reason that you want to make the change?

markstos:
@mfedderly `polyclip-ts` fixes about 20 bugs in `polygon-clipping`, bugs which Turf inherits through the dependency. The `polygclip-ts` project has gone so far as to take 20 bugs reports and turn them into test suite cases. There's more discussion about it here:

https://github.com/Turfjs/turf/issues/2409#issuecomment-1461511072

The solution may be a little slower, but at least it's correct and doesn't crash.

However I'm not going to be able to sort what I need to to solve support for the various JavaScript editions and module styles soon.

PR #2238, opened 2021-12-24
Fix crash on turf.intersect for touching polygons

pietervdvn:

Submitting a new TurfJS Module: N/A

Hi,

While testing turfjs in MapComplete, I discovered that `turf.intersect(p0, p1)` would crash if these polygons shared a border over multiple points.

I've added a testcase for this in the `turf-intersect`-package - which indeed crashed. Then I fixed it by checking that the result of the polygon-clipping would return a correct linear ring (at least for the outer ring).

At last, I've updated the documentation to clarify that polygons that touch will return null as well.


pietervdvn:
These are the test polygons: 
image

JamesLMilner:
Seems sensible to me and is always good to cover edge (literally) cases. One for @rowanwins?

pietervdvn:
Cool! Thx for merging, looking forward to the next release!

twelch:
Actually @pietervdvn this PR was not merged back to turf master, I did the opposite and merged latest turf master to this PR to update it and run the CI tests.  Can you reopen it?

pietervdvn:
oops!

rowanwins:
Thanks for this PR @pietervdvn 

So the main question mark here is what is the right output. Prior to v6 when we use `jsts` for performing the operation the returned feature could be a point, line or poly. That functionality kinda got lost as we experimented with and transitioned to `polygon-clipping`, but I think it's probably the better behaviour that we could now reinstate.

So perhaps we should rejig this PR so it can once again provide that response? That would be a breaking change so we'd need to include it in the next major release.

Extract from [v5.1.5](https://github.com/Turfjs/turf/blob/862515c0fab6ab4ed135d5f40263452fe49e882a/packages/turf-intersect/index.js) jsdoc
````
@returns {Feature|null} returns a feature representing the point(s) they share (in case of a {@link Point} or {@link MultiPoint}), 
the borders they share (in case of a {@link LineString} or a {@link MultiLineString}), 
the area they share (in case of {@link Polygon} or {@link MultiPolygon}). 
If they do not share any point, returns `null`.
````

Does that make sense?

pietervdvn:
The old behaviour might be better, but I do not have the time nor experience to fix it.

rowanwins:
I've done some looking at this. `polygon-clipping` doesn't actually return the full line that intersects along the edge, it's returning duplicate points. And it also doesn't return a point where corners intersect.
So it's a bit more problematic than a simple fix and will need some more thought :(

PR #2241, opened 2021-12-29
Remove objectAssign from all packages

### LINKED ISSUE #2235: Uncaught SyntaxError: object-assign in browser (using top-level @turf/turf package, @web/dev-server, Typescript)

- [6.5.0 ] The version of Turf you are using, and any other relevant versions.

I noticed that object-assign is still used at some places like turf-isolines and turf-isobands.
Is it possible to get rid of these? I believe nowdays we ca  already use the spread operator and there is not a real need anymore for that helper.

I am trying to use turf in a typescript project and getting bugged by 
`:index.js:5 Uncaught SyntaxError: The requested module './../../../../object-assign/index.js' does not provide an export named 'default'`
and unfortunately not yet an idea how to fix this..

JamesLMilner:
Please fill in this template.

The aim here is to remove `objectAssign` from the codebase - in the process we move `turf-isolines` and `turf-isobands` to TypeScript.
 
 Resolves #2235


rowanwins (approved):
This all looks good to me and given that we're implementing #2237 I think we should be fine for browser compatibility.

PR #2247, opened 2022-01-15
FeatureCollection args for @turf/difference @turf/intersect @turf/union

mfedderly:
Fixes https://github.com/Turfjs/turf/issues/2162 for all three packages where it is relevant
Supercedes https://github.com/Turfjs/turf/pull/2185

### DIFF
--- a/packages/turf-intersect/index.ts
+++ b/packages/turf-intersect/index.ts
@@ -39,7 +39,7 @@ import polygonClipping from "polygon-clipping";
  *   [-122.520217, 45.535693]
  * ]]);
  *
- * var intersection = turf.intersect(poly1, poly2);
+ * var intersection = turf.intersect(turf.featureCollection([poly1, poly2]));
  *
  * //addToMap
  * var addToMap = [poly1, poly2, intersection];
--- a/packages/turf-intersect/index.ts
+++ b/packages/turf-intersect/index.ts
@@ -1,15 +1,20 @@
-import { Feature, GeoJsonProperties, MultiPolygon, Polygon } from "geojson";
+import {
+  Feature,
+  GeoJsonProperties,
+  MultiPolygon,
+  Polygon,
+  FeatureCollection,
+} from "geojson";
 import { multiPolygon, polygon } from "@turf/helpers";
-import { getGeom } from "@turf/invariant";
+import { geomEach } from "@turf/meta";
 import polygonClipping from "polygon-clipping";
 
 /**
- * Takes two {@link Polygon|polygon} or {@link MultiPolygon|multi-polygon} geometries and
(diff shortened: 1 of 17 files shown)


twelch:
Thanks for putting this up @mfedderly.  I like the single argument, solves most use cases.

The one case I considered a second input argument would support is clipping each item in a FeatureCollection A, one at a time against an entire FeatureCollection B (something I do fairly regularly in my work).  It wouldn't be any faster performance, just a convenience to move the necessary traversing/merging/calling of A items with B into the function. At any rate, seems like could be a separate non-breaking extension to what you have now. 🤔

packages/turf-intersect/index.ts
```diff
@@ -34,24 +39,29 @@ import polygonClipping from "polygon-clipping";
...
- * var intersection = turf.intersect(poly1, poly2);
+ * var intersection = turf.intersect(turf.featureCollection([poly1, poly2]));
  *
  * //addToMap
  * var addToMap = [poly1, poly2, intersection];
  */
 export default function intersect<P = GeoJsonProperties>(
-  poly1: Feature<Polygon | MultiPolygon> | Polygon | MultiPolygon,
-  poly2: Feature<Polygon | MultiPolygon> | Polygon | MultiPolygon,
+  features: FeatureCollection<Polygon | MultiPolygon>,
```
twelch, line 48: I see, so by narrowing to a single input parameter that's a FeatureCollection, as opposed to an array in your last PR, we lose the ability to accept simple Geometries, and the caller has to wrap.  Your change to hex-grid is an example.  Gives me mixed feelings.

What if we accept a FC or an array?  Like:
```
features: FeatureCollection<Polygon | MultiPolygon> | Feature<Polygon | MultiPolygon[] | (Polygon | MultiPolygon)[]
```
If `geomEach` is being used in the functions, perhaps it would "just work"?  Food for thought.  I'm not going to block on it though.

twelch (approved):

packages/turf-intersect/index.ts
twelch: I guess technically, an array of geometries become just a GeometryCollection in this case.  And well if you have a Feature[] I guess the established pattern with many turf methods is to just accept a FeatureCollection, so I get that argument.

mfedderly: For some reason GeometryCollection isn't generic, so I'm hesitant to change this. `geomEach` does not accept arrays in its current for. We can also widen the parameter's type after we release 7.0 without breaking compatibility. I'm also not sure I totally understand your use case, any chance you want to make a PR to better suit your usecase? I'm going to try to get this merged as-is today, but I'm happy to follow up.

PR #2273, opened 2022-03-20
line-intersect additional type cleanup

twelch:
@rowanwins Please consider this bit of type cleanup for #2033, removing use of any.  Plus one more bit of cleanup while I was there.  Seemed easier than trying to comment on it.

I tried to add a default function signature for findIntersections to the empty sweepline-intersections module declaration file you created,  but could not get VSCode to pick up the types.  Failing that, I at least improved from use of any to a more explicit cast for result of findIntersections()


rowanwins (approved):
LGTM, thanks @twelch 👍

PR #2279, opened 2022-03-25
Fix docs generation

mfedderly:
The README.md files were falling out of date in the individual packages. This due to a combination of the generate-readmes script not supporting typescript, and it not being enforced at commit time by husky. Additionally we updated `documentation` a long time ago, so there's a lot of READMEs that got updated to meet the new standard (and possibly brought in changes from the `index.(js|ts)` files too.

First and 3rd commits are the manual changes, 2nd has the automatic changes.


scripts/generate-readmes
```diff
@@ -1,56 +1,76 @@
...
+packages.forEach((packagePath) => {
+  const directory = path.parse(packagePath).dir;
+  let indexPath = path.join(directory, "index.js");
+  const pckg = load.sync(packagePath);
+  const name = pckg.name;
+  const diagrams = glob
+    .sync(path.join(directory, "diagrams", "*"))
+    .filter(isImage);
+
+  // some of the packages are typescript instead
```
mfedderly, line 36: This file got hit by prettier when I committed, but 36-39 are the actual changes.

package.json
```diff
@@ -32,6 +32,9 @@
       "eslint --fix",
       "prettier --write"
     ],
+    "packages/*/index.{js,ts}": [
```
mfedderly, line 35: This block should keep us from falling out of date in the future.

rowanwins (approved):
LGTM 👍

PR #2280, opened 2022-04-02
Use robust-predicates for point-on-line

rowanwins:
This PR updates boolean-point-on-line to use the `robust-predicates` which should help us get around floating point errors.


rowanwins:
Closing for now, might revisit in the future

PR #2281, opened 2022-04-09
Fix lineOffset spherical geometry errors

Simage:
Project geometries to Azimuthal Equidistant projection prior to transformation and then back to WGS84. 

There is an earlier similar PR however it has a unique implementation whereas this implementation duplicates the projection code from the buffer implementation. Ideally, this projection code would eventually be refactored into a utility and maintained in a single location


JamesLMilner:
@rowanwins perhaps the best person to weigh in on this? It seems sensible to me but perhaps there's side effects I'm not seeing

rowanwins:
This PR might help in some circumstances but I don't think it will resolve some of the bigger underlying issues unfortunantly.

PR #2296, opened 2022-05-31
Set NearestPointOnLine type to always include dist, index, location

### LINKED ISSUE #2285: Would the result of nearestPointOnLine ever have a undefined `dist`?
On https://github.com/Turfjs/turf/blob/master/packages/turf-nearest-point-on-line/index.ts `NearestPointOnLine` is stated as having `dist` be possibly undefined. However, is that actually possible for it to be undefined? If not, then it seems better for the types to say that `dist` can never be undefined.

JamesLMilner:
Please fill in this template.

dist, index and location should always be returned for a valid point and lines. As such `NearestPointOnLine` should remove the optional nature of these paremeters.

Resolve #2285


packages/turf-nearest-point-on-line/index.ts
```diff
@@ -47,8 +47,17 @@ function nearestPointOnLine<G extends LineString | MultiLineString>(
...
 ): NearestPointOnLine {
-  let closestPt: any = point([Infinity, Infinity], {
+  if (!lines || !pt) {
+    throw new Error("lines and pt are required arguments");
+  }
+
+  let closestPt: Feature<
+    Point,
+    { dist: number; index: number; location: number }
+  > = point([Infinity, Infinity], {
```
twelch, line 57: I think you can replace NearestPointOnLine with this cleaner type definition (where you pass the properties into the generic, rather than extending the Feature), and then this can just become type NearestPointOnLine?

packages/turf-nearest-point-on-line/index.ts
```diff
@@ -47,8 +47,17 @@ function nearestPointOnLine<G extends LineString | MultiLineString>(
   pt: Coord,
   options: { units?: Units } = {}
 ): NearestPointOnLine {
-  let closestPt: any = point([Infinity, Infinity], {
+  if (!lines || !pt) {
+    throw new Error("lines and pt are required arguments");
+  }
```
twelch, line 52: Nice addition

twelch (approved):
Looks good, I made one comment to tighten up some type use if you agree.

packages/turf-nearest-point-on-line/index.ts
JamesLMilner: Yes, no need for the named type i think!

twelch (approved):

twelch:
Looks good

PR #2316, opened 2022-07-26
@turf/clone - Fix cloning of features with null geometries (#2314)

### LINKED ISSUE #2314: @turf/clone@6.5.0 - Unable to clone FeatureCollection containing features with null geometry
When trying to clone a GeoJSON FeatureCollection that contains `null` geometries, @turf/clone (v. 6.5.0) throws "Cannot read properties of null (reading 'type')". 

The error originates from the `cloneGeometry` function https://github.com/Turfjs/turf/blob/2dd2ef88c44d847989fa0b82089f3f81fe433b83/packages/turf-clone/index.ts#L61

which could be changed to something like the following
```js
if (geojson.geometry == null) {
  cloned.geometry = null;
} else {
  cloned.geometry = cloneGeometry(geojson.geometry);
}
```

in order to resolve the bug.
I can create a PR if desired!

Here's an Observable notebook reproducing the behaviour:
https://observablehq.com/@chrispahm/turf-clone-unable-to-clone-featurecollection-containing-f

chrispahm:
A simple fix allowing to clone features that contain null geometries (fixes #2314)


rowanwins (approved):
Looks good to me thanks @chrispahm

Issue #2323, opened 2022-08-16
Polygon-smooth crashes when dealing with MultiPolygons including more than one polygon with hole(s)


Pitouli:
For example, if you test on this MultiPolygon, it will crashes:

```
{
  "type": "MultiPolygon",
  "coordinates": [
    [
      [
        [102.0, 2.0],
        [103.0, 2.0],
        [103.0, 3.0],
        [102.0, 3.0],
        [102.0, 2.0]
      ]
    ],
    [
      [
        [100.0, 0.0],
        [101.0, 0.0],
        [101.0, 1.0],
        [100.0, 1.0],
        [100.0, 0.0]
      ],
      [
        [100.2, 0.2],
        [100.8, 0.2],
        [100.8, 0.8],
        [100.2, 0.8],
        [100.2, 0.2]
      ]
    ],
    [
      [
        [105.0, 0.0],
        [106.0, 0.0],
        [106.0, 1.0],
        [105.0, 1.0],
        [105.0, 0.0]
      ],
      [
        [105.2, 0.2],
        [105.8, 0.2],
        [105.8, 0.8],
        [105.2, 0.8],
        [105.2, 0.2]
      ]
    ]
  ]
}
```

I found the issue and I am sending a PR (I simply create this ticket to link the PR to it)

rowanwins:
Resolved by #2324

PR #2324, opened 2022-08-16
polygon-smooth: fix crash with MultiPolygons (#2323)

Pitouli:
polygon-smooth was crashing when dealing with MultiPolygons including more than one polygon with hole(s) (#2323)


rowanwins (approved):
This looks great thanks @Pitouli . I did add a commit reverting a couple of changes to files (around rollup and ts) as they get managed globally by the project.

PR #2344, opened 2022-10-06
Provide proper coord index

rowanwins:
I had this case recently where I needed the proper coordinate index for polygons with holes. Currently the callback advertises

```
turf.coordEach(features, function (currentCoord, coordIndex, featureIndex, multiFeatureIndex, geometryIndex) {
````
But `coordIndex` just continually increments up as you work through nested rings so you can't find the coordinate properly using indexing.

So there are two options
- Add another argument to the callback at the end, which doesn't require a breaking change (that's the approach taken here)
  - if we go down that route I'll probably rename coordIndex to make it more explicit
- Introduce a breaking change so that coordIndex returns the proper index

In preparing this PR I noticed some comments in the test.js like
https://github.com/Turfjs/turf/blob/9e20dd5ee91f4852f63dfee81bb0b0f7d5c5ed66/packages/turf-meta/test.js#L1178
and
https://github.com/Turfjs/turf/blob/9e20dd5ee91f4852f63dfee81bb0b0f7d5c5ed66/packages/turf-meta/test.js#L1190

So someone intended to go for the breaking change at some stage...

I don't have a strong preference either way.

### DIFF
--- a/packages/turf-meta/README.md
+++ b/packages/turf-meta/README.md
@@ -10,11 +10,11 @@ Type: [Function][1]
 
 ### Parameters
 
-*   `currentCoord` **[Array][2]<[number][3]>** The current coordinate being processed.
-*   `coordIndex` **[number][3]** The current index of the coordinate being processed.
-*   `featureIndex` **[number][3]** The current index of the Feature being processed.
-*   `multiFeatureIndex` **[number][3]** The current index of the Multi-Feature being processed.
-*   `geometryIndex` **[number][3]** The current index of the Geometry being processed.
+- `currentCoord` **[Array][2]<[number][3]>** The current coordinate being processed.
+- `coordIndex` **[number][3]** The current index of the coordinate being processed.
+- `featureIndex` **[number][3]** The current index of the Feature being processed.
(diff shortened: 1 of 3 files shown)


twelch (commented):
Nice. I like your approach taken.  coordIndex as-is seems like it has value so it makes sense to just leave it and add a new parameter at the end.

The name coordIndex also seems okay, I don't feel strongly a name change is needed.  Perhaps a clarification in the docs what it is if missing?

Consider ringIndex instead of indexOnRing for consistency with the other names.

rowanwins:
Re naming I was thinking `coordIndex` perhaps could become `index`, or `absoluteIndex`? 

> Consider ringIndex instead of indexOnRing for consistency with the other names.

I think `ringIndex` could be confusing because it could be interpreted as which ring (eg an outer ring or a hole)... maybe if we do change the name of the 2 argument this could be named `coordIndex`?

JamesLMilner (commented):
I think the approach here is good - would be good to update the docs to document indexInRing. I think the naming conventions used make sense and we probably don't need to change coordIndex (happy for others to disagree).

smallsaucepan:
Let's go for gold and do the breaking change next major release.

PR #2346, opened 2022-10-09
Add support for holes to @turf/mask

### LINKED ISSUE #2230: Incorrect @turf/intersect result with polygon holes
@turf/intersect version: 6.5.0
@turf/mask version: 6.5.0
@turf/helpers version: 6.5.0

Intersect does not produce a correct result when a polygon contains a hole, that hole contains an island, and the polygon to be intersected intersects that island. The part intersecting the island is ignored in the output.

As a simple example, here we create a large square with a small square hole in the middle, and use turf mask to invert it, then intersect that resulting mask with another large square of the same dimensions but without the hole. In this example, result is null, when in fact it should be a polygon with the same dimensions as the small square:

```
import {polygon} from '@turf/helpers';
import turfMask from '@turf/mask';
import turfIntersect from '@turf/intersect';

// coordinates of a 16 x 16 square:
const square = [[[0,16],[0,0],[16,0],[16,16],[0,16]]];
// coordinates of the same square, but with a 4 x 4 hole in the middle:
const withHole = [[[0,0],[16,0],[16,16],[0,16],[0,0]],[[6,6],[6,10],[10,10],[10,6],[6,6]]];
const pSquare = polygon(square); // the square polygon
const pWithHole = polygon(withHole); // the square with hole polygon
const mask = turfMask(pWithHole); // the mask (inverse) of the square with hole
// the intersection of the mask with the square:
const result= turfIntersect(mask, pSquare); // result is null, should be a 4 x 4 square
```

rowanwins:
Currently `@turf/mask` ignores holes. The PR adds an options argument to the operation with an `ignoreHoles` options, when set to `true` a `MultiPolygon` is returned with the holes written as outer-rings in the MultiPolygon.

Resolves #2230


smallsaucepan:
Hi @rowanwins. Will try to revive this PR. Would you agree (current behaviour aside) that the default behaviour of mask should be to _include_ holes? That seems most intuitive to me.

If you agree will keep the PR as is to maintain existing behaviour, however add a note to default ignoreHoles to `false` as a breaking change for v8.

PR #2349, opened 2022-10-12
Add support for partial overlaps and multiple overlaps to @turf/line-overlap

### LINKED ISSUE #901: lineOverlap on two polygons giving no output

### LINKED ISSUE #1670: lineOverlap for two MultiLineStrings calculated incorrectly if there is some overlapping inside each MultiLineString
jsfiddle:
http://jsfiddle.net/theptyza/w76tqdey/

Two MultiLineStrings have some overlapping in their components.
```javascript
var multiLine1 = turf.multiLineString(
    [[[0,0],[0,1],[0,2],[0,3],[1,3],[2,3]],
    [[1,3],[2,3],[2,2],[1,2],[0,2],[0,1],[0,0]]]
);

var multiLine2 = turf.multiLineString(
    [[[0,0],[0,1],[0,2],[0,3],[1,3],[2,3],[2,2]],
    [[2,3],[2,2],[1,2],[0,2],[0,1],[0,0]]]
);
```
In multiLine1 both lines share the [[1,3],[2,3]] part.
In multiLine2 both lines share the [[2,3],[2,2]] part.

These two multilines overlap completely, but overlap is displayed like this (multiLine1 in blue, multiLine2 in green, overlap in red):
image

If I remove the shared parts in multilines the overlap is displayed correctly
http://jsfiddle.net/theptyza/w76tqdey/3
image

### LINKED ISSUE #1890: lineOverlap producing holes
Hey all,

I'm trying to use lineOverlap. I am comparing a `candidateLine` to a `targetLine`, and in this example the entirety of `candidateLine` should be overlapping `targetLine` (especially when `tolerance: .1` is set). However, instead of getting back one LineString that captures the whole of `candidateLine`, I get back two LineStrings, with a big hole in the middle.

https://jsfiddle.net/vwh5z6sx/10/

(`candidateLine` and `targetLine` are commented out on L33 and L34, respectively)

Any assistance would be much appreciated and thank you for this great package!

### LINKED ISSUE #2580: @turf/turf-line-overlap lineOverlap does not recognize partially overlaps.
The lineOverlap function currently does not recognize when two lines partially overlap. This also affects booleanOverlap.

You can see it in this example: 
The blue and red lines clearly overlap in the green area. However, this is not recognized because the implementation can currently only recognize an overlap if a segment is completely overlapping.

image

- Example on [geojson.io](http://geojson.io/#data=data:application/json,%7B%22type%22%3A%22FeatureCollection%22%2C%22features%22%3A%5B%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%230F0%22%2C%22fill%22%3A%22%230F0%22%2C%22stroke-width%22%3A25%7D%2C%22geometry%22%3A%7B%22type%22%3A%22LineString%22%2C%22coordinates%22%3A%5B%5B120%2C-25%5D%2C%5B150%2C-25%5D%5D%7D%7D%2C%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%23F00%22%2C%22fill%22%3A%22%23F00%22%2C%22stroke-width%22%3A10%2C%22stroke-opacity%22%3A1%2C%22fill-opacity%22%3A0.1%7D%2C%22geometry%22%3A%7B%22type%22%3A%22LineString%22%2C%22coordinates%22%3A%5B%5B120%2C-25%5D%2C%5B160%2C-25%5D%5D%7D%7D%2C%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%2300F%22%2C%22fill%22%3A%22%2300F%22%2C%22stroke-width%22%3A3%2C%22stroke-opacity%22%3A1%2C%22fill-opacity%22%3A0.1%7D%2C%22geometry%22%3A%7B%22type%22%3A%22LineString%22%2C%22coordinates%22%3A%5B%5B110%2C-25%5D%2C%5B150%2C-25%5D%5D%7D%7D%5D%7D)
- Playground on [jsfiddle](https://jsfiddle.net/ag31men8/48/)
- GeoJson: 
``` json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "stroke": "#0F0",
        "fill": "#0F0",
        "stroke-width": 25
      },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [120, -25],
          [150, -25]
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {
        "stroke": "#F00",
        "fill": "#F00",
        "stroke-width": 10,
        "stroke-opacity": 1,
        "fill-opacity": 0.1
      },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [120, -25],
          [160, -25]
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {
        "stroke": "#00F",
        "fill": "#00F",
        "stroke-width": 3,
        "stroke-opacity": 1,
        "fill-opacity": 0.1
      },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [110, -25],
          [150, -25]
        ]
      }
    }
  ]
}
```

Rinze-Smits:
There are a number of open issues for @turf/line-overlap: #1890, #1670 , #901 . These are all fixed. It also fixes issues with polygons with holes for which no issue was filed. These are all resolved with this patch.

The module now also finds overlaps where only part of a line segment overlaps only a part of another line segment.
Multiple overlaps had been partially covered in a previous patch, but not for all cases, this is now fixed.
A small refactor was done to reduce duplicated code.
The module puts a bit more effort into making linestrings from found matches.

The test result for #901 was changed due to different handling of duplicates (the testcase has a larger tolerance than the length of some matching line segments, which causes them to be matched multiple times).

Resolves #1890
Resolves #1670
Resolves #901
Resolves #2580


rowanwins:
I've had a bit of a look - the output from issue-#901.geojson looks a little strange as it's including duplicate coordinates, like it's zigzagging backwards and forwards. So I think we still need some extra work on this @Rinze-Smits , but thank you for looking into things!

packages/turf-line-overlap/test/out/issue-#901.geojson
```diff
@@ -59,34 +36,11 @@
...
-        "stroke": "#0F0",
-        "fill": "#0F0",
-        "stroke-width": 25
-      },
-      "geometry": {
-        "type": "LineString",
-        "coordinates": [
+          [-113.3384152645109, 53.52244409344282],
           [-113.33832502043951, 53.52244398828247],
           [-113.3384152645109, 53.52244409344282],
```
rowanwins, line 42: We're getting some duplicate coords here, eg 0 & 2 are the same, 1 & 3 are the same...

Rinze-Smits: I agree that this is still an issue, but in the current master branch the output of this testcase also contains duplicate coords and in addition to that the current master duplicates those linestrings. So I do not consider this to be a regression but a slight improvement. As I built this patch mainly to fix bugs where output was missing entirely, I do not think completely solving the existing duplicate coordinate issues in the output belongs here, but should rather go in a new patch. I'm willing to create that as well in the near future.

Rinze-Smits:
Thank you for reviewing! I agree that this is still an issue, but in the current master branch the output of this testcase also contains duplicate coords and in addition to that the current master duplicates those linestrings. So I do not consider this to be a regression but a slight improvement. As I built this patch mainly to fix bugs where output was missing entirely, I do not think completely solving the existing duplicate coordinate issues in the output belongs here, but should rather go in a new patch. I'm willing to create that as well in the near future.

mfedderly:
@Rinze-Smits I'd like to merge this if we can. Progress over perfection. Can you reserve the merge conflicts please?
Thanks!

Issue #2401, opened 2023-02-25
booleanEqual reports false for equal coordinates


Tomalak:
I'm using Turf 6.5.0.

The [documentation for `booleanEqual`](https://turfjs.org/docs/#booleanEqual) states

> `booleanEqual`  
> 
> Determine whether two geometries of the same type have identical X,Y coordinate values. 

And yet:

```js
const p1 = turf.point([0, 0], {prop: 'A'});
const p2 = turf.point([0, 0], {prop: 'B'});

console.log(turf.booleanEqual(p1, p2)); // -> false
```

The way I understand it, properties should not be taken into account when comparing point equality.

That, or the documentation needs a clarification.

smallsaucepan:
Thanks for raising this @Tomalak. Fix for this should be in the latest v7 alpha if you'd like to try it out.

PR #2412, opened 2023-03-12
Fix boolean-equal docs and added new tests (#2401)

J-Guenther:
Please fill in this template.

- [x ] Use a meaningful title for the pull request. Include the name of the package modified.
- [x ] Have read [How To Contribute](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#how-to-contribute).
- [x ] Run `npm test` at the sub modules where changes have occurred.
- [x ] Run `npm run lint` to ensure code style at the turf module level.

I tried to fix the documentation issue mentioned in #2401 . I am not sure if booleanEqual should compare properties at all but currently it does. And so does the "geojson-equality" package used for comparing. It all depends on whether you compare two geometries or two features. I updated the documentation and added some tests to reflect the current behaviour.

If you say booleanEqual should always omit properties and only compare geometries regardless whether the input is a geometry or a feature, I could create a fix for that and update the pull request with it.

### DIFF
--- a/packages/turf-boolean-equal/README.md
+++ b/packages/turf-boolean-equal/README.md
@@ -4,7 +4,7 @@
 
 ## booleanEqual
 
-Determine whether two geometries of the same type have identical X,Y coordinate values.
+Determines whether two geometries or features of the same type have identical X,Y coordinate values and properties.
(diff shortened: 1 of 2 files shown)


rowanwins (approved):
Looks good thanks @J-Guenther

JamesLMilner (approved):

PR #2415, opened 2023-03-17
feature: add unit option to nearest-point

mwenko:
Please fill in this template.

### Feature description

- The `nearestPoint` function does not allow to specify the `units` option. It uses under the hood the `distance` function, that accepts such an option. This change adds that option as optional to the `nearestPoint()` function.
- This allows to specify the unit you wanna have returned in the result `properties.distanceToPoint`
- This saves the conversion step of a caller that needs the value in a different unit than kilometers (which is the default)

### Tests

- A test case was added and succeeds.


JamesLMilner:
@mwenko this seems like a nice improvement and think it is a good candidate for merging.

Can I request two small changes:

- Can you update `types.ts`  to include the units option. So basically this line goes from:

```
const nearest = nearestPoint(targetPoint, points);
````

to:

```
const nearest = nearestPoint(targetPoint, points, { units: "kilometers" });
````

- Can you please run `npm run docs`  to update the docs

After this I think I can approve

File: packages/turf-nearest-point/index.ts [hunk 1]
```diff
@@ -21,6 +21,7 @@ export interface NearestPoint extends Feature<Point> {
  * @name nearestPoint
  * @param {Coord} targetPoint the reference point
  * @param {FeatureCollection<Point>} points against input point set
+ * @param {Object} [options={}] Optional parameters
```
rowanwins, line 24: In here we can also add
````
 * @param {string} [options. units ='kilometers'] the units of the numeric result
````

rowanwins:
Looks really good @mwenko - just one final little thing and then we'll merge :)

packages/turf-nearest-point/index.ts
mwenko: ✅ sure, done

Issue #1726, opened 2019-07-07
length and nearestPointOnLine - accuracy


sweco-sekrsv:
I cant figure this on out. Have a look at this example:
https://jsfiddle.net/krillmcrawl/swe67ncv/10/

Move the mouse along the line and check the console. 

The actual lenght of the line is 4367.02(FME) or 4367.078(Qgis)
The lenght reported from Turf is 4353.6 (13 meter difference seems quite far off)

In the example I have generated stations lines every 10 meters(for easier debugging). As you can see the turf.nearestPointOnLine reports the correct measures in the beginning of the line but after a while it starts "drifting". At station line 60 it turf reports 50.8 and further in at 850 turf reports 847.3 and even further in at 2660 turf reports 2651. 

The accuracy it not quite there. Is there a setting somewhere that I can use to increase accuracy of the length and nearestPointOnLine calculations?

Thanks. Kristian

zazagag:
Not sure is it the same problem, but if we have point from "nearestPointOnLine" and check it with "booleanPointOnLine", we wil get the "false" mostly. Mayby I am missing something in the documentation, but think what both functions are should be connected. I mean, if we get "nearestPointOnLine", we can prove it by using "booleanPointOnLine".
I am changed @kristiansvenssonwsp [fiddle](https://jsfiddle.net/zazagag/1kpvxcwm/5/) a little bit to show this behaiver.

ngottlieb:
@zazagag I'm running into this same problem, kind of problematic as I need to generate a point on a multiline then figure out which segment of the multiline it's on.

Appkidd:
Also seeing this issue. `nearestPointOnLine` and `booleanPointOnLine` should surely be connected as described above.

smallsaucepan:
This could well be the difference between spherical and ellipsoidal calculations. Using the two different methods in geopy produces the results mentioned above: using great_circle gives (4353m) vs geodesic (4367).

Have opened discussion #2470 on the best way to go forward. One option would be to port geopy's distance formula to Turf. However am seeking advice on which technique should become the default if we offer multiple options.

smallsaucepan:
Hi @JamesLMilner @rowanwins @twelch . Will take a look at this if you'd like to assign my way.

smallsaucepan:
The length accuracy part of this should be addressed by #2476. Use needs to pass in the appropriate ellipsoid for underlying distance() calculation to use.

smallsaucepan:
[...]

PR #2476, opened 2023-08-26
Allow optional ellipsoid definition to be passed to distance allowing for calculation types other than great circle

smallsaucepan:
Allow user to pass an optional ellipsoid definition to distance(), and if they do perform an ellipsoidal distance calculation rather than the default great circle. Partially addresses #1726 - length comes out as expected when users passes a WGS84 datum. This is likely the datum being used by the verification tool i.e. Qgis

Accuracy of nearestPointOnLine is a separate issue, largely unrelated to calculation method used by distance.

Can't use the latest version of the geodesy library due to ESM / CJS module loading issues :(

Please fill in this template.

- [✓] Use a meaningful title for the pull request. Include the name of the package modified.
- [✓] Have read [How To Contribute](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#how-to-contribute).
- [✓] Run `npm test` at the sub modules where changes have occurred.
- [✓] Run `npm run lint` to ensure code style at the turf module level.


smallsaucepan:
Not sure if I need to request a review from yourselves. However noticed that there are several recent PRs that look like they're waiting on automated build results? From May onwards it seems. Maybe something amiss and blocking the CI chain? Thanks @JamesLMilner @twelch @stebogit @rowanwins

packages/turf-distance/index.ts
```diff
@@ -1,13 +1,61 @@
...
+import {
+  radiansToLength,
+  degreesToRadians,
+  datums,
+  Coord,
+  Datum,
+  Units,
+  convertLength,
+} from "@turf/helpers";
+const LatLon = require("geodesy").LatLonEllipsoidal;
```
twelch, line 12: Can this be switched to an import?

packages/turf-distance/index.ts
```diff
@@ -1,13 +1,61 @@
...
+ */
+function distance(
+  from: Coord | Point,
+  to: Coord | Point,
+  options: {
+    units?: Units;
+    datum?: Datum;
+  } = {}
+) {
+  if (options?.datum) {
```
twelch, line 49: Should this be checking for a specific `Datum` value rather than whether it's truthy?

packages/turf-distance/index.ts
twelch: I see, you're just using the same logic as was there before.

packages/turf-distance/index.ts
twelch: I see, you're just using the same logic as was there before.

packages/turf-distance/index.ts
```diff
@@ -1,13 +1,61 @@
...
+ * from.properties.distance = distance;
+ * to.properties.distance = distance;
+ */
+function distance(
+  from: Coord | Point,
+  to: Coord | Point,
+  options: {
+    units?: Units;
+    datum?: Datum;
+  } = {}
```
twelch, line 47: I'm seeing that this could become the standard way to pass a `datum` option to a turf function, if it supports it.  Seems like a good non-breaking start down this parth.

twelch (commented):
I like what you're doing here @smallsaucepan with exposing a `datum` option.  I would like to see others feedback on this and taking geodesy as a dependency in turf

smallsaucepan:
Thanks for that @twelch . Have added another commit to fix a post-commit CI issue (my fault).

packages/turf-distance/index.ts
smallsaucepan: Yep, this has been converted to an import in the latest commit.

smallsaucepan:
This PR should be good to go now. Uses geodesy 2.3.0. Bundle size increases in my build from 568083 to 584415 (about 3%).

packages/turf-distance/index.ts
```diff
@@ -1,25 +1,33 @@
...
+} from "@turf/helpers";
+import LatLon from "geodesy/latlon-ellipsoidal-vincenty.js";
 //http://en.wikipedia.org/wiki/Haversine_formula
 //http://www.movable-type.co.uk/scripts/latlong.html
 
 /**
- * Calculates the distance between two {@link Point|points} in degrees, radians, miles, or kilometers.
- * This uses the [Haversine formula](http://en.wikipedia.org/wiki/Haversine_formula) to account for global curvature.
+ * Calculates the distance between two {@link Point|points}.
+ * If a specific datum is passed, uses a geodesic ellipsoid calculation.
```
twelch, line 16: I think you are technically correct here in that if the user passes any valid datum object, that the code as it is now will run `geodesicEllipsoidDistance`.  But that doesn't seem clean/clear.  Technically, I could possibly make up a datum and it will still run that one function right? But why? It seems like it would be better to accept the one expected ellipsoid datum, and throw an error otherwise.  Also see my comment below about exposing just a string parameter to specify the datum.

twelch (commented):
Thanks @smallsaucepan for the thoughtful approach here and updates.  I accept the use of the geodesy module to offer this functionality and potentially more in the future.  I just have some questions/thoughts on cleaning up the interface/docs for using it.  I appreciate your feedback.

PR #2489, opened 2023-09-09
Replace ts-node with tsx in doc and test related build scripts

smallsaucepan:
This PR replaces ts-node with tsx in doc generation and test scripts wherever possible to allow us more flexible use of third party libraries.

Background: as part of working on issue #1726 and associated PR #2476 I ran in to some problems where ts-node fell over trying to run the turf unit tests. A third party library we would like to use introduced import mechanisms that ts-node wasn't coping with when running tests. This seems to be related to mixing CJS / ESM imports.

Note that the turf library itself was building fine - just the tests were failing.

Once this PR is merged it will clear the way for PR #2476 to progress.

n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "progress": "*",
     "rollup": "^2.34.2",
     "tape": "*",
-    "ts-node": "^9.0.0",
+    "tsx": "^3.12.8",
     "typescript": "^3.8.3",
     "yamljs": "*"
   }
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "progress": "*",
     "rollup": "^2.34.2",
     "tape": "*",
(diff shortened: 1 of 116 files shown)


package.json
```diff
@@ -65,6 +65,7 @@
     "rollup": "^2.34.2",
     "tape": "*",
     "ts-node": "^9.0.0",
```
mfedderly, line 67: Can we delete this ts-node reference as well now?

mfedderly (approved):

package.json
smallsaucepan: Would be good to tidy up. MRL relies on it at the moment though, so we need to keep it at least at the monorepo level. Will aim to remove (or at least update) in a subsequent PR.

PR #2505, opened 2023-10-05
Add the support/6.x branch to the CI config so it builds when pushed to.

smallsaucepan:
Configure github tools so CI builds the support/6.x branch when it is updated.


twelch (approved):

smallsaucepan:
Closing this as it looks like we're pushing ahead with v7 rather than maintenance releases of 6.5.0.

Issue #2504, opened 2023-10-04
@turf/unkink-polygon unkinking polygons with lots of points results in stack overflow


2chen:
version: 6.50

- we have a polygon that looks something like: ```{ type: "polygon", coordinates: [ Array(93031), Array(383), Array(79) ] }```
- calling unkinkPolygon on it results in `RangeError: Maximum call stack size exceeded`
- the relevant error is being thrown in this line: https://github.com/Turfjs/turf/blob/dd35b52725945b4fa29a98d9a550733e06cc222e/packages/turf-unkink-polygon/lib/simplepolygon.js#L47

- the underlying issue comes from trying to call `Array.push.apply` (or calling `apply` on any function) with a huge list of arguments (see https://stackoverflow.com/questions/22123769/rangeerror-maximum-call-stack-size-exceeded-why)

PR #2506, opened 2023-10-05
@turf/unkink-polygon - Fix a potentially unbounded array.push.apply

mfedderly:
See #2504 for the writeup.

### DIFF
--- a/packages/turf-unkink-polygon/test.js
+++ b/packages/turf-unkink-polygon/test.js
@@ -36,6 +36,28 @@ test("unkink-polygon", (t) => {
   t.end();
 });
 
+test("issue #2504", (t) => {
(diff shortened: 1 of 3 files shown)


packages/turf-unkink-polygon/test.js
```diff
@@ -36,6 +36,28 @@ test("unkink-polygon", (t) => {
   t.end();
 });
 
+test("issue #2504", (t) => {
+  // fill coords with a circle with an arbitrary number of points
+  const coords = [];
+  const points = 1000000;
```
mfedderly, line 42: This number has to be *really* large in node, I think the limit in Chrome is somewhat smaller

twelch (approved):

PR #2513, opened 2023-10-12
Export dist/es/index.d.ts files for use with Volar

duncanfinney:
**Description of Changes**
- Removed the `--declaration=false` to output `dist/es/index.d.ts` files for use with volar
- Volar is the vue + typescript toolchain for vscode: https://marketplace.visualstudio.com/items?itemName=Vue.volar

image

**Current Workaround**
- `yarn add --dev @types/turf`
- This is kind of confusing because the micro-packages like `@turf/intersect` should be in the typing package `@types/turf__intersect` package for the auto search to work

**Checklist**

### DIFF
--- a/.monorepolint.config.ts
+++ b/.monorepolint.config.ts
@@ -176,7 +176,7 @@ module.exports = {
             build: "npm-run-all --npm-path npm build:*",
             "build:js": "tsc",
             "build:es":
-              'tsc --outDir dist/es --module esnext --declaration false && echo \'{"type":"module"}\' > dist/es/package.json',
+              'tsc --outDir dist/es --module esnext && echo \'{"type":"module"}\' > dist/es/package.json',
           },
         },
         includePackages: TS_PACKAGES,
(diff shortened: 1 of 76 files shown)


smallsaucepan:
Hi @duncanfinney . Thanks for this. Are there any downsides you can see, or things we might need to mention in the documentation? For example would we need to advise anyone who hast the @types/turf workaround installed to remove it?

smallsaucepan:
Took a closer look at the error message and think we might need to look into this in more detail.

The message says:

> There are types at ‘/home/…/@turf/intersect/dist/js/index.d.ts, but this result could not be resolved when respecting package.json “exports”.

To summarise, the types are in the package, referenced from the "types" field in package.json which is as it should be. Volar finds them, but some other configuration prevents it resolving them confidently. Does that sound accurate to you?

Rather than duplicate the d.ts in the package, I'd lean more towards correcting the `exports` field (if that's the issue) so that Volar can be confident to use the types it does find. Are you open to doing a little hacking about in your local node_modules directory to try a few options?

smallsaucepan:
@duncanfinney going to retire this PR in favour of #2555 which should have Turf export declaration files more predictably and consistently. Once the next alpha drops (alpha.3 by my count) please give it a try and raise an issue if you run into problems.

PR #2519, opened 2023-10-21
Upgrade typescript, prettier, et al to continue to modernise build environment

smallsaucepan:
Upgrading typescript, prettier and a couple of other development libraries. Aims to modernise build environment. Tried to keep actual business logic code changes to a minimum. Largely succeeded as most changes are punctuation, or in a few cases tightening of type definitions.

Please fill in this template.

n/a


packages/turf-bbox-clip/index.ts
```diff
@@ -37,7 +37,7 @@ import { lineclip, polygonclip } from "./lib/lineclip";
  */
 export default function bboxClip<
   G extends Polygon | MultiPolygon | LineString | MultiLineString,
-  P = GeoJsonProperties
+  P extends GeoJsonProperties = GeoJsonProperties,
```
twelch, line 40: Is this duplicative? Does `P extends GeoJsonProperties` already imply default to `GeoJsonProperties`?

package.json
```diff
@@ -39,18 +39,19 @@
   },
   "devDependencies": {
     "@types/geojson": "7946.0.8",
-    "@types/node": "^17.0.21",
-    "@typescript-eslint/eslint-plugin": "^4.8.0",
-    "@typescript-eslint/parser": "^4.8.0",
+    "@types/node": "18.11.9",
```
twelch, line 42: Is the intention to lock the version for this particular dependency?  Just curious.

twelch (approved):
Thanks for keeping this PR slim.  I imagine there was more it could have taken on.  See my few comments but otherwise I approve.

packages/turf-bbox-clip/index.ts
AminoffZ: TypeScript complains without it ☹:
<pre>
export default function bboxClip<
  G extends Polygon | MultiPolygon | LineString | MultiLineString,
  P = GeoJsonProperties
>(feature: Feature<G, P> | G, bbox: BBox) {
  const geom = getGeom(<ins>feature</ins>);
  ...
}
</pre>
> Argument of type 'G | Feature<G, P>' is not assignable to parameter of type 'G | Feature<G, GeoJsonProperties>'.
>   Type 'Feature<G, P>' is not assignable to type 'G | Feature<G, GeoJsonProperties>'.
>     Type 'Feature<G, P>' is not assignable to type 'Feature<G, GeoJsonProperties>'.
>       Type 'P' is not assignable to type 'GeoJsonProperties'.
>         Type 'P' is not assignable to type '{ [name: string]: any; }'.ts(2345)
> index.ts(40, 3): This type parameter might need an `extends { [name: string]: any; }` constraint.
> index.ts(40, 3): This type parameter might need an `extends GeoJsonProperties` constraint.

package.json
smallsaucepan: Ran in to problems straying too far from this version. For example at time of committing 18.18.6 falls over. Could maybe loosen to patch ~ level updates, though want to understand more how the relationship works between this and whatever depends on it first.

smallsaucepan:
Thanks for taking a look @twelch

PR #2520, opened 2023-10-22
40% faster turf-area

AminoffZ:
This is a continuation on #1387, again only modifying the `function ringArea(coords: number[][]): number`. Fixed the issue with turf-isobands. Had to REGEN=true for turf-nearest-neighbor-analysis. There were some floating point differences on random-large-study-area.json where `area(studyArea)` would return 
> 1396158428.635545  

instead of

> 1396158428.6355453

## Bench

Before

> polygon x 5,999,571 ops/sec ±0.37% (94 runs sampled)

After

> polygon x 8,510,024 ops/sec ±0.28% (96 runs sampled)

Please fill in this template.


twelch:
Thanks @AminoffZ, full test suite in CI is running now.

After a quick initial review, when I read your comment "Fixed the issue with turf-isobands. Had to REGEN=true for turf-nearest-neighbor-analysis", I was expecting to see code in your PR somewhere related to it, but it wasn't clear to me.  Can you say more about this comment and any code changes for it?  Perhaps you possibly forgot to commit some of the code.  Thanks.

AminoffZ:
Hi, the fixes for turf-isobands is in 7c5d3824e2972b7012191be67e36549798fdcb24 along with some additional speed improvements. TBH I'm not entirely familiar with the testing process, but plainly, what I did was set REGEN=true when running the turf-nearest-neighbor-analysis test, which made it pass when it did not before. The result of running that was is in the last two commits.

twelch:
Thank you.  I guess I was expecting the REGEN=true to be a code change as well.  I can see now there is a level to the repo/testing I don't understand yet either.

twelch (approved):
Thank you @AminoffZ for this contribution, I approve.  I see now that you used REGEN=true to update the test outputs, which look very very close to the previous.

twelch:
@AminoffZ before I merge, can you go ahead and update the performance benchmark in `turf-area/bench.js`.  I know the speed will be relative to your machine, that's okay for now.

PR #2523, opened 2023-10-26
Relocate geojson-rbush module into turf monorepo for easier maintenance

smallsaucepan:
PR to bring the currently standalone geojson-rbush repo in to the Turf monorepo. Per the standalone project it's not meant to be used outside of Turf, so might as well make it part of the overall project for ease of integration and maintenance.

Not sure if it's been kept separate due to some project history. Did make attempts to find out though wasn't able to establish anything concrete.

This PR would be part one of a two step process. Once this PR is merged and the module published (presumably as part of 7.0.0-alpha.3) packages that depend on geojson-rbush can be switched over to use @turf/rbush for alpha.4

Please fill in this template.

Sort of? This is essentially an existing module. Retained README, etc as best as possible from standalone repo. Can revise if need be.


packages/turf-rbush/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@turf/rbush",
```
mfedderly, line 2: I think we should call this `@turf/geojson-rbush` for consistency and in case we ever want to pull the `rbush` dependency in. (And therefore the directory name may need to be changed too)

packages/turf-rbush/LICENSE.geojson-rbush
```diff
@@ -0,0 +1,20 @@
+The MIT License (MIT)
```
mfedderly, line 1: Is this license block the same as `LICENSE` with the exception of crediting Denis Carriere? We may need to just figure out how to combine the two and get rid of this one.

packages/turf-rbush/README.md
```diff
@@ -0,0 +1,217 @@
+# GeoJSON RBush used in TurfJS
+
+This fork of geojson-rbush is configured and published for the sole purpose of importing it to [TurfJS](https://github.com/Turfjs/turf).
+
+More information about the original library: https://github.com/DenisCarriere/geojson-rbush
+
+> This library is not intended to be used outside of TurfJS
```
mfedderly, line 7: I'm not sure we need to persist this warning? Its kind of a nice optimization on top of the rbush if you're dealing with geojson.

packages/turf/test.js
```diff
@@ -128,7 +128,7 @@ test("turf -- MIT license", (t) => {
     const { license } = pckg;
     if (license !== "MIT") t.fail(`${name} (license) must be "MIT"`);
     if (fs.readFileSync(path.join(dir, "LICENSE"), "utf8") !== text)
-      t.fail(`${name} (LICENSE) is different from @turf/turf`);
+      t.fail(`${name} (LICENSE) content is different from @turf/turf`);
```
mfedderly, line 131: I hope that this didn't take too long to figure out 🤦

mfedderly (commented):
I didn't review the code files too closely, since I assume they're copy/paste. If you want me to be more thorough let me know. I'd be happy to merge this after changing the package name to @turf/geojson-rbush

packages/turf-rbush/package.json
smallsaucepan: Sure thing. Will make that change.

packages/turf-rbush/LICENSE.geojson-rbush
smallsaucepan: Just being extra cautious to keep as hadn't heard from Denis, and wasn't 100% of the history. Will see if merging is possible.

packages/turf/test.js
smallsaucepan: No, easy enough to track down 😉

packages/turf-rbush/LICENSE.geojson-rbush
mfedderly: I think the package.json has a contributors field, but I'm not a lawyer so I don't fully understand the nuance of having him listed directly in the license file vs being there. I'm happy enough to just change the testing to omit this license file from the checks if we have to.

mfedderly (approved):

PR #2525, opened 2023-10-27
Fix turf-line-arc to generate the correct number of steps in the generated arc

### LINKED ISSUE #2524: turf-line-arc not providing requested number of steps
While looking at #2446 (already fixed I think) noted that the number of coordinates in the test case didn't match the number of segments expected (15).

The code in lineArc to add the next segment angle seems odd ([L69](https://github.com/Turfjs/turf/blob/e5afd5c38b0f8422172a3c64ca14118bd138b54e/packages/turf-line-arc/index.ts#L69C5-L69C47)):

`alfa = arcStartDegree + (i * 360) / steps;`

This seems to be adding a portion of a *full circle* each iteration, rather than a portion of the arc between start and end bearing. I suspect this causes the logic to take larger leaps between each iteration and produce fewer segments than expected.

Will investigate further, though please someone pipe up if I've misunderstood what the steps option is supposed to convey.

- [x] The version of Turf you are using, and any other relevant versions. - 7.0.0-alpha.2

smallsaucepan:
Resolves #2524 by fixing a bug in the way lineArc steps its way through the arc and generates coordinates for each segment. Lots of test fixture changes, however these are mostly arcs that are now being generated more smoothly with the default 64 segments, whereas before they would be generated with fewer than 64.

Please fill in this template.

n/a


packages/turf-line-arc/test.js
```diff
@@ -24,12 +24,54 @@ test("turf-line-arc", (t) => {
...
     const results = featureCollection([geojson, arc]);
-
     if (process.env.REGEN) write.sync(directories.out + filename, results);
     t.deepEquals(results, load.sync(directories.out + filename), name);
+    // Check the resulting arc geometry has the expected number of points.
+    // undefined or 0 steps should default to 64 + 1 points, 1 to 1 + 1,
+    // 2 to 2 + 1, ...
+    const expectedPoints = geojson.properties?.steps
+      ? geojson.properties?.steps + 1
+      : 64 + 1;
```
mfedderly, line 37: WIthout any context seeing that the arc would get N+1 points feels weird. The jsdoc on the function doesn't indicate that it would be +1 either.

packages/turf-line-arc/index.ts
```diff
@@ -57,21 +57,19 @@ export default function lineArc(
...
+  let alpha = arcStartDegree;
   const coordinates = [];
   let i = 0;
-
-  while (alfa < arcEndDegree) {
+  // How many degrees we'll swing around between each step.
+  const arcStep = (arcEndDegree - arcStartDegree) / steps;
+  // Add coords to the list, increasing the angle from our start bearing
+  // (alpha) by arcStep degrees until we reach the end bearing.
+  while (alpha <= arcEndDegree) {
```
mfedderly, line 67: I wonder if this gets weird with floating point logic with N steps and very occasionally skips the final step.

smallsaucepan: Had crossed my mind. I did some simple testing and if there are precision errors in the reverse direction they tend to go low:

```
let steps = 1234567890; // Tried various values for steps
const x = 1 / steps;
// x 8.10000007371e-10

// Re-multiplication never seems to *exceed* the original value
const t = 0 + x * steps;
// t 1
// t 0.9999999985906001
// t 0.9999999978858999
```
If the bearing tends a little short then the <= should let that loop get as close as possible to the endBearing without going a whole step over.

We could go back to a "calculate all the step angles except the last one and use the endAngle for the last" approach. From what I can see above though JS floating point shoudl be ok, and the code is simpler.

packages/turf-line-arc/test.js
mfedderly: I think you can 'understand' steps as either the number of points or the number of segments. It isn't documented so I suppose we're free to do whichever we want (and maybe document it?). We could maybe consider the previous behavior as the expected behavior. I'm not really tied to that as being important, but maybe we should document whatever the outcome is in the jsdoc at least.

packages/turf-line-arc/index.ts
mfedderly: With a slightly different construction I think we can go over, but I appreciate you digging in. I feel good about this and we can always substitute the arcEndDegree for the final point later if there's something we didn't forsee.

```
let steps = 1e5;
const x = 1/steps;
let v = 0;
for (let i=0; i<steps; i++) {
    v+=x;
}
v
0.9999999999980838


let steps = 1e10; // warning this is pretty slow
const x = 1/steps;
let v = 0;
for (let i=0; i<steps; i++) {
    v+=x;
}
v
1.0000000694756235
```

mfedderly (approved):

packages/turf-line-arc/test.js
smallsaucepan: I'll push a commit updating the docs, and then merge. Thanks for reviewing 👍

packages/turf-line-arc/index.ts
smallsaucepan: Sounds good.

PR #2527, opened 2023-11-01
Use marchingsquares lib for turf-isoband and turf-isoline, fixes "first break" bug

### LINKED ISSUE #1797: isobands fails on flat data

### LINKED ISSUE #2129: Isolines ignores first break value

- [ 6.3.0] The version of Turf you are using, and any other relevant versions.

Just use the sample code like this:

``` javascript
import * as turf from '@turf/turf';

var extent = [0, 30, 20, 50];
var cellWidth = 100;
var pointGrid = turf.pointGrid(extent, cellWidth, { units: 'miles' });

for (var i = 0; i < pointGrid.features.length; i++) {
  pointGrid.features[i].properties.temperature = Math.random() * 10;
}
var breaks = [5, 10];

var lines = turf.isolines(pointGrid, breaks, { zProperty: 'temperature' });
console.log(JSON.stringify(lines.features));
```

Then we get the result, notice the lines with break value '5' are not generated.

``` json
[{"type":"Feature","properties":{"temperature":10},"geometry":{"type":"MultiLineString","coordinates":[]}}]
```

And if the breaks are set as [0, 5, 10], the result will contain lines with break value '5', but lost lines with break value '0'. In a word, the isolines function seems lost the first break lines.

smallsaucepan:
Refactoring and fixes to turf-isoband and turf-isoline. Use third party marchingsquares module instead of hosting a copy of it internally. Resolves #1797 (sort of - it was already fixed, but was still open and we've added a regression test for it here) and also chose to resolve #2129 while we were in there.

Please fill in this template.

n/ahttps://docs.github.com/articles/closing-issues-using-keywords


packages/turf-isolines/bench.js
```diff
@@ -2,7 +2,7 @@ import fs from "fs";
 import path from "path";
 import load from "load-json-file";
 import Benchmark from "benchmark";
-import matrixToGrid from "matrix-to-grid";
+import matrixToGrid from "./lib/matrix-to-grid";
```
stebogit, line 5: what's the purpose of using these internal modules? That is, why not using the npm versions?

packages/turf-isolines/index.ts
```diff
@@ -1,8 +1,9 @@
 import bbox from "@turf/bbox";
+import clone from "@turf/clone";
 import { coordEach } from "@turf/meta";
 import { collectionOf } from "@turf/invariant";
 import { multiLineString, featureCollection, isObject } from "@turf/helpers";
-import isoContours from "./lib/marchingsquares-isocontours";
```
stebogit, line 5: you probably also want to remove the `./lib/marchingsquares-isocontours.js` file

packages/turf-isolines/test.js
```diff
@@ -54,6 +55,39 @@ test("isolines", (t) => {
...
+
+  lines.features.push(
+    lineString(getCoords(envelope(points))[0], {
+      description: "Debug line for testing",
+      stroke: "#F00",
+      "stroke-width": 1,
+    })
+  );
+
+  // Make sure an isoline is created for each break, and that it's
```
stebogit, line 82: > that ~it's~ its geometry isn't empty.

packages/turf-isolines/index.ts
```diff
@@ -151,11 +157,35 @@ function rescaleIsolines(
     point[1] = point[1] * scaleY + y0;
   };
 
-  // resize and shift each point/line of the createdIsoLines
+  // Ok. We need to *clone* the returned isolines due to the behaviour of
```
stebogit, line 160: > ~Ok.~ We need to clone the returned isolines due to the behaviour of the marchingsquare library ~behaviour~.

packages/turf-isolines/index.ts
```diff
@@ -151,11 +157,35 @@ function rescaleIsolines(
...
+  // run twice, leading to one point in the isoline rendering WAY out of
+  // position.
+  const clonedIsoLines: Feature<MultiLineString>[] = [];
   createdIsoLines.forEach((isoline) => {
+    clonedIsoLines.push(clone(isoline));
+  });
+
+  // resize and shift each point/line of the createdIsoLines
+  clonedIsoLines.forEach((isoline) => {
     coordEach(isoline, resize);
```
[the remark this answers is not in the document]
stebogit: on a second thought, maybe the resize function should just not mutate the values, but return a new array:
```js
const resize = (point: number[]) => [
  point[0] * scaleX + x0,
  point[1] * scaleY + y0,
];
```

stebogit (approved):
👍🏻

packages/turf-isolines/bench.js
smallsaucepan: For matrix-to-grid the file committed to turf differs from the one in npm. Looks like @DenisCarriere included a tweak when it was originally added? See https://github.com/Turfjs/turf/blame/2faafab9162bcadd35a79ec329f578881038b5ec/packages/turf-isolines/lib/matrix-to-grid.js#L45

More than happy to use the NPM version. Do you want to roll in that change and re-release? I'm not sure what the original justification was though.

For now will go with the internal modules and we can refactor this depending on what you'd like to do.

packages/turf-isolines/index.ts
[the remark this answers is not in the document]
smallsaucepan: I think the resize function has to mutate in place - coordEach discards the return value of the callback.

It's more a quirk of the marchingsquares library which seems to re-use the first coordinate array as the closing point as well. So the callback passed to coordEach is being unexpectedly run twice on a single set of numbers i.e.

What we get back from marchingsquares:

[0] -> [1, 1]
[1] -> [2, 2]
[2] -> [3, 3]
[3] -> [1, 1] // <-- not another array that also contains [1, 1], the **exact same** array referred to by [0]

What we end up with after coordEach (if callback appends an "x" for the sake of visualisation):

[0] -> [1xx, 1xx]
[1] -> [2x, 2x]
[2] -> [3x, 3x]
[3] -> [1xx, 1xx] // in our case making one point way off target

I'll tweak the workaround to maybe only clobber the last coord in each linestring with a cloned value, then run the resize function over that.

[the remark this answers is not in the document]
smallsaucepan: Looks like this might have been a red herring. Pulled out the cloning operation to rework it and the tests still passed. Suspect adding linearRing: false to marchingsquares.isoContours() may have fixed it already. CI tests have passed so will merge. Thanks for questioning it @stebogit 👍

PR #2531, opened 2023-11-09
Replace geojson-equality with a fresh implementation that fixes precision handling (boolean-equal / boolean-overlap)

### LINKED ISSUE #2530: booleanEqual check does not properly handle precision
Turf version: 6.5.0/7.0.0a

The way `booleanEqual` check handles precision is incorrect. The following example illustrates the problem:

```
const point1 = {"type": "Point", "coordinates": [100.55719328, 23.0365925]}
const point2 = {"type": "Point", "coordinates": [100.55719327926636, 23.03659249995325]}

booleanEqual(point1, point2, {precision: 7})   // returns true
booleanEqual(point1, point2, {precision: 6})   // returns false !
```

This is incorrect. If two points are equal to a precision of 7, then by definition they must be equal to a precision of 6 as well. The problem here arises from the underlying implementation of `GeojsonEquality`, which [uses toFixed](https://github.com/geosquare/geojson-equality/blob/master/dist/geojson-equality.js#L67C20-L69) to round the coordinates. The problem is that this introduces rounding errors that then cause the equality check to fail:

```
const a = 23.0365925;
const b = 23.03659249995325;

a.toFixed(6) // returns "23.036593"
b.toFixed(6) // returns "23.036592" 

a.toFixed(7) // returns "23.0365925"
b.toFixed(7) // returns "23.0365925" 

a.toFixed(6) === b.toFixed(6) // returns false
a.toFixed(7) === b.toFixed(7) // returns true
```

I think the correct behaviour here is not to use `toFixed()` but to check that both values fall within the acceptable precision, e.g., like this:

```
Math.abs(a - b) < (10**-6) // returns true, as expected
Math.abs(a - b) < (10**-7) // returns true, as expected
```

Which likely needs to be fixed upstream in GeojsonEquality ([I've filed a bug there too](https://github.com/geosquare/geojson-equality/issues/13)), or that logic replaced entirely in turf.

My suggestion is to drop the dependency on `geojson-equality` and implement it afresh here, which will allow updating the implementation to use typescript - more than happy to make a PR if the maintainers here agree.

solarissmoke:
Please fill in this template.

A fresh implementation of `geojson-equality`, to fix a precision error and resolve #2530 . 

I have put this new class in `turf-helpers`, and ported across all the tests from the upstream module, plus adding a new one to cover the precision handling. 

Please let me know if I should move anything around.

### DIFF
--- a/packages/turf-helpers/lib/geojson-equality.ts
+++ b/packages/turf-helpers/lib/geojson-equality.ts
@@ -1,0 +1,198 @@
+import {
+  Feature,
+  LineString,
+  Position,
+  GeoJSON,
+  Point,
+  Polygon,
+  GeometryCollection,
+  FeatureCollection,
+  MultiLineString,
+  MultiPoint,
+  MultiPolygon,
+} from "geojson";
+import equal from "deep-equal";
+
+/**
+
+ * GeoJSON equality checking utility.
+ * Adapted from https://github.com/geosquare/geojson-equality
+ *
+ * @memberof helpers
+ * @type {Class}
+ */
+export class GeojsonEquality {
+  private precision: number;
+  private direction = false;
+  private compareProperties = true;
+
+  constructor(opts?: {
+    precision?: number;
+    direction?: boolean;
+    compareProperties?: boolean;
+  }) {
+    this.precision = 10 ** -(opts?.precision ?? 17);
+    this.direction = opts?.direction ?? false;
+    this.compareProperties = opts?.compareProperties ?? true;
+  }
+
+  compare(g1: GeoJSON, g2: GeoJSON): boolean {
+    if (g1.type !== g2.type) {
+      return false;
+    }
+
+    if (!sameLength(g1, g2)) {
+      return false;
+    }
+
+    switch (g1.type) {
+      case "Point":
+        return this.compareCoord(g1.coordinates, (g2 as Point).coordinates);
+      case "LineString":
+        return this.compareLine(g1.coordinates, (g2 as LineString).coordinates);
+      case "Polygon":
+        return this.comparePolygon(g1, g2 as Polygon);
+      case "GeometryCollection":
+        return this.compareGeometryCollection(g1, g2 as GeometryCollection);
+      case "Feature":
+        return this.compareFeature(g1, g2 as Feature);
+      case "FeatureCollection":
+        return this.compareFeatureCollection(g1, g2 as FeatureCollection);
+      default:
+        if (g1.type.startsWith("Multi")) {
+          const g1s = explode(g1);
+          const g2s = explode(
+            g2 as MultiLineString | MultiPoint | MultiPolygon
+          );
+          return g1s.every((g1part) =>
+            g2s.some((g2part) => this.compare(g1part as any, g2part as any))
+          );
+        }
+    }
+    return false;
+  }
+
+  private compareCoord(c1: Position, c2: Position) {
+    return (
+      c1.length === c2.length &&
+      c1.every((c, i) => Math.abs(c - c2[i]) < this.precision)
+    );
+  }
+
+  private compareLine(
+    path1: Position[],
+    path2: Position[],
+    ind = 0,
+    isPoly = false
+  ): boolean {
+    if (!sameLength(path1, path2)) {
+      return false;
+    }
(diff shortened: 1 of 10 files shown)


smallsaucepan:
Thanks @solarissmoke. Have created a discussion #2537 to see if there is a better place than turf-helpers to put geojson-equality.

Apologies didn't address this before your PR - things are moving quite quickly. Will aim to merge this as is though (to avoid making work for you and get the bug resolved) and once we've decided we can refactor if need be.

Jump in on the discussion if you can. Would be good to get your thoughts.

packages/turf-helpers/lib/geojson-equality.ts
```diff
@@ -0,0 +1,192 @@
...
+  GeoJSON,
+  Point,
+  Polygon,
+  GeometryCollection,
+  FeatureCollection,
+  MultiLineString,
+  MultiPoint,
+  MultiPolygon,
+} from "geojson";
+import equal from "deep-equal";
```
smallsaucepan, line 14: Did a test publish and when trying to use GeojsonEquality in a test script got the error "Error: Cannot find module 'deep-equal'". You might need to add deep-equal as a dependency in packages/turf-helpers/package.json

smallsaucepan (changes requested):
Thanks for putting this together @solarissmoke. As noted I get an error when using turf-helpers as a standalone package, which I think is because deep-equal isn't included as a dependency in turf-helpers/package.json

This is a bit of a gotcha because while everything builds and runs locally, when the package is published to NPM the dependency isn't included.

If you want to reproduce this you can follow the instructions in the wiki at https://github.com/Turfjs/turf/wiki/Contributing#deploy-to-a-local-node-registry to do a test publish on your local machine

packages/turf-helpers/lib/geojson-equality.ts
solarissmoke: Thanks for this, and apologies for not checking this first time around. I have added the dependency. In the process I have:

- Upgraded to deep-equal 2.2.3 (previous version was 1.1.1 which is 4 years old)
- Removed the deep-equal dependency from turf-line-overlap, since it depends on turf-helpers

I don't _think_ this should have any negative consequences for turf-line-overlap, but would be grateful if you can sanity-check that change.

smallsaucepan: Good work getting deep-equal updated. Though as turf-line-overlap still uses deep-equal directly we should keep it as an explicit dependency in package.json. If you can reinstate that we should be good to go 👍

solarissmoke: Done

smallsaucepan:
Hi @solarissmoke. Have just finished a bunch of build tool changes in turf master branch. Would you be able to bring your branch up to date with those changes, and then we can get this PR merged back to Turf?

Feel I kept you waiting so if you want a hand with any conflicts let me know. I think most changes are to package.json files, etc rather than index.ts. And we now build with pnpm instead of yarn.

solarissmoke:
@smallsaucepan rebased - please double-check that the changes to the lock file look sensible.

smallsaucepan (approved):
Had some issues running locally @solarissmoke. Approving changes to get them into the CI pipeline and a consistent environment. We can tweak from there if need be 👍

smallsaucepan:
Hi @solarissmoke. The CI build had an error. Would you mind taking a look? Looks like a linting problem:

https://github.com/Turfjs/turf/actions/runs/7283948261/job/19881689391#step:7:40

Is ```pnpm test``` running for you locally?

solarissmoke:
@smallsaucepan please try running the workflow again - both test and lint issues should now be fixed.

smallsaucepan (approved):

solarissmoke:
@smallsaucepan I've no idea what is now causing the build to fail (I can't reproduce it locally - `pnpm install --frozen-lockfile` runs successfully for me). Would be grateful if you have any ideas.

smallsaucepan:
The tests have passed in CI. For future reference you also need to run ```pnpm test``` - that will run the linter and the unit tests.

Will get this merged and hopefully out in the next alpha! Thanks for sticking with this @solarissmoke. Appreciate your help with it.

PR #2532, opened 2023-11-09
Upgrade (almost) all outdated project dependencies

smallsaucepan:
Upgrading (almost) all project dependencies to latest versions. Lots of changes, though largely limited to package.json and test.js files. Possibly something we should pencil in to do occasionally to avoid big shifts like this.

Moved toward specifying explicit semver ranges in package.json for each package, rather than using "*" and deferring to the monorepo or some other package. Better to be explicit and use tools to keep things in sync. Monorepolint can get us some of the way, though once we move to a modern yarn that will allow an even cleaner approach to keeping things consistent.

Some code changes required though only to utility and test scripts, and then only function name changes e.g. load.sync() is now loadJsonFileSync() in the latest version. Shouldn't be any changes to module behaviour.

Removing chroma-js as a turf-isobands devDependency as it doesn't seem to be used.

Please fill in this template.

n/a


mfedderly (approved):

smallsaucepan:
Thanks @mfedderly

PR #2543, opened 2023-11-25
Migrating remaining JS modules to Typescript (part 1)

smallsaucepan:
Migrating remaining Javascript modules to Typescript (the main implementation file at least). More discussion in #2538 

This PR covers:

turf-difference
turf-dissolve
turf-ellipse
turf-envelope
turf-explode
turf-flatten
turf-flip
turf-point-on-feature
turf-points-within-polygon
turf-polygon-smooth
turf-polygon-tangents
turf-rewind
turf-sample
turf-sector
turf-shortest-path
turf-simplify
turf-tag

Please fill in this template.


packages/turf-rewind/index.ts
```diff
@@ -21,25 +33,24 @@ import { featureCollection, isObject } from "@turf/helpers";
...
-  if (!geojson) throw new Error("<geojson> is required");
-  if (typeof reverse !== "boolean")
-    throw new Error("<reverse> must be a boolean");
-  if (typeof mutate !== "boolean")
-    throw new Error("<mutate> must be a boolean");
-
-  // prevent input mutation
-  if (mutate === false) geojson = clone(geojson);
+  // Prevent input mutation if requested.
+  if (!mutate) geojson = clone(geojson);
```
mfedderly, line 50: Possibly a little sketchy but we can skip this whole copy for Point and MultiPoint geometries. Or maybe just don't even accept them into the function in the first place.

packages/turf-sample/index.ts
```diff
@@ -20,18 +21,18 @@ import { featureCollection } from "@turf/helpers";
  *   currentFeature.properties['marker-color'] = '#000';
  * });
  */
-function sample(featurecollection, num) {
-  if (!featurecollection) throw new Error("featurecollection is required");
-  if (num === null || num === undefined) throw new Error("num is required");
-  if (typeof num !== "number") throw new Error("num must be a number");
```
mfedderly, line 26: JS error handling

packages/turf-shortest-path/index.ts
```diff
@@ -40,104 +49,103 @@ import { Graph, astar } from "./lib/javascript-astar";
  * //addToMap
  * var addToMap = [start, end, options.obstacles, path];
  */
-function shortestPath(start, end, options) {
+function shortestPath(
+  start: Coord,
+  end: Coord,
+  options: {
+    obstacles?: Polygon | Feature<Polygon> | FeatureCollection<Polygon>;
```
mfedderly, line 56: I'm kind of surprised that this doesn't also allow MultiPolygons from the JSDoc above where it accepts `Feature`, but this matches the previous types from the deleted .d.ts.

packages/turf-shortest-path/index.ts
```diff
@@ -40,104 +49,103 @@ import { Graph, astar } from "./lib/javascript-astar";
...
-  var minDistance = options.minDistance;
-  var obstacles = options.obstacles || featureCollection([]);
+  let obstacles = options.obstacles || featureCollection([]);
+  let resolution = options.resolution || 100;
 
   // validation
   if (!start) throw new Error("start is required");
   if (!end) throw new Error("end is required");
-  if ((resolution && !isNumber(resolution)) || resolution <= 0)
+  if (resolution && (!isNumber(resolution) || resolution <= 0))
```
mfedderly, line 70: NIce bugfix swapping the `()` around

packages/turf-shortest-path/lib/javascript-astar.d.ts
```diff
@@ -0,0 +1,36 @@
+// Copied from @types/javascript-astar@0.0.34
```
mfedderly, line 1: We can probably just update definitely-typed directly, that's usually pretty quick to do.
Otherwise we probably need to preserve whatever license blob is relevant from upstream.

packages/turf-simplify/lib/simplify.d.ts
```diff
@@ -0,0 +1,14 @@
+// Copied from https://github.com/mourner/simplify-js/blob/master/index.d.ts
```
mfedderly, line 1: Can we just depend on simplify-js directly and remove our local copy?
Otherwise we probably need to preserve license text here as well.

packages/turf-dissolve/index.ts
```diff
@@ -25,57 +26,70 @@ import polygonClipping from "polygon-clipping";
...
+  if (!propertyName) {
     return flatten(
       multiPolygon(
         polygonClipping.union.apply(
           null,
+          // List of polygons expressed as Position[][][] a.k.a. Geom[]
           fc.features.map(function (f) {
             return f.geometry.coordinates;
-          })
+          }) as [Geom]
```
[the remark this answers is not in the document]
smallsaucepan: Have updated it to ```[Geom, ...Geom[]]```. The interface to union() expects an initial Geom, then any extra Geoms are spread in as rest parameters (rather than an array).

packages/turf-dissolve/index.ts
```diff
@@ -25,57 +26,70 @@ import polygonClipping from "polygon-clipping";
...
         )
       )
-    );
+    ) as FeatureCollection<Polygon>;
```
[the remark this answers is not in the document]
smallsaucepan: You're correct. Have removed.

packages/turf-sample/index.ts
smallsaucepan: Reinstated!

packages/turf-rewind/index.ts
```diff
@@ -21,25 +33,24 @@ import { featureCollection, isObject } from "@turf/helpers";
...
+    mutate?: boolean;
+  } = {}
+): Geometry | Feature | FeatureCollection {
   // Optional parameters
   options = options || {};
   if (!isObject(options)) throw new Error("options is invalid");
-  var reverse = options.reverse || false;
-  var mutate = options.mutate || false;
+  const mutate = options.mutate || false;
+  const reverse = options.reverse || false;
```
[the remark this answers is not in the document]
smallsaucepan: Fixed. Searched for pattern in other files as well and fixed those too.

packages/turf-rewind/index.ts
```diff
@@ -21,25 +33,24 @@ import { featureCollection, isObject } from "@turf/helpers";
...
 
-  // validation
-  if (!geojson) throw new Error("<geojson> is required");
-  if (typeof reverse !== "boolean")
-    throw new Error("<reverse> must be a boolean");
-  if (typeof mutate !== "boolean")
-    throw new Error("<mutate> must be a boolean");
```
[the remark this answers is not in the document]
smallsaucepan: Agreed. Reinstated.

packages/turf-rewind/index.ts
smallsaucepan: Have added some logic to only clone if not Point or MultiPoint.

packages/turf-points-within-polygon/index.ts
```diff
@@ -37,32 +47,42 @@ import { geomEach, featureEach, coordEach } from "@turf/meta";
...
         });
       });
       if (contained) {
-        results.push(multiPoint(pointsWithin, point.properties || {}));
+        results.push(
+          multiPoint(pointsWithin, point.properties) as Feature<MultiPoint, P>
+        );
       }
-    } else {
-      throw new Error("Input geometry must be a Point or MultiPoint");
```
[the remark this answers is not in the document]
smallsaucepan: Agree - logic reinstated.

packages/turf-flip/index.ts
```diff
@@ -18,11 +18,16 @@ import clone from "@turf/clone";
...
+  geojson: T,
+  options?: {
+    mutate?: boolean;
+  }
+): T {
   // Optional parameters
   options = options || {};
   if (!isObject(options)) throw new Error("options is invalid");
-  var mutate = options.mutate;
+  const mutate = options.mutate || false;
```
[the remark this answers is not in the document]
smallsaucepan: Done. And replaced in other areas where misuse appeared.

packages/turf-shortest-path/index.ts
smallsaucepan: Yeah, I haven't pulled too much at loose threads like this 😅

packages/turf-flatten/index.ts
```diff
@@ -19,10 +30,24 @@ import { featureCollection } from "@turf/helpers";
...
+  geojson: Feature<T> | FeatureCollection<T> | T
+): FeatureCollection<LineString>;
+
+function flatten<T extends Polygon | MultiPolygon>(
+  geojson: Feature<T> | FeatureCollection<T> | T
+): FeatureCollection<Polygon>;
+
+function flatten(geojson: AllGeoJSON): FeatureCollection<any>;
+
+function flatten(geojson: AllGeoJSON): FeatureCollection {
```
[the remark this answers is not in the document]
smallsaucepan: Personally I find the nested ternaries pretty difficult to parse, so the overloaded function definitions are more readable. Possibly also what the casual consumer would prefer?

packages/turf-flip/index.ts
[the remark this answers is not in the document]
smallsaucepan: @mfedderly do we also want to do this with the options object? E.g. a pattern like so:

```typescript
options?: {
  thisProp?: string,
  thatProp?: boolean,
}

// options could be anything e.g. undefined, empty object, string, etc
options = options ?? {} // defined but user could have still passed a string ...
if (!isObject(options)) throw new Error("options not an object"); // def got an object by now
const thisProp = options.thisProp ?? "some default"
const thatProp = options.thatProp ?? false;
```

packages/turf-shortest-path/lib/javascript-astar.d.ts
smallsaucepan: If you're happy to, I'll add a license blob as a temporary measure so this PR can proceed. In the background will see what can be done to fix the types at Definitely Typed and tidy up down the track.

packages/turf-flip/index.ts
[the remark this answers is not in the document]
mfedderly: Yeah I was kind of thinking about this. Perhaps we want something like this everywhere and then pull args from DefaultFooOptions as we need it.

```
export interface FooOptions {
   ...
}
const DefaultFooOptions: FooOptions = {};
export default (options: Partial<FooOptions>) => {}
```

packages/turf-flatten/index.ts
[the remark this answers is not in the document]
mfedderly: Yeah I maybe buy that argument. The ternaries aren't too bad if they're formatted appropriately it winds up like a big if/else, but its maybe harder to know how to read them as a consumer if you're getting some inferred type instead of something concrete.

packages/turf-rewind/index.ts
mfedderly: The thing that's sketchy is that if you use rewind assuming that its then safe to mutate the return because the rewind cloned it, you suddenly have different behavior on Point and MultiPoint which is confusing. Sorry we maybe should just always clone.

packages/turf-shortest-path/lib/javascript-astar.d.ts
mfedderly: I'm not a lawyer but that seems fine. Also this being A* explains why the pathing looks weird, I bet Dijkstra's would be slower but give a more intuitive looking answer.

mfedderly (approved):
looks like the force push hid all the changes from the github review window for me, but I'm going to just approve to unblock since we're definitely working different hours at the moment. Please clean up the license stuff before merging though.

I think we've been using `/*!` to denote license comments
https://github.com/Turfjs/turf/blob/master/packages/turf-great-circle/lib/arc.js#L1

smallsaucepan:
Whoops. Sorry Matt. To summarise, agreed with pretty much everything you pointed out and will address with another commit shortly.

Only things I'm not sure I could address are some overloads and local lib typings. For overloads will leave as is as they seem more readable to the casual user. For local lib typings will put license in as a temporary measure and if there's something published (or that we can update easily) on Definitely Typed will tidy up in a subsequent PR.

Thanks for approving!

packages/turf-simplify/lib/simplify.d.ts
smallsaucepan: Included license text for now.

packages/turf-polygon-smooth/index.ts
```diff
@@ -18,45 +25,72 @@ import { coordEach, geomEach } from "@turf/meta";
...
+  inputPolys:
+    | FeatureCollection<MultiPolygon>
+    | Feature<MultiPolygon>
+    | MultiPolygon,
+  options?: {
+    iterations?: number;
+  }
+): FeatureCollection<MultiPolygon>;
+
+function polygonSmooth(
```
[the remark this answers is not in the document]
smallsaucepan: Realised I was being too restrictive with the types. The implementation is totally fine with a mixture of both Polygon and MultiPolygon. Have removed the overload and now the one signature accepts all permutations.

packages/turf-shortest-path/lib/javascript-astar.d.ts
smallsaucepan: Included the license text as a workaround for now.

packages/turf-polygon-tangents/index.ts
```diff
@@ -94,31 +101,27 @@ function polygonTangents(pt, polygon) {
...
 }
 
-function processPolygon(polygonCoords, ptCoords, eprev, enext, rtan, ltan) {
-  for (var i = 0; i < polygonCoords.length; i++) {
-    var currentCoords = polygonCoords[i];
-    var nextCoordPair = polygonCoords[i + 1];
+function processPolygon(
+  polygonCoords: Position[],
+  ptCoords: Position,
+  eprev: number,
```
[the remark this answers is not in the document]
smallsaucepan: Yep, that should be fine. enext looked like it was being reused between the main function and a sub function, though only as a local var in the sub function.

markcarroll:
@smallsaucepan can we get a version bump of the 7.0 alpha so we can test these out?

PR #2550, opened 2023-12-01
 Convert tests to Typescript, change require to import, and add named exports to modules

smallsaucepan:
Converting all bench.js and test.js tests over to Typescript. In most cases this involved only renaming and converting require() to import(). There are still a lot of type warnings highlighted in IDEs, though they don't affect running the tests or the main build. Fixing all these type errors at once would be a mammoth task, so better to update as we go.

Replaced require() with import() in almost all cases within the module codespace. There are a few requires() left in utility scripts which can be addressed in a subsequent PR.

Added named exports for all modules. Eventually(?) default exports should be a thing of the past so laying the groundwork for that.

Both the import() and named exports work is in preparation for looking at rejigging the ESM / CJS approach used by the project.

Please fill in this template.


twelch (approved):
Appreciate the progress

mfedderly (commented):
Peeked at the first 50 or so files and this approach looks awesome

PR #2551, opened 2023-12-07
Upgrade to yarn v4

smallsaucepan:
Turf was using a very old version of yarn (v1), with v4 being the most recently released. Upgrading to a newer version (apart from v1 eventually not being supported) we get access to some monorepo sugar - "workspace:^" instead of "v7.0.0-alpha.2" etc, and first class yarn workspaces commands. There might be some npm deployment related changes, though this has been around long enough they should be well documented. Tagging @mfedderly for a review regardless.

Please fill in this template.


smallsaucepan:
Drat. Looks like the node-setup github action can't enable corepack, which we need to install the newer yarn version in the CI environment. Going to move this PR in to draft status until that is possible or I can find another way to press on 😩

Dependent on PR https://github.com/actions/setup-node/pull/901

mfedderly:
For what its worth, I don't think we are tied to yarn for any particular reason (I moved us there because we didn't even have a node package lock at the time). I've mostly moved over to pnpm if that helps make this PR easier.

packages/turf-boolean-within/index.ts
```diff
@@ -138,20 +138,18 @@ function isMultiPointOnLine(multiPoint: MultiPoint, lineString: LineString) {
...
     }
-    if (!oneInside) {
-      isInside = booleanPointInPolygon(multiPoint.coordinates[i], polygon, {
-        ignoreBoundary: true,
-      });
-    }
+
+    isInside = booleanPointInPolygon(multiPoint.coordinates[i], polygon, {
+      ignoreBoundary: true,
+    });
```
mfedderly, line 152: This change is a no-op (I guess?) but its confusing as part of the yarn upgrade. It also might be a step away from a better version where we can skip the second booleanPointInPolygon? Not really sure what the original intent was of the oneInside boolean.

mfedderly (commented):
Overall lgtm, happy to re-review once it builds.

packages/turf-boolean-within/index.ts
smallsaucepan: Me either. Yeah, it basically evaluated to ```if (true) ...```

Was an issue that didn't seem to justify its own PR, though happy to separate it out.

smallsaucepan:
> For what its worth, I don't think we are tied to yarn for any particular reason (I moved us there because we didn't even have a node package lock at the time). I've mostly moved over to pnpm if that helps make this PR easier.

Well, I might see how this goes with pnpm instead and if that's smoother scrap this PR for one based around that tool. Thanks for clarifying.

smallsaucepan:
Retiring in favour of PR #2555

PR #2555, opened 2023-12-17
Migration to pnpm and improved exports strategy

### LINKED ISSUE #2260: Several modules not exported in index.d.ts
I'm using @turf/turf with version 6.5.0.
It seems that several modules are not exported in index.d.ts, such as : booleanConcave,booleanIntersects.

Although they are indeed exported in "@turf/turf/dist/es/index.js" and "@turf/turf/dist/js/index.js",thus they can be actually imported and used in production mode.Still,the IDE can not find them during development mode, which may result in confusion for developers.

Would you please add export statement "booleanConcave","booleanIntersects" in index.d.ts? Thanks a lot.

### LINKED ISSUE #2307: Support Typescript's moduleResolution: "node16"
This is a bug in at least `@turf/envelope@6.5.0` and potentially other packages.

Typescript has recently released a new `moduleResolution` option that when set to `node16` or `nodenext` will use new ESM resolution logic. This includes an understanding of the new "exports" package.json field.

Any module in this monorepo that currently uses the exports field (like @turf/envelope) must now specify the types in the export config as well, otherwise Typescript will complain that it can't find the types file if you turn on `moduleResolution: "node16"`.

E.g. Turn this: 
```json
{
  "main": "dist/js/index.js",
  "module": "dist/es/index.js",
  "exports": {
    "./package.json": "./package.json",
    ".": {
      "import": "./dist/es/index.js",
      "require": "./dist/js/index.js"
    }
  },
  "types": "index.d.ts",
}
```

Into this:
```json
{
  "main": "dist/js/index.js",
  "module": "dist/es/index.js",
  "exports": {
    "./package.json": "./package.json",
    ".": {
      "types": "./index.d.ts",
      "import": "./dist/es/index.js",
      "require": "./dist/js/index.js"
    }
  },
  "types": "index.d.ts",
}
```

Until then, using this package in Typescript with the new module resolution scheme results in the error `TS7016: Could not find a declaration file for module '@turf/envelope'.`

### LINKED ISSUE #2553: Could not find a declaration file for module '@turf..."

"@turf/helpers": "^6.5.0",
"@turf/simplify": "^6.5.0"

```js
import simplify from "@turf/simplify";
const turf = require("@turf/helpers");
```

when using an import statement, i get `Could not find a declaration file for module '@turf/simplify'. ` ts error.
It's the same with all other modules even @turf/turf.

Any ideas/tips?

smallsaucepan:
Another tranche of changes designed to modernise the build:

* Migrating repo from old yarn to modern pnpm
* Improved our exports strategy to better support both CJS and ESM
* Using named exports / imports within Turf (keeping default exports for now)
* Dropping support for ie11

NOTE couldn't get monorepolint working with pnpm, so it is currently disabled. MRL project shows signs of recent activity so will revisit with a more recent release in the near future.

Fixes #2553 and fixes #2260.
Also fixes #2307 though that was really remained open for discussion purposes.

Please fill in this template.


.github/workflows/turf.yml
```diff
@@ -18,13 +18,20 @@ jobs:
         node-version: [16.x, 18.x, 20.x]
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout
+        uses: actions/checkout@v3
```
favna, line 22:
```suggestion
        uses: actions/checkout@v4
```

favna (changes requested):
> Migrating repo from old yarn to modern pnpm

Any particular reason to not go from old yarn to modern yarn? As in, Yarn v4. To use one of the copypasta tags from my Discord server:

<details><summary>Click to expand, includes all the info and copypastable commands for switching to it</summary>

Yarn v4 is a new version of Yarn that we recommend switching to as Yarn v1 has long since been deprecated.

> "But I don't see any update on [source]?"

That is correct. Yarn v4 is installed through Yarn itself. You configure Yarn v4 on a per-project basis. How you installed Yarn globally is largely irrelevant to this (corepack, volta, something else). How to install Yarn v4 for your project? Simply write:
```sh
yarn set version berry
```
This will download the new Yarn v4 binary and put it in `.yarn/releases`, you should push this to your Git repository. It will also create a `.yarnrc.yml` file which configures the path which you should also commit.

Next you probably also want to run the following 2 commands:
```sh
yarn config set enableGlobalCache true
yarn config set nodeLinker node-modules
```
This will add to your `.yarnrc` file:
```yaml
enableGlobalCache: true
nodeLinker: node-modules
```
This ensures you have a more traditional experience with `node_modules` and a global cache.

The next step is to nuke your `node_modules` and `yarn.lock` and run `yarn install`

Then some final adjustments. Put this in you `.gitignore`:
```properties
# Yarn files
.yarn/install-state.gz
.yarn/build-state.yml
```
And anywhere in your scripts in package.json where you use `*` you should wrap it in extra `"`
For example:
```json
{
	"format": "prettier --write \"src/**/*.ts\""
}
```
Mind you this last thing is a good thing to add regardless of script runner/package bundler because it ensures the glob is performed by the library and not by your shell, which may differ when people develop on different operating systems.

In short, the command to set everything up you can run:
```sh
yarn set version berry && yarn config set enableGlobalCache true && yarn config set nodeLinker node-modules && echo "" >> .gitignore && echo "# Yarn files" >> .gitignore && echo ".yarn/install-state.gz" >> .gitignore && echo ".yarn/build-state.yml" >> .gitignore
```


</details>

Ultimately Yarn v4 is
- Faster than Yarn v1, pretty much on par with pnpm
- Is very configurable
- Documentation is phenomenal
- I don't know how it works with pnpm as I don't use it but keep in mind that with Yarn you need to run `yarn npm publish` as opposed to `npm publish` so Yarn resolves `workspace:^` dependencies before publishing since that's obviously not valid for the npm registry. I assume for pnpm you'd need `pnpm publish` as well.
- Still sticks to node_modules. pnpm with their pnp module has some weird things from time to time in specific project stacks, such as when module augmenting with TypeScript of transitive dependencies and pnpm doesn't link the transitive dependencies to node_modules.. yeah it's a bit of a PITA sometimes. Probably doesn't affect Turf but personally it has really annoyed me about pnpm.

FWIW I'm not really sure what your status is in relation to the Turf org but I imagine the chances for the PR to get merged are far far higher by sticking to Yarn.

.github/workflows/turf.yml
lindenwells: context: github actions use node and I think checkout v4 uses node 20

smallsaucepan:
Thanks @favna. Ha ha I did have a yarn 4 PR almost ready though couldn't get it to work with GitHub CI (something corepack related). About the same time another of the Turf team mentioned we weren't necessarily married to yarn. So to get a release out asap just went with pnpm instead.

Perhaps we will revisit yarn down the track. Things like constraints would be nice, though might also take a look at syncpack.

Noted your comments about releasing too. Will keep an eye on that as we go.

.github/workflows/turf.yml
smallsaucepan: Will give the newer version a whirl.

favna:
> GitHub CI (something corepack related)

I'm curious to know what. This is not something I ever experienced. I always just use setup-node action and then just run `yarn install --immutable` (equivalent of `yarn install --frozen-lockfile`). Are you perhaps misunderstanding how Yarn v4 is installed and configured?

TL;DR of the collapsed content above, you install Yarn v4 on a per-project basis and the binary gets put in the `.yarn/releases` directory which you're supposed to check into Git. Then the global Yarn, whichever version that may be is irrelevant, knows to look for the `.yarnrc.yml` file and the `yarnPath` config in there and use the binary as specified by the path.

This has the great advantage of ensuring that every developer on the project uses the same toolchain and same version, thus ensuring that there are no discrepancies on that part when trying to help your co-workers*. As an anecdote of sorts, in the past I've had colleagues working on Linux while most of us worked on MacOS and whenever the Linux colleagues ran `npm install` the `package-lock.json` would change in terms of URLs, versions resolved and whether `http` or `https` was used even though we were using the same versions of Node and Npm. Thanks OS differences... never again for me, please.

\* I also do Java development and for this reason, I also much prefer Gradle over Maven because it has the exact same advantage

Aanyywaayy, ultimately do as you please of course. It's your project at all and it sounds like you guys talked about it well enough so I won't butt in further.

smallsaucepan:
> Are you perhaps misunderstanding how Yarn v4 is installed and configured?

Oh, that is _entirely_ possible 😅 The v4 PR was #2551 if that's of interest, and the CI corepack problem is being tracked in an open issue in actions/setup-node (https://github.com/actions/setup-node/issues/531)

> Aanyywaayy, ultimately do as you please of course. It's your project at all and it sounds like you guys talked about it well enough so I won't butt in further.

No, this is all good intel and welcome any time. Even if we don't use something right away it's good to know what sort of things to look out for.

package.json
```diff
@@ -1,22 +1,18 @@
...
-    "postlint": "documentation lint packages/turf-*/index.js",
-    "prepare": "lerna run build && node ./scripts/add-import-extensions.js",
-    "pretest": "npm run lint",
-    "test": "lerna run test",
-    "posttest": "lerna run --scope @turf/turf last-checks",
+    "lintDISABLED:mrl": "mrl check",
+    "lint:escheck-cjs": "es-check es8 packages/*/dist/cjs/index.cjs packages/turf/turf.min.js",
+    "lint:escheck-esm": "es-check --module es8 packages/*/dist/esm/index.mjs",
+    "lint:escheck-web": "es-check es5 packages/turf/turf.min.js",
+    "lint:docs": "documentation lint packages/turf-*/index.js packages/turf-*/index.mjs",
```
mfedderly, line 12: Does this work with .ts modules?

packages/turf-bezier-spline/lib/spline.ts
```diff
@@ -194,3 +194,6 @@ function B(t: number) {
     (1 - t) * (1 - t) * (1 - t),
   ];
 }
+
+export { Spline, Point };
+export default Spline;
```
mfedderly, line 199: We might consider marking these all as @deprecated (in another PR) if we're intending to remove them eventually

mfedderly (approved):
This is a huge amount of work. Thank you!

package.json
smallsaucepan: You mean "documentation lint"? Good question. I'll check and we can include turf-*/index.ts as a seperate change.

PR #2557, opened 2023-12-19
Upgrade monorepolint and rework its config to match the new pnpm work

mfedderly:
Also required an upgrade to husky to get the pre-commit hooks to work
Seems like it also regenerated all the READMEs (script had to get fixed) from the jsdocs in the various packages


packages/turf-clusters-dbscan/index.ts
```diff
@@ -121,14 +120,16 @@ function clustersDbscan(
...
+    return (tree.search(bbox) as ReadonlyArray<IndexedPoint>).filter(
+      (neighbor) => {
+        const neighborIndex = neighbor.index;
+        const neighborPoint = points.features[neighborIndex];
+        const distanceInKm = distance(point, neighborPoint, {
+          units: "kilometers",
+        });
+        return distanceInKm <= maxDistance;
+      }
+    );
```
mfedderly, line 132: Had to make changes in this file to get the build to pass? Not sure what's up with that.

twelch (approved):

packages/turf-clusters-dbscan/index.ts
twelch: Using @types/rbush in future may help resolve this

smallsaucepan:
Awesome! Thanks @mfedderly

PR #2560, opened 2023-12-24
Update release instructions for pnpm

mfedderly:
These are somewhat untested but I'm hoping they're right 🤞

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -1,0 +1,44 @@
+name: prerelease
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  build:
+    strategy:
+      fail-fast: true
+      matrix:
+        node:
+          - 18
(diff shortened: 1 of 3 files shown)


smallsaucepan:
@mfedderly is there more we can do than running through a CI build (that looks to have passed)? Besides merging the next PR how can we verify this will do what we expect?

mfedderly:
> @mfedderly is there more we can do than running through a CI build (that looks to have passed)? Besides merging the next PR how can we verify this will do what we expect?

I copypasted the action's skeleton from the monorepolint repo and made minor tweaks to it. I tested the publish commands along the way with `--dry-run` variants so 🤞 that it works after we merge. There's [some tooling](https://github.com/nektos/act) that maybe lets you run the actions locally, but I've never used it before.

smallsaucepan (approved):

PR #2562, opened 2023-12-30
Fix issues with prerelease builds

mfedderly:
- Added a fetch-depth: 0 config that should give us the git history and correctly calculate the canary versions
- Don't run scripts as part of the lerna release step, they were already run by the github actions steps
- Fix issue with reference to the node 'util' package that was preventing the browser version of @turf/turf from working

I think that the `undefined` references from the previous attempt are being caused by the root package.json not having a version. I think that's fine though?

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -18,6 +18,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


twelch (approved):
I would think no version at root level is fine. Not sure if lerna would care if you did add one

PR #2565, opened 2024-01-10
Converting packages to type: module by default

smallsaucepan:
Given we use the ES module import syntax almost everywhere, and are exporting CJS and ESM versions anyway, we should make our modules "ESM first".

Had to disable a few type tests, though these are only in packages where we import older third party libraries.

Didn't bother replacing node __dirname references in tests as they seem to work fine within each package without.


.monorepolint.config.mjs
```diff
@@ -88,21 +88,29 @@ export default {
...
+          publishConfig: {
+            access: "public",
+          },
           // @turf/turf is commonly consumed through CDNs, moving this output file is a breaking change for anyone
           // who has a hardcoded reference to this specific file, instead of letting the CDN pick the path.
           // Example of a URL that will break: https://unpkg.com/@turf/turf/dist/turf.min.js
           // Example of a URL that will keep working: https://unpkg.com/@turf/turf
           browser: "turf.min.js",
-          files: ["dist", "index.d.ts", "turf.min.js"],
+          files: ["dist", "turf.min.js"],
```
mfedderly, line 104: module in this package specifically might interact weirdly with turf.min.js here because it should be a commonjs output.

.monorepolint.config.mjs
```diff
@@ -88,21 +88,29 @@ export default {
...
             "./package.json": "./package.json",
             ".": {
               import: {
-                types: "./dist/esm/index.d.mts",
+                types: "./dist/esm/index.d.ts",
                 default: "./dist/esm/index.mjs",
               },
               require: {
-                types: "./dist/cjs/index.d.ts",
+                types: "./dist/cjs/index.d.cts",
```
mfedderly, line 113: I was a little surprised to see `.cts` as an extension here, but if that's what tsup gives us that seems fine.

packages/turf-along/test.ts
```diff
@@ -2,11 +2,11 @@ import path from "path";
 import test from "tape";
 import { loadJsonFileSync } from "load-json-file";
 import { Units, featureCollection } from "@turf/helpers";
-import { along } from "./index";
+import { along } from "./index.js";
 import { Feature, LineString } from "geojson";
 
 const line: Feature<LineString> = loadJsonFileSync(
-  path.join(__dirname, "test", "fixtures", "dc-line.geojson")
+  path.join("test", "fixtures", "dc-line.geojson")
```
mfedderly, line 9: Removing `__dirname` in the tests means that you can only run these tests with the cwd at the root of the package. I'm not sure that's a practical concern that's worth dealing with the `import.meta` nonsense.

packages/turf-hex-grid/test.ts
```diff
@@ -55,7 +55,7 @@ test("hex-grid", (t) => {
 });
 
 test("grid tiles count", (t) => {
-  const bbox1 = require(directories.in + "bbox1.json").bbox;
```
mfedderly, line 58: 😮

packages/turf/test.ts
```diff
@@ -4,10 +4,10 @@ import { glob } from "glob";
 import test from "tape";
 import camelCase from "camelcase";
 import documentation from "documentation";
-import * as turf from "./index";
+import * as turf from "./index.js";
 
 // Helpers
-const directory = path.join(__dirname, "..");
+const directory = path.join("..");
```
mfedderly, line 10: This is maybe a little extra sketchy and might be worth using `import.meta` since its reaching into other packages from here.

packages/turf-clusters-dbscan/lib/rbush-export.ts
```diff
@@ -0,0 +1,7 @@
+// Get around problems with moduleResolution node16 and some older libraries.
+// Manifests as "This expression is not callable ... has no call signatures"
+// https://stackoverflow.com/a/74709714
+
+import lib from "rbush";
+
+export const rbush = lib as unknown as typeof lib.default;
```
mfedderly, line 7: I'm a little surprised to see this being a problem.
https://github.com/mourner/rbush/blob/master/index.js has `export default class RBush`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/rbush/index.d.ts has `export default class RBush<T>`

Are we running an old rbush types or js package perhaps?

packages/turf-collect/lib/rbush-export.ts
```diff
@@ -0,0 +1,7 @@
+// Get around problems with moduleResolution node16 and some older libraries.
+// Manifests as "This expression is not callable ... has no call signatures"
+// https://stackoverflow.com/a/74709714
+
+import lib from "rbush";
```
mfedderly, line 5: Ditto for rbush

packages/turf-kinks/lib/sweepline-intersections-export.ts
```diff
@@ -0,0 +1,7 @@
+// Get around problems with moduleResolution node16 and some older libraries.
+// Manifests as "This expression is not callable ... has no call signatures"
+// https://stackoverflow.com/a/74709714
```
mfedderly, line 3: Same thing here, it appears to have default function in both the types and the code. Maybe its a weird build step from the package with the code in it?

https://github.com/rowanwins/sweepline-intersections/blob/master/src/main.js
https://github.com/rowanwins/sweepline-intersections/blob/master/types.d.ts

packages/turf-line-intersect/lib/sweepline-intersections-export.ts
```diff
@@ -0,0 +1,7 @@
+// Get around problems with moduleResolution node16 and some older libraries.
+// Manifests as "This expression is not callable ... has no call signatures"
+// https://stackoverflow.com/a/74709714
+
+import lib from "sweepline-intersections";
```
mfedderly, line 5: Ditto on `sweepline-intersections`

packages/turf-clusters-dbscan/types.ts
```diff
@@ -1,5 +1,8 @@
+/*
```
mfedderly, line 1: For this package, what import was causing problems?

packages/turf-kinks/types.ts
```diff
@@ -1,5 +1,8 @@
+/*
```
mfedderly, line 1: For this package, what import was causing problems?

packages/turf-hex-grid/test.ts
smallsaucepan: Thought this would be ok. Is there a better way to read in a test fixture?

.monorepolint.config.mjs
smallsaucepan: This is easy enough to fix. Will change to use more traditional file extensions in both flavour dist/ subdirectories.

.monorepolint.config.mjs
mfedderly: But won't that be treated as ESM now because the package is set to `module`?

packages/turf-hex-grid/test.ts
mfedderly: Yeah I like the change, I was surprised that we were using require for this.

smallsaucepan: It was a different time :)

.monorepolint.config.mjs
smallsaucepan: If we're talking about turf.min.js only, rollup will output UMD (CJS+AMD hybrid) regardless of what module is set to.

If you're talking about the overall @turf package, the exports section in package.json should allow node to pick the appropriate dist/ flavour for whatever environment the caller is trying to import into.

mfedderly: My understanding is that a consumer of @turf/turf would now see `type: "module"` and infer that turf.min.js is ESM and not UMD because of the .js extension. Is that incorrect?

smallsaucepan: I believe that's incorrect. Someone using @turf/turf from a node project shouldn't have any idea turf.min.js even exists. Node will choose the file to import (dist/esm/index.js or dist/cjs/index.js) based on the import mechanism their project employs. Us fleshing out the exports field in package.json allows that to work more predictably.

.monorepolint.config.mjs
smallsaucepan: Actually, can't do this just at the moment. tsup will override the .cjs extension to .js, but not .d.cts to .d.ts Depends on https://github.com/egoist/tsup/issues/939

packages/turf-along/test.ts
smallsaucepan: Easy enough to remedy. Will restore these references, and add a import.meta based __dirname where required.

.monorepolint.config.mjs
mfedderly: ok sweet. thanks for the education here

packages/turf-along/test.ts
mfedderly: Its probably only really useful in the packages/turf tests that look into other packages, just to be extra careful there

packages/turf/test.ts
smallsaucepan: Have reintroduced this __dirname reference.

packages/turf-along/test.ts
smallsaucepan: Reintroduced all these __dirname entries too just to be on the safe side.

packages/turf-kinks/types.ts
smallsaucepan: ```import { kinks } from "./index.js";``` bombs with a complaint about "Property 'default' does not exist on type ..." of whatever index.ts is trying to import. I can get IDE typing + the build to work, or the tests, but (seemingly) not both.

Can try to see if I can get the test re-enabled. Or if you're happy with the changes to the PR so far, we can merge and add a couple of issues to revisit this, possibly with PRs to the upstream projects.

.monorepolint.config.mjs
smallsaucepan: Hey @mfedderly remember a few days ago when you thanked me for educating you? Yeah, you are entitled to a refund on that. I was wrong. Half wrong, anyway.

Using turf.min.js should still be unimpacted by whether the monorepo package is type: module or commonjs

What doesn't work perfectly is using ```type: module``` and dist/cjs/index.js (.js extension), even if the exports section in package.json says _that's_ the file to use if you're using require. From the looks of arethetypeswrong, it also needs to have a .cjs extension, or we go back to dropping a ```type: commonjs``` package.json file into dist/cjs/ during the build.

packages/turf-kinks/types.ts
smallsaucepan: @mfedderly managed to get the type tests enabled again. Turns out the `tsc ... types.ts` command doesn't refer to the tsconfig.json in the current directory. Setting --module and --moduleResolution to node16 on the command line (to match tsconfig.shared.json) fixed that problem. Can't find a way to avoid the import wrappers entirely, however building, testing and type tests are working.

packages/turf-clusters-dbscan/types.ts
smallsaucepan: Problem was cropping up with the `import RBush from "rbush"` line.

smallsaucepan:
Hi @mfedderly. Have done the best as I can with this for now.

A summary of any item left unanswered seems to be something odd in the build or packaging of a third party library. Happy to chase these down one by one with the authors as it would be good to remove any wrapper files long term.

For now though, this should be good enough for us to push ahead with the goal of developing in an ESM world, and packaging ESM+CJS.

packages/turf-clusters-dbscan/types.ts
twelch: It looks like you're need to use ts-ignore and lose the rbush types in turf-clusters-dbscan?

Following the rabbit hole of one of the links you shared @smallsaucepan I found this module `default-import`

import { defaultImport } from "default-import";
const RBush = await defaultImport(rbushDefault);

smallsaucepan: Thanks for finding that Tim. It does do the trick though it looks like there may be some flow on effects. To use the `const RBush = await defaultImport(rbushDefault);` approach we need to change our JS target to at least es2022 to support the top level await. More so the umd package we generate for CDN usage can't have that top level await either, so it would be a change to how turf.min.js was included directly in the browser i.e. https://rollupjs.org/configuration-options/#output-format

Btw where were you seeing errors that required a ts-ignore? Previous approach seemed to work without complaining in my VSCode.

The lib/wrapper approach seems ok. We can use defaultImport, though I think that pushes us forward too quickly and I'm not sure the implications.

twelch: No problem if can't do it.  The migration I'm doing targets es2022 so I didn't think about top-level await.  One idea, wonder if can wrap the dynamic import into a little IIFE like:

```
import { defaultImport } from "default-import";
import rbushDefault from "rbush";

const RBush = await (async () => {
  const RBush = await defaultImport(rbushDefault);
  return RBush
})()

class RBushIndex extends RBush<Feature> {
...
```

At that point you might as well keep your wrapper if it's all the same.

I guess it was a ts-expect-error that I saw, not ts-ignore.  Maybe what I saw was just the original code you'd already replaced.
image

twelch (approved):
This all matches my expectations, based on an ESM migration I'm doing in another repo, which I'm updating to use the latest Turf.  I look forward to testing this out when it lands.

smallsaucepan:
Thanks for approving @twelch. Let me clean up (roll back the defaultImport change) and then I'll merge. Appreciate your input too @mfedderly given your earlier comments.

packages/turf-along/test.ts
```diff
@@ -1,10 +1,13 @@
 import path from "path";
+import { fileURLToPath } from "url";
 import test from "tape";
 import { loadJsonFileSync } from "load-json-file";
 import { Units, featureCollection } from "@turf/helpers";
-import { along } from "./index";
+import { along } from "./index.js";
 import { Feature, LineString } from "geojson";
 
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
mfedderly, line 9: I found a thing in my travels that tsup will polyfill these sorts of things for us.
https://tsup.egoist.dev/#inject-cjs-and-esm-shims

mfedderly (commented):
My only other followups would be the `rbush` and `sweepline-intersections` import issues.

Thanks for doing all of this!

PR #2571, opened 2024-01-17
Remove references to geojsonhint and geojson-fixtures

mfedderly:
There were transitively bringing in a weird dependency to a specific git commit of gerhobbelt/nomnom and was [breaking CI](https://github.com/Turfjs/turf/pull/2566) for dependabot.

### DIFF
--- a/packages/turf-explode/test.ts
+++ b/packages/turf-explode/test.ts
@@ -1,7 +1,6 @@
 import fs from "fs";
 import path from "path";
 import tape from "tape";
-import { all as fixtures } from "geojson-fixtures";
 import { loadJsonFileSync } from "load-json-file";
 import { writeJsonFileSync } from "write-json-file";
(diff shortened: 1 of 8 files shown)


packages/turf-explode/test.ts
```diff
@@ -11,14 +10,7 @@ const directories = {
   out: path.join(__dirname, "test", "out") + path.sep,
 };
 
-// Save input fixtures
-if (process.env.REGEN) {
-  Object.keys(fixtures).forEach((name) => {
-    writeJsonFileSync(directories.in + name + ".json", fixtures[name]);
-  });
-}
```
mfedderly, line 19: It only referenced the geojson-fixtures package to regenerate the inputs into the tests, so we already have what we need here.

smallsaucepan (approved):
Seems pretty straight forward 👍

Issue #2567, opened 2024-01-11
along with negative distance


leiflinse-trivector:
While porting the along method to dart_turf, I noticed that if you pass a negative distance to it, it will not return the first point in a line.

It is possible to solve this by a special condition at the start of the method, just after `coords` has been obtained. If distance is negative, then return the first coordinate.

Dart port: https://github.com/dartclub/turf_dart/pull/153

smallsaucepan:
Hi @leiflinse-trivector. Thanks for raising this. Could you please read #2148 and let us know how that compares to your request? It sounds like we have two requests for different behaviour from a negative offset.

leiflinse-trivector:
I think either clamping or a documented "from end" behavior is fine. But as it is now, it doesn't neither forbid negative input nor define what happens.

My reasoning is that when you input a distance that is longer then the total length of the line string, then it currently returns the last point, so in analogy of that was that if you "undershoot" the start by giving a negative value, it would clamp to the start.

But I also can see that the count from end behavior can be useful in some cases.

smallsaucepan:
My take on it would be, if the user passes an along value longer than the length of the line, it's probably accidental, and returning the farthest point on the line seems reasonable. If the user passes a *negative* value it's probably intentional and they are signalling they're specifically expecting different behaviour.

My gut feel is we would fix this according to #2148, so the behaviour is defined and no error is thrown. Is that a behaviour you'd be happy to align dart_turf with?

PR #2573, opened 2024-01-20
Fix along with negative distance was undefined

Leffe108:
This changes so that if distance is negative, then it measures along the line in reverse. Before, it would use bearing and return a point before the start.

Also document what happens if abs(distance) is larger than the length of the line.

Related issue: https://github.com/Turfjs/turf/issues/2567

---

Please fill in this template.


magnusburton (approved):
Much needed change, thanks! It might be breaking for some who've come to expect this pattern however.

smallsaucepan:
Thanks for taking a look at this @Leffe108. Have made a request above if you could expand the tests to cover some more of the functionality that would be fantastic.

PR #2583, opened 2024-02-28
include original id in line-slice

andrewharvey:
Please fill in this template.

### DIFF
--- a/packages/turf-line-slice/index.js
+++ b/packages/turf-line-slice/index.js
@@ -54,7 +54,7 @@ function lineSlice(startPt, stopPt, line) {
     clipCoords.push(coords[i]);
   }
   clipCoords.push(ends[1].geometry.coordinates);
-  return linestring(clipCoords, line.properties);
+  return linestring(clipCoords, line.properties, { ...('id' in line && { id: line.id }) });
 }
 
 export { lineSlice };


mfedderly:
Are you still interested in getting this merged? I'm not sure I understand what the use case would be for this. Can you write up more details about why this has to be implemented inside the method instead of adding the `id` property to the returned feature?

Thanks!

andrewharvey:
> Are you still interested in getting this merged? I'm not sure I understand what the use case would be for this. Can you write up more details about why this has to be implemented inside the method instead of adding the id property to the returned feature?

I wasn't sure if this was a good change, so left it as a draft. At the moment the sliced line retains the original lines `properties` but not the `id` and I got caught out by that change.

The argument for retaining the `id` is slicing a line returns an exact clone of the original line but with a mutated geometry only.

The argument for dropping the `id` is that the sliced line is no longer the same feature, and shouldn't have the same `id`.

I think either way it's going to depend on the downstream use to determine if they want to retain the ID, allocate a new one or drop the ID, so regardless we should improve the documentation to make it clear what the user will get back.

I'm slightly leaning in favour of the first option, to retain the `id` so the side affects of this method are reduced, and it's clear the method is only doing one thing, slicing the geometry.

I'll update the change description.

PR #2600, opened 2024-05-02
Backport typescript fixes from #2555 to v6.5.0

RobinVdBroeck:
Please fill in this template.

----

Doing an upgrade to 7.0.0 might be a big endeavor for some users of this library, so this would allow them to use modern Typescript features (like node16 module resolution) or use a bundler like vite.

I've backported the changes done in #2555 to v6.5.0. I haven't backported the pnpm changes, since that seems to be out of scope for this backport.

I've bumped the version of typescript here to 4.7.2, since that is the first one to introduce the `node16` module resolution algorithm.

I've manually verified the changes by setting up a minimal vite project, which was previously incompatible with v6.5.0, and using my local build (by yarn link). I've also setup a small script (`const foo = require("@turf/union").default; console.log(foo != undefined`) to make sure that commonjs is not broken.

tsup requires that there is a `tsconfig.json` file in every package, so I've added a script `sync-tsconfigs` that writes the tsconfig.json to every package.


tsconfig.shared.json
```diff
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
-    "target": "es5",
-    "module": "commonjs",
+    "target": "es2017",
```
RobinVdBroeck, line 3: Do I need to bring down to es5? I've taken this over from the master branch, but I'm not sure about the implications.

smallsaucepan: This should be fine. Where we landed was only the CDN bundle (in @turf/turf) needs to be es5 so it can be imported directly into the browser.

tsup.config.ts
```diff
@@ -0,0 +1,30 @@
...
+  dts: true,
+  entry: ["index.?s"], // while we have a mix of TS and JS packages
+  minify: false,
+  skipNodeModulesBundle: true,
+  sourcemap: true,
+  target: "es2017",
+  tsconfig: "tsconfig.json",
+  keepNames: true,
+  // treeshake: true, causes "chunk.default" warning, breaks CJS exports?
+  cjsInterop: false, // putting this to true will break backwards compatability
```
RobinVdBroeck, line 15: Currently, when importing `require("@turf/foo")`, you will get back an object containing a property called `default`, which is the default export. If the property above to false, this will not be the case and the default export will be the actual export if there is only 1 export. However, this is a breaking change, so I had to but this to false.

.monorepolint.config.ts
```diff
@@ -163,63 +129,28 @@ module.exports = {
...
-            "test:tape": "node -r esm test.js",
-          },
-        },
-        includePackages: JS_TAPE_PACKAGES,
-      },
-      {
-        options: {
-          scripts: {
-            "test:tape": "ts-node -r esm test.js",
+            bench: "tsx bench.js",
```
RobinVdBroeck, line 149: Everything is handled by tsx now. I tried to get this to work with `-r esm`, but I got really weird errors.

tsconfig.shared.json
```diff
@@ -5,6 +5,6 @@
     "declaration": true,
     "esModuleInterop": true,
     "strict": true,
-    "moduleResolution": "node"
+    "moduleResolution": "bundler"
```
RobinVdBroeck, line 8: Since we bundle our code for the consumers of this library, "bundler" is what we need. It uses the module resolution algorithm of node16 without requiring the file extensions.

RobinVdBroeck:
Hey, this PR has been open for a while. Is there any way I can help accelerate this?

smallsaucepan:
Hi @RobinVdBroeck. Thanks for submitting the PR. Happy to work with you to get it merged.

I do agree with your there is value in backporting some fixes to the v6 branch as well. Please be aware though that all our time and energy at the moment is dedicated to getting v7 out.

To cover off an additional testing step, have you tried using arethetypeswrong to check the generated JS and typedefs?

RobinVdBroeck:
arethetypeswrong report no problems, both for the bundle in `@turf/turf` and the subpackages.

image
image

elvince:
Hi,

Can we expect a new release with thoses changes soon?
it will be really valuable to get typings working correctly in v6.5.0

Thanks,

packages/turf-along/tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
   "extends": "../../tsconfig.shared.json",
   "compilerOptions": {
-    "outDir": "dist/js"
+    "allowJs": true
```
smallsaucepan, line 4: @RobinVdBroeck are these changes the same for all tsconfig.json files? If so, can we push them into the tsconfig.shared.json file and simplify these to just extend that?

packages/turf-angle/test.js
```diff
@@ -3,7 +3,7 @@ const path = require("path");
 const glob = require("glob");
 const load = require("load-json-file");
 const write = require("write-json-file");
-const sector = require("@turf/sector");
+const sector = require("@turf/sector").default;
```
smallsaucepan, line 6: Do we know why this import out of all of them is the only one to need .default added?

RobinVdBroeck: I think it has something to do with the other packages being written in typescript, which was built using `tsc` directly instead of using `rollup`. (We need .default when requiring because of https://tsup.egoist.dev/#interop-with-commonjs why this is needed)

I think I can fix this by only enabling cjsInterop if it's a typescript project, but that that migth complicate the setup. Let me know if this is something we want and I can try to get something working this weekend.

smallsaucepan:
Hi @RobinVdBroeck. Revisiting this PR and noted for it to build with our current CI setup it would need to be pnpm based. Started looking at that and it's getting a bit "whack-a-mole"-ish. Every problem I solve leads to two more appearing.

Will persist a bit longer. In the meantime, what would be your thoughts if instead we ported the latest build system and packaging infrastructure (what we released v7 with) over to the support/6.x branch, leaving the actual Turf code as is?

In other words a 6.x release cut using contemporary v7 infrastructure and improved packaging, but with 6.5.0 functionality as the Javascript level.

I'm guessing you want to stick with 6.5.0 because you're not free to commit to whatever API changes came with v7?

RobinVdBroeck:
@smallsaucepan Could you please clarify what you need from me? I think you're asking me to open the following PRs, but I'm not entirely sure:

1. Use pnpm instead of yarn so it matches v7 (this is not currently in this PR).
2. Update lerna so it matches v7 (this is not currently in this PR).
3. Update TypeScript so we can use the correct bundle resolution.
4. Introduce tsx + TypeScript fix.

From a JavaScript point of view, nothing has changed in this PR. I haven't changed any source code.

I believe there's value in handling steps 1-3 in a separate PR to make this one easier to understand.

smallsaucepan:
Sorry I was vague @RobinVdBroeck. Not asking you to do anything or open new PRs at the moment. More asking for your thoughts on a possible alternate way to a typescript friendly 6.x release.

I was hoping to add on to your PR an upgrade to pnpm as well so we could build and publish the 6.x bundle with our current build script. This proved problematic, which lead me to wonder if we could approach from a totally different direction.

Instead of upgrading parts of the v6 plumbing as you've done in this PR, could we instead effectively take a copy of the v7 branch, and copy in the v6 business logic only, packaging that up instead.

I'm experimenting with that now, though would be interested in what you think of that approach?

smallsaucepan:
Ok, have managed to add to this PR so that it will build with our current github CI (minimal pnpm and lerna changes).

@RobinVdBroeck do you mind if I push those changes here?

RobinVdBroeck:
Yeah no problem, feel free to do that. I guess it comes down to the same thing.

smallsaucepan:
Will review the PR with the combined changes and see if it makes it through a build 🤞

tsconfig.shared.json
smallsaucepan: Swapped this back to "node" to minimise changes to 6.5.0. Was able to get it to build ok, though will see how we go with a mock publish.

smallsaucepan (approved):

mfedderly:
I have a couple things to think about for this PR:

1) Upgrading to 7.0.0 is hopefully not *that* big of a change, I migrated two large monorepos with lots of separate Turf consumers. The changes mostly fell into a few categories: using @types/geojson instead of @turf/helpers, using the named exports instead of the defaults, and some small breaking changes to the APIs of @turf/union and @turf/difference. Hopefully you aren't stuck on 6.x for too much longer.
2) I didn't look at this PR too closely, but I'd be worried about introducing breaking changes to packages unintentionally. So just be careful around things like the default exports.

smallsaucepan:
Thanks @mfedderly. Both good points.

1. We don't necessarily have to do this. I guess I'm guided by what little feedback we have i.e. someone raised an issue and it got two thumbs up.
2. Definitely. Goal at the moment will be to see if we can get it to build in CI. Then will do a side by side comparison of the 6.5.0 generated code and this.

PR #2602, opened 2024-05-04
Pin pnpm version using corepack

RobinVdBroeck:
Please fill in this template.

I had pnpm 9 (the latest version) installed globally, which resulted in changes to package-lock.json when running pnpm install. This pins the version of pnpm to pnpm v8 when using [corepack](https://nodejs.org/api/corepack.html). When not using corepack, it will still use the global installed version.

I've updated the documentation about this in CONTRIBUTING.md, let me know if I should also add this information in another place.


twelch:
Thanks for the PR @RobinVdBroeck , I noticed that my fresh pnpm version 9 also drastically changed the package-lock.  What do you think @smallsaucepan are you okay with locking the pnpm version, so we don't thrash the lockfile?

If so, this PR would need some cleanup with the doc update I just merged.

CONTRIBUTING.md
```diff
@@ -28,7 +28,7 @@ Once you've cloned a Turf repository, and have a terminal open and currenty in t
   - enable pnpm as a package manager. Requires Node 16+.  Alternatively just `npm install -g pnpm`.
 - `pnpm install`
   - install dependencies and build modules
-- `npm test`
+- `pnpm test`
```
twelch, line 31: @RobinVdBroeck I was under the impression that by running `corepack enable pnpm` as the Turf Contributing doc suggests, that users then won't need to use `pnpm` directly, they can just use `npm` and it will use the symlink it setup to pnpm.  That's what I found in my use.  In that case, can you revert your doc changes?

package.json
```diff
@@ -28,6 +28,7 @@
     ],
     "**/*": "prettier --write --ignore-unknown"
   },
+  "packageManager": "pnpm@8.15.8+sha512.d1a029e1a447ad90bc96cd58b0fad486d2993d531856396f7babf2d83eb1823bb83c5a3d0fc18f675b2d10321d49eb161fece36fe8134aa5823ecd215feed392",
```
twelch, line 31: Is the sha code essential here?  Is that to be extra sure? Why not just lock to the semver version `8.15.8`?

twelch (changes requested):
Thanks @RobinVdBroeck see my comments.  Appreciate any feedback.

CONTRIBUTING.md
twelch: Actually I see now with your addition of `packageManager` to the package.json, that may just replace the need to run `corepack enable pnpm`.

twelch: And I see that my assumption about use of `npm` running `pnpm` is wrong.

From the corepack docs:
> It's always possible to run a `npm` command within a project configured to be used with another package manager, since Corepack cannot intercept it.

twelch: I also tested and you do need to `corepack enable pnpm` in order for `pnpm install` to work (unless you already have it installed globally), and for it to pick up the version to use from package.json.  So CONTRIBUTING docs look correct as-is.  I'll resolve this now.

package.json
twelch: Okay I see now that's just what running `corepack use pnpm@latest` does, it populates `packageManagers` in package.json.  And will do next time when we want to update the version used by the project.

twelch (approved):
I'm ready to approve this PR.  It follows the process to pin version recommended by pnpm to prevent lockfile thrashing - https://pnpm.io/installation#using-corepack.  We can update to pnpm v9 in a follow-on.  Any final thoughts/reservations @smallsaucepan @mfedderly?  I'll merge soon otherwise.

PR #2609, opened 2024-05-22
Update changelog for 7.0 release

mfedderly:


twelch (approved):
Consider clarifying that Typescript users will need to migrate their code to use @types/geojson also.  That seems to be a common question/comment that comes up with people using the 7.x prereleases.

mfedderly:
> Consider clarifying that Typescript users will need to migrate their code to use @types/geojson also. That seems to be a common question/comment that comes up with people using the 7.x prereleases.

`- Move to @types/geojson package instead of declaring our own (#2158)` <- this is the first entry under breaking changes, is there more wording you want to add?

twelch:
Thanks @mfedderly , yea I read that first entry as describing turfs internal change.  How about adding "base geojson type definitions are no longer re-rexported by Turf.  Typescript users will need to import them from @types/geojson directly"

PR #2612, opened 2024-06-03
Upgrade to Lerna 8

mfedderly:
Can someone double check the changes from running `lerna repair`?

https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes


lerna.json
```diff
@@ -5,5 +5,6 @@
     "publish": {
       "registry": "https://registry.npmjs.org"
     }
-  }
+  },
+  "$schema": "node_modules/lerna/schemas/lerna-schema.json"
```
RobinVdBroeck, line 9: This is a nitpick, but normally a "$schema" is the first property in a JSON file. This makes it more discoverable when reading the file without any context.

RobinVdBroeck (approved):
LGTM

twelch (approved):
@mfedderly can/did you confirm that caching is still working for the 3 tasks as part of the top-level `prepare` and `test` scripts?

mfedderly:
Yes they appear to still cache as expected.

Issue #2552, opened 2023-12-07
Return type of `turf.buffer` is wrong since it can return `undefined`


martinhath:
[Right here](https://github.com/Turfjs/turf/blob/d83e5801ac9d8bba237501668fa1a86a15a377bd/packages/turf-buffer/index.d.ts#L34).  One example of where it returns `undefined` [is here](https://github.com/Turfjs/turf/blob/d83e5801ac9d8bba237501668fa1a86a15a377bd/packages/turf-buffer/index.js#L117).

PR #2613, opened 2024-06-03
Fix @turf/buffer types to allow for undefined returns when the inputs result in an invalid geometry

mfedderly:
Fixes https://github.com/Turfjs/turf/issues/2552


braeden:
Linked issue: https://github.com/Turfjs/turf/issues/2552

PR #2625, opened 2024-06-17
Move @turf/helpers code into individual files to help with tree shaking

mfedderly:
From looking at https://github.com/Turfjs/turf/pull/2623 I'm hoping we can avoid marking things as `@__PURE__` and just use multiple files instead.

I'm open to moving all of these files out of the root and into lib (or another directory?). Just went for a quick proof of concept for this initial commit though.


mfedderly:
This is unnecessary

PR #2628, opened 2024-06-21
Reduce number of dependencies

benmccann:
`deep-equal` is a terribly heavy library: https://npmgraph.js.org/?q=@turf/helpers

also updates to pnpm 9 and updates the versions tested on the CI accordingly

Screenshot from 2024-06-21 07-53-10


smallsaucepan:
Thanks for putting the work into this @benmccann. Removing deep-equal from turf-helpers is currently underway in #2623. Once that's merged would you be happy to pull those changes into here and then we can convert any remaining instance/s (turf-line-overlap only?)?

Also noticed changes to the versions of node we build Turf against and other workflow changes. All for keeping things modern. Could you please make mention in the description though the PR includes this?

packages/turf-helpers/package.json
```diff
@@ -69,7 +68,7 @@
     "typescript": "^5.2.2"
   },
   "dependencies": {
-    "deep-equal": "^2.2.3",
+    "dequal": "^2.0.3",
```
mfedderly, line 71: Did you consider `fast-deep-equal` at all? I hadn't heard of `dequal` but it seems plenty popular. I didn't compare final bundle sizes with both, but they're both much smaller than `deep-equals`

.github/workflows/turf.yml
```diff
@@ -15,21 +15,21 @@ jobs:
...
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
 
       - name: Install pnpm
-        uses: pnpm/action-setup@v2
+        uses: pnpm/action-setup@v4
         with:
-          version: 8
+          version: 9
```
mfedderly, line 32: I would strongly prefer that the build tooling changes go into another PR. Technically removing supported node versions is a break, but also 16 is EOL'd. We might consider just adding 22 and not removing 16 unless we want to 'sneak' that break into 7.x.

packages/turf-helpers/package.json
benmccann: they both seem pretty good. I don't really have much of a preference for one vs the other

.github/workflows/turf.yml
benmccann: yeah, makes sense. I just only have pnpm 9 installed on my machine and unfortunately they've dropped support for older node versions

if a maintainer would like to grab any changes you'd like to keep from this PR and send it, please feel free to do so

https://github.com/Turfjs/turf/pull/2623 is already accomplishing most of what I was aiming for, so I'd also be happy to close this one if that's easier

mfedderly:
Ok I opened #2631 to remove deep-equals from the other package that had included it. We need to merge #2623 first and then I'll fix #2631 up so that we remove the dependency everywhere. If you feel strongly that we should make the workflow changes lets continue that here. Otherwise feel free to close this.

benmccann:
thank you!!

PR #2635, opened 2024-06-26
Stop turf-mask mutating by default, make it an option

### LINKED ISSUE #2634: v6.5.0 introduced a breaking change to `turf.mask`: it now mutates the mask input
On version `v6.5.0`, `turf.mask(polygon, mask)` now mutates `mask` in place. In version `v6.3.0`, it did not.

Since this change is not documented, it can lead to some hard-to-diagnose bugs.

Here's a simple test to verify:
```js
const poly1 = turf.polygon([[[-73.2, 42],[-73.2, 42.2],[-73, 42.2],[-73, 42],[-73.2, 42]]]);
const poly2 = turf.polygon([[[-73.1, 42],[-73.1, 42.2],[-73.3, 42.2],[-73.3, 42],[-73.1, 42]]]);
const poly2Clone = turf.clone(poly2);
const _masked = turf.mask(poly1, poly2);
console.assert(turf.booleanEqual(poly2, poly2Clone)); // passes on v6.3.0, fails on v6.5.0
```

This is due to the changes in #2130, which rewrote `turf.mask()`.

The simplest solution is probably to add a `mutate` option and default it to `false` (like `turf-simplify` does). I'll make a PR with that approach.

farkmarnum:
Resolves #2634.

This PR reverts the behavior of `turf.mask()` back to how it was pre-`v6.5.0`, but allows users to opt-in to input mutation via a new option. Also adds tests for this behavior.

Performance difference is very small:
```
basic (mutate = false) x 294,373 ops/sec ±0.25% (95 runs sampled)
basic (mutate = true) x 307,397 ops/sec ±0.13% (97 runs sampled)
mask-outside (mutate = false) x 100,575 ops/sec ±0.55% (97 runs sampled)
mask-outside (mutate = true) x 103,180 ops/sec ±0.40% (94 runs sampled)
```


farkmarnum:
@smallsaucepan looks good! I also just pushed a few tweaks: adding the new arg to the `types.ts` test file and fixing a type error in `bench.ts`.

Should be good to go

packages/turf-mask/bench.ts
```diff
@@ -1,13 +1,21 @@
...
+import clone from "@turf/clone";
+
+// type guard to narrow the type of the fixtures
+const isPolygonFeature = (
+  feature: Feature<Polygon | MultiPolygon>
+): feature is Feature<Polygon> => feature.geometry.type === "Polygon";
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
+const SKIP = ["multi-polygon.geojson", "overlapping.geojson"];
```
mfedderly, line 17: Do these still need to be skipped?

mfedderly (approved):

packages/turf-mask/bench.ts
mfedderly: Ah they fail the 'Fixtures should have two features' check 🤦

PR #2644, opened 2024-07-07
Converted turf-mask to Typescript and added support for geometry parameters

smallsaucepan:
Converts turf-mask from Javascript to Typescript.

Fixes a bug where even though parameter types suggest passing a Geometry should be ok, the code doesn't actually deal with that case.

Please fill in this template. Use a meaningful title for the pull request. Include the name of the package modified.

- [x] bug fx / maintenance


packages/turf-mask/index.ts
```diff
@@ -0,0 +1,112 @@
...
+    // though they should be equivalent.
+    polygonOuters = createGeomFromPolygonClippingOutput(
+      polygonClipping.union(polygon.geometry.coordinates as Geom)
+    );
+  } else {
+    // Geometry
+    // Need to cast below as Position[][] isn't quite as strict as Geom, even
+    // though they should be equivalent.
+    polygonOuters = createGeomFromPolygonClippingOutput(
+      polygonClipping.union(polygon.coordinates as Geom)
```
twelch, line 49: Gotta love that type friction

twelch (approved):
This looks good, thank you @smallsaucepan.  Not sure if it would have been cleaner, but with a code conversion that moves a file *and* has additional code fixes, for scannability I like to use `git mv` so that you keep a finer-grained changelog.

twelch:
Oh, out of curiosity, any change in benchmark numbers?  Might be worth doing in these conversions

smallsaucepan:
It _is_ a bit slower. It's not type related though. One of our linting rules is [prefer-spread](https://eslint.org/docs/latest/rules/prefer-spread). Rolling back only that part of the code brought the performance back up.

Would your preference be to disable that rule, or cop the performance hit?

packages/turf-mask/index.ts
smallsaucepan: In a mind to submit a PR to DefinitelyTyped to tighten up the Position definition. That would hopefully remove the need for concessions like this.

twelch:
My instinct is to disable the prefer-spread rule for a performance-minded library like Turf.

packages/turf-mask/index.ts
twelch: I wonder can you use the new  `satisfies` operator here inline before casting to check it matches?

`satisfies Geom as Geom`

smallsaucepan:
Cool. Will disable (for this case only for now) and proceed with the merge.

packages/turf-mask/index.ts
```diff
@@ -0,0 +1,112 @@
...
+ * const addToMap = [masked]
+ */
+function mask<T extends Polygon | MultiPolygon>(
+  polygon: T | Feature<T> | FeatureCollection<T>,
+  mask?: Polygon | Feature<Polygon>
+): Feature<Polygon> {
+  // Define mask
+  const maskPolygon = createMask(mask);
+
+  var polygonOuters = null;
```
twelch, line 35: Consider a few more uses of `var` that could change to `let`

packages/turf-mask/index.ts
smallsaucepan: Gave this a shot and it doesn't seem to work. I think because Position is number[] (possibly with zero elements) and Geom is [number, number] (exactly two elements). Typescript isn't willing to make the call, ultimately barfing with
```
Type 'Position' is not assignable to type 'Pair'.
Target requires 2 element(s) but source may have fewer.
```

Issue #1987, opened 2020-12-18
turf.area() produces different results when redundant points are added


reyemtm:
If polygon A shares a partial boundary of a larger polygon B, but is otherwise completely inside polygon B, turf.area() will give a different result for the dissolved result of A and B than B alone. 

```
turf.area(turf.dissolve(A,B)) != turf.area(B)
```

This can be fixed by running turf.cleanCoords() on the dissolved polygon, at least in initial testing. I will put up an Observable notebook for testing soon.

reyemtm:
https://observablehq.com/@reyemtm/turfjs-issue-with-turf-area-and-redundant-points

reyemtm:
https://observablehq.com/@reyemtm/turfjs-issue-with-turf-area-and-redundant-points

stebogit:
@reyemtm I think a better (more correct) way of highlighting/documenting the issue would be to get rid of the dissolve step and just compare the results of `@turf/area` on the two features.
Would you please update your example?

reyemtm:
Ok, I updated it. I have not looked into the code as to why this might be happening. I am also not sure how much additional compute is needed for cleanCoords().

stebogit:
Ok there seems to be an issue, but it is not obvious what's going on.
For some reason only a specific point, the one at the top, is causing the two areas to be different, otherwise they coincide:

```js
const box = turf.polygon([[
  [-90,40], // NW
  [-89.9,40], // NE
  [-89.9,39.9], // SE
  [-90,39.9], // SW
  [-90,40],
]]);

const boxRedundantPoints = turf.polygon([[ 
  [-90,40], // NW
  [-89.95,40.0], // N <-- issue
  [-89.9,40], // NE
  [-89.9,39.95], // E
  [-89.9,39.9], // SE
  [-89.95,39.9], // S
  [-90,39.9], // SW
  [-90,39.95], // W
  [-90,40],
]]);

const points = box.geometry.coordinates[0].map(p => turf.point(p, {'marker-color': '#F00', 'marker-size': 'small'}))
const redundantPoints = boxRedundantPoints.geometry.coordinates[0].map(p => turf.point(p, {'marker-color': '#0F0'}))

const areaDiff = turf.area(box) - turf.area(boxRedundantPoints);
console.log(areaDiff);
```

So basically this is OK (**`areaDiff === 0`**):
[...]

PR #2665, opened 2024-07-30
turf-area: bug fix: off-by-one bug in area calculation

Abdumbo99:
- [Y] bug fix
- [Y] Have read and followed the steps for [preparing a pull request](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#preparing-a-pull-request).

First of all, I would like to thank the contributors behind Mapbox and all of its subprojects for their amazing work. I want to point out a small mistake that I found while re-writing some of the geo calculations for a library in GO, using Mapbox, amongst other projects, as a reference.
As per [**The GeoJSON Specification (RFC 7946)**](https://tools.ietf.org/html/rfc7946). A Polygon consists of Linear Rings and a linear ring ["... is a closed LineString with _four_ or more positions"](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6), the reason for the four positions is that ["The first and last positions are equivalent, and they MUST contain identical values; their representation SHOULD also be identical"](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6), i.e. to ensure the linear ring is closed the first and last position must be identical, deeming the first, or the last for that matter, position redundant and opens the door for some miscalculations similar to the one that I am attempting to fix in this PR.

The "-1" added in this PR solved two problems one is much simpler than the other but it is the main focus of this PR. The issues are:

1. The `if (coordsLength <= 2) return 0` in line 102 **does not comply with the four positions requirement** mentioned above, so a polygon with the invalid coordinates below would pass it, even though it is not a proper polygon (has less than four coordinates). `[[[46.78401786031469,24.672371366746333 ], 
[46.78401786031469, 24.69016675081427 ],
[46.78401786031469, 24.672371366746333]]]`

2. The second issue is in the calculations. The code cites the source ` * Reference:
  Robert. G. Chamberlain and William H. Duquette, "Some Algorithms for Polygons on a Sphere", JPL Publication 07-03, Jet propulsion Laboratory, Pasadena, CA, June 2007 http://trs-new.jpl.nasa.gov/dspace/handle/2014/40409` then uses it properly, almost, at least. From this source we can see the calculation of the area:
  "Assume we are dealing with a polygon with N vertices", after some work the equation becomes:
  
Screenshot from 2024-07-26 00-17-56

In this equation we have the same conditions as Geojson's polygons; N and 0 are identical. Thus the equation stops at N-1, i.e. the point just before the duplicate 0th point. 
In the code, however, this is not the case, as it stops at the Nth point instead. T**his extra iteration on the Nth point results in an inconsistent behavior.** For example, if we use this multipolygon:
`{"type":"MultiPolygon","coordinates":[[[[28.321755510202507,16.35627490376781],[20.424575867090823,1.7575215418945476],[48.254218513706036,20.42650462625916],[36.310934132380964,14.226760576846956],[28.321755510202507,16.35627490376781]]]]}
`
We will get the good and rather accurate approximated area of: 1139359621777.495, but if we rotate the coordinates upwards, while maintaining the order and right-hand rule we will get the following calculations: **139359621777.4954**, 1139359621777.495, 1139359621777.495, and 1139359621777.495. The coordinates of each rotation are respectively:
`[
  [ 20.424575867090823, 1.7575215418945476 ], [ 48.254218513706036, 20.42650462625916 ], [ 36.310934132380964, 14.226760576846956 ], [ 28.321755510202507, 16.35627490376781 ], [ 20.424575867090823, 1.7575215418945476 ]
]`

`[[ 48.254218513706036, 20.42650462625916 ], [ 36.310934132380964, 14.226760576846956 ], [ 28.321755510202507, 16.35627490376781 ], [ 20.424575867090823, 1.7575215418945476 ], [ 48.254218513706036, 20.42650462625916 ]]
`

`[[ 36.310934132380964, 14.226760576846956 ], [28.321755510202507, 16.35627490376781 ], [ 20.424575867090823, 1.7575215418945476 ], [ 48.254218513706036, 20.42650462625916 ], [ 36.310934132380964, 14.226760576846956 ]]`

`[[ 28.321755510202507, 16.35627490376781 ], [ 20.424575867090823, 1.7575215418945476 ], [ 48.254218513706036, 20.42650462625916 ], [36.310934132380964, 14.226760576846956 ], [ 28.321755510202507, 16.35627490376781 ]]`

Now while the difference in the calculations is trivial, the inconsistency is, definitely, problematic and is caused by the the extra iteration on the Nth point. 
I hope my points are clear, I am ready to clarify any unclear points if needed.
Since this concerns last, first and second point at some point in the calculations I am suspicious that this possibly Fixes (edit:NOT) #1987
NOTE: similar PR can be found here: https://github.com/mapbox/geojson-area/pull/13


Abdumbo99:
Just a quick note that this does not test against #1987 because it was not in response to it, but the issue could be resolved  as a side effect of this PR.

twelch:
Thank you @Abdumbo99 for submitting this pull request and the background.  To help get this accepted, would you be willing to add a unit test demonstrating the issue and that it's fixed?

Abdumbo99:
Thank you @twelch for the reply, I committed my tests.  If there are any style issues in the changes please let me know (I am not fluent in TS). 
The tests fail on ` const coordsLength = coords.length;`  but pass on `  const coordsLength = coords.length - 1;`
**Test `"turf-area-length-check"`** checks for the length properly, i.e. if it is shorter than 4 positions, now the code does not pass a polygon with 3 positions anymore. Now keep in mind that the constructor does indeed prohibit the creation of polygons that contain less than 4 positions, however, the `area` function double-checks for this criterion, and if the `coordinates` member variable was modified after creation (as done in the test) it can break the rule and the length check in the `area`  function will _pass polygons that it should not pass_. 
The old code would fail on this test, returning an actual calculation instead of returning 0. The new code passes this test.

**Test `"turf-area-rotation-consistency"`** checks for consistency in the calculation of the area if the polygon positions were to be rotated, in a way that does not violate the right-hand rule, i.e. counter-clockwise. 
The old code would fail on the very first iteration of the rotation, while it passes on the others, the logs message is included below:
```
# turf-area-rotation-consistency
not ok 3 should be strictly equal
  ---
    operator: equal
    expected: 1139359621777.4954
    actual:   1139359621777.495
```
The new code passes this test.

Abdumbo99:
Also as an extra note this PR _**does not resolve**_ https://github.com/Turfjs/turf/issues/1987

packages/turf-area/test.ts
```diff
@@ -21,6 +21,37 @@ const fixtures = fs.readdirSync(directories.in).map((filename) => {
...
+]);
+
+// compiler does not pass coords less than 4 point long, so have to re-assign
+invalidPoly.coordinates = [
+  [
+    [101.0, 0.0],
+    [101.0, 0.5],
+    [101.5, 0.5],
+  ],
+];
```
twelch, line 43: I would suggest creating an invalid Polygon geojson object directly if the helper is not allowing you to.

```
import { Polygon } from 'geojson'
const invalidPoly: Polygon = {
  type: "Polygon",
  coordinates: [
    [
      [101.0, 0.0],
      [101.0, 0.5],
      [101.5, 0.5],
    ],
  ]
};
```

packages/turf-area/test.ts
```diff
@@ -21,6 +21,37 @@ const fixtures = fs.readdirSync(directories.in).map((filename) => {
...
+
+const rotatingPoly = polygon([
+  [
+    [28.321755510202507, 16.35627490376781],
+    [20.424575867090823, 1.7575215418945476],
+    [48.254218513706036, 20.42650462625916],
+    [36.310934132380964, 14.226760576846956],
+    [28.321755510202507, 16.35627490376781],
+  ],
+]);
```
twelch, line 53: Not required, but if these test inputs are only used in a single test, consider moving the definitions down into the test they are used, so that the tests are more self-contained.  I think it makes the tests easier to read.

twelch (changes requested):
Nice work adding the tests @Abdumbo99, very thorough with the rotation.  I have one request to improve how you create the invalid polygon and then I'll be happy to approve and merge.  Let me know if you run into any issue.

Abdumbo99:
Alright, I did as you suggested, please let me know if there are any other issues.

twelch:
@Abdumbo99 it looks like your area function fix is causing one of the test outputs in turf-isobands to change and the continuous integration testing is now failing.  If you run `pnpm test` at the top-level of the repo you should see it.

I checked and turf-isobands uses the area function to sort its resulting isoband polygons by area, so the sort order for the `matrix2` test output specifically has changed, I'm sure for the better.

The solution is to update the turf-isobands test fixtures to the new correct output in your PR.  Would you be willing to give it a try?  Commands are as follows.  I verified it worked for me.

```bash
cd packages/turf-isobands
pnpm test                             // verify matrix2 test fails
REGEN=true pnpm test        // regenerate the 4 test fixtures for turf-isobands package in test/out directory
pnpm test                             // verify all tests in package now pass
cd ../..
pnpm test                             // verify all tests in all packages now pass
```

Once finished, you can add the updated fixtures to your Pull Request.  Ping me and I will rerun the CI testing, and hopefully we'll be good to go.

Abdumbo99:
@twelch,  Should be good to go now I guess?

twelch (approved):
Thank you for seeing this through @Abdumbo99 .

twelch:
matrix2 is the only fixture that had a slightly different order on 2 very small polygons.  The other fixtures appear to be the same, only the json formatting changed, adding a lot of additional lines

Abdumbo99:
Thank you for the help @twelch . Happy to know nothing major changed!

PR #2668, opened 2024-08-05
Upgrade dependencies

mfedderly:
Upgrade some dependencies that dependabot seems unable to manage automatically. Lots of pnpm-lock surgery to get targeted fixed, but some package.json upgrades as well.


twelch (approved):
@mfedderly I'm guessing the lockfile changes you made manually will stay in as long as someone doesn't rebuild it from scratch?

mfedderly:
> @mfedderly I'm guessing the lockfile changes you made manually will stay in as long as someone doesn't rebuild it from scratch?

Yeah or we introduce something that has an explicit dependency on something older? Its relatively stable though.

PR #2671, opened 2024-08-05
Update PR/develop action to cache pnpm files

mfedderly:
Comments in-line, but hoping this is part of increasing the perf of our CI builds.


.github/workflows/turf.yml
```diff
@@ -22,14 +22,14 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
+        cache: "pnpm"
```
mfedderly, line 26: ^ there's a new version out, so I grabbed that, and you can tell it that we're using pnpm for it to cache things for us automatically. I'm hoping this speeds up the `pnpm install` part of our CI on subsequent runs.

.github/workflows/turf.yml
```diff
@@ -22,14 +22,14 @@ jobs:
...
         with:
           node-version: ${{ matrix.node-version }}
 
       - name: Install pnpm
-        uses: pnpm/action-setup@v2
+        uses: pnpm/action-setup@v4
```
mfedderly, line 31: Another new version. v4 will throw an error if there is a mismatch between the action runner pnpm and the one defined in package.json's `packageManager` field. Luckily `version` below is optional and it will use the one in package.json by default 🥳

twelch (approved):
Good incremental improvement. Bummer the speedup isn't more.

PR #2673, opened 2024-08-06
Added support/6.x branch to CI config

smallsaucepan:
Add support/6.x branch to CI config so related activity prompts a build and test cycle.

Not directly related to any release, just housekeeping.

### DIFF
--- a/.github/workflows/turf.yml
+++ b/.github/workflows/turf.yml
@@ -2,9 +2,13 @@ name: CI build
 
 on:
   push:
-    branches: [master]
+    branches:
+      - master
+      - support/6.x
   pull_request:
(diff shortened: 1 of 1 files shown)


smallsaucepan:
@mfedderly does this look ok? Part of getting #2600 out.

mfedderly (approved):

mfedderly:
Before we do a release from support/6.x we should make sure that the publish part of the release knows to put a `--dist-tag`  on there, so we don't flip flop between 6.x and 7.x sharing the default tag in npm.

smallsaucepan:
Will definitely talk through the release process with you first @mfedderly.

Issue #2617, opened 2024-06-12
Cannot find module 'geojson' or its corresponding type declarations.


BenJackGill:
Using `@turf/helpers` version 7.0.0 and I am seeing this TypeScript error during deployment:

```
../../node_modules/.pnpm/@turf+helpers@7.0.0/node_modules/@turf/helpers/dist/esm/index.d.ts:1:205 - error TS2307: Cannot find module 'geojson' or its corresponding type declarations.

1 import { GeoJSON, Feature, Point, Position, LineString, MultiLineString, Polygon, MultiPolygon, FeatureCollection, Geometry, GeometryCollection, GeometryObject, GeoJsonProperties, BBox, MultiPoint } from 'geojson';
                                                                                                                                                                                                              ~~~~~~~~~


Found 1 error in ../../node_modules/.pnpm/@turf+helpers@7.0.0/node_modules/@turf/helpers/dist/esm/index.d.ts:1
```

BenJackGill:
Looking inside `node_modules` it seems that the `package.json` file for `@turf/helpers` is missing `geojson`. It is not installed under `devDependencies` or `dependencies`.

**Workaround:** I can overcome this when installing all of turf using `pnpn add @turf/turf` instead of just the individual package `pnpn add @turf/helpers` package. But I would prefer to only use individual packages so I still believe this needs fixing.

smallsaucepan:
That's no good. Will take a look at this as a priority @BenJackGill. Thanks for letting us know.

smallsaucepan:
@BenJackGill could you please share a bit more about your dev process and deployment environment? Would like to reproduce this locally so I can confirm any change fixes it.

BenJackGill:
I have a monorepo.

Internal packages of the monorepo are [just-in-time packages](https://turbo.build/repo/docs/core-concepts/internal-packages#just-in-time-packages) that directly export the typescript. And then I also have apps which consume those internal packages.

I have Turf installed in both the internal packages and the apps.

The app that is throwing the error is for Firebase Functions, which happens to have Turf installed inside it.

It's late here but I will try to get a minimal reproduction up for you later if that would help.

BenJackGill:
Here is the monorepo reproduction: https://stackblitz.com/edit/vitejs-vite-pbqflx?file=packages%2Fschema-geo-database%2Fpackage.json

There is an internal package named `@seoturbo/schema-geo-database` that is used within the app `functions-admin`.

[...]

PR #2676, opened 2024-08-06
Add explicit geojson types dependency

mfedderly:
Fixes https://github.com/Turfjs/turf/issues/2617

I struggled a bit trying to figure out what type of dependency this actually represents. I considered defining it as a peer dependency, but in the end it seems like several other packages declare it as a normal dependency (including `maplibre-gl`).

PR done in two commits, the first one is the manual changes and the second is just running monorepolint and pnpm to get everything in sync with the new spec.

### DIFF
--- a/.monorepolint.config.mjs
+++ b/.monorepolint.config.mjs
@@ -250,5 +250,14 @@ export default {
       },
       includePackages: [MAIN_PACKAGE],
     }),
+
+    requireDependency({
+      options: {
+        dependencies: {
+          "@types/geojson": "^7946.0.0",
+        },
+      },
+      includePackages: [MAIN_PACKAGE, ...TS_PACKAGES, ...JS_PACKAGES],
+    }),
   ],
 };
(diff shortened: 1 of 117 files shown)


twelch (approved):

smallsaucepan:
@mfedderly did we intentionally go from 7946.0.8 back to 7946.0.0 in package.json?

mfedderly:
@smallsaucepan yes. Internally we're still using the pnpm-lock version of 7946.0.14 within turf itself. Is there some incompatibility across the @types/geojson that I should know about (between 7946.0.0 and .14)?

twelch:
@mfedderly I don't quite get it, why not use the latest?  If we want to communicate that a lower version is supported, is peerDependencies the way to do that?

```

**Comment by smallsaucepan** [comment]
@mfedderly yes, looks like there were incompatible changes between .0 and .14. They haven't done us any favours with their versioning scheme.

In Feature interface the geometry and properties fields both became more restrictive:

```diff
147c146
<     geometry: G | null;
---
>     geometry: G;

156c155
<     properties: P | null;
---
>     properties: P;
```

mfedderly:
@smallsaucepan oy ok. Updated to .10 minimum version and explained some reasoning https://github.com/Turfjs/turf/pull/2688

PR #2677, opened 2024-08-06
Rework prettier setup

mfedderly:
Since I was already here I started by upgrading husky prettier and lint-staged.

It turns out that we [intentionally dialed down the prettier aggressiveness](https://github.com/Turfjs/turf/pull/2519/files#diff-b640b344ee7f3f03d2a443795a5d0708ef50e2e6e34214109ab2aad13ad6ba98R13-R19) because the formatting of test fixtures looked pretty ugly with the existing prettier settings. I think that if we increase the target printWidth for prettier, we can more confidently re-enable it for all of the test fixtures.

I think that even imperfect formatting of the fixtures is better than inconsistent formatting because of disabling it.

Its probably easiest to review this commit by looking at the first one to understand the changes, and the second commit for getting a sense of how the changes impacted the formatting across the codebase.


packages/turf-ellipse/test/out/rotation-degrees.json
```diff
@@ -58,9 +54,7 @@
     {
       "type": "Feature",
       "bbox": [
-        -74.11421824800942,
-        40.64244465235168,
-        -73.8809366917385,
```
mfedderly, line 63: we could maybe use printWidth of `90` for geojson and json files to avoid this awkward wrapping

mfedderly: But maybe its just fine to leave it at 80, it usually wraps correctly and stuff like bigMatrix.json seems fine.

twelch: Yeah this version with 80 looks better.  I'm open to it going higher for geojson and json as you said.

mfedderly: I updated this with 90, if you're happy I'll plan on merging this one and closing #2670

twelch: It looks fine to me.  Any further optimization can happen as follow-on

smallsaucepan:
Personally would rather see code stay where it is (80 right?). Changing that is going to lead to unnecessary churn in pretty much every file.

For geojson, as we can set a different width, how about we experiment repository wide with a couple of different ones? For example, if 90 results in the majority of our single line [1, 2, 3, 2, 1] arrays wrapping, try a larger value until it's just a handful.

mfedderly:
@smallsaucepan this PR does exactly what you suggest, everything stays at 80 except .json and .geojson which are closer to 90 (printWidth is not a hard cap, just a suggestion). I think that 90 seems to strike a pretty nice balance for the geojson not wrapping things too aggressively.

mfedderly:
Going to merge this now. If we have outstanding PRs with .geojson or .json changes we should make sure to merge in `master` to get a fresh build and make sure we don't merge a prettier break.

Issue #1176, opened 2017-12-12
Custom earth radius


dhivehi:
I think it would be better to be able to set earth's radius as a global parameter in turf. Because many people like me would be using google earth's radius which is also accepted by many professional organaisations. As for now, for every update, i download turf js and edit radius and use.

stebogit:
Sorry, just realized you said set and not get... 😅

dhivehi:
I want that feature mostly in distance calculation modules. But would be better if can be used in all modules.
`turf.pointToLineDistance(pt, line, {units: 'meters', earthRadius: 6378137});`
Adding support for extra parameter "earthRadius" would be great i guess.

dhivehi:
Attached screenshot shows what troubles me for every turf update.
![capture](https://user-images.githubusercontent.com/18715293/33921608-c3dcb4aa-dfe6-11e7-83a4-a31fede47994.PNG)
(Web application made with openlayers & TURF)

smallsaucepan:
Perhaps reconsider using options to override the radius. That might work ok for using distance directly. However any other modules that _in turn_ use distance would need to also accept earthRadius as an option so everything matches up. For example, the midpoint function.

It may be better to pass the new radius from the inside out globally, rather than outside in on a case by case basis. Something like having @turf/helpers check for an environment variable override which applies the desired radius globally (at least within turf).

twelch:
I agree with your approach @smallsaucepan if it can work. I think you brought it up in a discussion.

What does this mechanism look like? Needs to work in browser and node/other runtimes.

It could be a global variable that helpers looks for on init?  Falling back to a default value of mean radius.
```
export const earthRadius = TURF_EARTH_RADIUS || 6371008.8;
```
How the variable gets set in user space code would be up to the user.  Simple browser app it can be set in the base script, or top of a module.  If user bundles their app with Vite, they can pass in the value via an environment variable at build time and set the variable - https://vitejs.dev/guide/env-and-mode


https://vitejs.dev/guide/env-and-mode

PR #2683, opened 2024-08-07
Clarify behavior of some existing turf modules

### LINKED ISSUE #2638: turf V7.0.0 squareGrid result is not square，It can be judged solely by the naked eye.

- [ ] Verify this issue hasn't already been reported, or resolved in the latest alpha pre-release.

### LINKED ISSUE #2643: Calculate area with higher accuracy
I use the latest version of Turf to calculate the area of ​​a geojson polygon.
The area result from Turf is: 22.13 m2 and Geographic Lib (https://geographiclib.sourceforge.io/) 22.06 m2. I also checked on QGIS the area result is 22.03 m2

So Geographic Lib has higher accuracy than Turf Lib. Can you improve this issue?

Thank you so much.

Geojson data source: 
`{"type":"Polygon","coordinates":[[[106.66416522963165,14.579907369919486],[106.6641609900968,14.579919026608949],[106.66415731659645,14.579932022843553],[106.66414116820758,14.579927094033418],[106.66413907381092,14.579900138182057],[106.66411935271576,14.579868331778414],[106.66412579474931,14.5798490375216],[106.66414456560182,14.579868433595195],[106.66415768266069,14.579875838613062],[106.66416266097694,14.579878985318168],[106.66416175432456,14.579893036877686],[106.66416522963165,14.579907369919486]]]}`

twelch:
- [x] doc update

Resolves #2638 (squareGrid/rectangleGrid/triangleGrid/pointGrid)
Resolves #2643 (area, helpers/earthRadius)

In addition, to clarifying the behavior of some modules, it takes a shot at replacing stale lists of units in JSDocs, with a link to another page with the full list of Turf units accepted.

Right now I have it link directly to the turf/helpers source file, which isn't optimal, so I'm open to feedback.  The link should be durable.  One idea is to create a Units page in turf-www that provides a maintained full list of units.  I think you may have originally had this idea @smallsaucepan, thoughts?


packages/turf-helpers/index.ts
```diff
@@ -57,15 +57,17 @@ export type AllGeoJSON =
  */
 
 /**
- * Earth Radius used with the Harvesine formula and approximates using a spherical (non-ellipsoid) Earth.
- *
+ * The {@link https://en.wikipedia.org/wiki/Earth_radius#Arithmetic_mean_radius arithmetic mean radius} of the Earth.
+ * Used by Turf modules that model the Earth as a sphere. This radii value
+ * was chosen over others because it is {@link https://rosettacode.org/wiki/Haversine_formula#:~:text=This%20value%20is%20recommended recommended for use} with the Haversine formula,
+ * used by turf/distance.
```
twelch, line 63: I think this is useful background.  The intention is to follow this with #1176 and allow users to specify a custom earth radius in modules that support it (area, distance, etc.) similar to OpenLayers.  This will just become the default earthRadius.

smallsaucepan:
There are a handful of miscellaneous classes defined in turf e..g Units, QuadratAnalysisResult. I'll take a look to see if there's a way to get documentation.js to generate a shared page for those automatically. Otherwise, yeah we might just have to add a manually updated page to the website and link to that. Will get back to you.

twelch:
Not sure if it's overkill, but @smallsaucepan I've started playing a little bit with a tool for querying the abstract syntax tree of a Typescript file.  Try to access the string literals for the type declaration.  Haven't figured it all the way out, but potential.

```
import fs from "fs-extra";
import { ast, query } from '@phenomnomnominal/tsquery';

(async () => {
    try {
      const helpersTs = fs.readFileSync("packages/turf-helpers/index.ts", "utf8").toString()
      const theAst = ast(helpersTs);
      console.log(theAst.statements);
      const tNodes = query(theAst, 'TypeAliasDeclaration');
      console.log(tNodes.length); // 7
    } catch (e) {
      console.warn(e)
    }
})();

```

smallsaucepan:
Thanks @twelch that's actually a handy debug tool.

packages/turf-triangle-grid/index.ts
```diff
@@ -10,13 +10,13 @@ import { intersect } from "@turf/intersect";
...
+ * Creates a grid of triangular polygons.
  *
  * @name triangleGrid
- * @param {Array<number>} bbox extent in [minX, minY, maxX, maxY] order
- * @param {number} cellSide dimension of each cell
+ * @param {Array<number>} bbox extent of grid in [minX, minY, maxX, maxY] order
+ * @param {number} cellSide dimension of each grid cell.  Two triangles are created in each cell.
  * @param {Object} [options={}] Optional parameters
- * @param {string} [options.units='kilometers'] used in calculating cellSide, can be degrees, radians, miles, or kilometers
+ * @param {string} [options.units='kilometers'] used in calculating cellSide.  Supports all valid Turf {@link https://github.com/Turfjs/turf/blob/master/packages/turf-helpers/index.ts Units}
```
smallsaucepan, line 19: {Units} here too.

packages/turf-helpers/index.ts
```diff
@@ -57,15 +57,17 @@ export type AllGeoJSON =
...
  * @memberof helpers
  * @type {number}
  */
 export const earthRadius = 6371008.8;
 
 /**
- * Unit of measurement factors using a spherical (non-ellipsoid) earth radius.
+ * Unit of measurement factors based on equatorial earthRadius.
```
smallsaucepan, line 70: Should this be "arithmetic mean radius"?

packages/turf-point-grid/index.ts
```diff
@@ -12,13 +12,13 @@ import { distance } from "@turf/distance";
 import { point, featureCollection, Units } from "@turf/helpers";
 
 /**
- * Creates a {@link Point} grid from a bounding box, {@link FeatureCollection} or {@link Feature}.
+ * Creates a grid of points
  *
  * @name pointGrid
- * @param {Array<number>} bbox extent in [minX, minY, maxX, maxY] order
- * @param {number} cellSide the distance between points, in units
+ * @param {Array<number>} bbox extent of grid in [minX, minY, maxX, maxY] order
```
smallsaucepan, line 18: Could make type here {BBox} same as function definition.

smallsaucepan (changes requested):

packages/turf-rectangle-grid/index.ts
```diff
@@ -14,15 +14,14 @@ import {
   Units,
 } from "@turf/helpers";
 /**
- * Creates a grid of rectangles from a bounding box, {@link Feature} or {@link FeatureCollection}.
+ * Creates a grid of rectangular polygon cells of equal width and height in degrees.
```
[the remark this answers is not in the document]
twelch: My wording was confusing I see.  No the width and height won't be equal for rectangles.  And yes degrees is significant.  I've reworked this to convey better

twelch:
@smallsaucepan thanks for your feedback.  I think this PR is ready. I've cleaned up the JSDocs further.  And something new, as a stopgap, I've also manually created a `README_UNITS.md` file in turf-helpers with all of the supported turf UNITS and AREA_UNITS.  I link to this readme is now in each of the 4 grid functions in this PR.

Follow-on work:
* I do think the unit name extraction can be automated, as part of generate_readmes.ts, and inserted into the README markdown.  I have some initial work in the `extract-units` branch but it's WIP, the AST queries aren't quite working right.  * * Swap a link to this file into all of the other modules that support units.

packages/turf-point-grid/index.ts
```diff
@@ -12,13 +12,13 @@ import { distance } from "@turf/distance";
...
+ * Creates a grid of points
  *
  * @name pointGrid
- * @param {Array<number>} bbox extent in [minX, minY, maxX, maxY] order
- * @param {number} cellSide the distance between points, in units
+ * @param {BBox} bbox extent of grid in [minX, minY, maxX, maxY] order
+ * @param {number} cellSide the distance between points
  * @param {Object} [options={}] Optional parameters
- * @param {string} [options.units='kilometers'] used in calculating cellSide, can be degrees, radians, miles, or kilometers
+ * @param {Units} [options.units='kilometers'] the units of the cellSide value.  Supports all valid Turf {@link https://github.com/Turfjs/turf/blob/master/packages/turf-helpers/README_UNITS.md Units}
```
mfedderly, line 21: is this link going to work from the public docs site? (I think its going to send them to the github markdown page instead of staying on the docs website) Maybe that's fine for now, but it won't be versioned cleanly if we do it this way.

mfedderly (commented):
Probably needs a run of `pnpm run docs` to get the README.md's generated. Otherwise I'm a big fan of all of these changes.

packages/turf-point-grid/index.ts
twelch: When clicking the link from the public docs website, they will be sent to the markdown page, for now.  #2694 is meant to take it the rest of the way and extract the units into the turf-helpers README

mfedderly: sounds good. happy to merge this as is and give you some breathing room to get that in a good place.

mfedderly (approved):

smallsaucepan:
@twelch and @mfedderly, has this actually resolved #2638? I believe the implementation needs to be rolled back. It's not just a documentation tweak.

twelch:
I chose doc clarity to resolve that issue.  Not as a solution but an improvement over confusing docs. Rollback could certainly happen but wont fully solve user needs either as the change was made for a reason. I had started a discussion on grids that I think might yield the right path forward.

smallsaucepan:
@twelch, I don't agree with this approach. The updates to the documentation don't clear up the situation or explain what is happening or how to compensate for it.

The original change was made to cover an edge case where someone was trying to create a grid covering the entire planet (#2079). The fix for that (#2106) now produces visibly deficient results for what is probably a high proportion of users.

We should reopen #2638 until we at least decide how grids should work. I did reply to your discussion with follow up questions, though have not seen any activity since.

twelch:
@smallsaucepan I appreciate your thoughts.  My intention was to create a new issue with backlinks, It's been a busy few days for me I just hadn't gotten to yet, but I'm happy to reopen the previous if you think that's a better fit.

smallsaucepan:
Thanks @twelch. Our best short term solution may be to roll back to the 6.5.0 grid implementation, which would close #2638 but reopen #2079. Might be the lesser of two evils.

PR #2682, opened 2024-08-07
Merge turf:master into experimental dx branch

smallsaucepan:
Please fill in this template. Use a meaningful title for the pull request. Include the name of the package modified.

- [ ] Is this a bug fix, new functionality, or a breaking change?
- [ ] Have read and followed the steps for [preparing a pull request](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#preparing-a-pull-request).

Submitting a new TurfJS Module.

- [ ] Overview description of proposed module.
- [ ] Include JSDocs with a basic example.
- [ ] Execute `./scripts/generate-readmes` to create `README.md`.
- [ ] Add yourself to **contributors** in `package.json` using "Full Name <@GitHub Username>".

PR #2684, opened 2024-08-07
Upgrade typescript

mfedderly:
Supposedly this should give us performance gains, but I'm not seeing it in the PR build timings. I'm a little hesitant to upgrade Typescript just for maintenance sake because we've seen correctness issues internally with a bunch of our Typescript upgrades recently.

We may be able to leverage some of the newer features to clean up tsconfigs perhaps?


twelch:
I'm okay with upgrading just for maintenance sake, I think it reduces the amount of change at any one time, spreads it out.

twelch:
Do you think this PR is ready and useful to merge now?  Or perhaps land after next release.

mfedderly:
Lets land it after #2682 gets merged, that's probably the highest priority PR outstanding.

smallsaucepan:
> Lets land it after https://github.com/Turfjs/turf/pull/2682 gets merged

For the archives, have chatted over on #2702 (successor to #2682) and will merge this upgrade first instead.

smallsaucepan (approved):

PR #2685, opened 2024-08-07
Update nx config for tests

mfedderly:
This fixes an issue where nx would cache test runs even if the underlying files have changed. This only happens if you execute tests like so:

```
pnpm lerna run test --scope=@turf/meta
```

This invokes the lerna/nx caching layer, and this will miss the fact that the actual code itself changed which should cause the tests to rerun. To fix this I extracted the source file inputs into a named input, and then used that in both build and test so that everything stays in sync.


mfedderly:
This might not be worth merging if we're taking #2682 instead, or it would at least need to be revisited.

mfedderly:
I reworked this to just not conflict with that PR, we can merge this safely now.

twelch (approved):

smallsaucepan:
Thanks for this @mfedderly 👍

PR #2686, opened 2024-08-07
Change typescript to depend on source instead of built files

mfedderly:
Minimal version of #2682

There's an issue where it doesn't leave the imports in the built code.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -10,6 +10,7 @@ packages/turf-directional-mean/test/in/bus_route_utm.json
 
 # is actually output
 packages/turf/turf.min.js
+packages/turf/test.example.js
 
 # Ignore test fixture json in case intentional line breaks help with coord
 # readability.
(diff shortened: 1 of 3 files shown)


packages/turf/test.ts
```diff
@@ -331,7 +331,7 @@ const turfTypescriptPath = path.join(__dirname, "..", "turf-*", "index.d.ts");
 
 // Test Strings
 const requireString = `import test from 'tape';
-import * as turf from './dist/esm/index.js';
+import * as turf from './index.ts';
```
mfedderly, line 334: Jumping into the built output gets us away from the tsconfig paths settings, so we have to hit the index.ts instead.

.prettierignore
```diff
@@ -10,6 +10,7 @@ packages/turf-directional-mean/test/in/bus_route_utm.json
 
 # is actually output
 packages/turf/turf.min.js
+packages/turf/test.example.js
```
mfedderly, line 13: I think this will help with rerunnability locally

PR #2693, opened 2024-08-08
Revert @turf/kinks to 6.5.0 version

### LINKED ISSUE #2627: @turf/kinks v7.0.0 doesn't report self-intersection points in polygon (v6.5.0 does)
I've been using `@turf/kinks` v6.5.0 to find whether the polygon has some self-intersecting lines inside.

The geometry that on v6.5.0 returned 3 points as a result of the `kinks(geometry)` call, in v7.0.0 returns no points.

Here is a gist with the GeoJSON(s): https://gist.github.com/arkus7/bbd0bad3a4d81eb6edc5d24f43ad9f7c

Example on JSFiddle using v7.0.0 https://jsfiddle.net/abefhm5o/1/
This shows that the newest version doesn't report the kinks the same as v6.5.0: https://jsfiddle.net/o1L8sd4u/

mfedderly:
Fixes #2627

This manually reverts our @turf/kinks implementation to just before we migrated it to `sweepline-intersections` but it keeps the tests added by that PR and adds the new ones from the most recent issue.


packages/turf-kinks/test/out/switzerlandKinked.geojson
```diff
@@ -6,15 +6,15 @@
...
+        "coordinates": [9.219017027587402, 47.657889465138915]
       }
     },
     {
       "type": "Feature",
       "properties": {},
       "geometry": {
         "type": "Point",
-        "coordinates": [9.219017027587402, 47.657889465138915]
+        "coordinates": [9.182061729855718, 47.67348418083509]
```
mfedderly, line 17: These are the same just reversed in ordering.

packages/turf-kinks/test/out/triple.geojson
```diff
@@ -22,7 +22,7 @@
       "properties": {},
       "geometry": {
         "type": "Point",
-        "coordinates": [-45, 2.5]
+        "coordinates": [-45, -7.5]
```
mfedderly, line 25: Again the same just reordered in the output.

packages/turf-kinks/package.json
```diff
@@ -66,7 +66,6 @@
   "dependencies": {
     "@turf/helpers": "workspace:^",
     "@types/geojson": "^7946.0.10",
-    "sweepline-intersections": "^1.5.0",
```
mfedderly, line 69: Does this mean we should also revert the `sweepline-intersections` change in `@turf/line-intersect` too?

twelch (approved):

PR #2694, opened 2024-08-09
auto-generate list of Units and AreaUnits and link 

### LINKED ISSUE #1298: Documentation: Supporting "meters" unit
Hi,
All operations that accepts units of "degrees, radians, miles, or kilometers" can also accept "meters".
One of our developers called `distance` with unit "kilometers" and divided the result by 1000, because the documentation did not specify that it supports meters too.

Minor issue, not critical...

twelch:
- [x] bug fix

Resolves #1298
Builds on https://github.com/Turfjs/turf/pull/2683

ToDo
- [ ] auto-generate list of Units and AreaUnits (Need to figure out how to structure the queries properly, help welcome)
- [ ] insert into turf-helpers/README.md as part of generate-readmes.ts
- [ ] migrate all modules with `units` parameters to link to this README.  Make sure the functions actually support all units.

Feedback is welcome on approach of using the [tsquery](https://github.com/phenomnomnominal/tsquery) module to query and build a master list of supported Units and AreaUnits, as part of the generate-readmes.ts script run by CI.  tsquery is a relatively easy-to-use AST/query tool that is different from the AST that documentation.js produces, which I think using Babel.  I think this will be easier to maintain.


package.json
```diff
@@ -33,6 +33,7 @@
     "@monorepolint/config": "0.5.0-alpha.132",
     "@monorepolint/core": "0.5.0-alpha.132",
     "@monorepolint/rules": "0.5.0-alpha.132",
+    "@phenomnomnominal/tsquery": "6.1.3",
```
mfedderly, line 36: `ts-morph` would be the alternative library here if you decide you don't like `tsquery`, I don't have a strong opinion here

PR #2702, opened 2024-08-16
Introduced new developer experience efficiencies

smallsaucepan:
Reasonably major attempt to streamline the Turfjs development environment for contributors. Overall concept is to avoid generating Javascript build artefacts unless we absolutely have to.

Summary of changes. Some may be contentious, so happy to debate or justify in more depth:

1. Uses `paths` config (tsconfig.shared.json) to help build and test tools / editors resolve TS dependencies between packages more directly. Used to have to build dependent packages to JS first and load those
1. Removed need to build project before testing. Major time saver for Turf org developers
1. Adopted newer npm lifecycle targets to avoid building things before we need them. Doesn't automatically generate two versions of JS on `npm install` for example
1. Added static type checking for all projects. TS packages type check index.ts directly, JS packages test a types.ts scaffold file as the entry point. Detects errors introduced in locally edited dependent packages
1. Removed most existing types.ts files. Approach above hits the index.ts file directly, and most types.ts file weren't doing much more than serving as an entry point for tsc
1. Rejigged rollup + last-tests targets in @turf/turf as `cdnBundle`. Felt this more accurately describes what that's doing
1. ESM and CJS Javascript, and minified CDN bundle, aren't generated until we are actually preparing to publish a release
1. Updated module and moduleResolution to esnext and bundler respectively (tsconfig.shared.json). This didn't seem to affect the generated JS output, as the target remains at es2017
1. At end of CI run all checked in code will have been unit tested and type checked. Artefacts generated for a release (EJS, CJS, etc) will no longer be tested every CI run. However, those are the products of automatic processes so testing every time would rarely uncover problems not already seen earlier in unit tests.

Further background discussion #2629 and some relevant comments on now superceded PR #2682.

- [x] project infrastructure


.monorepolint.config.mjs
```diff
@@ -50,6 +46,56 @@ const JS_TAPE_PACKAGES = TAPE_PACKAGES.filter(
 
 export default {
   rules: [
+    fileContents({
```
mfedderly, line 49: https://monorepolint.com/docs/rules/standard-tsconfig/ maybe we should use `standardTsconfig` now instead of fileContents. My biggest issue with the current monorepolint is that we've got hand-managed lists of files. Although most of our packages only use a single file, its definitely unexpected if I add another file and the import doesn't work until it gets added in here (and further extends the proliferation of special cases in this file.

packages/turf-boolean-touches/types.ts
```diff
@@ -1,9 +0,0 @@
-import * as helpers from "@turf/helpers";
-import { booleanTouches } from "./index.js";
-
-const pt = helpers.point([0, 0]);
-const line = helpers.lineString([
-  [0, 0],
-  [10, 10],
-]);
-booleanTouches(pt, line);
```
mfedderly, line 9: I think the point of these was to help guard against API breaks in the types? Kind of like what DefinitelyTyped does. I'm not sure that that's a super compelling argument here with a relatively basic amount of syntax expressed in this file. The `last-checks` step is grabbing the code in the examples and checking that the example compiles. Thoughts? (I do kind of want each package to test its own example, documentation generation, etc).

packages/turf-clusters-dbscan/index.ts
```diff
@@ -2,7 +2,7 @@ import { GeoJsonProperties, FeatureCollection, Point } from "geojson";
 import { clone } from "@turf/clone";
 import { distance } from "@turf/distance";
 import { degreesToRadians, lengthToDegrees, Units } from "@turf/helpers";
-import { rbush as RBush } from "./lib/rbush-export.js";
+import rbush from "rbush";
```
mfedderly, line 5: Its nice that we can get rid of the import hacks, I'm still trying to fix the underlying typings still.

packages/turf-clusters-dbscan/types.ts
```diff
@@ -1,30 +0,0 @@
...
-const minPoints = 3;
-const units = "miles";
-clustersDbscan(points, maxDistance);
-clustersDbscan(points, maxDistance, { units });
-clustersDbscan(points, maxDistance, { units, minPoints });
-
-// Custom Properties
-clustered.features[0].properties.cluster;
-clustered.features[0].properties.dbscan;
-clustered.features[0].properties.foo;
```
mfedderly, line 30: This is an example of a type test that might be more likely to catch typings breaks than just the example for the package.

packages/turf-geojson-rbush/tsconfig.testTypes.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.shared.json",
+  "compilerOptions": {
+    "noEmit": true
+  },
+  "files": ["types.ts"]
+}
```
mfedderly, line 7: If you wanted to consolidate this further, you can make a root `tsconfig.testTypes.json` which has all of this and then each package can just be an `extends` directive. With `${configDir}` things like `files` are now possible to define once. https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#the-configdir-template-variable-for-configuration-files

packages/turf-geojson-rbush/tsconfig.testTypes.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../../tsconfig.shared.json",
+  "compilerOptions": {
+    "noEmit": true
+  },
+  "files": ["types.ts"]
```
mfedderly, line 6: I'm actually a little surprised this doesn't need to declare `index.js`, I must be missing something about how the paths directive works or something.

packages/turf-line-slice-along/types.ts
```diff
@@ -0,0 +1 @@
+import { lineSliceAlong } from "./index";
```
mfedderly, line 1: Probably needs some usage if this file exists.

packages/turf-line-slice/types.ts
```diff
@@ -0,0 +1 @@
+import { lineSlice } from "./index";
```
mfedderly, line 1: Probably needs some usage if this file exists.

.monorepolint.config.mjs
mfedderly: I wonder if we can get away with something like this and be a little more forgiving of adding extra files.

```json
{
  "include": ["index.js", "index.ts", "lib/*", "*.d.ts"],
  "exclude": ["index.d.ts"]
}
```

packages/turf-boolean-touches/types.ts
smallsaucepan: Do you mean like a form of API locking between major versions? I thought so too at first. However, I could easily narrow the signature of booleanTouches to Feature<Point>, Feature<LineString> and this test would still pass. As is, I don't think they really protect us much.

> I do kind of want each package to test its own example, documentation generation, etc).

Definitely. That is still the case. The `tsc --noEmit` in each package is checking to see if the module compiles. We don't need to actually call our method to have tsc examine it though. Just importing seems to be enough.

We can add a bunch of example invocations if we want. All we're doing though is proving we know how to call our own library. It's not testing the type safety of the implementation any better.

packages/turf-geojson-rbush/tsconfig.testTypes.json
smallsaucepan: That would be great! Wasn't able to find a way to avoid these until now. I will refactor.

packages/turf-geojson-rbush/tsconfig.testTypes.json
smallsaucepan: I don't think tsc can type check the JS implementation at all. So for the JS projects I believe it uses types.ts as an entry point, loads up index.d.ts (infers that's what you really mean when asking to import from index.js), and then checks what it can.

packages/turf-line-slice-along/types.ts
smallsaucepan: Unless I'm missing your point, this is probably as good as we can make it, at least for a JS project. We could add example calls to lineSliceAlong, but all that's will confirm is we've written the example calls to match the signatures defined in index.d.ts

Alternative tack: do we not bother doing static type checking in our remaining JS packages? I don't think it's really adding much.

.monorepolint.config.mjs
smallsaucepan: Will check this idea out. Thanks 👍

smallsaucepan:
Thanks for that feedback @mfedderly. Let me know what you think about the type test comments. Everything else I'll refactor in coming days and push up.

packages/turf-geojson-rbush/tsconfig.testTypes.json
mfedderly: https://github.com/Turfjs/turf/pull/2684 is the Typescript upgrade for Turf

packages/turf-geojson-rbush/tsconfig.testTypes.json
mfedderly: Right but shouldn't `index.d.ts` have to be listed in files in that case?

packages/turf-line-slice-along/types.ts
mfedderly: Yeah if we aren't getting benefit from types.ts testing, maybe we can just remove it entirely. Something like https://api-extractor.com/ might be better for making sure we don't actually break APIs than these manual tests.

mfedderly: Ok so I dug in on one of the types.ts files and it originally came from here which is part of the early Typescript implementation, so maybe in 2024 they are less relevant.

https://github.com/Turfjs/turf/commit/74140e862cc7ba44082353de824cb76e0024a9c1

packages/turf-geojson-rbush/tsconfig.testTypes.json
smallsaucepan: `files` seems more to be the top level entry points where you want tsc to start compiling. From there it uses whatever the module resolution policy is to find all the dependencies.

packages/turf-geojson-rbush/tsconfig.testTypes.json
smallsaucepan: Shall we get #2684 merged first then, I'll bring that into this PR, and then rejig to use the configDir stuff?

packages/turf-line-slice-along/types.ts
smallsaucepan: So, sounds like we're happy to ditch the types.ts for remaining JS projects, knowing that we'll have actually useful static type checking once they're converted to TS?

API extractor worth investigating too 👍

packages/turf-geojson-rbush/tsconfig.testTypes.json
mfedderly: https://www.typescriptlang.org/tsconfig/#files I guess it doesn't say that it has to be exhaustive, but I always read that as needing to list every file. 🤷

packages/turf-line-slice-along/types.ts
mfedderly: I'd like to get more consensus around this decision, especially since I wasn't around when the decision was made to include them. I don't want to delay this PR indefinitely though.

packages/turf-geojson-rbush/tsconfig.testTypes.json
mfedderly: Yeah feel free to merge it whenever you want, I hit the update button. Actually though this brings up a bit of an interesting point, a new typescript version (or relevant package.json change?) should force everything to rebuild/test.

smallsaucepan: If you mean in CI, then yes it did.

packages/turf-line-slice-along/types.ts
smallsaucepan: Sure thing. I'll proceed as if we will retire them for the remaining JS projects, and if justification comes up to keep them, it's easy to revert before we merge this PR.

packages/turf/package.json
```diff
@@ -67,11 +67,11 @@
...
   ],
   "scripts": {
-    "build": "tsup --config ../../tsup.config.ts && rollup -c rollup.config.js",
-    "last-checks": "npm-run-all last-checks:testjs last-checks:example",
-    "last-checks:example": "tsx test.example.js",
-    "last-checks:testjs": "tsx test.ts",
-    "test": "echo '@turf/turf tests run in the last-checks step'"
+    "build": "tsup --config ../../tsup.config.ts",
+    "cdnBundle": "npm-run-all cdnBundle:minify cdnBundle:checks",
+    "cdnBundle:checks": "tsx test.ts && tsx test.example.js && tsx test.ts",
```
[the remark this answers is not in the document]
smallsaucepan: Fixed

packages/turf-line-slice-along/types.ts
smallsaucepan: The above is done. To further simplify configuration (MRL mainly) all packages now run tsc. For the JS projects though it's pretty much a no-op as it can't make any guarantees about the type safety of the JS code.

PR #2714, opened 2024-09-15
Fixed turf-angle to consistently measure angles clockwise

### LINKED ISSUE #2703: `turf.angle` does not return signed angle as claimed
I'm trying to get the behavior of PostGIS [`ST_Angle`] in client-side JavaScript, so I turned to [`@turf/angle`]. Compare docs:

> **`ST_Angle`:** Computes the clockwise angle […] enclosed by the points P1-P2-P3

> **`@turf/angle`:** Finds the angle formed by two adjacent segments defined by 3 points. The result will be the (positive clockwise) angle with origin on the `startPoint`–`midPoint` segment

These both look like they do what I want: in particular, they both specify that they return the *clockwise* angle. So I can determine whether a point `B` lies to the right or the left of my line `O`→`A` by checking whether the result of `angle(A, O, B)` is in [0°, 180°] or [180°, 360°]:

```
$ psql -tc "SELECT degrees(ST_Angle('POINT(0 1)', 'POINT(0 0)', 'POINT(1 0)'))"
      90

$ psql -tc "SELECT degrees(ST_Angle('POINT(0 1)', 'POINT(0 0)', 'POINT(-1 0)'))"
     270
```

But Turf gives the same answer for both queries:

```
> turf = require("@turf/turf")  // v7.0.0
> turf.angle([0, 1], [0, 0], [1, 0])  // okay
90
> turf.angle([0, 1], [0, 0], [-1, 0])  // wrong!
90
```

I tested on **version 7.0.0**. I also see this behavior represented in test cases:

https://github.com/Turfjs/turf/blob/6889237200a088fa5539f92f361fc96135e09ac1/packages/turf-angle/test.ts#L95-L96

Isn't this in error? The positive-clockwise angle from *A*=(0, 1) to *B*=(−1, 0) with origin *O*=(0, 0) is not 90°; it is 270° or −90°. The only points 90° clockwise of *A* are (*x*, 0) for *x* > 0.

I see #1192, which discusses this issue and raises a similar question (see the second diagram in particular, with the blue solid and dashed angles). But that PR "just added a few test cases to confirm and highlight the current behavior" and did not actually either change the default behavior or add an `absolute` option.

What is the best way to get the documented functionality within Turf? I can compute `bearing(O, B) - bearing(O, A)` and normalize to [−180°, 180°], but it seems like that's roughly what `@turf/angle` does before it calls `Math.abs` and throws away the useful sign information that it promises to return, which seems strange to me so I want to check if there's something that I'm missing.

[`ST_Angle`]: https://postgis.net/docs/ST_Angle.html
[`@turf/angle`]: https://turfjs.org/docs/api/angle

---

- [x] The version of Turf you are using, and any other relevant versions: v7.0.0
- [x] GeoJSON data: not necessary; see repros above
- [x] Snippet of source code: see above
- [x] Verify this issue hasn't already been reported: searched for `clockwise angle` and read all issues

smallsaucepan:
Additionally, reviewed the implementations of azimuthToBearing, radiansToDegrees, and degreesToRadians to be easier to follow. Small discrepancies in the angle() geojson test fixtures caused by the bearing be taken from the wrong end of each line needed to be fixed. These tests included additional illustrative info meaning they were overly fragile though, so retired those in favour of vanilla unit tests which should be sufficient for a simple function like angle.

Resolves #2703.

- [x] bug fix


mfedderly (approved):
Very nice!

smallsaucepan:
Thanks for reviewing @mfedderly

PR #2715, opened 2024-09-15
Reversed the order in which random polygon points are generated to satisfy the right hand rule

### LINKED ISSUE #2605: Randomly generated polygons do not comply with the right-hand rule.
I am currently testing some functionalities in my app, of which one is to check if an object is a valid GeoJSON. I'm using  [@mapbox/geojsonhint](https://github.com/mapbox/geojsonhint) for this.

```typescript
import { hint } from '@mapbox/geojsonhint';
import { randomPolygon } from '@turf/random';

const feature = randomPolygon(1).features[0];

function isValidGeoJSON(geojson: any): { valid: boolean; errors: any[] } {
	const errors = hint(geojson);
	return {
		errors,
		valid: errors.length === 0
	};
}

isValidGeoJSON(feature);
```

To test this function I generate random polygons with `@turf/random`. The problem is that all polygons randomly generated by the `randomPolygon` function fail due to the right-hand rule.

The RFC 7946 states [here](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6):

> A linear ring MUST follow the right-hand rule with respect to the area it bounds, i.e., exterior rings are counterclockwise, and holes are clockwise.

[@mapbox/geojsonhint](https://github.com/mapbox/geojsonhint) is not the only one complaining about this, tools like [geojson.io](http://geojson.io/) and [geojsonlint](https://geojsonlint.com/) also throw the same error.

I'm using @turf/random v6.5.0.

Random feature generated by `randomPolygon`:
```
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [
          -158.66509293153055,
          -73.08040644678542
        ],
        [
          -158.97038620807854,
          -74.68046996596513
        ],
        [
          -153.1212960672334,
          -72.72257839813791
        ],
        [
          -155.32806726918324,
          -76.44993243660309
        ],
        [
          -159.07024019343282,
          -75.21488784692389
        ],
        [
          -159.01662503629552,
          -78.17553374285981
        ],
        [
          -160.06913205531168,
          -75.78662828914291
        ],
        [
          -167.94347021160297,
          -79.55249722610885
        ],
        [
          -163.71367843734242,
          -73.78563223508077
        ],
        [
          -164.07509676069895,
          -67.48211236028651
        ],
        [
          -158.66509293153055,
          -73.08040644678542
        ]
      ]
    ]
  }
}
```

To reproduce generate a new feature using the `randomPolygon` function from `@turf/random` or copy the example above and paste it into [geojson.io](http://geojson.io/) or [geojsonlint](https://geojsonlint.com/).

smallsaucepan:
Resolves #2605.

- [x] bug fix


packages/turf-random/index.ts
```diff
@@ -165,9 +165,9 @@ function randomPolygon(
     vertices[vertices.length - 1] = vertices[0]; // close the ring
 
     // center the polygon around something
-    vertices = vertices.map(
-      vertexToCoordinate(randomPositionUnchecked(paddedBbox))
-    );
+    vertices = vertices
+      .reverse() // Make counter-clockwise to adhere to right hand rule.
```
mfedderly, line 169: This seem fine to do in the module's current state, but the performance of the module with all of the `array.map()` calls seems like it'd probably be easy to make quite a bit faster. If we went to that trouble we'd probably want to just make vertices just populate in the reverse order so that it was correct without the reversing step here.

Happy to merge this as is, because I haven't heard of any performance-sensitive random polygon generation hitting issues here, and this reads clearly in its current form.

mfedderly (approved):

packages/turf-random/index.ts
smallsaucepan: Let me take a look and see if there's a way to avoid reverse() ...

smallsaucepan: Some performance results. Ops/sec for a 100000 vertex random polygon, average of three runs:

| Original  | reverse()  |
| ------------- | ------------- |
| 47.98  | 46.75  |

Agree with you - the added code complexity isn't worth it. Will merge.

Issue #1669, opened 2019-05-02
discrepancy between code and documentation for union


andrewharvey:
The JSDoc for turn.union at https://github.com/Turfjs/turf/blob/master/packages/turf-union/index.ts says:

> * Takes two {@link (Multi)Polygon(s)} and returns a combined polygon. If the input polygons are not contiguous, this function returns a {@link MultiPolygon} feature.

However the online documentation at http://turfjs.org/docs/#union says:

> Takes two or more polygons

which is wrong, as union only takes two polygons, as two arguments, not any number of polygons as the live documentation suggests.

I'm not sure where I need to place the PR to fix this, given it's already fixed in master.

lostpebble:
I've been making use of union recently and passing in multiple polygons like so:

```typescript
const newGeometryFeature = union(...geometryFeatures);
```

It seems to be combining more than two polygon features together. But looking at the code in master, it seems like it only takes two? Kinda weird... Does the master code and released version not align?

I'm using ` "@turf/turf": "5.1.6"`.

andrewharvey:
> It seems to be combining more than two polygon features together. But looking at the code in master, it seems like it only takes two? Kinda weird... Does the master code and released version not align?

Yeah that's what I'm very confused about too!

rowanwins:
Hi @andrewharvey & @lostpebble 

Sorry for the confusion. Some of those docs are probably reflectively of the change from `JSTS` which was used in v5 and earlier which took any number of arguments. In v6 `martinez` was used which only takes 2 arguments. 

In v7 I've implemented [`polygon-clipping`](https://github.com/mfogel/polygon-clipping) which goes back to taking any number of arguments. The latest hasn't been published yet as a turf module yet but the 0.14.0 release of `polygon-clipping` looks really good and passes all the turf tests.

brylie:
This almost seems like a regression. Is there a way to get a union of a MultiPolygon with Turf.js?

rowanwins:
Docs and code are now in sync. It's still on my radar to relook at adding support for more than 2 polys.

PR #2732, opened 2024-10-15
Fixed confusing turf-union documentation

### LINKED ISSUE #1900: union@5.x supports multiple polygon param but union@6.x not support

### LINKED ISSUE #2377: Union of multiple Features/polygons yet in version 6
`turf-union` in all its docs mentions the union of only _two_ (multi)polygons or geojson's

https://www.npmjs.com/package/@turf/union

It would be nice and useful if we could unite several (multi)polygons or geojson's

```js
let union
union = turf.union(poly1, poly2)
union = turf.union(poly1, poly2, poly3)
union = turf.union(poly1, poly2, poly3, ..., polyn)

const polygons = [poly1, poly2, poly3, ..., polyn]
union = turf.union(...polygons)
```

### LINKED ISSUE #2657: @turf/union : ambiguity in the documentation
turf.union() can only merge 2 polygons at once (https://stackoverflow.com/questions/68005988/geojson-turf-merge-multiple-polygons-to-one-polygon-keeping-hole). It should be specified in the documentation (https://www.npmjs.com/package/@turf/union) and in the TypeScript interface file to avoid invalid merges with more than 2 polygons.

smallsaucepan:
There has been some confusion about what parameters union expects - see #1669, #1900, #2377 and #2657.

Docs are now in sync with the signature of the union function, which definitely _does_ accept multiple polygons or multipolygons.

Resolves #1900 
Resolves #2377
Resolves #2657

### DIFF
--- a/packages/turf-union/README.md
+++ b/packages/turf-union/README.md
@@ -4,47 +4,64 @@
 
 ## union
 
-Takes input [(Multi)Polygon(s)][1] and returns a combined polygon. If the input polygons are not contiguous, this function returns a [MultiPolygon][2] feature.
+Takes a collection of input polygons and returns a combined polygon. If the
+input polygons are not contiguous, this function returns a multi-polygon
+feature.
 
 ### Parameters
 
-*   `features` **[FeatureCollection][3]<([Polygon][1] | [MultiPolygon][2])>**&#x20;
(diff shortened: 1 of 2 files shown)


twelch (approved):
I see the switch in the jsdoc from Feature to FeatureCollection.  That looks correct.

Issue #2722, opened 2024-09-24
Defaulting to `ignoreSelfIntersections: false` changes behavior of a number of Turf methods in v7


pm0u:
Introduced in #2033

Defaulting to _include_ self intersections in the `@turf/line-intersect` module has introduced breaking changes to a number of consuming modules as well as creating unexpected behaviors within those modules. Previously, I believe this algorithm did not / was not able to check for self intersections however now it does by default.

## Examples
[6.5.0](https://codesandbox.io/p/sandbox/turfjs-6-5-0-m4hz96)
[7.1.0](https://codesandbox.io/p/sandbox/turfjs-7-1-0-s8x273) (includes sweepline module example)

## Related issues
https://github.com/Turfjs/turf/issues/2707
https://github.com/Turfjs/turf/issues/2700
https://github.com/Turfjs/turf/issues/2633
https://github.com/Turfjs/turf/issues/2585

pm0u:
While booleanIntersects can be worked around by adding an explicit `{ ignoreSelfIntersections: true }`, some methods such as `booleanCrosses` do not expose this option and do not explicitly exclude self intersections themselves essentially rendering them broken.

smallsaucepan:
Thanks for pulling this together @pm0u. We're taking a look now.

pm0u:
@smallsaucepan - thanks, happy to help with a PR if desired but I think where it stands now there are still some decisions to be made around how to handle this change before I could know where to go with that. let me know if I can help

smallsaucepan:
Hi @pm0u. The current plan is to revert Turf to the ignore behaviour of 6.5.0 - across the board. These look to be undocumented breaking changes that we should fix as soon as possible.

Have created discussion https://github.com/Turfjs/turf/discussions/2728 to work through the details of a PR. Are you in a position to join in over there?

smallsaucepan:
Hi @pm0u. Looks like discussion over at #2728 has concluded. Maintainers have agreed we should revert ignoreSelfIntersections behaviour to 6.5.0. Which I _think_ means setting ignoreSelfIntersections default value to true wherever it appears in the current API.

Would you like to prepare a PR for this?

pm0u:
I can work on that some this week - I am thinking some associated tests for this would be good as well. Some that include self intersections and are expected to return false for results at the very least (The other side may be useful as well?). I should be able to get the changes together but may reach out for some help on the PR as far as testing.

pm0u:
[...]

PR #2741, opened 2024-10-29
Revert self intersection behavior: `boolean-disjoint`

pm0u:
Purpose: Bug fix

Partially Addresses: 
https://github.com/Turfjs/turf/discussions/2728
https://github.com/Turfjs/turf/issues/2722

`boolean-disjoint`
- Sets `ignoreSelfIntersection` option to default to `true`
- Adds minimal types to geojson in test files to correct TS errors
- Adds a self intersecting polygon to test cases
- Updates test cases for expected values based on the flip of "default"
- Adds test cases for the self intersecting polygon


pm0u:
moving this back to draft as I think it would be best to update this PR after changes in https://github.com/Turfjs/turf/pull/2742

I think we should determine if we want dependent packages to just rely on the default or to be explicit about a setting (meaning - default to `undefined` if unprovided which would lead to `true` OR default to `true` if unset by the args).

PR #2748, opened 2024-11-19
Added a cautionary message to Units about measuring length in radians or degrees.

smallsaucepan:
Added a cautionary message to Units about measuring length in radians or degrees. Results differ based on latitude.


twelch (approved):

smallsaucepan:
Ace. Thanks @twelch

PR #2749, opened 2024-11-19
Added a regression test for clone() bug that was probably fixed a while ago.

### LINKED ISSUE #1621: [@turf/clone] - object with key "length" is interpreted as array due to value.length in js code
I have a GeoJSON feature with a key "length":
```js
myGeoJSON = {
    geometry: [...],
    bbox: [...],
    properties: {
        processed: {
            length: {
                orig: 123.456,
                opti: 100.00,
            },
        }
    }
};
```
at some point, `clone` is called, which interprets the processed object as an array due to line
```js
} else if (value.length) {
```
in
```js
/**
 * Clone Properties
 *
 * @private
 * @param {Object} properties GeoJSON Properties
 * @returns {Object} cloned Properties
 */
function cloneProperties(properties) {
    var cloned = {};
    if (!properties) return cloned;
    Object.keys(properties).forEach(function (key) {
        var value = properties[key];
        if (typeof value === 'object') {
            if (value === null) {
                // handle null
                cloned[key] = null;
            } else if (value.length) {
                // handle Array
                cloned[key] = value.map(function (item) {
                    return item;
                });
            } else {
                // handle generic Object
                cloned[key] = cloneProperties(value);
            }
        } else cloned[key] = value;
    });
    return cloned;
}
```
and throws an error...

looking at the TypeScript code, it looks as this is handled properly (source: https://github.com/Turfjs/turf/blob/master/packages/turf-clone/index.ts#L78):

```ts
} else if (Array.isArray(value)) {
```

there might be a reason for this, but it would be nice if JS array identification would be done in a more robust way.

p.s.: meanwhile I change my key :|

smallsaucepan:
Highly likely this was resolved when we updated our JS target for individual packages from es5 to es20xx.

Resolves #1621

### DIFF
--- a/packages/turf-clone/test.ts
+++ b/packages/turf-clone/test.ts
@@ -3,6 +3,7 @@ import {
   point,
   lineString,
   polygon,
+  feature,
   featureCollection,
   geometryCollection,
 } from "@turf/helpers";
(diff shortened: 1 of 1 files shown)


twelch (approved):

smallsaucepan:
Thanks @twelch

PR #2751, opened 2024-11-20
Added special handling to booleanPointOnLine for zero length lines

### LINKED ISSUE #2750: @turf/boolean-point-on-line incorrectly returns true in some cases when line is of zero length
turf 7.1.0

In cases where the line is of zero length (start and end points the same) booleanPointOnLine incorrectly returns true for any other point on the same longitude.

```
const line = turf.lineString([[1, 1], [1, 1]]);
let pt;

pt = turf.point([2, 13]);
turf.booleanPointOnLine(pt, line);
//false, cool

pt = turf.point([1, 13]);
turf.booleanPointOnLine(pt, line);
// true, oh no
```

Technically (I guess?) this function should only return true if the three points - start, end, and point - are the same.

smallsaucepan:
In cases where the point was on the same longitude as a zero length line, even if the point was far away, booleanPointOnLine would return true.

Resolves #2750 

- [x] bug fix


smallsaucepan:
Merging this will clear the way to getting a couple of cleanCoords bugs fixed as well.

packages/turf-boolean-point-on-line/index.ts
```diff
@@ -97,6 +97,25 @@ function isPointOnLineSegment(
   } else if (cross !== 0) {
     return false;
   }
+
+  // Special cases for zero length lines
+  // https://github.com/Turfjs/turf/issues/2750
+  if (Math.abs(dxl) === Math.abs(dyl) && Math.abs(dxl) === 0) {
```
mfedderly, line 103: Does this math need to be done with some small epsilon for the float comparison instead of `===`?

smallsaucepan: It should be ok without. The new section is guarding against the old (lower) code getting too many `0`s for dxl, etc. They don't seem to behave if they do.

If they get any value above zero, no matter how small, they cope. For example, did some experiments changing the test line string to:

```typescript
lineString([
  [1, 1],
  [1 + Number.EPSILON, 1],
])
```

and all is well.

smallsaucepan: I could try incorporating additional checks into the existing ```return dxl > 0 ? x1 <= x && x <= x2 : x2 <= x && x <= x1;``` type code but keeping them separate seemed neater.

mfedderly (approved):
thank you!

PR #2760, opened 2024-12-04
Build housekeeping for December

smallsaucepan:
Doing chores 🧹

* Updated build dependencies: tape, tsx, tsup, tslib, babel, rollup, monorepolint, and eslint.
* Removed a monorepolint rule that wasn't adding much value. Enforcing rollup version in only one package.
* eslint largely migrated automatically. Had to broaden some rule disabling for types.ts and test.ts. Unsure though if that was a config migration issue or eslint getting harder to please in its old age.


smallsaucepan:
This probably isn't going to pass CI until we drop building on node 16 - https://github.com/eslint/eslint/issues/18851#issuecomment-2328602881

smallsaucepan:
Made a meal of this in my fork. Will start again.

smallsaucepan:
There was a conflict between eslint and the node 16 build. Now that we've dropped 16 from CI these changes pass. Ready for review!

JamesLMilner (approved):
Generally this looks okay to me. You also eradicated 'ts-node' as a dev dependency from the top level package.json, assumedly its behavior is replaced with `tsx`? 

I don't know if the monorepolint piece has been superseded by some other actions/library? i.e. is there any programmatic enforcement of consistency between packages?

smallsaucepan:
Thanks for that. Yes, I'm going to guess we left ts-node in to be safe when revamping the build last year. Couldn't find any remaining references to it now though.

>  don't know if the monorepolint piece has been superseded 

On the whole monorepolint still runs - haven't changed that. All I removed was a rule that verified the rollup version in a single package. The rule wasn't adding any value except making someone update the version in _two_ places should they upgrade rollup.

MRL is still making sure we use the same version of typescript, etc in all packages.

PR #2761, opened 2024-12-04
Removed node 16 from CI and added node 22 instead

smallsaucepan:
Drop node 16 from CI (end of life was a year ago). Add node 22 instead.

### DIFF
--- a/.github/workflows/turf.yml
+++ b/.github/workflows/turf.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x]
+        node-version: [18.x, 20.x, 22.x]
 
     steps:
       - name: Checkout


mfedderly (approved):
This is technically a breaking change because we're dropping node16 official support, but I also think that's fine.

JamesLMilner (approved):

smallsaucepan:
Thank you both for looking at this.

> we're dropping node16 official support

@mfedderly gotcha. In the interests of tidying up, where do we define that? Can't see anything in the readme.

mfedderly:
@smallsaucepan I don't think we define it anywhere, but in other projects I've seen that they make a major semver break when they drop an old node version from support even if nothing else changes. For Node support I think that "We support currently supported versions of Node" is a reasonable policy, with browser support that gets trickier because I kind of expect people using modules other than @turf/turf to be doing their own transpile/polyfilling, but I know not everyone is being careful of every package in node_modules, for instance.

PR #2768, opened 2024-12-13
Added bbox parameter to @turf/interpolate

### LINKED ISSUE #1031: Add `bbox` parameter to @turf/interpolate
Currently `@turf/interpolate` defines the area on which to interpolate the values as the bounding box surrounding the input points.
It would be useful to be able to define your area of interest, which might be different than the current (then default)`bbox(points)`, like this:

screen shot 2017-10-19 at 8 11 05 am

or this:
screen shot 2017-10-19 at 8 19 22 am

lemonig:
Please fill in this template. Use a meaningful title for the pull request. Include the name of the package modified.

- [x] https://github.com/Turfjs/turf/pull/2766

>>>

I created three new test files with bbox, and the generated results were in line with expectations after my testing. But I don't know if every test file needs to create a new file with bbox in it

Resolves #1031


packages/turf-interpolate/index.js
```diff
@@ -55,7 +67,8 @@ function interpolate(points, cellSize, options) {
   gridType = gridType || "square";
   weight = weight || 1;
 
-  var box = bbox(points);
+  box = box || bbox(points);
```
smallsaucepan, line 70: Better to use `??` here.

packages/turf-interpolate/index.js
```diff
@@ -21,6 +31,7 @@ import { collectionOf } from "@turf/invariant";
  * @param {string} [options.property='elevation'] the property name in `points` from which z-values will be pulled, zValue fallbacks to 3rd coordinate if no property exists.
  * @param {string} [options.units='kilometers'] used in calculating cellSize, can be degrees, radians, miles, or kilometers
  * @param {number} [options.weight=1] exponent regulating the distance-decay weighting
+ * @param {BBox}   [options.bbox] assigned interpolation region
```
smallsaucepan, line 34: Would be helpful to expand on this documentation a little. Maybe something additional like "If not provided here, defaults to bbox of input points."?

packages/turf-interpolate/index.js
```diff
@@ -6,10 +6,20 @@ import { centroid } from "@turf/centroid";
 import { squareGrid } from "@turf/square-grid";
 import { triangleGrid } from "@turf/triangle-grid";
 import { clone } from "@turf/clone";
-import { featureCollection } from "@turf/helpers";
+import { featureCollection, validateBBox } from "@turf/helpers";
 import { featureEach } from "@turf/meta";
 import { collectionOf } from "@turf/invariant";
 
+function checkBBox(bbox) {
```
smallsaucepan, line 13: Is this providing extra protection over and above just calling validateBBox? Also, why the check for `bbox.bbox`?

smallsaucepan (changes requested):
Hi @lemonig. Thanks for this PR. Have a couple of minor changes to request and a question about checkBbox if you could please take a look.

packages/turf-interpolate/index.js
lemonig: I looked at the documentation and it mentioned that instead of using ES6, `??` is an ES2020 feature, so I used `||` , if ES6 can be used that would be great.
[Docs](https://turfjs.org/docs/contributing)

packages/turf-interpolate/index.js
lemonig: Oh, I overlooked that. I’ll add it.

packages/turf-interpolate/index.js
smallsaucepan: Would say that documentation is well out of date, and we use `??` successfully in other places. Go for it.

packages/turf-interpolate/index.js
lemonig: Yes, I overthought it. This check is redundant since the validateBBox method already handles the validation、

smallsaucepan:
Thanks heaps for that following through on this @lemonig 👍

PR #2771, opened 2024-12-16
Fixed cleanCoords to remove points with appropriate tenacity

### LINKED ISSUE #2305: Cleancoords on edge case  lineString results in 'invalid polygon' error
Turf 6.5.0

`turf.cleanCoords(turf.lineString([[0, 0], [0, 1], [0, 0]]))` causes an invalid polygon error to occur though interestingly `turf.lineString([[0, 0], [0, 1], [0, 0]])` is just fine.

Expecting the output to be `{"type":"Feature","properties":{},"geometry":{"type":"LineString","coordinates":[[0,0],[0,1]]}}` as there is no point in a lineString which goes to a location and returns on exactly the same path?

Another example that can trigger the invalid polygon error: `turf.cleanCoords(turf.lineString([[0, 0], [0, 0], [0, 2], [0,2], [0, 0]]))`

### LINKED ISSUE #2406: booleanEqual() sometimes outputs false by internal cleanCoords()
Both of these polygons have the same shape and are in the same location. The length of the coordinates is also the same. The only difference is the order of the coordinates.

A:
```json
{
  "type": "Feature", 
  "properties": {},
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [1, 3], [3, 3], [3, 1], [3, -3], [-3, -3], [-3, 3], [1, 3]
      ]
    ]
  }
}
```
B:
```json
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [
        [-3, -3], [-3, 3], [1, 3], [3, 3], [3, 1], [3, -3], [-3, -3]
      ]
    ]
  }
}
```
cleanCoords() drops [3,1] from A, whereas it does [1,3] and [3,1] from B. Therefore, the lengths of the coordinates are different and the result of booleanEqual is false. In polygon A, [1,3] is at the starting point, so it seems to make that difference.

- Turf Version: 6.5.0
- Snippet: https://jsfiddle.net/jLpdb61t/9/

### LINKED ISSUE #2740: cleanCoords removes points too aggressively from retraced vertical lines
cleanCoords is too aggressive when dealing with straight vertical lines that start and stop at the same point.

Working clockwise:

🟥 North [0,0],[0,2],[0,0] -> **[0,0],[0,0]**
🟩 East [0,0],[2,0],[0,0] -> [0,0],[2,0],[0,0]
🟥 South [0,0],[0,-2],[0,0] -> **[0,0],[0,0]**
🟩 West [0,0],[2,0],[0,0] -> [0,0],[2,0],[0,0]

Adding extra points does not help:

🟥 [0,0],[0,1],[0,2],[0,3],[0,0] -> **[0,0],[0,0]** 

Going off axis or deviating the line does help:

🟩 [0,0],[0,2],[1,2],[0,2],[0,0] -> [0,0],[0,2],[1,2],[0,2],[0,0] // retraced but not straight
🟩 [0,0],[0,2],[1,2],[0,0] -> [0,0],[0,2],[1,2],[0,0] // not retraced a.k.a. triangle
🟩 [0,0],[1,2],[0,0] -> [0,0],[1,2],[0,0] // retraced but off axis

Different start and end points are fine:

🟩 [0,0],[0,2] -> [0,0],[0,2]

v7.1.0

smallsaucepan:
In certain cases cleanCoords was being _too_ aggressive when it removed points.

In other cases, if the start point of a polygon was a valid point to remove, it _wasn't_ being discarded.

Resolves #2305
Resolves #2406 
Resolves #2740 

- [x] Summary of the changes.


pnpm-lock.yaml
```diff
@@ -17029,6 +17056,49 @@ packages:
       - yaml
     dev: true
 
+  /tsup@8.3.5(tsx@4.19.2)(typescript@5.5.4):
```
mfedderly, line 17059: ^ this caught my eye. We can probably just run `pnpm dedeup` in its own PR to clean all of this up.

mfedderly (approved):

PR #2814, opened 2025-01-15
Support (Multi)Point,FeatureCollection and GeometryCollection in bboxClip

### LINKED ISSUE #2813: Suggestion: crop
There is not really an easy way to crop a featureCollection to a polygon or bbox.

There are times when I just want to do:

`const newFC = turf.crop(oldFC, [xmin, ymin, xmax, ymax])

or 

`const newFC = turf.crop(oldFC, cropPolygon)`

It's relatively easy to use intersect to crop `Polygon`s, and `booleanPointInPolygon` to filter out `Point`s, but I'm not even sure if there's a straightforward way to crop `LineString`s currently.

stevage:
Fixes #2813 

Also, small change to existing behaviour to return Multi* objects with no coordinates instead of defective Polygon and LineString objects. That is, instead of a Polygon with `coordinates: []` which is in violation of the GeoJSON spec, you get a `MultiPolygon` with `coordinates: []` which is ok.

Possibly this would count as a breaking change? But would anyone really rely on that?

- [x] Summary of the changes. (mostly explained in #2813)


stevage:
Oh, a note: it would be nice to have the type signature capture this logic:
- if you pass a Point, LineString or Polygon, you get the same thing back, or the Multi version
- otherwise you get the same thing back

I tried. TypeScript defeated me though.

smallsaucepan:
> instead of a Polygon with coordinates: [] which is in violation of the GeoJSON spec

[3.1](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1) suggests that is actually ok? "GeoJSON processors MAY interpret Geometry objects with empty "coordinates" arrays as null objects."

Not saying that's any more intuitive. Just seems odd to go from 0 results = multipoint, 1 result = point, 2 or more results = back to multipoint. Would we be better off returning null? Would be easier to check in calling code.

> Possibly this would count as a breaking change? But would anyone really rely on that?

It would be breaking if we started returning a wider variety of types than before.

> I tried. TypeScript defeated me though.

Where did this come unstuck? Were you trying function overloading? Or some other mechanism?

stevage:
>[3.1](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1) suggests that is actually ok? "GeoJSON processors MAY interpret Geometry objects with empty "coordinates" arrays as null objects."

Yeah, but since Turf is generating the data, that would essentially be forcing all apps using Turf to adopt the non-standard interpretation.

> Not saying that's any more intuitive. Just seems odd to go from 0 results = multipoint, 1 result = point, 2 or more results = back to multipoint. Would we be better off returning null? Would be easier to check in calling code.

It's intuitive if you think of the rules of English:

- 0 pointS
- 1 point
- 2 pointS
- ...

The Multi- strategy is imho the best for consistency.

- Return null for a polygon with no points does not work for a FeatureCollection, because it can't contain `null` values. So Polygons in a FeatureCollection would have to behave differently from a single Polygon.
- It's pretty questionable whether a Point can have a `[]` coordinates array. The spec in 3.1.1 says "one position in the case of a Point geometry, " and it is clear that a position must be at least two numbers. So potentially the Point case would have to behave differently from LineString and Polygon.

There is theoretically also the option of returning a feature with a null geometry, which would be legal, but weird in its own way.

What I like about my approach:
- the output features are 100% legal and extremely likely to be handled correctly by any application without special handling
- it is already the case that LineStrings and Polygons might get turned into Multi- geometries, so any application author already has to consider that possibility.

I did consider that perhaps this behaviour could be an option, like `{ returnNulls: true}` as a third argument. If true, it would return null for a single feature outside the Bbox, or just drop them from a FeatureCollection. That could be a future enhancement though.

> It would be breaking if we started returning a wider variety of types than before.

It's already true that you can pass a Polygon and get back a MultiPolygon. By coincidence, if anyone is using logic like `if (returnedPolygon.geometry.coordinates.length === 0)` that will actually still work.

> Where did this come unstuck? Were you trying function overloading? Or some other mechanism?

It was a big chain of conditional generics (`... ? ... : ... `). I should have made a copy of the attempt, oops.

stevage:
Pinging @mourner as the original author in case he has any opinions.

smallsaucepan:
All fair points about null geometries in features, etc @stevage.

> It's intuitive if you think of the rules of English:
> 
> 0 pointS
> ...

Hadn't thought of it that way. I suppose I'm looking at it from the POV of a user who wants to know "was there anything returned in the clip area?" They would have to do this right?

```
if (result.type === "Feature" &&
  result.geometry.type = "MultiWhatever" &&
  result.geometry.coordinates.length === 0) {}
```

I understand the approach you're suggesting makes sense if you just feed the result straight into a loop to process the points somehow. If it's empty the loop runs zero times, no dramas.

If the user wants to make a decision at that point though they need to know how to detect an empty result. That first test above would require a pretty good understanding of geojson to get right.

How about returning `null` for the entire result e.g.
```
if (result === null) {}
```
We do that in other places (e.g. intersect, difference) so there is a precedent.

>> It would be breaking if we started returning a wider variety of types than before.

> It's already true that you can pass a Polygon and get back a MultiPolygon.

True, however now we're also talking about returning Points, MultiPoints, and FeatureCollections. If client code isn't handling those already (and I don't think they would as 7.2.0 doesn't mention them) that would probably start generating type errors.

> > Where did this come unstuck? Were you trying function overloading? Or some other mechanism?

> It was a big chain of conditional generics (... ? ... : ... ). I should have made a copy of the attempt, oops.

Oh god. I hope you've recovered. Something like this could be an option (generics omitted for line length purposes):

```typescript
// Overloaded typedefs
function bboxClip(feature: Feature<Point | MultiPoint>, bbox: BBox): Feature<Point | MultiPoint> | FeatureCollection;
function bboxClip(feature: Feature<Line | MultiLine>, bbox: BBox): Feature<Line | MultiLine> | FeatureCollection
function bboxClip(feature: Feature<Polygon | MultiPolygon>, bbox: BBox): Feature<Line | MultiLine> | FeatureCollection;
// plus options for FeatureCollections and GeometryCollections ...

// Actual function implementation is a superset of all the above
function bboxClip(feature: Feature<Point | MultiPoint | Line | MultiLine | Polygon | MultiPolygon>, bbox: BBox): Feature<Point | MultiPoint | Line | MultiLine | Polygon | MultiPolygon> | FeatureCollection {
  // do the stuff
}
```
It's reasonably readable - definitely more so than 15 levels of nested ternary statements!

Big picture, thinking of Turf-wide consistency, I think it's useful to have a tidy way to indicate to users there was no result, whatever that means. I believe this PR with its current scope will be a breaking change (cause of other return types), so wondering if that changes the conversation at all?

stevage:
>I suppose I'm looking at it from the POV of a user who wants to know "was there anything returned in the clip area?" 

It's a pity we don't also have a function that more directly answers that question, a "booleanCrossesBbox" or something.

> They would have to do this right?

```
if (result.type === "Feature" &&
  result.geometry.type = "MultiWhatever" &&
  result.geometry.coordinates.length === 0) {}
```

I think they just have to do something like this:

```
if (! result?.geometry?.coordinates?.length > 0) {  }
```

>If the user wants to make a decision at that point though they need to know how to detect an empty result. That first test above would require a pretty good understanding of geojson to get right.

I don't think this is true.

> How about returning null for the entire result e.g.

This is much more problematic, because now we force the user to do the check, because it's very likely any subsequent code will throw errors. I really don't see "nothing in the clipped area" as a special case warranting special handling or thrown exceptions, and I think the function should behave accordingly.

>True, however now we're also talking about returning Points, MultiPoints, and FeatureCollections. If client code isn't handling those already (and I don't think they would as 7.2.0 doesn't mention them) that would probably start generating type errors.

Those types are never returned because those types aren't accepted as arguments currently. If we can get the type signature of the function correct, it won't cause any generated type errors, I think.

>Something like this could be an option (generics omitted for line length purposes):

Ah yes, overloads looks definitely like the right approach here.

>I believe this PR with its current scope will be a breaking change (cause of other return types), so wondering if that changes the conversation at all?

Again, if we get the type signature right, I don't think that's the case.

Before: Pass a Polygon, get either a MultiPolygon or Polygon back. After, the same. But also if you happen to pass a Point, now you get either a Point or a MultiPoint instead of it just being a type error. I don't think that's a breaking change?

Incidentally, what do you think about the idea of an options flag to request nulls instead of Multis?

stevage:
>Ok I'm willing to be optimistic, though think we'll have to take the overloading approach to stand a chance.

I think I have achieved this now in my latest commit.

I haven't thoroughly tested the type resolution, but my basic tests seem to give the right result. I'm guessing we don't have any actual TypeScript type testing in Turf?

The new type definitions are more specific than previously. Before there wasn't even a defined return type.

A wrinkle I hadn't considered was that if you pass a non-specific geometry type in, you should just get a non-specific geometry out. Fortunately, that case seems to work without losing the specificity of the other cases.

>And the user adjusts their "empty result test" to suit. That cover all the bases?

I don't think think the user has to adjust anything. If they are doing this:

```js
const out = bboxClip(myPolygon, bbox);
if (out.geometry.coordinates.length === 0) console.log('nothing in clip area')
```

it will still work.

If for some reason they have written:

```js
const out = bboxClip(myPolygon, bbox);
if (out.geometry.type === 'Polygon' && out.geometry.coordinates.length === 0) console.log('nothing in clip area')
```

then this will not work.

Worth noting that the documentation did not make any promises about what happens when a polygon is not within the clip area. So now it is much clearer.

stevage:
Btw after doing a bit more close reading, I should correct some of my statements:

- a `Polygon` with coordinates `[]` is not in violation of the spec, as far as I can tell.
- a `LineString` with coordinates `[]` *is*: "[3.1.4](https://stevage.github.io/geojson-spec/#section-3.1.4). LineString
For type "LineString", the "coordinates" member is an array of two or more positions."

FWIW, geojson.io takes the same interpretation: a LineString with [] coordinates is invalid.

Also, the current behaviour of `bboxClip` is that if you pass a `MultiLineString` containing a single `LineString` it will return a `LineString` back. (Just an oversight in the original implementation I think). So maintaining the original geometry type does not appear particularly sacrosanct.

stevage:
So...after all that (sorry!), I've reverted back to returning a `Polygon` with empty coordinates array when the Polygon is outside the clip region. I think I was confusing each ring (which needs at least 4 members) with the coordinates array itself (which doesn't seem to require more than 0 rings).

I did change the behaviour so that if you pass a single-item MultiLineString, you get a single-item MultiLineString back. That just seemed like a bug.

smallsaucepan:
Lol. All good. Taking a fresh look at the latest commit.

> A wrinkle I hadn't considered was that if you pass a non-specific geometry type in, you should just get a non-specific geometry out.

Would you mind dropping an example of what that would look like? Had a go locally and TS is refusing to accept anything I'm doing to prompt the situation.

> So maintaining the original geometry type does not appear particularly sacrosanct.

Not sure we could even if we wanted to. For example, what would be your intuition for the return type of the below (passing in a Polygon)?

Screenshot 2025-01-22 at 22 29 45

Btw, seems like the current behaviour merges the bbox with the clipped multiple polygons to be able to return a polygon. Not sure if that was intentional to keep the same return type or some other reason. See #2816.

stevage:
> > A wrinkle I hadn't considered was that if you pass a non-specific geometry type in, you should just get a non-specific geometry out.
> 
> Would you mind dropping an example of what that would look like? Had a go locally and TS is refusing to accept anything I'm doing to prompt the situation.

Well, this:

```typescript
const test = { type: "Point", coordinates: [0, 0] } as Geometry;
const out = bboxClip(test, bbox);
```

Hovering over `out` in VS Code gives 

image

  

> 
> > So maintaining the original geometry type does not appear particularly sacrosanct.
> 
> Not sure we could even if we wanted to. For example, what would be your intuition for the return type of the below (passing in a Polygon)?

Indeed. To be more precise, I should have said "maintaining the original geometry type, in cases where it's possible, .."

> Btw, seems like the current behaviour merges the bbox with the clipped multiple polygons to be able to return a polygon. Not sure if that was intentional to keep the same return type or some other reason. See #2816.

Yeah, I made some assumptions about the current behaviour in this whole PR, but I suspect the current behaviour produces defective results in a lot of places. It's really just a very simplistic wrapper around `lineClip` which doesn't deal with the complexities of clipping polygons.

smallsaucepan:
> Well, this:

Of course. Thanks.

> I suspect the current behaviour produces defective results in a lot of places

Guess we cross the bridge when we come to it. Feel the return types are generous enough now that when we fix the behaviour the types won't have to change again.

Do you have any thoughts on how to adapt the JSDoc to handle the multiple overloads? There should be a params / returns combo for each to make clear the output types are constrained by the inputs e.g. put in a Polygon and you know you're never going to get a FeatureCollection.

Screenshot 2025-01-23 at 13 01 31

stevage:
Hmm, I didn't think about JSdoc. I doubt it can handle anything as complex
as the overloads, but it should have the other types added at least.

I'm away for a few days, so feel free to make that doc change otherwise
I'll get to it in a bit.

smallsaucepan:
Will take a look. Thanks for working through it 👍

stevage:
Ok, I have updated the doc.

Also, since it seems that currently polygons are *not* split into MultiPolygons I have updated the doc to reflect that, and added test cases. I do think it should be fixed, but that's covered in #2816.

stevage:
Btw, just checking - is there anything else I need to do? Or is it just a question of waiting until someone has time to review and merge? Happy to be patient if so.

.gitignore
```diff
@@ -33,4 +33,4 @@ node_modules
 
 .vscode
 
-.nx/
```
smallsaucepan, line 36: Was this an intentional commit?

packages/turf-bbox-clip/README.md
```diff
@@ -4,13 +4,14 @@
 
 ## bboxClip
 
-Takes a [Feature][1] and a bbox and clips the feature to the bbox using
-[lineclip][2].
-May result in degenerate edges when clipping Polygons.
+Takes a [Feature][1], [Geometry][2] or [FeatureCollection][3] and a bbox and returns the part of the object within the bbox.
+Lines and polygons are clipped using [lineclip][4].
+If a Point or LineString geometry is entirely outside the bbox, a [MultiPoint][5] or [MultiLineString][6] with empty `coordinates` array is returned.
+LineString geometries may also become MultiLineString if the clipping process cuts them into several pieces.
```
smallsaucepan, line 10: Did we get to a point where we have a complete description of what might go in and what might go out? Is there something we can say here about polygons too? For example, If the input geometry is entirely outside the bbox a corresponding multi-geometry is returned with an empty coordinates array. Similarly, if a geometry is split into multiple pieces a corresponding multi-geometry is returned. If a multi-geometry is split and only a single item remains, the corresponding singular-geometry is returned. 

Probably don't need to quote the particular lineclip library.

smallsaucepan (changes requested):

smallsaucepan:
Hi @stevage. Having a go at getting this one closed out. Couple of minor changes / confirmations requested. Thanks!

packages/turf-bbox-clip/README.md
stevage: >Probably don't need to quote the particular lineclip library.

That's just the status quo.

As for the rest, I really don't remember at this point. Whatever you decide is fine with me.

PR #2821, opened 2025-01-18
booleanPointInPolygon - Perf improvement + fix issue with multiple polygons when the point is on the boundary of a polygon

### LINKED ISSUE #3050: booleanPointInPolygon possible false negative with an overlapping MultiPolygon
booleanPointInPolygon can possibly return a false negative under certain circumstances. It effectively returns early without considering the point might yet be in one of the other polygon geometries.

From @Pitouli 's comments on PR #2821:

> ### Fix edge case
> If the point was on the edge of a polygon, the function was returning immediately "true" if the option ignoreBoundary was false, and false otherwise.
> 
> But the RFC does not say that polygons inside a multipolygon cannot overlap, it only says a multipolygon is an array of polygons.
> https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.7
> 
> So in case the point is on the border of one of the polygon and ignoreBoundary is true, the loop should continue to see if the point is inside an another polygon.

Pitouli:
### Perf improvement

No breaking changes; it simply return early once it confirmed that the point is inside at least one polygon, instead of checking all the other polygons.

### Fix edge case

If the point was on the edge of a polygon, the function was returning immediately "true" if the option ignoreBoundary was false, and false otherwise. 

But the RFC does not say that polygons inside a multipolygon cannot overlap, it only says a multipolygon is an array of polygons.
https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.7

So in case the point is on the border of one of the polygon and ignoreBoundary is true, the loop should continue to see if the point is inside an another polygon. 

Note that some libraries seems to consider that multipolygon must not overlap
https://github.com/Leaflet/Leaflet/issues/6173#issuecomment-790615562
But this PR is not even a breaking change if we consider this definition, since for a non overlapping multipolygons the result of the function will stay the same ; the only difference is that it will indeed loop over all the polygons instead of exiting early.

### Tests

The tests have been ran for this package, and a new case has been created to specifically test the situation with a multipolygon with a point on the borders of a small polygon inside a bigger one. 

I let Github CI run the full tests.

Resolves #3050 3050


mfedderly (approved):
Thank you!

PR #2829, opened 2025-01-23
Run pnpm dedupe to clean up the lockfile

mfedderly:
Minor housekeeping for us to pull in fewer duplicate packages when working on Turf.

We could consider enforcing this so we don't reintroduce duplicates, but it also feels fine to just occasionally run a dedupe and commit that instead.


smallsaucepan (approved):
Looks good. Thanks @mfedderly

PR #2830, opened 2025-01-25
Fixed simplify to not get stuck in an infinite loop on certain geometries

### LINKED ISSUE #1788: @turf/simplify gets stuck in an infinite loop on certain geometries
Under certain circumstances the `turf/simplify` function never terminates.

The issue appeared with a geometry resulting from a union of 2 adjacent geometries that left a small gap in between the original polygons.

[Full geometry on geojson.io](http://geojson.io/#id=gist:senritsu/931da76e12035ff28b42195bd3765523&map=19/50.10820/11.66136) and [the problematic ring by itself on geojson.io](http://geojson.io/#id=gist:senritsu/a481522e875fa900eacd3d026323ef18&map=19/50.10810/11.66219)

[This fiddle](https://jsfiddle.net/vcg2mt8f/) reproduces the issue using only the problematic ring itself, not the full geometry.

The freeze occurs in the commented part of this function here in `simplify/index.js`
```js
function simplifyPolygon(coordinates, tolerance, highQuality) {
    return coordinates.map(function (ring) {
        var pts = ring.map(function (coord) {
            return {x: coord[0], y: coord[1]};
        });
        if (pts.length < 4) {
            throw new Error('invalid polygon');
        }
        var simpleRing = simplifyJS(pts, tolerance, highQuality).map(function (coords) {
            return [coords.x, coords.y];
        });
        //remove 1 percent of tolerance until enough points to make a triangle
        while (!checkValidity(simpleRing)) {
            tolerance -= tolerance * 0.01;
            simpleRing = simplifyJS(pts, tolerance, highQuality).map(function (coords) {
                return [coords.x, coords.y];
            });
        }
        if (
            (simpleRing[simpleRing.length - 1][0] !== simpleRing[0][0]) ||
            (simpleRing[simpleRing.length - 1][1] !== simpleRing[0][1])) {
            simpleRing.push(simpleRing[0]);
        }
        return simpleRing;
    });
}
```

For the given example geometry the `simplifyJS` call returns an invalid ring (with only 3 points: `[[11.662180661499999, 50.1081498005], [11.662192661499999, 50.108041800500004], [11.662180661499999, 50.1081498005]]`) no matter how small the tolerance gets. This means the loop never terminates, and the page hangs.

smallsaucepan:
simplify was getting stuck in an infinite loop when it couldn't reduce the tolerance far enough to generate a valid polygon. This often happened with very small rings e.g. couple of centimetres to a side.

Resolves #1788 

- [x] Summary of the changes.

### DIFF
--- a/packages/turf-nearest-neighbor-analysis/README.md
+++ b/packages/turf-nearest-neighbor-analysis/README.md
@@ -30,7 +30,7 @@ Type: [object][1]
 
 ## nearestNeighborAnalysis
 
-Nearest Neighbor Analysis calculates an index based the average distances
+Nearest Neighbor Analysis calculates an index based on the average distances
 between points in the dataset, thereby providing inference as to whether the
 data is clustered, dispersed, or randomly distributed within the study area.
 
--- a/packages/turf-simplify/index.ts
+++ b/packages/turf-simplify/index.ts
@@ -6,19 +6,18 @@ import { AllGeoJSON, isObject } from "@turf/helpers";
 import { simplify as simplifyJS } from "./lib/simplify.js";
 
 /**
- * Takes a {@link GeoJSON} object and returns a simplified version. Internally uses the 2d version of
- * [simplify-js](http://mourner.github.io/simplify-js/) to perform simplification using the Ramer-Douglas-Peucker algorithm.
- *
+ * Simplifies the geometries in a GeoJSON object. Uses the 2d version of
+ * [simplify-js](https://mourner.github.io/simplify-js/).
  *
  * @function
- * @param {GeoJSON} geojson object to be simplified
+ * @param {GeoJSON} geojson GeoJSON object to be simplified
  * @param {Object} [options={}] Optional parameters
- * @param {number} [options.tolerance=1] simplification tolerance
- * @param {boolean} [options.highQuality=false] whether or not to spend more time to create a higher-quality simplification with a different algorithm
- * @param {boolean} [options.mutate=false] allows GeoJSON input to be mutated (significant performance increase if true)
- * @returns {GeoJSON} a simplified GeoJSON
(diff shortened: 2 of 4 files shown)


packages/turf-nearest-neighbor-analysis/README.md
```diff
@@ -30,7 +30,7 @@ Type: [object][1]
 
 ## nearestNeighborAnalysis
 
-Nearest Neighbor Analysis calculates an index based the average distances
+Nearest Neighbor Analysis calculates an index based on the average distances
```
mfedderly, line 33: Looks like this was already in the index.ts comment, just not populated into the readme. I thought we had pre-commit hooks to make sure that didn't happen. Oh well.

packages/turf-simplify/index.ts
```diff
@@ -147,11 +146,20 @@ function simplifyPolygon(
     }
     let ringTolerance = tolerance;
     let simpleRing = simplifyJS(ring, ringTolerance, highQuality);
-    // remove 1 percent of tolerance until enough points to make a triangle
-    while (!checkValidity(simpleRing)) {
+
+    // If simplified ring isn't valid (has been over simplified) reduce the
+    // tolerance by 1% and try again.
+    while (!checkValidity(simpleRing) && ringTolerance >= Number.EPSILON) {
```
mfedderly, line 152: I was worried about this leaving impossibly tiny polygons with a large number of points, but EPSILON is so much smaller than any meaningful coordinate delta that this is probably a nonissue.

```
> area({type: "Polygon", coordinates: [[0,0], [0, Number.EPSILON], [Number.EPSILON, Number.EPSILON], [Number.EPSILON, 0], [0, 0]]})
0
```

mfedderly (approved):
Thank you!

Issue #2824, opened 2025-01-20
pointToPolygonDistance with hole return unsigned values


ata-electronics:
Hello,
The pointToPolygonDistance function return positive value for a point in polygon with an hole:

How to reproduce :
Use this file in https://turf-sandbox.netlify.app/
[turfSandbox_scipt_pointToPolygonDistance.txt](https://github.com/user-attachments/files/18479900/turfSandbox_scipt_pointToPolygonDistance.txt)

Fix:
I think that the issue is on line 72 of file https://github.com/Turfjs/turf/blob/master/packages/turf-point-to-polygon-distance/index.ts
`return Math.min(smallestInteriorDistance, Math.abs(exteriorDistance));`
should be replaced by
`return Math.max(smallestInteriorDistance*-1, exteriorDistance);`

smallsaucepan:
Thanks for raising this @ata-electronics. Documentation definitely states "A hole is treated as the exterior of the polygon." so that value should be positive.

smallsaucepan:
Well done tracking that down @ata-electronics.

Would you mind explaining what was happening and how the different logic fixes the problem?

And also, do you have any idea why the existing test in multi-polygon.geojson didn't catch this issue?

LHBruneton-C2C:
Hello,

I've looked up this issue. I agree with @ata-electronics suggestion, and I can explain why.

You can have a look at this PR https://github.com/Turfjs/turf/pull/2845

And the existing `multi-polygon.geojson` didn't catch this issue because it only happens with polygons. Multi-polygons are correctly re-signed on line 56 with the call to `booleanPointInPolygon`.

smallsaucepan:
Thank you both @ata-electronics and @LHBruneton-C2C. Sorry took so long to get to it. Merging now.

PR #2845, opened 2025-02-24
Fix issue 2824 on turf-point-to-polygon-distance

LHBruneton-C2C:
Fixes https://github.com/Turfjs/turf/issues/2824 

A single case of the the computation was returning an incorrectly signed result:
- polygon
- with holes
- inside the polygon, outside the holes

A fixture has been added to pinpoint this exact case.

This case has been fixed to return the correctly signed result.


smallsaucepan (approved):
Fantastic! Thanks for tracking down and fixing the problem @LHBruneton-C2C

PR #2846, opened 2025-02-24
boolean-valid - Allow MultiLineStrings with only a single LineString

### LINKED ISSUE #2837: Why does `@turf/boolean-valid` require `MultiLineString`s to have at least two component `LineString`s?
I assume the OGC Simple Features spec mandates this but I am unable to find a source:

`boolean-valid` [will reject](https://github.com/Turfjs/turf/blob/7ee0b1d8b8af25ba88041afe3cc81db6ca44a4ad/packages/turf-boolean-valid/index.ts#L45) a `MultiLineString` which has `coordinates` with only one `LineString`, e.g. `"coordinates": [ [...] ]` 

The GeoJSON spec is frustratingly [silent](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.5) on whether or not `MultiLineString` coordinates can be empty; however, we frequently encounter GeoJSON `MultiLineString`s composed of a single `LineString`, typically from GPS devices.

Curiously, for the other multipart geometries, `boolean-valid` [does not](https://github.com/Turfjs/turf/blob/7ee0b1d8b8af25ba88041afe3cc81db6ca44a4ad/packages/turf-boolean-valid/index.ts#L65) [enforce this](https://github.com/Turfjs/turf/blob/7ee0b1d8b8af25ba88041afe3cc81db6ca44a4ad/packages/turf-boolean-valid/index.ts#L34). Why is that?

A `GeometryCollection` with an empty `"geometries": []` is similarly rejected, though the GeoJSON spec [explicitly allows this](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.8). Again, I assume the reason is the OGC spec, but I can't find a direct source.

sharmaB01:
Fixes #2837

Why does @turf/boolean-valid require MultiLineStrings to have at least two component LineStrings? #2837

because multiline will be like
```
{
  "type": "MultiLineString",
  "coordinates": [
    [
      [100.0, 0.0],
      [101.0, 1.0],
      [102.0, 2.0]
    ]
  ]
}
```


mfedderly (approved):
Did [some more digging](https://github.com/Turfjs/turf/issues/2837#issuecomment-5013025819) and I think we need to expect 1 or more LineStrings.

If I had to guess the `2` probably came from a copy/paste error from a Position coordinates check (for a [lng, lat] pair).

Thank you!

Issue #2441, opened 2023-05-31
Using booleanContains cannot accurately determine the inclusion result of line segments and concave convex polygons


Steve245270533:
Using booleanContains cannot accurately determine the inclusion result of line segments and concave convex polygons.

version: 6.5.0

```javascript
import { lineString, polygon, booleanContains } from '@turf/turf';

const turf_polygon = polygon([
  [
    [0, 0],
    [4, 0],
    [4, 1],
    [3, 1],
    [3, 2],
    [2, 2],
    [2, 1],
    [1, 1],
    [1, 3],
    [0, 3],
    [0, 0],
  ],
]);

const turf_line = lineString([
  [1, 2],
  [4, 0],
]);

console.log(booleanContains(turf_polygon, turf_line));
```

The final result is true, and it is evident from the image that this line segment is not completely contained within the polygon.  

I'm not sure if there's a problem with my usage. Is there any way to correctly determine the inclusion of line segments and polygons?

![image](https://github.com/Turfjs/turf/assets/46118247/60add397-34f2-463f-af17-aed2e3688cf8)

oscar9:
I'm also having the same issue with a very similar case. Thanks in advance.

samuelarbibe:
Hey! added a PR to fix this.  
#2848

Janickvw:
Having the same issues. Is there any updates on getting https://github.com/Turfjs/turf/pull/2848 merged in @samuelarbibe? Thanks for getting a PR up.

samuelarbibe:
Waiting for a review from a maintainer :)

smallsaucepan:
Related issue with different geometries at #2595 (poly / poly). The issues are related in as much as they return false positives if all the _points_ are inside, without noting whether the lines between the points venture outside the other polygon.

Not suggesting this PR needs to fix that as well - just that the issues (and fixes) might have similarities.

planted-mreinstein:
Can we get a release please, now that this has been resolved?  7.3.2 or 7.4?   I don't care about typescript, but geometry bugfixes are critical.

PR #2848, opened 2025-02-27
turf-boolean-contains + turf-boolean-within: Fix line in polygon

samuelarbibe:
This PR fixes a false-positive when running `boolean-contains` (or `boolean-within`) of `lineString` inside a `polygon`, as demonstrated in #2441.   

The current implementation of `boolean-contains` checks if all the midpoints of the line segments are inside the polygon, and results in a false-positive in the following cases:
  
Line midpoint in Polygon

Line midpoint in Polygon

This PR proposes the following logic:
1. The line and polygon's `bboxes` must overlap. (optimization)
2. All line points must be contained by the polygon.
3. At least one of the line's points must be contained **inside** of the polygon, and not on the boundary.
4. If line intersects with polygon - all segments must be contained the polygon. (optimization - midpoint is sufficient)

This "waterfall" approach makes sure no extra computation is performed - unless necessary. Only quirky cases will require the extra computation of steps 3 and 4 (a single iteration of step 3 and 4 will suffice).
The only exception is that the line's points are checked to be contained instead of the midpoints - but it returns false-positives for pretty basic cases, so I find this to be a reasonable compromise.

A test case for the second image has been added.


Janickvw:
Hey @samuelarbibe i've found an issue with this fix. I'll test out a fix and get back to you.

samuelarbibe:
> Hey @samuelarbibe i've found an issue with this fix. I'll test out a fix and get back to you.

I'm curious. Let me know :)

Janickvw:
@samuelarbibe  I think you focused too much on advanced use cases and forgot the simple ones 😂 
This implementation fails for this use case:
image

Sandbox link: https://turf-sandbox.netlify.app/?gist=6ce3e4e6d674e6dd9c92c5265b6a25e0

The issue is this line not returning any segments when there are no intersections.
```js
const lineSegments = turf.lineSplit(turf.feature(linestring), turf.feature(polygon));
```

Here is a potential fix:
```js
// ...
  const lineSegmentsOrLineString = lineSegments.features.length > 0 ? lineSegments.features : [turf.feature(linestring)];

  for (const lineSegment of lineSegmentsOrLineString) {
    const midpoint = turf.midpoint(
      lineSegment.geometry.coordinates[0],
      lineSegment.geometry.coordinates[1]
    );
// ...
```

Fix sandbox: https://turf-sandbox.netlify.app/?gist=57378d936e774ac353de86e975060bf9

Janickvw (changes requested):
As commented above

samuelarbibe:
Thank you.... got too focused on the complicated cases 😅. 
Implemented a fix similar to you suggestion, with a little fix (The problem is that if the line sting has 2 or more vertices on the boundary of the polygon, the midpoint might be contained by the polygon, so I split the linestring into multiple linestring, one per segment).

packages/turf-boolean-contains/index.ts
```diff
@@ -179,14 +179,26 @@ function isLineOnLine(lineString1: LineString, lineString2: LineString) {
...
-  const segments = coords
-    .slice(1)
-    .map((coord, i) => lineString([coords[i], coord]));
+  const outputSegments: Feature<LineString>[] = [];
+
+  for (let i = 0; i < coords.length - 1; i++) {
+    const seg = lineString([coords[i], coords[i + 1]]);
+    const split = lineSplit(seg, feature(polygon));
+
+    if (split.features.length === 0) {
```
smallsaucepan, line 193: Could you please add in some broad comments about why / what each section is doing? For example, what the parts of the commit mentioned in https://github.com/Turfjs/turf/pull/2848#issuecomment-3589032345 are there for.

Nothing too verbose. Just a line or two here and there to help orient the reader.

smallsaucepan:
Would you mind doing some rewording of the documentation while you're here? The "exact opposite result" phrasing isn't very satisfying (and I'm not sure it's true anyway if taken literally).

From
> Boolean-contains returns True if the second geometry is completely contained by the first geometry. The interiors of both geometries must intersect and, the interior and boundary of the secondary (geometry b) must not intersect the exterior of the primary (geometry a). Boolean-contains returns the exact opposite result of the @turf/boolean-within.
> 
> returns true/false

To
> Tests whether geometry _a_ contains geometry _b_. The interiors of both geometries must intersect, and the interior and boundary of geometry _b_ must not intersect the exterior of geometry _a_.
> 
> `booleanContains(a, b)` is equivalent to `booleanWithin(b, a)`
> 
 > returns true if geometry _a_ contains geometry _b_, false otherwise

Happy to hear it if you have other suggestions as well!

packages/turf-boolean-contains/index.ts
samuelarbibe: Added some comments that align the code with the logic described in the documentation 📖

smallsaucepan (approved):
Awesome. Thanks @samuelarbibe 👍

smallsaucepan:
Thanks for your work on this too @Janickvw ⭐

Janickvw:
> fix similar to you suggestion, with a little fix (The problem is that if the line sting has 2 or more vertices on the boundary of the polygon, the midpoint might be contained by the polygon, so I split the linestring into multiple linestring, one per segment).

Sorry to be a pain, since I know this is already merged. But i've read this comment a few times and still don't understand it. Seems redundant to split a line segment that completely overlaps a boundary? Do you have a graphic or picture you could use to demonstrate?

samuelarbibe:
> Sorry to be a pain, since I know this is already merged. But i've read this comment a few times and still don't understand it. Seems redundant to split a line segment that completely overlaps a boundary? Do you have a graphic or picture you could use to demonstrate?

Hey! no problem.

the problem is that the `split` function isn't sufficient.
For example, this scenario:

Screenshot 2025-12-09 at 14 27 36


The `split` function will return in this case a single line string feature:
Screenshot 2025-12-09 at 14 37 53
And using a midpoint on the whole line string may result in faulty cases.

The new function created, `splitLineIntoSegmentsOnPolygon`, splits the line string on the polygon - but makes sure to return individual segments.

```tsx
  const lineSegmentsOrLineString = lineSegments.features.length > 0 ? lineSegments.features : [turf.feature(linestring)];
```

is insufficient, as it assumes the `split` function splits in a certain way.
For reference - the `LineIsContainedByPolygonWithSegmentOnBoundary` test fails when using this logic.

Janickvw:
> > Sorry to be a pain, since I know this is already merged. But i've read this comment a few times and still don't understand it. Seems redundant to split a line segment that completely overlaps a boundary? Do you have a graphic or picture you could use to demonstrate?
> 
> Hey! no problem.
> 
> the problem is that the `split` function isn't sufficient. For example, this scenario:
> 
> Screenshot 2025-12-09 at 14 27 36
> The `split` function will return in this case a single line string feature: Screenshot 2025-12-09 at 14 37 53 And using a midpoint on the whole line string may result in faulty cases.
> 
> The new function created, `splitLineIntoSegmentsOnPolygon`, splits the line string on the polygon - but makes sure to return individual segments.
> 
> ```tsx
>   const lineSegmentsOrLineString = lineSegments.features.length > 0 ? lineSegments.features : [turf.feature(linestring)];
> ```
> 
> is insufficient, as it assumes the `split` function splits in a certain way. For reference - the `LineIsContainedByPolygonWithSegmentOnBoundary` test fails when using this logic.

Ah ok interesting. Thanks for the reply @samuelarbibe!

smallsaucepan:
Thanks for doing that followup @Janickvw and @samuelarbibe

Issue #2808, opened 2025-01-11
nearestPointOnLine throws Error: coordinates must contain numbers


kamami:
I have a curious problem. I am using turf.nearestPointOnLine with two different lines. It is working perfectly fine with **line1**, but **line2** is throwing me this Error: `Error: coordinates must contain numbers`

Honestly I don't see the difference...

Here is my code: 

```
          const line1 = turf.lineString([
            [-77.031669, 38.878605],
            [-77.029609, 38.881946],
            [-77.020339, 38.884084],
            [-77.025661, 38.885821],
            [-77.021884, 38.889563],
            [-77.019824, 38.892368],
          ]);
          const line2 = turf.lineString([
            [10.57846, 49.8463959],
            [10.57846, 49.8468386],
            [10.57846, 49.8468386],
            [10.57846, 49.8468386],
            [10.57846, 49.8472814],
            [10.57846, 49.8472814],
          ]);

          const point = turf.point([-122.3102, 47.6634]);

          const nearest = turf.nearestPointOnLine(line2, point);  //throws error
          const nearest = turf.nearestPointOnLine(line1, point);  //works fine

```

twelch:
Looks similar to #2807.  I see the last two positions in your second line are redundant (the same). Can you try using the turf `cleanCoords` helper function and see if that addresses the error? I wonder if this is a new issue as of 7.2.0.

kamami:
Yes, that did the trick! I used this to find the nearest point next to the mouse cursor on the LineString. If the zoom was too low, it threw the error. Now with cleanCoords, it is working also on low zoom levels! 

```
    const handleMouseMove = (e) => {
      const mainLayerId = "route-ptv-segments-buffer"; // Stelle sicher, dass dies der korrekte Layer-Name ist

      if (!mapRef.current.getLayer(mainLayerId)) {
        return;
      }

      const features = mapRef.current.queryRenderedFeatures(e.point, {
        layers: [mainLayerId],
      });

      if (features.length > 0) {
        const mousePoint = turf.point([e.lngLat.lng, e.lngLat.lat]);
        let minDistance = Infinity;
        let closestFeature = null;
        let closestPoint = null;

        const feature = features[0];

        const line = turf.cleanCoords(feature);

        const nearest = turf.nearestPointOnLine(line, mousePoint);
        const distance = turf.distance(mousePoint, nearest);

        if (distance < minDistance) {
          minDistance = distance;
          closestFeature = feature;
[...]

PR #2849, opened 2025-03-02
turf-nearest-point-on-line: Fix issue #2808

insideGen:
Add a short circuit if the start position is equal to the stop position to avoid the “Error: coordinates must contain numbers” problem.

Fix issue #2808


twelch (approved):
@insideGen thank you for your contribution!  Test added using the original example.

Issue #2841, opened 2025-02-17
hexGrid says it accepts a multipolygon but the typing only allows polygon


SimonsUnikaNamn:
As you can see in the [docs](https://turfjs.org/docs/api/hexGrid), it specifically says that multipolygon is allowed, but the [typing](https://github.com/Turfjs/turf/blob/master/packages/turf-hex-grid/index.ts#L24) says otherwise

I have tested disregarding the typing and it seems to work

Will fix a PR for this

SimonsUnikaNamn:
Wasn't able to create a new branch, not sure how I am supposed to open a PR so will just depend on someone else fixing this minor thing

smallsaucepan:
Thanks @SimonsUnikaNamn.

> Wasn't able to create a new branch, not sure how I am supposed to open a PR

You will need to fork Turfjs/turf first, create and work in a branch in your forked repo, and then raise a PR to merge _that_ branch back in to turf.

> I have tested disregarding the typing and it seems to work

If you're still interested in doing a PR, thanks for offering to tidy this up.

Would you mind also confirming any calls hexGrid makes in turn accept MultiPolygons? Just to make sure all possible flows are "MultiPolygon safe". Maybe also add a simple test case as well?

smallsaucepan:
This was resolved by #2856

PR #2856, opened 2025-03-17
turf-hexgrid: fix issue #2841

SimonsUnikaNamn:
Summary of changes:
Allowed mask to be of type `Polygon | MultiPolygon`
Typing of mask does not allow for MultiPolygon while the documentation obviously states that this is allowed

I have read the `preparing a pull request` section. One unrelated test failed. so I hope my local setup is incorrect and the CI will let this pass


twelch (approved):
Changes look good @SimonsUnikaNamn, tests are passing.  I'm not seeing any calls within that need to be changed to accept Multipolygon too.

twelch:
Thanks for your contribution. This will become available in the next release

PR #2861, opened 2025-03-19
Improve NearestPoint type for @turf/nearest-point so that it keeps the properties of the points in the FeatureCollection

EmilJunker:

This PR improves the `NearestPoint` type and the type definition for the `nearestPoint` function.

Consider the following example in TypeScript:

```ts
const coordinatesList = [[0, 0], [1, 1], [0, 1]];
const pointsList = coordinatesList.map((coordinate, i) => {
    return turf.point(coordinate, {
        name: `point${i}`
    });
});
const pointsCollection = turf.featureCollection(pointsList);
const myNearestPoint = turf.nearestPoint(turf.point([2, 2]), pointsCollection);
```

With the old behaviour `myNearestPoint.properties.name` is of type `any` according to TypeScript.

With the changes in this PR, the type of `myNearestPoint.properties.name` is correctly inferred as `string`.

This change should be non-breaking.

### DIFF
--- a/packages/turf-nearest-point/index.ts
+++ b/packages/turf-nearest-point/index.ts
@@ -4,12 +4,11 @@ import { clone } from "@turf/clone";
 import { distance } from "@turf/distance";
 import { featureEach } from "@turf/meta";
 
-interface NearestPoint extends Feature<Point> {
+interface NearestPoint<P = GeoJsonProperties> extends Feature<Point> {
   properties: {
     featureIndex: number;
     distanceToPoint: number;
-    [key: string]: any;
-  };
+  } & P;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


twelch (changes requested):
Thanks for the PR @EmilJunker!  Your addition looks really close.  I see a couple of errors if you run `pnpm install` at the top-level of your local repo.

` index.ts(7,28): error TS2304: Cannot find name 'GeoJsonProperties'.`

It looks like if you add an import at the top
`import { Feature, FeatureCollection, Point, GeoJsonProperties } from "geojson";`

and additionally extend P as well as assigning a default value:
`function nearestPoint<P extends GeoJsonProperties = GeoJsonProperties>(`

Then you should be in business.  Could you make these changes and ping back for another check?

Be sure to run both `pnpm install` and `pnpm test` at the top level to make sure everything is passing per the [guidelines](https://github.com/Turfjs/turf/blob/master/docs/CONTRIBUTING.md#preparing-a-pull-request)

EmilJunker:
@twelch Thank you for your patience and guidance. I have opened a new pull request with the changes you requested: #2866

PR #2863, opened 2025-03-26
Pin the exact SHA1's of GitHub Actions we use

mfedderly:
This is as recommended in [the security guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), and as [demonstrated recently](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised) in an action we don't use.

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -17,17 +17,17 @@ jobs:
     runs-on: ${{matrix.platform}}
 
     steps:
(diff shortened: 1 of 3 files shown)


smallsaucepan:
Dig it. Thanks @mfedderly

smallsaucepan (approved):

PR #2866, opened 2025-04-07
Improve NearestPoint type for @turf/nearest-point so that it keeps the properties of the points in the FeatureCollection

EmilJunker:
This PR improves the NearestPoint type and the type definition for the nearestPoint function.

Consider the following example in TypeScript:

```ts
const coordinatesList = [[0, 0], [1, 1], [0, 1]];
const pointsList = coordinatesList.map((coordinate, i) => {
    return turf.point(coordinate, {
        name: `point${i}`
    });
});
const pointsCollection = turf.featureCollection(pointsList);
const myNearestPoint = turf.nearestPoint(turf.point([2, 2]), pointsCollection);
```

With the old behaviour myNearestPoint.properties.name is of type any according to TypeScript.

With the changes in this PR, the type of myNearestPoint.properties.name is correctly inferred as string.

This change should be non-breaking.

This PR replaces #2861.

- [x] Summary of the changes.

### DIFF
--- a/packages/turf-nearest-point/index.ts
+++ b/packages/turf-nearest-point/index.ts
@@ -1,15 +1,15 @@
-import { Feature, FeatureCollection, Point } from "geojson";
+import { Feature, FeatureCollection, GeoJsonProperties, Point } from "geojson";
 import { Coord, Units } from "@turf/helpers";
 import { clone } from "@turf/clone";
 import { distance } from "@turf/distance";
 import { featureEach } from "@turf/meta";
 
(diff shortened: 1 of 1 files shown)


twelch (approved):
Confirmed in light testing that user-provided properties are properly flowing through to the returned type, while also picking up featureIndex and distanceToPoint.

twelch:
Thank you @EmilJunker for improving this.  It will become available in the next release, timing TBD.

PR #2867, opened 2025-04-07
Compute multiFeatureLocation in addition to location property in @turf/nearest-point-on-line

### LINKED ISSUE #2753: nearestPointOnLine unexpectedly sums up lengths of multiLineStrings for location property
turf 7.1.0

Consider the following example:

```js
const multiLineString = turf.multiLineString([
  [[-122.3125, 47.6632], [-122.3102, 47.6646]],
  [[-122.3116, 47.6623], [-122.3091, 47.6636]]
]);

const pointA = turf.point([-122.3106, 47.6638], {name: 'A'});
const pointB = turf.point([-122.3102, 47.6634], {name: 'B'});

const nearestToA = turf.nearestPointOnLine(multiLineString, pointA, { units: "meters" });
const nearestToB = turf.nearestPointOnLine(multiLineString, pointB, { units: "meters" });

console.log(nearestToA.properties.location); // 150.293316294307
console.log(nearestToB.properties.location); // 389.9028953092881
```

<details>
<summary>Expand for GeoJSON FeatureCollection of all the above</summary>

```
{
    "type": "FeatureCollection",
    "features": [
        {
            "type": "Feature",
            "properties": {},
            "geometry": {
                "type": "MultiLineString",
                "coordinates": [
                    [
                        [
                            -122.3125,
                            47.6632
                        ],
                        [
                            -122.3102,
                            47.6646
                        ]
                    ],
                    [
                        [
                            -122.3116,
                            47.6623
                        ],
                        [
                            -122.3091,
                            47.6636
                        ]
                    ]
                ]
            }
        },
        {
            "type": "Feature",
            "properties": {
                "name": "A"
            },
            "geometry": {
                "type": "Point",
                "coordinates": [
                    -122.3106,
                    47.6638
                ]
            }
        },
        {
            "type": "Feature",
            "properties": {
                "name": "B"
            },
            "geometry": {
                "type": "Point",
                "coordinates": [
                    -122.3102,
                    47.6634
                ]
            }
        },
        {
            "type": "Feature",
            "properties": {
                "dist": 45.90981630012612,
                "multiFeatureIndex": 0,
                "location": 150.293316294307,
                "index": 0
            },
            "geometry": {
                "type": "Point",
                "coordinates": [
                    -122.31101109378298,
                    47.6641062907408
                ]
            }
        },
        {
            "type": "Feature",
            "properties": {
                "dist": 32.74106993655739,
                "multiFeatureIndex": 1,
                "location": 389.9028953092881,
                "index": 0
            },
            "geometry": {
                "type": "Point",
                "coordinates": [
                    -122.30993283032643,
                    47.66316692823026
                ]
            }
        }
    ]
}
```
</details>

`pointA` gets snapped onto a point on the northern line while `pointB` gets snapped onto a point on the southern line of the `multiLineString`. Both snapped points are located approximately 150 meters from the respective line start. However, while the `location` property of `nearestToA` has the expected value (`150.293316294307`), the `location` property of `nearestToB` is unexpectedly big (`389.9028953092881`). Apparently, turf has added the full length of the northern line to it which is not what I want.

With the current behavior, there is no easy way to know the location of the `nearestToB` point along the southern line (which should be a value around `150`). Please consider to add an extra property for this (maybe you could call it `multiFeatureLocation`) or add an option that allows the user to choose which behavior they want.

image

EmilJunker:
This adds a `multiFeatureLocation` property to the points returned by the `nearestPointOnLine` function. If there is just one LineString, then this will be the same as the existing `location` property. But if you pass a MultiLineString, this will tell you the distance along the line between the start of the MultiLineString where the closest point was found and the closest point itself. See #2753 for an example why this is useful.

Closes #2753.

- [x] Summary of the changes.

### DIFF
--- a/packages/turf-nearest-point-on-line/index.ts
+++ b/packages/turf-nearest-point-on-line/index.ts
@@ -18,7 +18,7 @@ import { getCoord, getCoords } from "@turf/invariant";
  * @param {Geometry|Feature<Point>|number[]} pt point to snap from
  * @param {Object} [options={}] Optional parameters
  * @param {string} [options.units='kilometers'] can be degrees, radians, miles, or kilometers
- * @returns {Feature<Point>} closest point on the `line` to `point`. The properties object will contain four values: `index`: closest point was found on nth line part, `multiFeatureIndex`: closest point was found on the nth line of the `MultiLineString`, `dist`: distance between pt and the closest point, `location`: distance along the line between start and the closest point.
+ * @returns {Feature<Point>} closest point on the `line` to `point`. The properties object will contain four values: `index`: closest point was found on nth line part, `multiFeatureIndex`: closest point was found on the nth line of the `MultiLineString`, `dist`: distance between pt and the closest point, `location`: distance along the line between start and the closest point, `multiFeatureLocation`: distance along the line between start of the `MultiLineString` where closest point was found and the closest point.
  * @example
  * var line = turf.lineString([
  *     [-77.031669, 38.878605],
@@ -47,6 +47,7 @@ function nearestPointOnLine<G extends LineString | MultiLineString>(
     index: number;
     multiFeatureIndex: number;
     location: number;
+    multiFeatureLocation: number;
     [key: string]: any;
   }
 > {
@@ -58,18 +59,33 @@ function nearestPointOnLine<G extends LineString | MultiLineString>(
 
   let closestPt: Feature<
     Point,
-    { dist: number; index: number; multiFeatureIndex: number; location: number }
+    {
+      dist: number;
+      index: number;
+      multiFeatureIndex: number;
+      location: number;
+      multiFeatureLocation: number;
+    }
   > = point([Infinity, Infinity], {
     dist: Infinity,
     index: -1,
     multiFeatureIndex: -1,
     location: -1,
+    multiFeatureLocation: -1,
   });
 
   let length = 0.0;
+  let multiFeatureLength = 0.0;
+  let currentMultiFeatureIndex = -1;
   flattenEach(
     lines,
     function (line: any, _featureIndex: number, multiFeatureIndex: number) {
+      //reset multiFeatureLength at each changed multiFeatureIndex
+      if (currentMultiFeatureIndex !== multiFeatureIndex) {
+        currentMultiFeatureIndex = multiFeatureIndex;
+        multiFeatureLength = 0.0;
+      }
+
       const coords: any = getCoords(line);
 
       for (let i = 0; i < coords.length - 1; i++) {
@@ -107,15 +123,22 @@ function nearestPointOnLine<G extends LineString | MultiLineString>(
         let intersectPt:
           | Feature<
               Point,
-              { dist: number; multiFeatureIndex: number; location: number }
+              {
+                dist: number;
+                multiFeatureIndex: number;
+                location: number;
+                multiFeatureLocation: number;
+              }
             >
           | undefined;
 
         if (intersectPos) {
+          const lineLocationDist = distance(start, intersectPos, options);
           intersectPt = point(intersectPos, {
             dist: distance(pt, intersectPos, options),
             multiFeatureIndex: multiFeatureIndex,
-            location: length + distance(start, intersectPos, options),
+            location: length + lineLocationDist,
+            multiFeatureLocation: multiFeatureLength + lineLocationDist,
           });
         }
 
@@ -134,8 +157,9 @@ function nearestPointOnLine<G extends LineString | MultiLineString>(
           };
         }
 
-        // update length
+        // update length and multiFeatureLength
         length += sectionLength;
+        multiFeatureLength += sectionLength;
       }
     }
   );
--- a/packages/turf-nearest-point-on-line/test.ts
+++ b/packages/turf-nearest-point-on-line/test.ts
@@ -398,6 +398,37 @@ test("turf-nearest-point-on-line -- multifeature index", (t) => {
   t.end();
 });
 
+test("turf-nearest-point-on-line -- issue 2753 multifeature location", (t) => {
+  const multiLine = multiLineString([
+    [
+      [-122.3125, 47.6632],
+      [-122.3102, 47.6646],
+    ],
+    [
+      [-122.3116, 47.6623],
+      [-122.3091, 47.6636],
+    ],
+  ]);
+
+  const ptA = point([-122.3106, 47.6638], { name: "A" });
+  const ptB = point([-122.3102, 47.6634], { name: "B" });
+
+  const nearestToA = nearestPointOnLine(multiLine, ptA, { units: "meters" });
+  const nearestToB = nearestPointOnLine(multiLine, ptB, { units: "meters" });
+
+  t.equal(
+    Number(nearestToA.properties.multiFeatureLocation.toFixed(6)),
+    150.293316,
+    "nearestToA multiFeatureLocation"
+  );
+  t.equal(
+    Number(nearestToB.properties.multiFeatureLocation.toFixed(6)),
+    157.736676,
+    "nearestToB multiFeatureLocation"
+  );
+  t.end();
+});
+
 test("turf-nearest-point-on-line -- issue 1514", (t) => {
   const pt = point([-40.01, 56]);
   const line = lineString([


EmilJunker:
@smallsaucepan I just resolved a merge conflict so this PR can be cleanly merged again. I would appreciate if this was included in the next release. Please let me know if anything is missing from it or holding it back.

EmilJunker:
@smallsaucepan I have again resolved any merge conflicts of this PR. It would be great if this feature could be included in the upcoming 7.3.0 release.

smallsaucepan:
Oh @EmilJunker! Missed the release by just a few minutes. Will review and make sure we get into the next version 👍

smallsaucepan:
Have been reviewing your PR (looks good!) and I think I gave you bad advice earlier. I'm not sure any more multiFeatureLOCATION is the best choice. Other functions use _distance_ or a variation when describing how far along a line you've gone (see _along_).

I started wondering if any other existing properties don't make sense ...

- index: closest point was found on nth line part
- multiFeatureIndex: closest point was found on the nth line of the MultiLineString
- dist: distance between pt and the closest point
- location: distance along the line between start and the closest point

If we could start fresh, do you think the options below would be more clear or meaningful?

- multiFeatureIndex: **unchanged** closest point was found on the nth line of the MultiLineString
- index: **unchanged** closest point was found on nth line part
- multiFeatureDistance: **renamed** what location currently provides, distance from the absolute start of the multilinestring
- distance: **new** your parameter, distance from the start of the current linestring
- ptDistance: **renamed** what dist currently provides, distance to the input point

Trying to establish a plan before we go further, and would appreciate your thoughts on it.

EmilJunker:
Thank you for the review. I agree with you that the current naming of the properties doesn't really make sense and that this would be a good opportunity to fix that. The fact that the _location_ describes the distance along the line while _dist_ is the distance to the closest point really confused me when I started using the nearestPointOnLine function.

I also mostly agree with your suggested new names. I also agree with you that (going by my current implementation) multiFeatureLocation and location should basically switch roles since the former is probably more useful in practice than the latter.

However, I think I would prefer "pointDistance" over "ptDistance" as a new name for "dist", because I'm generally not too fond of acronyms in source code. We aren't writing C++ here 😉 .

Maybe the function parameter name could also be changed from "pt" to "point" while we are at it. But turf functions seem to be a bit inconsistent in that regard, e.g. [pointToPolygonDistance](https://turfjs.org/docs/api/pointToPolygonDistance) has a "point" param while [pointToLineDistance](https://turfjs.org/docs/api/pointToLineDistance) has a "pt" param.

To make it even clearer that the distance is measured along the line, we could also call the other properties "lineDistance" and "multiFeatureLineDistance".

In conclusion, I think the closest point should have the following properties:

- multiFeatureIndex: closest point was found on the nth line of the MultiLineString
- index: closest point was found on the nth line part
- multiFeatureLineDistance: distance along the line from the absolute start of the MultiLineString
- lineDistance: distance along the line from the start of the line where the closest point was found
- pointDistance: distance to the input point

I'd be happy to make the necessary naming changes as part of this pull request. My only concern is that this would be a breaking change, thus possibly deferring the release of this feature. Can you guarantee that it would still be included in the next turf release even with this breaking change?

smallsaucepan:
Oh it would totally be a breaking change. What we can do here though is come up with an end goal, and the plan for a smooth transition. This PR could be the first non-breaking step, and then we deprecate and phase out at a major release with later PRs. From the looks of it we'd be able to introduce new properties without changing the behaviour of existing ones.

Going back to a totally clean slate, what does each parameter really refer to?

- multiFeatureIndex is the **lineStringIndex** (the index into the possible MultiLineString input, always 0 for a simple LineString)
- index is the **segmentIndex** of the linestring feature (referred to by featureIndex)
- location is the **multiLineStringDistance** (same as below for a single LineString input)
- your new prop is the **lineStringDistance** (or the distance along the segment identified by lineStringIndex)
- another theoretical prop could then be **segmentDistance** (or the distance along the segment identified by segmentIndex. Not suggesting adding it, just that the naming scheme leaves room for it)

This naming is a departure from the way it's been done so far, though here's my thinking - the index is named after the type of item it points to, rather than the name of the container.

team of people -> personIndex
bag of apples -> appleIndex
multiLineString of lineStrings -> lineStringIndex

To visualise:

issue-2867-terminology

Agree it would be good to have some consistency around pt / point too. Considerations might be elements where we pass in multiple points, and possible clashes with the point() function. Nothing else I can think of better describes the intent here though - target point? Origin point?

EmilJunker:
I see you've put a lot of thought into this. I like your idea to name the index after the type of item instead of the container. At the moment, the property names are a bit confusing because the function accepts both LineStrings and MultiLineStrings, so some of the properties are called multiXXX. But this is of course confusing for users who call the function with just a LineString and then still see multiXXX properties. Your suggestions for lineStringIndex and segmentIndex would definitely help to clear up that confusion because they make sense regardless of how many LineStrings there are (one or many).

By the same logic, maybe the old _location_ property could be renamed to totalLineStringDistance instead of multiLineStringDistance. Not sure, just an idea. Do you think that would be better? 

I'm not strongly advocating for the parameter to be called point instead of pt. My main point (no pun intended) was that I think pointDistance would be a better name than ptDistance for the old _dist_ property.

I would be happy to adapt this pull request in any way you see fit. Just tell me how to proceed and I'll do my best to get it done quickly.

smallsaucepan:
Thanks @EmilJunker. Proposing the mashup below. Might ask @JamesLMilner and @mfedderly to review as well.

tl;dr feel we could do better naming the properties in the return object from nearestPointOnLine. Suggesting we take the opportunity with this PR to introduce _new_ property names, and deprecate the old ones somewhere down the track. Existing properties would continue to work (be duplicated in the response) until at least the next major release.

| Current | New | Description |
|-|-|-|
|multiFeatureIndex| lineStringIndex| closest point was found on the nth line of the MultiLineString|
|index|segmentIndex|closest point was found on nth line segment of above|
|location|totalDistance |what location currently provides, distance from the absolute start of the multilinestring|
|<blank>|lineDistance|new parameter, distance from the start of the current linestring|
|<blank>|segmentDistance|another new parameter if we want to, distance from the start of the current segment|
|dist|pointDistance|distance from the nearest point to the original input point|

There's no overlap so it wouldn't change the behaviour of existing code. Do these seem internally consistent, and also in line with terminology in other functions?

mfedderly:
Re `pt` vs `point`: If I had to guess, we probably just never considered parameter names as part of the public api and so they wound up inconsistent. Changing them won't even be an API break unless its a key inside an object, so that's low regret to make those a bit more explicit, since they are part of the way we document things.

Re return property names: I agree the existing ones are inconsistent, and I've found them confusing. The [proposed names](https://github.com/Turfjs/turf/pull/2867#issuecomment-3588515490) seem reasonable to me. If I can get greedy I'd love for the implementation to change to putting the proposed new properties into the object, and then a second block below that copying to the old names, with a comment above that block saying that it is deprecated.

EmilJunker:
@smallsaucepan @mfedderly Thank you for your input. I have implemented and pushed the necessary changes. Everything looks good from my side. Please let me know if this is what you had in mind.

I added the six new properties as discussed above, including segmentDistance. The old properties remain unchanged for now, but they can be easily removed later.

I also renamed the `pt` parameter to `inputPoint`.

EmilJunker:
If necessary, I could explicitly mark the old properties as deprecated using TypeScript's `@deprecated` tag:

```ts
// deprecated properties START
/** @deprecated use `lineStringIndex` instead */
multiFeatureIndex: number;
/** @deprecated use `segmentIndex` instead */
index: number;
/** @deprecated use `totalDistance` instead */
location: number;
/** @deprecated use `pointDistance` instead */
dist: number;
// deprecated properties END
```

Compatible editors like VS Code would then cross out those properties in the code suggestions:

{8F2AA8F6-7F3A-4EA4-AD15-77721612EE0F}

And wherever one of those properties is accessed in the source code, it would also be crossed out:

{740C1019-5C6B-4D17-B917-E3B9D237A9CF}

But that's probably something for a future release and should not be part of this pull request.

mfedderly:
This all lgtm, I'd be in favor of adding the `@deprecated` tags now. Can you also run `pnpm docs` to regenerate that README.md file from the new jsdoc that got updated? I thought that was supposed to happen automatically but perhaps you didn't get the pre-commit hooks installed by default.

EmilJunker:
Alright, I added the deprecation annotations and updated the README.

Generating the README with `pnpm run docs` didn't quite work for me though. I had to make some edits to the generate-readmes script to make it even do something. And afterwards I had to manually put some line breaks back into the README that got removed for some reason.

Furthermore, it tried to add weird sections to the README based on the `/** @deprecated */` comments in the code. I also manually removed them, I hope that's not a problem.

Anyway, it's all good to go now from my side.

smallsaucepan:
Thanks Emil. There are a couple of conflicts due to another PR I merged so I'll clean those up (only fair). Will take a look at the documentation README generation too. If I can I'll push to your PR branch (this PR) and then merge once it's building.

smallsaucepan:
Conflicts are resolved, though not having much luck finding a way to generate the deprecated tag on individual properties of a return object. To start with we need something like this in the doc comments

```
 * @returns {Feature<Point, GeoJsonProperties & { lineStringIndex: number; ... } 
```
but the documentation generator barfs on that even before we add deprecated tags.

How about this added to the description instead?

> We have begun the process of migrating to different return properties for this function. The new properties we recommend using as of v7.4 are:
> - lineStringIndex - point was found on the nth LineString of an input MultiLineString. Previously `multiFeatureIndex`
> - segmentIndex - point was found on the nth segment of the above LineString
> - totalDistance - distance from the start of the input MultiLineString. Previously `location`
> - lineDistance - distance from the start of the relevant LineString
> - segmentDistance - distance from the start of the relevant segment
> - pointDistance - distance between found point is from input reference point. Previously `dist`
> 
> multiFeatureIndex, location, and dist will continue to work as they do currently until at least the next major release.

EmilJunker:
> How about this added to the description instead?

Yes, I think that would work. The important thing is that it should be easy to remove both the old properties from the source code and the migration hint from the documentation when the time is ready. Thus, a single block of text in the description would be ideal.

Oh and don't forget to mention that `segmentIndex` replaces the old `index` property, and that the latter will also be removed in the future.

packages/turf-nearest-point-on-line/index.ts
```diff
@@ -17,6 +17,17 @@ import { getCoord, getCoords } from "@turf/invariant";
...
+ * ⚠️ We have begun the process of migrating to different return properties for
+ * this function. The new properties we recommend using as of v7.4 are:
+ * - lineStringIndex - point was found on the nth LineString of an input MultiLineString. Previously `multiFeatureIndex`
+ * - segmentIndex - point was found on the nth segment of the above LineString. Previously `index`
+ * - totalDistance - distance from the start of the overall MultiLineString. Previously `location`
+ * - lineDistance - distance from the start of the relevant LineString
+ * - segmentDistance - distance from the start of the relevant segment
+ * - pointDistance - distance between found point is from input reference point. Previously `dist`
+ *
+ * multiFeatureIndex, location, and dist continue to work as previously until at least the next major release.
```
EmilJunker: and also `index` ;)

packages/turf-nearest-point-on-line/index.ts
```diff
@@ -53,16 +64,10 @@ function nearestPointOnLine<G extends LineString | MultiLineString>(
     lineDistance: number;
     segmentDistance: number;
     pointDistance: number;
-    // deprecated properties START
```
EmilJunker: Unless it's causing problems, I'd be in favor of keeping the "deprecated properties START" and "deprecated properties END" comments to let contributors know which code sections will be removed at some point.

packages/turf-nearest-point-on-line/index.ts
smallsaucepan: Oops, yes. Thank you

packages/turf-nearest-point-on-line/index.ts
smallsaucepan: Also done 👍

PR #2871, opened 2025-04-22
fix nearestPointOnLine for points far away from the segment endpoints

### LINKED ISSUE #2870: nearestPointOnLine can return wrong endpoint when the target point is very far from the line
Given a `LineString` like this, with one endpoint in the North Atlantic ocean and one in the South Pacific:
```json
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "LineString",
    "coordinates": [
      [-96.80974, -58.32062],
      [-50.2464, 38.36643]
    ]
  }
}
```

and a target `Point` like this, in Europe,
```json
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Point",
    "coordinates": [13.6023, 46.50646]
  }
}
```

the `nearestPointOnLine` function should return `[-50.2464, 38.36643]`, the segment endpoint in the North Atlantic ocean, as it is much closer than the other endpoint (5173km vs 15348km), and no other point on the segment is closer.

However, instead, turf returns a point very close to (withing floating-point-precision error of) the South Pacific endpoint.

mlc:
Fixes #2870; see that ticket for a description of the observed bug.

This patch fixes the described bug by adjusting how `nearestPointOnSegment` picks an endpoint when it determines that the nearest intersection of the great circles is not on the segment. Rather than doing work with angles, it just computes the distance to the two endpoints and picks the closer one.

I think this is safe—it does not break any existing tests—and should not lead to a performance reduction (there are just as many calls to `distance` as there were before, and some vector math has been removed), but the algorithms here are a little bit outside of my wheelhouse, so I would certainly appreciate any feedback if there is something wrong. A more natural way of fixing the issue might be to correctly choose the right value of `I` in the section above the one I modified, but I can not quite figure out how to make that work.

Thanks so much for turf and for your attention to this!

- [x] Summary of the changes.
- [ ] Heads up if this is a breaking change.

### DIFF
--- a/packages/turf-nearest-point-on-line/index.ts
+++ b/packages/turf-nearest-point-on-line/index.ts
@@ -250,13 +250,10 @@ function nearestPointOnSegment(
   // If angle AI or BI is greater than angleAB, I lies on the circle *beyond* A
   // and B so use the closest of A or B as the intersection
   if (angle(A, I) > angleAB || angle(B, I) > angleAB) {
-    if (
-      distance(vectorToLngLat(I), vectorToLngLat(A)) <=
-      distance(vectorToLngLat(I), vectorToLngLat(B))
-    ) {
-      return [vectorToLngLat(A), true, false];
+    if (distance(posC, posA) <= distance(posC, posB)) {
+      return [posA, true, false];
     } else {
-      return [vectorToLngLat(B), false, true];
+      return [posB, false, true];
     }
   }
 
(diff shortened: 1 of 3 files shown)


bratter:
Ah sorry @mlc - I missed this when creating #2940. I had the same issue but took a bit of a different approach to fixing and also found a few other edge cases in the process. #2940 will conflict with this but fixes additional issues and also speeds up the algorithm. Happy to discuss.

mlc:
@bratter I don't feel strongly about my solution—if yours is better, I'm all for it.

smallsaucepan:
Hi @mlc and @bratter. #2940 has been merged so I will close this one. Thanks heaps @mlc for putting this PR together.

PR #2879, opened 2025-05-04
Updated build tools - pnpm, lerna, typescript, tsup, etc housekeeping

smallsaucepan:
Semi-annual build tools housekeeping. Updated typescript, lerna, pnpm, etc to their latest versions. Should not be any changes to any package's functional dependencies.


mfedderly (approved):
Thank you!

Issue #2893, opened 2025-06-02
randomLineString does not take max_rotation option into account


claustres:
It seems to be related to the fact that the previous angle variable is actually a [local variable](https://github.com/Turfjs/turf/blob/master/packages/turf-random/index.ts#L237) of the inner loop over the vertices so that it is incorrectly reset at each vertex. Here is a suggested modification of the code, which also appears to be more simple:
```
for (var i = 0; i < count; i++) {
    var startingPoint = turf.randomPosition(bbox);
    var vertices = [startingPoint];
    var priorAngle = Math.random() * 2 * Math.PI;
    for (var j = 0; j < num_vertices - 1; j++) {
        var angle = priorAngle + (Math.random() - 0.5) * max_rotation * 2;
        priorAngle = angle;
        var distance = Math.random() * max_length;
        vertices.push([
            vertices[j][0] + distance * Math.cos(angle),
            vertices[j][1] + distance * Math.sin(angle),
        ]);
    }
    features.push(turf.lineString(vertices));
}
```

I would be pleased to try to create a PR if you think this fix is correct, althought it might take some time to me for a first contribution to study how to write a relevent test, update the build, etc.

smallsaucepan:
Thanks @claustres. If it works better and reads better, more than happy for you to put together a PR with your revisions. If you haven't already, perhaps first take a look through Turf's bug history to make sure it's not that way for a reason.

More than happy for you to submit a PR. The [contributing guide](https://github.com/Turfjs/turf/blob/05494d856f2a9e1fccc292268b87eef4284fbc32/docs/CONTRIBUTING.md) might help you get started. Sing out if you have any questions!

Also, if you could attach some example geojson of the incorrect behaviour to this issue that might be of use to future visitors.

bratter:
@smallsaucepan is your intent to merge #2896? I'm happy to work on #2894, but that will conflict with this.

smallsaucepan:
Sounds good thanks @bratter. Yes aim to merge #2896, so will take a look.

Alternately @claustres would you be happy to work with @bratter to have your PR address both issues?

claustres:
Hello, sorry for the delay, I would be happy to help and can update my PR accordingly. Is it only to rename all optional parameters from snake case to lower camel case or is there anything else to perform ?

smallsaucepan:
[...]

PR #2896, opened 2025-06-06
fix: randomLineString does not take max_rotation option into account (#2893)

claustres:
Fix for issue https://github.com/Turfjs/turf/issues/2893.

The code has been changed from what has been suggested first in comment to use internal turfjs functions like `destination`.

Tests have been added for the `randomLineString` function as they did not exist previously.

For coherence with the `randomPolygon` function of the same package I suggest that the `max_rotation` option should be given in decimal degrees as well. This is also more coherent with the internal use of `destination` requiring a bearing angle in degrees.


packages/turf-random/index.ts
```diff
@@ -233,20 +234,18 @@ function randomLineString(
...
-              (vertices[j][1] - vertices[j - 1][1]) /
-                (vertices[j][0] - vertices[j - 1][0])
-            );
-      const angle = priorAngle + (Math.random() - 0.5) * max_rotation * 2;
-      const distance = Math.random() * max_length;
-      vertices.push([
-        vertices[j][0] + distance * Math.cos(angle),
-        vertices[j][1] + distance * Math.sin(angle),
-      ]);
+    var priorBearing = Math.random() * 360;
```
smallsaucepan: Reckon we would prefer sticking with let and const. Unless there is a specific use for var here?

mfedderly (commented):
This PR is pretty hard to review because it intermixes the options renaming with the bugfix. Is there any chance you'd be able to rework this to just the fix (for v7.4), and then a second version that does the API renames (for v8)? The first commit seems to be most of the changes for the bugfix, but some of the later commits also seems to change some of the behavior.

Issue #2809, opened 2025-01-13
Error: coordinates must contain numbers


Felankia:
Version: 7.2.0

Code:
`
let line = {}; //GeoJSON contant in [https://github.com/user-attachments/files/18392267/geojson.txt]
line = turf.cleanCoords(line);
let grid = turf.polygon([[[105.82001,38.30005],[105.87001,38.30005],[105.87001,38.350049999999996],[105.82001,38.350049999999996],[105.82001,38.30005]]])
turf.lineSplit(line, grid);`

Stacktrace:
`Error: coordinates must contain numbers
    at point (../node_modules/@turf/helpers/dist/cjs/index.cjs:78:11)
    at ../node_modules/@turf/nearest-point-on-line/dist/cjs/index.cjs:69:40
    at../node_modules/@turf/meta/dist/cjs/index.cjs:255:13
    at geomEach (../node_modules/@turf/meta/dist/cjs/index.cjs:197:15)
    at flattenEach (../node_modules/@turf/meta/dist/cjs/index.cjs:248:3)
    at nearestPointOnLine (../node_modules/@turf/nearest-point-on-line/dist/cjs/index.cjs:42:21)
    at ../node_modules/@turf/turf/node_modules/@turf/line-split/dist/cjs/index.cjs:112:53
    at featureEach (../node_modules/@turf/meta/dist/cjs/index.cjs:148:11)
    at findClosestFeature (../node_modules/@turf/turf/node_modules/@turf/line-split/dist/cjs/index.cjs:111:21)
    at ../node_modules/@turf/turf/node_modules/@turf/line-split/dist/cjs/index.cjs:56:27`

petrot:
Same problem here with the "nearestPointOnLine" function.

paulomesquita-wsn:
I discovered that the problem is related to a line with two consecutive equal coordinates

smallsaucepan:
Likely to be closely related to #2807

maciejgoscinski-latrace:
Same issue for me recently, and I arrived at the same conclusion as @paulomesquita-wsn - this is what happens when two adjacent coords are identical.

Maybe there's a good reason for the function to require deduplicated coords on the input, but in such case it should print a clearer error message, and document this requirement.

Duct-tape-fixed on our end by sanitizing the input geojsons to filter out duplicate coords.

colinperepelken:
[...]

Issue #2807, opened 2025-01-09
pointToPolygonDistance if polygon's last points are redundant


mwenko:
Please provide the following when reporting an issue:

- [x] Description of the problem, and how it differs from what you expected.
- [x] Version of Turf you are using, and of any other relevant software.
- [x] GeoJSON data as a [gist file](https://gist.github.com/) or [geojson.io](https://geojson.io/) (filename extension must be `.geojson`). Simple reproducible examples are preferrable.
- [x] Snippet of source code for complex examples using [jsfiddle](https://jsfiddle.net/).
- [x] Confirmation this issue hasn't already been reported, or is resolved and just hasn't been released yet.

### Problem

**Version: 7.2.0**

**pointToPolygonDistance**
I get an error that says `coordinates must contain numbers`. It definitely has something to do with the redundant points at the end of the polygon. (Look at the last 2 entries of the polygon, they are equal). If I remove them one of them, the method works.

When debugging, both coordinates appear to be `NaN`:
<img width="939" alt="image" src="https://github.com/user-attachments/assets/f6e4188d-68d2-4dbf-9a8f-8538ff01d40f" />

**Code:**
```
// point & polygon can be used from GeoJSON below
const distance = turf.pointToPolygonDistance(point, polygon, {
                units: "meters",
            });
```

**Stacktrace**: 
```
at point (../node_modules/@turf/helpers/index.ts:269:11)
      at ../node_modules/@turf/nearest-point-on-line/index.ts:113:25
      at ../node_modules/@turf/meta/index.js:748:11
      at geomEach (../node_modules/@turf/meta/index.js:597:13)
      at flattenEach (../node_modules/@turf/meta/index.js:739:3)
      at nearestPointOnLine (../node_modules/@turf/nearest-point-on-line/index.ts:70:3)
      at distanceToSegment (../node_modules/@turf/point-to-line-distance/index.ts:113:3)
      at ../node_modules/@turf/point-to-line-distance/index.ts:78:15
      at ../node_modules/@turf/meta/index.js:941:13
      at coordEach (../node_modules/@turf/meta/index.js:119:15)
      at ../node_modules/@turf/meta/index.js:913:7
      at ../node_modules/@turf/meta/index.js:748:11
      at geomEach (../node_modules/@turf/meta/index.js:597:13)
      at flattenEach (../node_modules/@turf/meta/index.js:739:3)
      at segmentEach (../node_modules/@turf/meta/index.js:898:3)
[...]

PR #2903, opened 2025-06-12
@turf/nearest-point-on-line TESTS: add test for duplicated points on line string affecting v7.2.0

alacret:
### Summary of Changes

- Adding unit tests that fail on version 7.2.0
- This test will prevent regressions affecting this package

### Issues related:

- https://github.com/Turfjs/turf/issues/2809
- https://github.com/Turfjs/turf/issues/2807


smallsaucepan (approved):
Thanks for contributing this @alacret 👍

PR #2916, opened 2025-07-08
boolean-valid - Add support for GeometryCollection geometries

ltkum:
-  Added a check to validate Geometry Collections in the booleanValid function
-  We enforce the recommendations of not nesting Geometry Collections, as it will allow us to avoid infinite potential infinite validation loops
- Added tests files to ensure this work as expected.


packages/turf-boolean-valid/index.ts
```diff
@@ -83,6 +83,31 @@ function booleanValid(feature: Feature<any> | Geometry) {
...
+      }
+
+      // To avoid situations where Geometry Collections nested within one another
+      // could lead to a situation where there is an infinite validation loop,
+      // we enforce the GeoJSON RFC recommendations
+      // (https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.8)
+      // which states that we should not have Nested Geometry Collections
+      return (
+        Array.isArray(geom.geometries) &&
+        geom.geometries.length > 0 &&
```
mfedderly, line 107: Contrary to RFC7946's "It is possible for this array to be empty.", the Simple Features Specification doc specifies "A GeometryCollection is a geometry that is a collection of 1 or more geometries."

Because this package is doing validation against the SFS, I believe this is the correct check to do here.

mfedderly (approved):
Thank you!

PR #2924, opened 2025-07-24
Use newer @types/rbush that fixes the need for the reexport shims

mfedderly:
Also flowing from https://github.com/Turfjs/turf/pull/2921, I noticed that the rbush re-export shims aren't necessary now that the rbush types are more correct.


JamesLMilner (approved):
Assuming this doesn't break the build and all works as expected then this all looks good to me

Issue #2081, opened 2021-05-14
Use point FeatureCollection can not create isolines


shizukanaqun:
Please provide the following when reporting an issue:

- [6.3.0 ] The version of Turf you are using, and any other relevant versions.
- [<script src="https://gist.github.com/shizukanaqun/a9abfc5644566c3bfe61487c111c5fe0.js"></script>] GeoJSON data as a [gist file](https://gist.github.com/) or [geojson.io](http://geojson.io/) (filename extension must be `.geojson`).
- [ ] Snippet of source code or for complex examples use [jsfiddle](https://jsfiddle.net/).

shizukanaqun:
var epsg_84 = 'PROJCS["WGS_1984_UTM_Zone_51N",GEOGCS["GCS_WGS_1984",DATUM["D_WGS_1984",SPHEROID["WGS_1984",6378137.0,298.257223563]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Transverse_Mercator"],PARAMETER["False_Easting",500000.0],PARAMETER["False_Northing",0.0],PARAMETER["Central_Meridian",123.0],PARAMETER["Scale_Factor",0.9996],PARAMETER["Latitude_Of_Origin",0.0],UNIT["Meter",1.0]]';
                        var arr = succ.split("\r\n");
                        var newcoord = proj4(epsg_84, [121.37622190181578, 36.708180574436966]);
                        //var newcoord = proj4('EPSG:3857', [121.37622190181578, 36.708180574436966]);
                        for (var i = 0; i < arr.length; i++) {
                            var newline = arr[i];
                            var resultlist = newline.split(",");
                            for (var c = 0; c < resultlist.length - 1; c++) {
                                var x = newcoord[0] - (40 - (c)) * 250;
                                var y = newcoord[1] + (40 - (i)) * 250;
                                //var coord4326 = proj4('EPSG:3857', 'EPSG:4326', [x, y]);
                                var coord4326 =  proj4(epsg_84, 'EPSG:4326', [x, y]);
                                var gaussvalue = parseFloat(resultlist[c]);
                                pointFeatures.push(turf.point([coord4326[0], coord4326[1]], { gaussResult: gaussvalue }));
                                //console.log('x:' + coord4326[0] + '----y:' + coord4326[1] + '----v:' + gaussvalue);
                            }
                        }
                        var collection = turf.featureCollection(pointFeatures);
[...]

PR #2927, opened 2025-07-29
@turf/isobands and @turf/isolines check pointGrid input is actually gridded already

### LINKED ISSUE #2509: @turf/isobands-In some cases, calculations can lead to dead cycles
version [6.5]
Here is an example data, and I am preparing to convert the grid point to isoLine

when i use ⬇️
gridToIsoGeojson(gridJsonToGrid(gridObj), valueList, colorList, smoothIterations, simplifyTolerance)

in isobands.js >> isoBands >> BandGrid2AreaPaths

In this method, you will fall into a dead cycle (do while)
Currently, there is only a small amount of data that may have issues

I'm not very good at math😭I just located a certain value in this method to calculate the polygon
isoLineData1 >> success
isoLineData3 >> error
[isoLineData3.json.zip](https://github.com/Turfjs/turf/files/12854430/isoLineData3.json.zip)
[isoLineData1.json.zip](https://github.com/Turfjs/turf/files/12854502/isoLineData1.json.zip)

mfedderly:
We were seeing some issues filed around these methods not working as expected. After looking at the data it appears that the pointGrid input is not always evenly spaced gridded points. When this happens `gridToMatrix()` results in an unusable result. The matrix is either not large enough to ever make any cells for marchingsquares, or of inconsistent dimensions due to not reusing the same latitudes across points (even just a slight discrepancy in the values will cause an issue).

This adds explicit checks for this condition, and slightly clarifies the docs on what we're expecting to be passed in. I'm open to improvements in the wording here for clarity.

Fixes #2509, #2081


packages/turf-isobands/index.ts
```diff
@@ -35,7 +35,7 @@ type GroupedRings =
  * value breaks and generates filled contour isobands.
  *
  * @function
- * @param {FeatureCollection<Point>} pointGrid input points - must be square or rectangular
+ * @param {FeatureCollection<Point>} pointGrid input points - must be square or rectangular and already gridded
```
smallsaucepan: Is there a clearer way to express this? Maybe cribbing from the comment where we actually do the check:

> must be square or rectangular, and already gridded. That is, have consistent x and y dimensions, and be at least 2x2 in size.

packages/turf-isolines/index.ts
```diff
@@ -69,6 +69,19 @@ function isolines(
 
   // Isoline methods
   const matrix = gridToMatrix(pointGrid, { zProperty: zProperty, flip: true });
+
```
smallsaucepan, line 72: We do really need to find a way to share common code internally 😅

PR #2937, opened 2025-10-10
[turf/buffer] Expose BufferOp's endCapStyle overload for line-type buffers

plymer:
Hello all, this is my first PR and contribution to an open-source project!

I love this library and I've been using it in a big project at work. One thing that always bugged me was that there was seemingly no way to opt out of the default rounded ends on line-type buffers. I've written my own line-buffer logic and I've tried using multiple turf line offsets to build 'buffered' geometries, all with the goal of achieving flat end caps that don't overshoot the end vertices' extents.  I've struggled with reprojection issues, and I've always felt that `turf/buffer` was missing something.

I did some digging, and it turns out that the JSTS library's BufferOp actually supports specifying and `endCapStyle` internally, but it was never exposed as an option in `turf/buffer` for some reason.

This PR introduces an optional `endCapStyle` buffer option that defaults its behaviour to the current, always-round implementation. There are no breaking changes with this PR, and I wrote what I think are a fairly comprehensive set of tests to make sure that is the case.

Please point out if I've overlooked anything but I think this is a pretty simple enhancement without any backwards compatibility issues.

Thanks for taking a look!


mfedderly:
I'm working through the open PRs and I'm going to close this PR for now.

I've added https://github.com/Turfjs/turf/issues/639 to the v8 milestone to keep tracking this and to remind myself to add this functionality to the buffer API after we know that the new buffer implementation works for everyone and won't need to get pulled back out.

PR #2951, opened 2025-11-16
Fixed a bug in nearestPointOnLine causing duplicate points to be returned from lineSlice

### LINKED ISSUE #2946: lineSlice returns line with duplicate coordinates
Using the following arguments, you get the result specified, which contains a duplicate coordinate. In turn, if you call `lineSlice` again with the result, it throws an error like that reported in https://github.com/Turfjs/turf/issues/2939 and related issues.
```
{
        startPt: {
          type: 'Feature',
          properties: {},
          geometry: { type: 'Point', coordinates: [1.9999999999999998, 0] }
        },
        endPt: {
          type: 'Feature',
          properties: {},
          geometry: { type: 'Point', coordinates: [1.9999999999999991, 2] }
        },
        line: {
          type: 'Feature',
          properties: {},
          geometry: {
            type: 'LineString',
            coordinates: [
              [2.9998475636392152, 0.00015236670332290198],
              [2, -1],
              [1.9999999999999991, 2]
            ]
          }
        },
        result: {
          type: 'Feature',
          properties: {},
          geometry: {
            type: 'LineString',
            coordinates: [
              [1.9999999999999998, 6.294812763123119e-32],
              [1.9999999999999991, 2],
              [1.9999999999999991, 2]
            ]
          }
        }
      }
```

smallsaucepan:
Fixed a bug in nearestPointOnLine where index of returned point would sometimes refer to a non-existent line string segment. This was having a flow on effect in lineSlice where an additional duplicate point would be added as a result.

Fix adds a check to make sure nearestPointOnLine isn't returning the index of a line segment that doesn't exist.

Resolves #2946


EmilJunker:
@smallsaucepan @mfedderly Hi, I hope you are doing well. I just wanted to let you know that the change made in this PR broke my code and gave me quite a headache. 😅 I was really looking forward to using the new turf 7.3.2 version because it included my new nearestPointOnLine properties from #2867. You can imagine my frustration when the upgrade broke my code and I realized that it's because the new version also includes this unexpected change.

It's obviously not a huge problem and I have since implemented a workaround. But I would appreciate it very much if this special case could be handled better by turf itself.

I understand that this change was made to fix a problem in the lineSlice function. And after thinking about it for some time, I also agree that it makes sense that the maximum possible value for the `segmentIndex` property is `line.length - 2` and never `line.length - 1` since the property description clearly reads "closest point was found on nth line segment of the LineString" and there is no line segment after the last line point.

However, large parts of my code relied on the previous behavior, and regardless of which behavior you prefer, I hope you agree that the nearest point falling on the last point of the line is a special case that should be handled accordingly.

I therefore propose to add a boolean property to the point returned by the nearestPointOnLine function that is true if and only if the nearest point exactly falls on the last point of the input line. We could name it `isLastLinePoint` or something like that.

If you give the green light on my proposal (or have a better idea), I would be happy to make a pull request.

smallsaucepan:
@EmilJunker would you mind raising a seperate issue on this, and include some example geometries? What used to work but doesn't now? Refer to this PR and we'll take a look right away, and weigh up the options (including rolling this change back). Thank you and sorry it took the enjoyment out of using those new properties.

PR #2952, opened 2025-11-16
Preparing for NPM trusted publishing plus node version housekeeping

smallsaucepan:
NPM is recommending avoiding using NPM tokens for publishing, instead favouring [trusted publishers](https://docs.npmjs.com/trusted-publishers) (e.g. a particular github workflow). This change:

- updates node versions (housekeeping)
- adds the required configuration item to release workflow act as a trusted publisher in future
- renames turf.yml to ci.yml to be more specific about what that workflow does


.github/workflows/release.yml
```diff
@@ -6,6 +6,7 @@ on:
       - "v*.*.*"
 
 permissions:
+  id-token: write
```
mfedderly, line 9: Were you able to configure a trusted publisher on npmjs? Otherwise I can dig into doing that.

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [20.x, 22.x, 24.x]
```
mfedderly: I've seen other packages consider dropping support for old versions of nodejs as a breaking change.
We aren't technically breaking support here, just not testing it which would make it more likely for a breaking change to sneak in.

I'm happy to either merge this as a non-major change, or push it off for later if you'd prefer.

smallsaucepan: Happy to tread cautiously. So add 18 back in and keep 24 too? Or cap it at 3 - 18, 20, 22?

.github/workflows/release.yml
smallsaucepan: I haven't set up that side of it yet, so be my guest!

.github/workflows/ci.yml
mfedderly: Yeah I think we can just add 24. Once this merges we need to make 24 required.

.github/workflows/release.yml
mfedderly: Oof, I don't see a way to do it for all of the packages at once. I'll probably click through all of them one at a time later today 😩.

I'm going to configure it like this. Note that I added a permissive `release` environment in GitHub now, which we can configure later without having to go to all 1xx packages and reconfigure it later.

image

After that, per the docs guidance, I will then also set this:

image

smallsaucepan:
Putting on hold until I can confirm pnpm support for trusted publishers ...

smallsaucepan:
@mfedderly have reinstated this, adding node 18.x back in during CI. Left it at node 20 though for building the releases. This enough to press on with a trusted publishing test?

.github/workflows/release.yml
```diff
@@ -14,7 +15,7 @@ jobs:
       fail-fast: true
       matrix:
         node:
-          - 18
+          - 20
```
mfedderly, line 18: Yeah I think this can pretty reasonably be any version we want for the release workflow

smallsaucepan:
Thanks for taking a look @mfedderly

PR #2964, opened 2025-11-28
Upgrade Lerna

mfedderly:
Lerna from 8 -> 9, picks up some transitive dependency upgrades that remove Dependabot warnings.
I don't think any of the [deprecations](https://github.com/lerna/lerna/releases/tag/v9.0.0) matter to us (`lerna add`, `lerna bootstrap`, and `lerna link`), and they also dropped support for Node 18.

There were a few small .*ignore changes from `pnpm lerna repair` which is suggested to be run by the Lerna upgrade guide.


mfedderly:
As an aside, `lerna` + `nx` seems to have a massive amount of dependencies. I think about 1/4 of all of our pnpm-lock.yaml is coming from `lerna` itself. I wonder if there's a simpler setup that retains the caching benefits for better devex, now that pnpm can take over most of the `lerna` work itself.

smallsaucepan:
This overlaps with the draft PR I was looking at a while back (#2702). Still very much of the mind developing the turf library itself shouldn't require compiling the whole repo to JS first.

Maybe where we aim to get to is:
1. native typescript development
2. pnpm for monorepo consistency (jettisoning lerna)
3. nx (or alternative) to cache test results (no need to cache build results because we don't build any)
4. only ever worry about cjs-ing, esm-ing, and rollup-ing to in the CI workflows

smallsaucepan (approved):

PR #2733, opened 2024-10-16
Typescript-ifying turf-great-circle

thomas-hervey:
This change is intended to typescript-ify the `turf-great-circles` module following the typescript-ifying discussion https://github.com/Turfjs/turf/discussions/2538.

Summary of changes:
- Converted `/index.js` into `/index.ts`
- Added additional tests to `/test.ts` to cover different `start` and `end` input types
- Added `/lib/arc.d.ts` including class GreatCircle

Template requirements:
- [x] Is this a bug fix, new functionality, or a breaking change?
  - This is an update of the turf-great-circles module to use typescript.
- [x] Have read and followed the steps for [preparing a pull request](https://github.com/Turfjs/turf/blob/master/CONTRIBUTING.md#preparing-a-pull-request).

PR #2965, opened 2025-12-08
Migrate @turf/great-circle to TypeScript

mfedderly:
In the spirit of consolidating our build infrastructure, I took a few minutes to migrate one of the remaining packages.

I considered moving us back to the upstream `arc` npm module, but it benchmarked slightly slower when I was done. (ops/s fluctuated but seemed to be vaguely in these neighborhoods)

Before: 154,482 ops/s
This PR: 172,181 ops/s
Un-vendored: 121,306 ops/s

There were also minor swings in the turf.min.js size (with this PR being slightly smaller than before, and the un-vendored version being slightly larger).


packages/turf-great-circle/lib/arc.js
```diff
@@ -1,315 +0,0 @@
...
-  this.lat = lat;
-  this.x = D2R * lon;
-  this.y = D2R * lat;
-};
-
-Coord.prototype.view = function () {
-  return String(this.lon).slice(0, 4) + "," + String(this.lat).slice(0, 4);
-};
-
-Coord.prototype.antipode = function () {
```
mfedderly, line 41: This was unused and removed.

packages/turf-great-circle/lib/arc.js
```diff
@@ -1,315 +0,0 @@
...
-  var anti_lat = -1 * this.lat;
-  var anti_lon = this.lon < 0 ? 180 + this.lon : (180 - this.lon) * -1;
-  return new Coord(anti_lon, anti_lat);
-};
-
-var LineString = function () {
```
mfedderly, line 47: This conflicted with GeoJSON's LineString, and was just a wrapper on an array anyhow.

packages/turf-great-circle/lib/arc.js
```diff
@@ -1,315 +0,0 @@
...
-    return {
-      geometry: { type: "MultiLineString", coordinates: multiline },
-      type: "Feature",
-      properties: this.properties,
-    };
-  }
-};
-
-// TODO - output proper multilinestring
-Arc.prototype.wkt = function () {
```
mfedderly, line 89: This was unreferenced and therefore removed

packages/turf-great-circle/lib/arc.js
```diff
@@ -1,315 +0,0 @@
...
-    arc.geometries.push(line);
-    var points = poMulti[m];
-    for (var j0 = 0; j0 < points.length; ++j0) {
-      line.move_to(points[j0]);
-    }
-  }
-  return arc;
-};
-
-export { Coord, Arc, GreatCircle };
```
mfedderly, line 309: I changed the exporting to only export the named GreatCircle and dropped the other named ones AND the defaults

packages/turf-great-circle/lib/arc.ts
```diff
@@ -0,0 +1,289 @@
...
+  constructor(public properties = {}) {}
+
+  public json() {
+    if (this.geometries.length <= 0) {
+      return {
+        geometry: {
+          type: "LineString",
+          // Although null coordinates aren't part of RFC7946 (A feature with a null geometry would be preferred)
+          // we leave this as-is for now in order to preserve backwards compatibility
+          coordinates: null as unknown,
```
mfedderly, line 60: I'm not sure if we can even trigger this path? I didn't touch it though, even though it isn't spec.

packages/turf-great-circle/lib/arc.ts
```diff
@@ -0,0 +1,289 @@
+/*!
```
mfedderly, line 1: Sadly the rename didn't get marked as a move because of the necessary changes. I didn't change too much, but I did delete several things which I noted in the arc.js file in this PR.

packages/turf-great-circle/lib/arc.ts
```diff
@@ -0,0 +1,289 @@
...
+ * NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { Feature, LineString, MultiLineString } from "geojson";
+
+var D2R = Math.PI / 180;
+var R2D = 180 / Math.PI;
+
+class Coord {
```
mfedderly, line 33: At first glance its pretty annoying to create an object for every lng, lat pair coming in, but it saves us doing the D2R multiplication constantly in the actual algorithm so I eventually gave in and left it.

mfedderly:
https://github.com/Turfjs/turf/pull/2733 did a nice job of cleaning up additional syntax and using the `arc` npm upstream again, so lets merge that instead. Happy to dig this back up if anyone's workflow is broken by the slightly slower perf of the upstream code vs what we had vendored.

PR #2966, opened 2025-12-08
Migrate @turf/interpolate to TypeScript

mfedderly:
Benchmarks were mixed, but performance is pretty much the same before/after.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [18.x, 20.x, 22.x]
+        node-version: [18.x, 20.x, 22.x, 24.x]
 
     steps:
       - name: Checkout
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,7 @@ on:
       - "v*.*.*"
 
 permissions:
+  id-token: write
   contents: write
 
 jobs:
@@ -14,7 +15,7 @@ jobs:
       fail-fast: true
       matrix:
         node:
-          - 18
+          - 20
         platform:
           - ubuntu-latest
     name: "${{matrix.platform}} / Node.js ${{ matrix.node }}"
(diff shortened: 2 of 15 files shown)


packages/turf-interpolate/index.ts
```diff
@@ -36,29 +37,44 @@ import { collectionOf } from "@turf/invariant";
...
+  var property = options.property ?? "elevation";
+  var weight = options.weight ?? 1;
+  var box = options.bbox ?? bbox(points);
+
+  // validation post options-parsing
+  if (weight !== undefined && typeof weight !== "number") {
+    throw new Error("weight must be a number");
+  }
 
-  box = box ?? bbox(points);
```
mfedderly, line 60: ^ the defaults got pushed above, but the bbox(points) call has to be after the falsy check for points or we lose the useful error message from the validation.

packages/turf-interpolate/index.ts
```diff
@@ -80,31 +96,44 @@ function interpolate(points, cellSize, options) {
...
+      if (d === 0) {
+        zw = zValue;
+      }
       var w = 1.0 / Math.pow(d, weight);
       sw += w;
       zw += w * zValue;
     });
     // write interpolated value for each grid point
     var newFeature = clone(gridFeature);
+    newFeature.properties ??= {};
```
mfedderly, line 130: On the below line, the `properties` could have been null from the types, so we just make sure its there before proceeding.

packages/turf-interpolate/index.ts
```diff
@@ -36,29 +37,44 @@ import { collectionOf } from "@turf/invariant";
  * //addToMap
  * var addToMap = [grid];
  */
-function interpolate(points, cellSize, options) {
+function interpolate<T extends "point" | Grid>(
```
smallsaucepan: What do you reckon about changing this to:

`function interpolate<T extends Grid = "square">(`

"point" is an element of the Grid union anyway, so no need to list it separately. This still seems to allow the conditional return type determination at the end of the function to work.

And I noticed if gridType is left empty, with the current definition the return type is Point | Polygon. However our runtime default is `square` which unambiguously infers Polygon. Setting that for the template default lets TS narrow the return type.

mfedderly: Good catch that's a nice improvement

PR #2969, opened 2025-12-14
Migrate @turf/line-chunk to TypeScript

mfedderly:
Another quick package migration to work towards unifying the build experience


packages/turf-line-chunk/index.ts
```diff
@@ -22,30 +29,47 @@ import { featureCollection, isObject } from "@turf/helpers";
...
-    if (reverse)
+    if (reverse) {
       feature.geometry.coordinates = feature.geometry.coordinates.reverse();
+    }
 
-    sliceLineSegments(feature, segmentLength, units, function (segment) {
-      results.push(segment);
-    });
+    sliceLineSegments(
+      feature as Feature<LineString>,
```
mfedderly, line 66: I think this is a bit of a sketchy cast, but I can't change this without changing the input signature which also takes `GeometryCollection` and `Feature<GeometryCollection>`, which theoretically allow more geometry types. I think we just pass through non-line geometries though 🤷 . Kind of a weird API.

packages/turf-line-chunk/index.ts
```diff
@@ -80,7 +111,7 @@ function sliceLineSegments(line, segmentLength, units, callback) {
       segmentLength * (i + 1),
       { units: units }
     );
-    callback(outline, i);
+    callback(outline);
```
mfedderly, line 114: I just standardized on having no index arg since it wasn't being used and the calls conflicted a bit.

packages/turf-line-chunk/index.ts
smallsaucepan: Quickly checked and the code throws at runtime if passed a polygon, and fails silently if passed a point. Do we add a runtime check to avoid the throw for polys (fixing a bug), and print a warning for everything that won't be handled e.g."Passed a <geometry.type> to lineChunk, probably in your GeometryCollection, ignoring."?

Longer term I'm pretty sure we can narrow to `GeometryCollection<T>` and `Feature<GeometryCollection<T>>`

packages/turf-line-chunk/index.ts
smallsaucepan: Is it possible JS people are using `i` at runtime?

smallsaucepan (approved):
Looks great thanks @mfedderly. Suggested a runtime check / warning we could add if you think it'd be a good idea.

packages/turf-line-chunk/index.ts
mfedderly: Ah I tested `Point` and it didn't throw an error, so I was hesitant to change the runtime behavior by starting to throw one. Lets land the js -> ts migration and I'll come behind and make the GeometryCollection typings more strict and look at what to do for the unsupported geometry types in a separate PR (for easier revert, if need be 🤞).

smallsaucepan: > starting to throw one 

Wasn't suggesting that. Leave point as is and _guard_ against polygon throwing.

mfedderly: Yeah lets continue discussing here: https://github.com/Turfjs/turf/pull/2970#discussion_r2617661286 :)
Happy to do either way though.

PR #2974, opened 2025-12-16
Migrate @turf/line-offset to TypeScript

mfedderly:
Another PR getting us closer to a single build system for the whole project

### DIFF
--- a/packages/turf-helpers/index.ts
+++ b/packages/turf-helpers/index.ts
@@ -219,27 +219,26 @@ export function geometry<
   type: T,
   coordinates: any[],
   _options: Record<string, never> = {}
-): {
-  Point: Point;
-  LineString: LineString;
-  Polygon: Polygon;
-  MultiPoint: MultiPoint;
-  MultiLineString: MultiLineString;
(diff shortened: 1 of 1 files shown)


packages/turf-line-offset/index.ts
```diff
@@ -26,25 +28,29 @@ import { intersection } from "./lib/intersection.js";
...
+function lineOffset<T extends LineString | MultiLineString>(
+  geojson: Feature<T> | T,
+  distance: number,
+  options: { units?: Units } = {}
+) {
   // Optional parameters
   options = options || {};
   if (!isObject(options)) throw new Error("options is invalid");
-  var units = options.units;
+  const { units = "kilometers" } = options;
```
smallsaucepan, line 39: Was the units default value not being set? 😬

smallsaucepan (approved):

packages/turf-line-offset/index.ts
mfedderly: It's eventually set in @turf/helpers' `lengthToRadians`

PR #2978, opened 2025-12-17
@turf/line-slice-along to TypeScript

mfedderly:


packages/turf-line-slice-along/bench.ts
```diff
@@ -1,51 +1,54 @@
...
+var route1: Feature<LineString> = loadJsonFileSync(
+  path.join(__dirname, "test", "fixtures", "route1.geojson")
 );
-var route2 = JSON.parse(
-  fs.readFileSync(__dirname + "/test/fixtures/route2.geojson")
+var route2: Feature<LineString> = loadJsonFileSync(
+  path.join(__dirname, "test", "fixtures", "route2.geojson")
 );
 
+const options = { units: "miles" } as const;
```
mfedderly, line 20: I think this had been broken when we changed the options to an object

smallsaucepan: Huh. So why wasn't the isObject call catching this? Or was bench just not being run by anyone?

packages/turf-line-slice-along/index.ts
```diff
@@ -26,25 +27,29 @@ import { lineString, isObject } from "@turf/helpers";
...
-function lineSliceAlong(line, startDist, stopDist, options) {
+function lineSliceAlong(
+  line: Feature<LineString> | LineString,
+  startDist: number,
+  stopDist: number,
+  options: { units?: Units } = {}
+): Feature<LineString> {
   // Optional parameters
-  options = options || {};
   if (!isObject(options)) throw new Error("options is invalid");
```
smallsaucepan, line 37: Reckon we should set default units explicitly even if it effectively happens a lower level. Maybe we destructure similar to #2974?

`const { units = "kilometers" } = options;`

That'd also make sure we're not passing any rando options further down to distance or destination.

packages/turf-line-slice-along/index.ts
```diff
@@ -26,25 +27,29 @@ import { lineString, isObject } from "@turf/helpers";
...
 
-  var coords;
-  var slice = [];
+  var coords: Position[];
```
smallsaucepan, line 40: Another couple of vars we could jettison.

packages/turf-line-slice-along/bench.ts
mfedderly: Yeah I think it was just broken. There's no actual testing making sure bench works. I don't think we really want to just burn CI time benchmarking though.

packages/turf-line-slice-along/index.ts
mfedderly: Good point about accidentally sending extra options down.

packages/turf-line-slice-along/index.ts
mfedderly: Lets just apply eslint's [no-var](https://eslint.org/docs/latest/rules/no-var) and [prefer-const](https://eslint.org/docs/latest/rules/prefer-const) (in another PR)

smallsaucepan (approved):

PR #2981, opened 2025-12-18
Migrate @turf/meta to TypeScript

mfedderly:
Steps:
- Condense index.js and index.d.ts into index.ts
- Fix type signatures around *Each methods being able to return false
- Annotate some type information manually where TypeScript doesn't handle our branching on `type` very well
- Mark some `bbox` and `id` callback args as potentially `undefined`, because they are
- Add some incorrect-but-good-enough typings around the *Reducer `Reduce` generic to get us to the next major version where we can fix things
- Make some incorrect casts to allow us to be a bit more permissive at runtime and stricter in TypeScript

### DIFF
--- a/packages/turf-meta/index.ts
+++ b/packages/turf-meta/index.ts
@@ -1133,7 +1133,7 @@ function flattenReduce<
 function segmentEach<P extends GeoJsonProperties = GeoJsonProperties>(
   geojson: AllGeoJSON,
   callback: (
-    currentSegment?: Feature<LineString, P>,
+    currentSegment: Feature<LineString, P>,
     featureIndex?: number,
     multiFeatureIndex?: number,
     segmentIndex?: number,
@@ -1266,8 +1266,8 @@ function segmentReduce<
     | Feature<GeometryCollection, P>
     | GeometryCollection,
   callback: (
-    previousValue?: Reducer,
-    currentSegment?: Feature<LineString, P>,
+    previousValue: Reducer,
+    currentSegment: Feature<LineString, P>,
     featureIndex?: number,
     multiFeatureIndex?: number,
     segmentIndex?: number,
@@ -1290,7 +1290,7 @@ function segmentReduce<
         previousValue = currentSegment as Reducer;
       else
         previousValue = callback(
-          previousValue,
+          previousValue as Reducer,
           currentSegment as Feature<LineString, P>,
           featureIndex,
           multiFeatureIndex,
--- a/packages/turf-meta/test.ts
+++ b/packages/turf-meta/test.ts
@@ -10,8 +10,22 @@ import {
   geometryCollection,
   featureCollection,
   lineStrings,
+  Lines,
 } from "@turf/helpers";
 import * as meta from "./index.js";
+import {
+  BBox,
+  Feature,
+  FeatureCollection,
+  Geometry,
+  LineString,
+  MultiLineString,
+  MultiPoint,
+  MultiPolygon,
+  Point,
+  Polygon,
+  Position,
+} from "geojson";
 
 const pt = point([0, 0], { a: 1 });
 const pt2 = point([1, 1]);
@@ -80,7 +94,7 @@ const geomCollection = geometryCollection(
   { a: 0 }
 );
 const fcNull = featureCollection([feature(null), feature(null)]);
-const fcMixed = featureCollection([
+const fcMixed = featureCollection<Point | LineString | MultiLineString>([
   point([0, 0]),
   lineString([
     [1, 1],
@@ -98,8 +112,10 @@ const fcMixed = featureCollection([
   ]),
 ]);
 
-function collection(feature) {
-  const featureCollection = {
+function collection<G extends Geometry>(
+  feature: Feature<G>
+): [Feature<G>, FeatureCollection<G>] {
+  const featureCollection: FeatureCollection<G> = {
     type: "FeatureCollection",
     features: [feature],
   };
@@ -107,14 +123,16 @@ function collection(feature) {
   return [feature, featureCollection];
 }
 
-function featureAndCollection(geometry) {
-  const feature = {
+function featureAndCollection<G extends Geometry>(
+  geometry: G
+): [G, Feature<G>, FeatureCollection<G>] {
+  const feature: Feature<G> = {
     type: "Feature",
     geometry: geometry,
     properties: { a: 1 },
   };
 
-  const featureCollection = {
+  const featureCollection: FeatureCollection<G> = {
     type: "FeatureCollection",
     features: [feature],
   };
@@ -144,7 +162,7 @@ test("coordEach -- Point", (t) => {
 
 test("coordEach -- LineString", (t) => {
   featureAndCollection(line.geometry).forEach((input) => {
-    const output = [];
+    const output: Position[] = [];
     let lastIndex;
     meta.coordEach(input, (coord, index) => {
       output.push(coord);
@@ -161,7 +179,7 @@ test("coordEach -- LineString", (t) => {
 
 test("coordEach -- Polygon", (t) => {
   featureAndCollection(poly.geometry).forEach((input) => {
-    const output = [];
+    const output: Position[] = [];
     let lastIndex;
     meta.coordEach(input, (coord, index) => {
       output.push(coord);
@@ -197,9 +215,9 @@ test("coordEach -- Polygon excludeWrapCoord", (t) => {
 
 test("coordEach -- MultiPolygon", (t) => {
   const coords = [];
-  const coordIndexes = [];
-  const featureIndexes = [];
-  const multiFeatureIndexes = [];
+  const coordIndexes: number[] = [];
+  const featureIndexes: number[] = [];
+  const multiFeatureIndexes: number[] = [];
   meta.coordEach(
     multiPoly,
     (coord, coordIndex, featureIndex, multiFeatureIndex) => {
@@ -218,9 +236,9 @@ test("coordEach -- MultiPolygon", (t) => {
 
 test("coordEach -- FeatureCollection", (t) => {
   const coords = [];
-  const coordIndexes = [];
-  const featureIndexes = [];
-  const multiFeatureIndexes = [];
+  const coordIndexes: number[] = [];
+  const featureIndexes: number[] = [];
+  const multiFeatureIndexes: number[] = [];
   meta.coordEach(
     fcMixed,
     (coord, coordIndex, featureIndex, multiFeatureIndex) => {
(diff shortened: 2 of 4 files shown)


packages/turf-meta/index.ts
```diff
@@ -34,7 +57,17 @@ import { feature, point, lineString, isObject } from "@turf/helpers";
...
+  geojson: AllGeoJSON,
+  callback: (
+    currentCoord: number[],
+    coordIndex: number,
+    featureIndex: number,
+    multiFeatureIndex: number,
+    geometryIndex: number
+  ) => void | false,
+  excludeWrapCoord?: boolean
+): void | false {
```
mfedderly, line 70: Public API change!

The `index.d.ts` version of this had just `void` for coordEach and the callback return value.
I had to add `| false` to every *Each method's callback, and also to some of the *Each functions themselves.

This is the first instance of the change, but it happens throughout the file on most *Each functions.

I think this is ok to do without a major version because returning void is still allowed.

packages/turf-meta/index.ts
```diff
@@ -50,7 +83,9 @@ function coordEach(geojson, callback, excludeWrapCoord) {
     type = geojson.type,
     isFeatureCollection = type === "FeatureCollection",
     isFeature = type === "Feature",
-    stop = isFeatureCollection ? geojson.features.length : 1;
+    stop = isFeatureCollection
+      ? (geojson as FeatureCollection).features.length
+      : 1;
```
mfedderly, line 88: This is also an example of another really common change in this PR. TypeScript was not narrowing the types enough and so we require some manual annotation of types to make everything work out. I don't think I'm doing anything in here that isn't being checked at runtime.

packages/turf-meta/index.ts
```diff
@@ -511,7 +661,24 @@ function coordAll(geojson) {
...
+    | FeatureCollection<G, P>
+    | G
+    | GeometryCollection
+    | Feature<GeometryCollection, P>,
+  callback: (
+    currentGeometry: G,
+    featureIndex: number,
+    featureProperties: P,
+    featureBBox: BBox | undefined,
+    featureId: Id | undefined
```
mfedderly, line 684: Public api change!

These are just being pulled from the underlying features, which may or may not have a bbox or id. This change can probably cause new TypeScript errors, but these things are potentially already runtime errors so we can probably push it without a major version? Worst case people can revert to the existing behavior by using `!` and move along with the upgrade.

packages/turf-meta/index.ts
```diff
@@ -678,7 +852,27 @@ function geomEach(geojson, callback) {
...
+    | G
+    | GeometryCollection
+    | Feature<GeometryCollection, P>,
+  callback: (
+    previousValue: Reducer,
+    currentGeometry: G,
+    featureIndex: number,
+    featureProperties: P,
+    featureBBox: BBox | undefined,
+    featureId: Id | undefined
```
mfedderly, line 872: Another instance of the bbox/id getting new `undefined` options as above

packages/turf-meta/index.ts
```diff
@@ -264,7 +360,7 @@ function coordReduce(geojson, callback, initialValue, excludeWrapCoord) {
     },
     excludeWrapCoord
   );
-  return previousValue;
+  return previousValue as Reducer;
```
mfedderly, line 363: This block of three `as Reducer` casts are a bit of a compromise to avoid making breaking changes right now.

In the first cast, `currentCoord` is almost certainly not a `Reducer` (its the first coord, kind of in the way `Array.prototype.reducer` will fill in `accumulator` with `arr[0]` if an initial value is unspecified).

The second cast in theory could be a `!` because of previousValue being set to `currentCoord`, except that's slightly wrong because `Reducer` *could* extend `undefined`, so I think `as Reducer` is in the right direction?

The last cast in the return is *also* wrong because `previousValue` can remain unset if `initialValue` is unspecified and we don't actually have any coords to reduce in `geojson`.

So all of these casts are slightly wrong, but I don't actually think we can change it without changing the runtime behavior to make this simpler to type. I think my preference would be to make `initialValue` mandatory, and remove the code that sets previousValue if it was not already set. That would make the types cleaner and users could bring the behavior back by checking the `*Index` value in the callback.

This is just the first instance of this change, but the logic basically holds across all of the rest of reducers below as well.

packages/turf-meta/index.ts
```diff
@@ -1230,9 +1546,9 @@ function findSegment(geojson, options) {
       properties = properties || geojson.properties;
       geometry = geojson.geometry;
       break;
-    case "Point":
-    case "MultiPoint":
-      return null;
+    case "Point" as any:
+    case "MultiPoint" as any:
+      return null as any;
```
mfedderly, line 1551: The Point and MultiPoint cases here aren't technically reachable, but I'm afraid to change runtime behavior.

packages/turf-meta/index.ts
```diff
@@ -1244,16 +1560,19 @@ function findSegment(geojson, options) {
   }
 
   // Find SegmentIndex
-  if (geometry === null) return null;
+  if (geometry === null) return null as any;
```
mfedderly, line 1563: `findPoint` and `findSegment` both have `null` return paths that aren't documented in the type signature

mfedderly:
While working on this, I found a bunch of stuff that probably needs to get sorted as part of a major version.

```
propEach needs stricter inference for Props generics
propReduce accepts Geometry but then uses propEach which doesn't support that
*Reduce needs to get better type inference around its initialValue, callback arg's inferred initial values, and the return type possibly being undefined without an initial value
findPoint doesn't seem to handle GeometryCollections
findPoint can return null but it isn't part of the TypeScript signature
findSegment can return null but it isn't part of the TypeScript signature
flattenEach and flattenReduce callback signatures don't infer from MultiFoo down to Foo
The *Each and *Reducer callback args sometimes have indexes marked as optional which might be incorrect
```

I'm considering the wisdom of actually just deprecating several of these methods instead of trying to fix them.

packages/turf-meta/test.ts
```diff
@@ -333,22 +351,22 @@ test("Array.reduce() -- previous-coordinates+initialValue", (t) => {
     lastIndex = index;
     coords.push(currentCoords);
     return currentCoords;
-  }, line[0]);
+  }, undefined as any);
```
mfedderly, line 354: I think this was a legitimate bug in the tests

packages/turf-meta/test.ts
```diff
@@ -333,22 +351,22 @@ test("Array.reduce() -- previous-coordinates+initialValue", (t) => {
...
   t.equal(lastIndex, 1);
   t.equal(coords.length, 2);
   t.end();
 });
 
 test("unknown", (t) => {
   t.throws(function () {
-    meta.coordEach({});
+    meta.coordEach({} as any, () => undefined);
```
mfedderly, line 362: It throws an error on the invalid `{}` arg, but TypeScript requires the 2nd arg to be passed anyhow now

packages/turf-meta/test.ts
```diff
@@ -395,7 +413,7 @@ test("geomEach -- multiGeometryFeature-properties", (t) => {
 
 test("flattenEach -- MultiPoint", (t) => {
   featureAndCollection(multiPt.geometry).forEach((input) => {
-    const output = [];
+    const output: MultiPoint[] = []; // this is actually Point[] but flattenEach's types are bad
```
mfedderly, line 416: flattenEach and flattenReduce both have pretty gnarly typing bugs that need to be fixed. MultiPoint/LineString/Polygon should all get transformed to non-multi Point/LineString/Polygon.

packages/turf-meta/test.ts
```diff
@@ -490,7 +511,7 @@ test("flattenReduce -- previous-feature+initialValue", (t) => {
       features.push(current);
       return current;
     },
-    null
+    null as any
```
mfedderly, line 514: The generic is inferring exactly `null` for the type, which would make the `return current` invalid without this `any` cast

packages/turf-meta/test.ts
```diff
@@ -568,7 +589,7 @@ test("null geometries -- index", (t) => {
     meta.flattenReduce(
       fc,
       (prev, feature, featureIndex) => prev.concat(featureIndex),
-      []
+      [] as number[]
```
mfedderly, line 592: Same with these, but its a `never[]` instead

packages/turf-meta/test.ts
```diff
@@ -659,10 +682,10 @@ const geojsonSegments = featureCollection([
...
 
 test("segmentEach -- index & subIndex", (t) => {
-  const featureIndexes = [];
-  const multiFeatureIndexes = [];
-  const geometryIndexes = [];
-  const segmentIndexes = [];
+  const featureIndexes: (number | undefined)[] = [];
+  const multiFeatureIndexes: (number | undefined)[] = [];
+  const geometryIndexes: (number | undefined)[] = [];
+  const segmentIndexes: (number | undefined)[] = [];
```
mfedderly, line 688: There's a bunch of these functions in meta that pass optional *Index values to the callback, when I think they're always specified and not optional. We need to do a pass and clean that up so this sort of thing can be `number[]` as much as possible.

packages/turf-meta/test.ts
```diff
@@ -995,10 +1022,10 @@ test("lineEach & lineReduce -- assert", (t) => {
...
     /* no-op */
   };
-  meta.lineEach(pt, noop); // Point geometry is supported
-  meta.lineEach(multiPt, noop); // MultiPoint geometry is supported
-  meta.lineReduce(pt, noop); // Point geometry is supported
-  meta.lineReduce(multiPt, noop); // MultiPoint geometry is supported
+  meta.lineEach(pt as any, noop); // Point geometry is supported
+  meta.lineEach(multiPt as any, noop); // MultiPoint geometry is supported
+  meta.lineReduce(pt as any, noop); // Point geometry is supported
+  meta.lineReduce(multiPt as any, noop); // MultiPoint geometry is supported
```
mfedderly, line 1028: More instances of TypeScript being annoying when we try to test runtime behavior that TypeScript would have blocked.

packages/turf-meta/types.ts
```diff
@@ -73,9 +73,15 @@ const customLineStrings = featureCollection([customLineString]);
 /**
  * meta.coordEach
  */
-const coordEachValue: void = meta.coordEach(pt, (coords) => coords);
-coordEach(pt, (coords, index) => coords);
-meta.coordEach(pt, (coords, index) => coords);
```
mfedderly, line 78: I think that this was only kinda-sorta a working syntax before, but once we change the package to TypeScript we start getting errors about it. coordEach's callback should return `void | false`, but specifying `coords` like this makes is a return value. I just add `{}s` around each of these to just make it a no-op reference inside the function instead of a return value. This gets repeated all the way down the file.

I think that this would not have been allowed if you were already running TypeScript against this file before, and this isn't a real break. It was pretty annoying to have to eat this syntax change though, so if you think it would have been allowed before we should figure out how to not cause this.

PR #2983, opened 2025-12-19
Example typings for featureReduce

mfedderly:
Related: #2982 #2981

### DIFF
--- a/packages/turf-area/index.ts
+++ b/packages/turf-area/index.ts
@@ -3,37 +3,114 @@ import {
   FeatureCollection,
   GeoJsonProperties,
   Geometry,
-  GeometryCollection,
-  GeometryObject,
+  Point,
 } from "geojson";
 import { earthRadius } from "@turf/helpers";
 import { geomReduce, featureEach } from "@turf/meta";
 
 // would be in @turf/meta
-export function featureReduce<
-  Reducer,
-  G extends GeometryObject,
-  P extends GeoJsonProperties = GeoJsonProperties,
->(
-  geojson:
-    | Feature<G, P>
-    | FeatureCollection<G, P>
-    | Feature<GeometryCollection, P>,
+
+// with initial value specified, things are relatively straightforward
+export function featureReduce<R, T extends Feature | FeatureCollection>(
+  geojson: T,
   callback: (
-    previousValue: Reducer,
-    currentFeature: Feature<G, P>,
+    previousValue: R,
+    currentFeature: T extends FeatureCollection<infer G, infer P>
+      ? Feature<G, P>
+      : T,
     featureIndex: number
-  ) => Reducer,
-  initialValue?: Reducer
-) {
-  var previousValue = initialValue;
+  ) => R,
+  initialValue: R
+): R;
+
+// with no initial value specified, previousValue can sometimes be Feature<G,P>
+// the return value may also be undefined if we get a FeatureCollection without a guaranteed first element
+export function featureReduce<T extends Feature | FeatureCollection>(
+  geojson: T,
+  callback: (
+    previousValue: T extends FeatureCollection<infer G, infer P>
+      ? Feature<G, P>
+      : T,
+    currentFeature: T extends FeatureCollection<infer G, infer P>
+      ? Feature<G, P>
+      : T,
+    featureIndex: number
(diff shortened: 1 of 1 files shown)


packages/turf-area/index.ts
```diff
@@ -1,6 +1,120 @@
...
+  FeatureCollection,
+  GeoJsonProperties,
+  Geometry,
+  Point,
+} from "geojson";
 import { earthRadius } from "@turf/helpers";
-import { geomReduce } from "@turf/meta";
+import { geomReduce, featureEach } from "@turf/meta";
+
+// would be in @turf/meta
```
mfedderly, line 11: Mostly just pay attention to the two function typings, not the one with the actual implementation itself

mfedderly: Reference from lib.es5.d.ts on how Array.prototype.reduce is typed. Note that you can either specify an initial value and deal in that type for the reduce callbacks, or you can omit it and you are forced to use the shape of whatever is in the array.

```
    /**
     * Calls the specified callback function for all the elements in an array. The return value of the callback function is the accumulated result, and is provided as an argument in the next call to the callback function.
     * @param callbackfn A function that accepts up to four arguments. The reduce method calls the callbackfn function one time for each element in the array.
     * @param initialValue If initialValue is specified, it is used as the initial value to start the accumulation. The first call to the callbackfn function provides this value as an argument instead of an array value.
     */
    reduce(callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: T[]) => T): T;
    reduce(callbackfn: (previousValue: T, currentValue: T, currentIndex: number, array: T[]) => T, initialValue: T): T;
    /**
     * Calls the specified callback function for all the elements in an array. The return value of the callback function is the accumulated result, and is provided as an argument in the next call to the callback function.
     * @param callbackfn A function that accepts up to four arguments. The reduce method calls the callbackfn function one time for each element in the array.
     * @param initialValue If initialValue is specified, it is used as the initial value to start the accumulation. The first call to the callbackfn function provides this value as an argument instead of an array value.
     */
    reduce<U>(callbackfn: (previousValue: U, currentValue: T, currentIndex: number, array: T[]) => U, initialValue: U): U;
 ```

File: packages/turf-area/index.ts [hunk 2]
```diff
@@ -1,6 +1,120 @@
...
+    featureIndex: number
+  ) => R,
+  initialValue?: R
+): T extends Feature ? R : R | undefined {
+  var previousValue: R | T extends FeatureCollection<infer G, infer P>
+    ? Feature<G, P>
+    : T = initialValue as any;
+  featureEach(geojson, function (currentFeature, featureIndex) {
+    if (featureIndex === 0 && initialValue === undefined)
+      previousValue = currentFeature as any;
```
mfedderly, line 65: Note that this is actually *different* from how Array.prototype.reduce works when you don't specify an initial value. Without an initial value, you need to set previousValue to the first feature, but then only iterate 1...n in the array.

I think in order to know whether or not we were passed an initial value, we need to use `arguments.length` to see if we got an explicit undefined for initialValue vs a parameter that wasn't specified at all.

packages/turf-area/index.ts
```diff
@@ -1,6 +1,120 @@
...
+    else
+      previousValue = callback(
+        previousValue,
+        currentFeature as any,
+        featureIndex
+      ) as any;
+  });
+  return previousValue as any;
```
mfedderly, line 73: I made a bunch of sloppy `as any` casts inside here, but the implementation would need to change to match up with the new behaviors to align with Array.prototype.reduce anyhow. Obviously I'd try to avoid them in a final implementation.

packages/turf-area/index.ts
```diff
@@ -1,6 +1,116 @@
...
+  };
+
+// a FeatureCollection where we know that there is at least one feature (somehow)
+const populatedFeatureCollection: {
+  type: "FeatureCollection";
+  features: [Feature<Point, GeoJsonProperties>];
+} = {
+  type: "FeatureCollection",
+  features: [feature],
+};
```
mfedderly, line 100: This bit is a little unfortunate. You cannot omit the explicit type definition here and use `as const` instead, because that marks the whole thing as readonly and breaks the typings of featureReduce. We should probably handle readonly inputs as well and just make the relevant things downstream of that also readonly.

PR #2984, opened 2025-12-19
@turf/buffer TypeScript migration

mfedderly:
This is almost ready to go, but requires types to be [upstreamed](https://github.com/Turfjs/turf-jsts/pull/4) into @turf/jsts first.


mfedderly:
Lets actually just solve this by migrating to clipper2-ts instead (see linked PR)

PR #2985, opened 2025-12-20
Migrate @turf/line-split to TypeScript

mfedderly:


packages/turf-line-split/index.ts
```diff
@@ -43,21 +61,38 @@ function lineSplit(line, splitter) {
   // to avoid possible approximation issues in rbush
   var truncatedSplitter = truncate(splitter, { precision: 7 });
 
+  // Ensure we're consistently sending Feature<LineString> into the splitLine methods.
+  if (line.type !== "Feature") {
+    line = feature(line);
+  }
```
mfedderly, line 67: Its hard to spot in the diff, but `splitLineWithPoint` has this snippet, which only works with Features. Shout out to catching this issue at compile time instead of runtime.
```
  var startPoint = getCoords(line)[0];
  var endPoint = getCoords(line)[line.geometry.coordinates.length - 1];
```

PR #2986, opened 2025-12-20
@turf/invariant getTypes cleanup

mfedderly:
Ran into some issues when [typing @turf/line-split](https://github.com/Turfjs/turf/pull/2985) that warranted a small cleanup PR.

- Make type inference work a lot better
- Simplify logic
- Add a test for Feature<null> behavior, which is surprising and I don't like it
- Remove an unused parameter

I'm kind of unsure how much of a breaking change this is. Removing a parameter that was never used breaks the signature, but also I'm not sure how anyone would've been passing in anything for this parameter that appears to have been unused since we started shipping types.

We're also narrowing the return type from `string` to a specific string (or string union, depending on input). This could cause a new break with something like the (very contrived) snippet below:

```
let t = getType(pointFeature); // t is of type `"Point"` now, previously it would have been `string`
t = "foo"; // previously that's fine, but now it would error with "foo" isn't "Point"
```

Overall I think I'd push for merging this in 7.x, but happy for pushback there if others feel strongly against.

### DIFF
--- a/packages/turf-invariant/README.md
+++ b/packages/turf-invariant/README.md
@@ -132,8 +132,6 @@ Get GeoJSON object's type, Geometry type is prioritize.
 ### Parameters
 
 *   `geojson` **[GeoJSON][7]** GeoJSON object
-*   `_name` **[string][8]?**&#x20;
-*   `name` **[string][8]** name of the variable to display in error message (unused) (optional, default `"geojson"`)
 
 ### Examples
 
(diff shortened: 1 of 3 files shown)


smallsaucepan:
Would lean toward saving this for a major release. No point risking it if there is no pressure.

Out of curiosity what does the documentation for the return type end up looking like?

mfedderly:
> Out of curiosity what does the documentation for the return type end up looking like?

I tried to generate the docs again locally and ... have no idea how to do it. If its just parsing the README.md files, it will just say `string` 🤷. Happy to take a look again if you want to hold my hand getting the docs setup locally for me.

smallsaucepan:
All good for it to stay as string. Will flesh out turf-www with a better readme though.

PR #2988, opened 2025-12-22
Speed up development builds and drop tsup

smallsaucepan:
Changed multiple build and test processes to enable dev and testing (as maintainers) without having to build to ESM+CJS Javascript first.

In short:
- heading in the direction of TS composite builds, which require capturing more explicit inter package dependencies so that eventually TS can work its magic with incremental builds and the like
- each packages gets one or two new tsconfig files - one for building and one for type testing - and a tsup config file. There are a lot of them, though they are almost exclusively automatically generated (by monorepolint
- thus able to remove the `build` pre-requisite from the `test` target - test TS directly, including dependent packages
- updated some tsconfig options such as module and moduleResolution, though kept target at es2017 to be cautious between major versions

Vastly reduces the amount of uncessary JS we need to compile to test TS code. Developer workflow can be as quick as:

- git clone ...
- cd turf/
- pnpm install
- cd packages/turf-whatever/ - and edit a TS file to fix a bug
- pnpm test - to run only that package's unit tests (and type tests if they exist)
- cd ../.. && pnpm test - to run unit and type tests for all packages

Shouldn't result in the generation of a single line of Javascript!

To eventually generate JS for publishing there are new top level targets `prerelease` and `release`. Part of this PR review can be deciding what a `ci` target needs to do as well.

File by file checks on the contents of dist/ indicate no changes to generated ESM or CJS other than a handful of lines removed because we were able to remove an old node16 shim. Debugging map file symbols do change though that's somewhat expected.

Scaffolding changes committed first, and then the generated files separately to help with reviewing.

- [x] Summary of the changes.


tsconfig.shared.json
```diff
@@ -8,6 +8,134 @@
     "moduleResolution": "bundler",
     "importHelpers": true,
     "skipLibCheck": true,
-    "erasableSyntaxOnly": true
-  }
+    "erasableSyntaxOnly": true,
+    "composite": true,
+    "allowImportingTsExtensions": true,
+    "outDir": "dist",
```
mfedderly: I'm kind of surprised this doesn't need to be "${configDir}/dist", but it seems to work 🤷

.prettierignore
```diff
@@ -12,6 +12,8 @@ pnpm-lock.yaml
 
 /.nx/workspace-data
 
+# Auto generated by mrl, which produces a slightly clashing format.
 tsconfig.json
 tsconfig.build.json
 tsconfig.shared.json
+tsconfig.types.json
```
mfedderly, line 19: What is the nature of this conflict? In my mind MRL should write the .json files, and the prettier pre-commit hook should clean them up before finalizing the commit, and then MRL should just accept the reformatted version.

I'm guessing it didn't work that way?

package.json
```diff
@@ -2,18 +2,24 @@
   "private": true,
   "funding": "https://opencollective.com/turf",
   "scripts": {
+    "build": "pnpx lerna run build",
     "docs": "tsx ./scripts/generate-readmes.ts",
+    "escheck": "pnpm run /escheck:.*/",
+    "escheck:cjs": "es-check es8 packages/*/dist/cjs/index.cjs packages/turf/turf.min.js",
+    "escheck:esm": "es-check --module es8 packages/*/dist/esm/index.js",
+    "escheck:web": "es-check es5 packages/turf/turf.min.js",
```
mfedderly, line 11: lmao we're still shipping es5 turf.min.js. That should definitely get bumped way forward.
https://github.com/Turfjs/turf/issues/2989

package.json
```diff
@@ -2,18 +2,24 @@
...
     "lint:prettier": "prettier --check .",
     "preinstall": "npx only-allow pnpm",
-    "prepare": "husky && lerna run build",
-    "test": "pnpm run lint && lerna run test && lerna run --scope @turf/turf last-checks"
+    "prepare": "husky",
+    "prerelease": "pnpm run build && pnpm run rollup",
+    "release": "pnpm run test:release && pnpm publish -r",
+    "rollup": "lerna run --scope @turf/turf rollup",
+    "test": "pnpx lerna run test",
+    "test:release": "pnpm run lint && pnpm run escheck && lerna run --scope @turf/turf last-checks"
```
mfedderly: I would be in favor of just keeping this all one `test` script, instead of breaking it out differently for releases.
We can probably entirely get rid of the escheck step once we swap TypeScript to targeting `esnext`.

packages/turf-line-intersect/index.ts
```diff
@@ -9,7 +9,7 @@ import {
   Polygon,
 } from "geojson";
 import type { Intersection } from "sweepline-intersections";
-import { sweeplineIntersections as findIntersections } from "./lib/sweepline-intersections-export.js";
+import findIntersections from "sweepline-intersections";
```
mfedderly: I didn't realize this shim stuff was still around. 🧹

tsup.config.ts
```diff
@@ -7,22 +7,11 @@ const baseOptions: Options = {
   sourcemap: true,
   // treeshake: true, causes "chunk.default" warning, breaks CJS exports?
   minify: false,
-  skipNodeModulesBundle: true,
   cjsInterop: true,
   splitting: true,
+  external: [
+    /^@turf\//, // Externalize all @turf workspace packages
+  ],
```
mfedderly: Is this required to basically reverse the `paths` changes when we're doing the tsup step?
They should be already excluded by default if I'm reading the docs right.

https://tsup.egoist.dev/#excluding-packages

packages/turf-along/tsconfig.json
```diff
@@ -1,3 +1,24 @@
 {
-  "extends": "../../tsconfig.shared.json"
+  "extends": "../../tsconfig.shared.json",
+  "include": [
+    "index.ts",
+    "lib/**/*.ts"
+  ],
```
mfedderly, line 6: I think this can just be moved into tsconfig.shared.json like so:

```
"include": [
  "${configDir}/index.ts",
  "${configDir}/lib/**/*.ts"
]
```

packages/turf-along/tsconfig.build.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "composite": false
```
mfedderly: I'm a little surprised you can't use the default tsconfig.json for the build step as well.
It might be a little extra work if you do the build step in every package as it will re-check any shared dependencies have already been built, but I think that a build of @turf/turf should just recursively build all packages for free if we're doing it in CI.

.prettierignore
smallsaucepan: Sadly, it did not. mrl doesn't accept the reformatted version. There might be a prettier config that doesn't correct these (or an mrl switch that makes it more forgiving). Added to the ignore so we could press on in the meantime.

packages/turf-along/tsconfig.json
smallsaucepan: Will give it a try.

smallsaucepan:
Putting this on ice until we convert remaining JS in turf packages. Otherwise adding in complexity only to remove it again in the near future.

PR #2992, opened 2025-12-31
Clean up .js files in various TypeScript-first packages

mfedderly:
Turns out the migration steps to TypeScript in certain packages just added typings to the .js files and moved on. This converts the rest of the supporting .js files, so we don't have to rely on the `tsup` build behavior of copying these into the dist directory for us.

I'll add a few inline notes of things I noticed along the way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
       "eslint --fix"
     ],
     "packages/*/index.{js,ts}": [
-      "pnpm tsx ./scripts/generate-readmes"
+      "pnpm tsx ./scripts/generate-readmes",
+      "git add ./packages/*/README.md"
     ],
     "*": "prettier --write --ignore-unknown"
   },


packages/turf-isobands/lib/grid-to-matrix.ts
```diff
@@ -1,6 +1,7 @@
 import { getCoords, collectionOf } from "@turf/invariant";
 import { featureEach } from "@turf/meta";
 import { isObject } from "@turf/helpers";
+import { Feature, FeatureCollection, Point } from "geojson";
```
mfedderly, line 4: isobands and isolines share these two matrix transform files, they're identical copies.

packages/turf-isobands/lib/matrix-to-grid.ts
```diff
@@ -57,6 +67,9 @@ function matrixToGrid(matrix, origin, cellSize, options) {
   for (var r = 0; r < matrixRows; r++) {
     // create first point in the row
     var first = rhumbDestination(origin, cellSize * r, 0, { units: units });
+    if (first.properties == null) {
+      first.properties = {};
+    }
```
mfedderly, line 72: To satisfy the type strictness.

packages/turf-unkink-polygon/lib/geojson-polygon-self-intersections.ts
```diff
@@ -1,56 +1,68 @@
 // Find self-intersections in geojson polygon (possibly with interior rings)
+import { Feature, Polygon, Position } from "geojson";
```
mfedderly, line 2: This file and simplepolygon felt really rough to migrate 😩

packages/turf-unkink-polygon/lib/geojson-polygon-self-intersections.ts
```diff
@@ -1,56 +1,68 @@
...
+        ifIsectAddToOutput(ringA, edgeA, ring1, edge1);
+      });
     }
   }
 
-  if (!filterFn)
-    output = {
-      type: "Feature",
-      geometry: { type: "MultiPoint", coordinates: output },
-    };
```
mfedderly, line 49: `filterFn` is always defined in our usage, so I removed this.

packages/turf-unkink-polygon/lib/simplepolygon.ts
```diff
@@ -100,19 +103,27 @@ function simplepolygon(feature) {
...
+      Polygon,
+      {
+        index: number;
+        parent: number;
+        winding: number;
+        netWinding: number | undefined;
+      }
+    >;
+    determineParents(output);
+    setNetWinding(output);
```
mfedderly, line 116: `var output` was declared twice and the types disagreed, and then the determineParents and setNetWinding calls were referencing it quasi-implicitly because of the closure. Gnarly stuff.

packages/turf-unkink-polygon/lib/simplepolygon.ts
```diff
@@ -225,7 +242,7 @@ function simplepolygon(feature) {
   for (var i = 0; i < pseudoVtxListByRingAndEdge.length; i++) {
     for (var j = 0; j < pseudoVtxListByRingAndEdge[i].length; j++) {
       for (var k = 0; k < pseudoVtxListByRingAndEdge[i][j].length; k++) {
-        var coordToFind = pseudoVtxListByRingAndEdge[i][j][k].coord;
+        let coordToFind = pseudoVtxListByRingAndEdge[i][j][k].coord;
```
mfedderly, line 245: ^ multiple `var` defs bypassed by turning them into `let`s

packages/turf-unkink-polygon/lib/simplepolygon.ts
```diff
@@ -347,13 +364,15 @@ function simplepolygon(feature) {
         walkingRingAndEdge = isectList[nxtIsect].ringAndEdge2;
         isectList[nxtIsect].ringAndEdge2Walkable = false;
         if (isectList[nxtIsect].ringAndEdge1Walkable) {
-          var pushing = { isect: nxtIsect };
+          var pushing: { isect: number; parent: number; winding: number } = {
+            isect: nxtIsect,
+          } as any; // as any because parent and winding are filled in below
```
mfedderly, line 369: Leaving the code as-is, but I'd write this differently if we were TypeScript first. The isect/parent/winding tuple type is also defined in several places instead of giving it a name.

packages/turf-unkink-polygon/lib/simplepolygon.ts
```diff
@@ -563,23 +634,20 @@ function equalArrays(array1, array2) {
...
+  var u: Record<string, number> = {};
   var isUnique = 1;
   for (var i = 0, l = array.length; i < l; ++i) {
-    if (Object.prototype.hasOwnProperty.call(u, array[i])) {
+    if (Object.prototype.hasOwnProperty.call(u, array[i].toString())) {
       isUnique = 0;
       break;
     }
-    u[array[i]] = 1;
+    u[array[i].toString()] = 1;
```
mfedderly, line 650: This is probably horrible for perf `[number, number]`.toString() gets used as the key into `u`

packages/turf-unkink-polygon/lib/geojson-polygon-self-intersections.ts
```diff
@@ -74,9 +86,9 @@ function geojsonPolygonSelfIntersections(feature, filterFn, useSpatialIndex) {
     if (frac0 >= 1 || frac0 <= 0 || frac1 >= 1 || frac1 <= 0) return; // require segment intersection
 
     var key = isect;
-    var unique = !seen[key];
+    var unique = !seen[key.toString()];
```
mfedderly, line 89: this is `[number, number].toString()` being used as a key. This has to be bad for perf.

packages/turf-unkink-polygon/lib/geojson-polygon-self-intersections.ts
mclaeysb: Hi @mfedderly . I'm the author of simplepolygon and migrate the code to typescript around the same time (two weeks earlier actually) of your commit. Just found out about your work too. There's some useful feedback here I'll try to incorporate in the near future, thanks!

I'm curious if the vendoring of simplepolygon and geojson-polygon-self-intersections is a strategic choice, or if you'd consider to add them as dependencies?

Manuel

mfedderly: @mclaeysb in this PR I just blindly migrated the remaining .js files to .ts since we're trying to get off of the split js/ts build system that we've got going on right now. These few files definitely felt like I was more likely to make a mistake in the translation compared to the other ones.

I'm not actually sure why this file was originally vendoered (edit: see below about circular dependencies onto Turf packages). I'd be open to transitioning this to a dependency on your original NPM package though. Using the original package would be good for overall bundle sizes, on the off chance that someone pulls in `simplepolygon` and `@turf/unkink-polygon` at the same time.

One thing that's been on my mind is that Turf does tend to generate ~500k downloads a week, and so I feel some responsibility to guard against supply-chain attacks. Turf just published its first release with [NPM's Trusted Publishing](https://docs.npmjs.com/trusted-publishers), as we already were publishing through GitHub actions.

I'm still thinking about the exact details but I would prefer our dependencies to meet the following standards:
- Have a npm/pnpm/yarn lockfile checked in
- Have relatively few entries in `dependencies`
- Have a relatively low number of issues flagged by Dependabot in your package and its transitive dependencies (ignoring anything coming from devDependencies, because it doesn't impact Turf and its consumers). Basically: do you increase OUR Dependabot list if we take a dependency on your package.
- Use trusted publishing (I know this is new and lots of our existing packages don't satisfy this, I think it's probably valuable but understand that its annoying to set up)

In `simplepolygon`'s case, there's another complication. Turf cannot safely take dependencies on a package that itself has dependencies on `@turf/*`. The reason being is that our published packages depend on all of the other Turf packages at *at least* the same version. For example your package currently depends on `"@turf/area": "7.3.1"`, which will already pull in a second version of `@turf/area` with the current Turf 7.3.3. Sneakily, even if you depended on `"@turf/area": "^7"`, it would give us these same issues when we release v8.

Separately, I suspect you don't need the dependency on `geojson`, and just want the `@types/geojson` devDependencies.

mfedderly: Oh another thing, even for single-maintainer packages, it makes me more comfortable if there are pull requests that publicly run some amount of testing before being merged.

mclaeysb: Thanks for the reply (and the work on Turf!). I understand the risk management perspective. I just wanted to offer help if a dependency solution was preferred, but for me it all ok.

Going forward I'll try to find some time to include your suggestions on the `geojson` dependencies and other remarks in this PR (like those on speed and efficiency) in the latest version of simplepolygon. I don't see myself setting up a more complicated publishing or CI though.

You can then consider to update the vendored version to the latest simplepolygon (with TS and improvements) just once (or add a dependency, but reading your requirement, maybe that won't happen).

I'll keep you updated
Cheers!

mfedderly: @mclaeysb sounds good. Thanks to you also also for simplepolygon in the first place! Its been working well for Turf since before my time on the project, and as far as I know nobody has complained about performance. It was just something I noticed along the way as I migrated the code.

Happy to lend another set of eyes on the simplepolygon side if you'd like, otherwise feel free to ping me and I can re-vendor the updated code. We appreciate the continued maintenance here, always nice to have some confidence that things haven't been abandoned.

PR #2993, opened 2026-01-02
Migrated turf-meta to typescript (bare bones edition)

smallsaucepan:
Minimally intrusive conversion of turf-meta to typescript. Rather than try to untangle internal type inconsistencies, leave the public interface and runtime as is and sprinkle a liberal dose of @ts-expect-error statements. Once decided on a resolution strategy we can remove those as we go.

Note the `excludeWrapCoord` parameter on coordReduce wasn't defined in the d.ts. It is documented and a parameter at runtime, so have added it as an optional fourth param. Believe this would be non-breaking.

- [x] Summary of the changes.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4139,6 +4139,9 @@ importers:
       '@types/geojson':
         specifier: ^7946.0.10
         version: 7946.0.14
+      tslib:
+        specifier: ^2.8.1
+        version: 2.8.1
     devDependencies:
       '@turf/random':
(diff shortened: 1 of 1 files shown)


mfedderly (approved):
Thanks!

For those curious about all of the `@ts-expect-error` introduced here, this is a compromise to avoid introducing breaking changes in a minor release. If you try to start removing these errors things start to get [out of hand](https://github.com/Turfjs/turf/pull/2981) quickly.

PR #2991, opened 2025-12-24
Migrate @turf/buffer to TypeScript

mfedderly:
Instead of [messing around](https://github.com/Turfjs/turf-jsts/pull/4) [with @turf/jsts needing types](https://github.com/Turfjs/turf/pull/2984), there's actually a new entrant into the underlying library space that solves a lot of problems at once.

[clipper2-ts](https://github.com/countertype/clipper2-ts) seems pretty exciting for us. I happened to find it because of our desire to drop our jsts dependency, but it also has intersect, union, difference, xor. So perhaps we can rework the other packages to depend on this library (polyclip-ts is slow with its bignumber implementation, and we'd have to fork it if we wanted to drop that). Because it's a port of a mature C library, it already comes with a ready-made test suite.

In the current state of this PR, the test suite fails to run, and 3 of the output fixtures look weird/broken (issue#783, negative-buffer, polygon-with-holes), but I'm pretty sure this is just me not understanding how to use the library.

Here's some numbers I grabbed with the work in progress version. Dropping @turf/jsts saves us ~26 of the total turf.min.js size, and it benchmarks faster (sometimes by a lot!).

```
before (turf.min.js 537847 bytes)
antimeridian x 54,630 ops/sec ±1.83% (94 runs sampled)
feature-collection-points x 10,855 ops/sec ±0.56% (91 runs sampled)
geometry-collection-points x 16,381 ops/sec ±0.53% (96 runs sampled)
issue-#783 x 19,396 ops/sec ±0.53% (97 runs sampled)
issue-#801-Ecuador x 28,810 ops/sec ±0.63% (96 runs sampled)
issue-#801 x 28,521 ops/sec ±0.80% (98 runs sampled)
issue-#815 x 43,879 ops/sec ±0.62% (94 runs sampled)
issue-#900 x 9,404 ops/sec ±0.54% (99 runs sampled)
issue-#916 x 24,485 ops/sec ±0.61% (95 runs sampled)
linestring x 37,909 ops/sec ±0.72% (98 runs sampled)
multi-linestring x 7,135 ops/sec ±0.65% (98 runs sampled)
multi-point x 22,585 ops/sec ±0.43% (95 runs sampled)
multi-polygon x 13,294 ops/sec ±0.57% (99 runs sampled)
negative-buffer x 47,031 ops/sec ±1.59% (93 runs sampled)
north-latitude-points x 15,976 ops/sec ±1.05% (96 runs sampled)
[...]

PR #2994, opened 2026-01-05
Migrate @turf/buffer to TypeScript, ignoring errors

mfedderly:
I'd like to do this [more properly](https://github.com/Turfjs/turf/pull/2991), but that will require a major version release.

In the meantime, we can do a low-effort job that gets us migrated away from Javascript-first packages and consolidate on TypeScript instead to simplify some of the build tooling.


smallsaucepan (approved):
Thanks @mfedderly. One more hurdle down.

plymer:
Hey, this is a fantastic effort. I know the plan is to move away from jsts, but any chance that in the meantime #2937 could get rolled into this update? I'd be happy to re-do the PR in whatever typescript file comes out of your work 🙂

mfedderly:
@plymer I'm trying to focus on https://github.com/Turfjs/turf/pull/2991 instead of this, and I'm hesitating to expose these configurations until we know we're settled on a library that *does* provide this. But I am intending to add this eventually.

Unfortunately I'm less convinced that clipper2-ts is a library we want to depend on going forward, so I might go and port clipper2 from scratch myself before we can land that.

plymer:
Cheers, I appreciate the update and the roadmap ❤️

bratter:
Is your lack of desire to use clipper2 as-is an ergonomics/efficiency thing with all the transformation?

mfedderly:
Sadly no, I have concerns about accuracy past `Number.MAX_SAFE_INTEGER`, and after some rounds of issues/pr's in that repo I'm not sure its reliable enough to bring on as a dependency for us.

smallsaucepan:
Converting to draft as it's likely we'll favour #2991 instead

mfedderly:
Going to officially close in favor of #2991

mfedderly:
#2991 actually requires a breaking API change (steps -> arcTolerance), so for v7 we probably just need to do this to unblock some other infrastructure work we'd like to do (this is the last javascript-first package, everything else is TypeScript).

plymer:
Hey there, is the feeling still that you want to hold off exposing the jsts endCapStyles? I recognize that you intend for breaking changes and might have a different API surface planned.

I'd still be happy to port my PR to your TypeScript implementation if you think it would be alright in the interim but totally understandable if you'd prefer not to go that route.

mfedderly:
Unfortunately we can't merge #2991 because of the breaking api change involved with the exposed implementation detail of `steps`, which is not supported in geoclipper2 (it uses `arcTolerance` instead and there isn't a real way to map between the two as far as I know). So the new buffer implementation is stuck waiting for a major version (no ETA), and I'm reluctant to change the current version because I don't want to add another breaking change.

Sorry I know this is unfortunate @plymer

plymer:
Ah no worries! I very much appreciate all the work going on here... it's made my job much simpler and I'm still rocking the patch I made for this niche case :)

PR #2995, opened 2026-01-05
Documentation updates for v8

mfedderly:
Based on how things are going with v7, I'd like to make several policy changes when we finally make the next major release. These changes should hopefully let us ship new code without being quite so restricted by compatibility demands.

Specifically:
- turf.min.js should have a rolling target of some reasonable level of browser support
- All other published code is declared as targeting esnext, and therefore requires transpilation and polyfilling by consumers
- Even though we're targeting esnext, we won't break nodejs compatibility without a major and therefore shouldn't be hitting the most bleeding edge syntax anyhow
- We officially state that we do not promise more than 6 digits of precision (~10cm error), which gives us better guidance when tuning our own implementations

This is intended as a starting point for all of the wording, happy to receive feedback and make this better.

cc @bratter


README.md
```diff
@@ -26,18 +26,39 @@ Or explore the Turf API and examples at [turfjs.org](https://turfjs.org/).
...
-Turf is currently published to work with any reasonably modern Javascript runtime.
+Turf is currently published to work with any reasonably modern JavaScript runtime.
 
 Node is a first class citizen, and we recommend using an [Active or Maintenance LTS](https://nodejs.org/en/about/previous-releases) release.
 
 Other runtimes, such as Deno or Bun, are not officially supported. We would be very interested to hear your experiences though.
 
+## Precision
+
+Although we make a reasonable effort for accuracy in our implementations, we are ultimately limited by the fact that we are handling GeoJSON data.
```
bratter: I would always use "precision" rather than accuracy given that the two concepts are subtly different.

README.md
```diff
@@ -26,18 +26,39 @@ Or explore the Turf API and examples at [turfjs.org](https://turfjs.org/).
...
+For lots of consumer devices, GPS accuracy is within a few meters of the actual point. When using latitude and longitude, 6 digits after the decimal
+works out to an [error around 10cm](https://datatracker.ietf.org/doc/html/rfc7946#section-11.2). Although libraries exist to provide arbitrary
+precision math, we try to use JavaScript's built in numeric type (64 bit floating point) for performance reasons.
```
bratter: Consider "we use JavaScript's built in types for performance reasons."

I also wonder if we could add a suggestion along the lines of: "Where operations don't depend directly on lng/lat coordinates, it is often possible to increase the *effective* precision on small shapes by first transforming to a zoomed in projection, running an operation, then re-projecting to lng/lat." i.e., help people understand that there are alternatives. Just a thought.

bratter (commented):
Couple of thoughts on the precision topic. Don't feel well equipped to comment on the versioning question given my use cases, but seems reasonable.

mfedderly:
We should also make it clear that we may bring in libraries that *also* need to be handled. Interestingly the first candidate here would be clipper2-ts, which can't really safely be polyfilled with the BigInt syntax.

Issue #721, opened 2017-05-08
Getting error while finding difference between a multipolygon and a polygon


amishas157:
I am getting following error while finding the difference betweeen a polygon and a multipolygon.

```
/Users/AmishaSingla/Documents/osm-adiff-compare/node_modules/@mapbox/osm-compare/node_modules/@turf/difference/node_modules/jsts/dist/jsts.min.js:14
ln.checkValid(this.edgeList.getEdges()),this.graph.addEdges(this.edgeList.getEdges()),this.computeLabelling(),this.labelIncompleteNodes(),this.findResultAreaEdges(t),this.cancelDuplicateResultEdges();var n=new Sn(this.geomFact);n.add(this.graph),this.resultPolyList=n.getPolygons();var i=new wn(this,this.geomFact,this.ptLocator);this.resultLineList=i.build(t);var r=new Ln(this,this.geomFact,this.ptLocator);this.resultPointList=r.build(t),this.resultGeom=this.computeGeometry(this.resultPointList,this.resultLineList,this.resultPolyList,t)},labelIncompleteNode:function(t,e){var n=this.ptLocator.locate(t.getCoordinate(),this.arg[e].getGeometry());t.getLabel().setLocation(e,n)},copyPoints:function(t){for(var e=this.arg[t].getNodeIterator();e.hasNext();){var n=e.next(),i=this.graph.addNode(n.getCoordinate());i.setLabel(t,n.getLabel().getLocation(t))}},findResultAreaEd
```
The polygon and multipolygon are following: 
- http://bl.ocks.org/d/44385641e432c52cdfd548218454fd42
- http://bl.ocks.org/d/28768027d930ce93f0798fa6f6463fad

Source Code:
https://jsfiddle.net/tboq1cv5/

Not sure if the input to difference method is right or not. 

cc @rowanwins @DenisCarriere

rowanwins:
Hi @amishas157 

This is likely due to the same issue previously around `jsts` not supporting multipolygons very well.

Also as an FYI - in jsfiddle you can add an external resource like the minified version of turf with the url `https://npmcdn.com/@turf/turf/turf.min.js`. then rather the `require` failing you can just say `turf.difference(feature, feature2)`. Check out [this example](https://jsfiddle.net/yjctqtc6/) and open the console to see the result, In this case I just took one part of your multipoly and it works fine, so yeah Im pretty sure its the multipoly thing.

Cheers

amishas157:
[...]

PR #2997, opened 2026-01-10
Ported union, difference, and intersect to clipper2-ts to improve performance

### LINKED ISSUE #2833: turf.intersect(poly1, poly2) crashes with an OOME
Running this code on v7.2.0 hangs and eventually ends in a OOME

```
import * as turf from '@turf/turf';
const poly1coords = [
  [-2.232359, 53.349714],
  [-2.23244, 53.349848],
  [-2.232526, 53.349986],
  [-2.23262, 53.350125],
  [-2.232699, 53.350257],
  [-2.232757, 53.350349],
  [-2.232762, 53.350357],
  [-2.232771, 53.350365],
  [-2.232777, 53.350369],
  [-2.232784, 53.350372],
  [-2.23279, 53.350374],
  [-2.232837, 53.350365],
  [-2.232833, 53.35036],
  [-2.232802, 53.350318],
  [-2.232359, 53.349714],
];
const poly2coords = [
  [-2.232359, 53.349714],
  [-2.23244, 53.349848],
  [-2.232526, 53.349986],
  [-2.23262, 53.350125],
  [-2.232699, 53.350257],
  [-2.232757, 53.350349],
  [-2.232762, 53.350357],
  [-2.232771, 53.350365],
  [-2.232777, 53.350369],
  [-2.232784, 53.350372],
  [-2.23279, 53.350374],
  [-2.232834, 53.350366],
  [-2.23283, 53.350359],
  [-2.232808, 53.350326],
  [-2.23281, 53.350329],
  [-2.232802, 53.350318],
  [-2.232359, 53.349714],
];

const poly1 = turf.polygon([poly1coords]);
const poly2 = turf.polygon([poly2coords]);

const intersection = turf.intersect(turf.featureCollection([poly1, poly2]));
```

The polygons are thin strips of land, almost entirely overlapping apart from the top-left corner:

Image

I've tried running this in different environments (in a Node app, on a browser) and it hits the same error

### LINKED ISSUE #2851: Performance regression in @turf/union from v7.1.0 to v7.2.0
There is a significant performance regression in the @turf/union module when upgrading from version 7.1.0 to 7.2.0. Union operations are more than 10 times slower in 7.2.0.

**Reproduction Steps**
1. Create a test file (e.g., test-union.js) with the following content:

```
import { circle } from '@turf/circle';
import { featureCollection } from '@turf/helpers';
import { union } from '@turf/union';

const circles = [];
for (let i = 0; i < 1000; i++) {
  circles.push(circle([i * 0.1, 0], 200));
}

console.time('union');
union(featureCollection(circles));
console.timeEnd('union');
```
2. Install Dependencies with v7.1.0:
`npm install @turf/circle @turf/helpers @turf/union@7.1.0`
Run the test file:
`node test-union.js`
Observe the timing output (around 500 ms).

3. Upgrade to version 7.2.0:
`npm install @turf/union@7.2.0`
Run the test file again:
`node test-union.js`
Note the increased timing output (around 5.5 s).

**Expected Behavior:**
The union operation should complete in a similar timeframe across these minor versions (around 500 ms for the provided test).

Observed Behavior:
	•	v7.1.0: ~500 ms
	•	v7.2.0: ~5.5 s

This performance degradation is critical, especially for workflows involving large numbers of geometries.

**Environment:**
	•	Node.js version: 22
	•	OS: macOS, Ubuntu

**Additional Context:**
This slowdown might be due to changes in the union implementation or dependencies in v7.2.0. Any investigation or fixes to address this regression would be greatly appreciated.

### LINKED ISSUE #2955: turf.difference doesn't terminate (infinite loop)
Facing a similar issue to a previous issue I raised
[https://github.com/Turfjs/turf/issues/2892](https://github.com/Turfjs/turf/issues/2892)

Upgrading to
```
"@turf/turf": "^7.2.0",
```
Fixed that one case of inputs however for this it doesn't resolve

```
import * as turf from '@turf/turf';
import { Feature, GeoJsonProperties, MultiPolygon, Polygon } from 'geojson';

const poly1 = {"type":"Feature","properties":{},"geometry":{"type":"Polygon","coordinates":[[[-92.47571,42.71498],[-92.47532,42.71495],[-92.47535,42.71421],[-92.47515,42.71349],[-92.47469,42.71323],[-92.47452,42.71135],[-92.47335,42.71131],[-92.47306,42.71123],[-92.473,42.71104],[-92.47285,42.70978],[-92.47336,42.70849],[-92.47344,42.70545],[-92.47322,42.70482],[-92.47232,42.70426],[-92.47207,42.70387],[-92.47219,42.70043],[-92.45653,42.70042],[-92.45618,42.70054],[-92.45617,42.70811],[-92.45637,42.70858],[-92.46089,42.70862],[-92.46087,42.71126],[-92.46057,42.71161],[-92.46044,42.71215],[-92.4607,42.71283],[-92.46089162382165,42.714197499528396],[-92.46233,42.71402],[-92.46312,42.71395],[-92.46392,42.7139],[-92.46865,42.71389],[-92.46870945733048,42.7138914499349],[-92.4688,42.71365],[-92.46887,42.71364],[-92.46887,42.71351],[-92.46888,42.71344],[-92.4689,42.71352],[-92.46888,42.71362],[-92.4689,42.71374],[-92.46899,42.71388],[-92.46939,42.71391],[-92.46972,42.71398],[-92.46954554,42.71394299],[-92.46984019,42.71398775],[-92.47003,42.714],[-92.47024026,42.71404852],[-92.47025,42.71405],[-92.47047394300176,42.71409653854567],[-92.47075,42.71413],[-92.47098495,42.71420272],[-92.47102,42.71421],[-92.47133963,42.7142857],[-92.47117,42.71426],[-92.4715,42.71431],[-92.4717302922405,42.71437823473792],[-92.47182101859093,42.71439971782684],[-92.47219,42.71447],[-92.47274,42.7146],[-92.47296401,42.71466637],[-92.47321483,42.71470267],[-92.47314,42.7147],[-92.47342,42.71471],[-92.47367,42.71476],[-92.47371545000003,42.71477818],[-92.4738516,42.7148096],[-92.47394,42.71482],[-92.47409,42.71486],[-92.47452012,42.71491103],[-92.47487,42.71492],[-92.47499884,42.71493741],[-92.47511,42.71494],[-92.47518205445736,42.71496217060226],[-92.47524,42.71497],[-92.47546367,42.71496791],[-92.47551603,42.71496508],[-92.47571,42.71498],[-92.47571,42.71498]]]}} as Feature<Polygon, GeoJsonProperties>
const intersection = {"type":"Feature","properties":{},"geometry":{"type":"MultiPolygon","coordinates":[[[[-92.47571,42.71498],[-92.47532,42.71495],[-92.47535,42.71421],[-92.47515,42.71349],[-92.47469,42.71323],[-92.47452,42.71135],[-92.47335,42.71131],[-92.47306,42.71123],[-92.473,42.71104],[-92.47293778625954,42.71051740458015],[-92.47293,42.71086],[-92.47301,42.71186],[-92.47298,42.71201],[-92.4728,42.71231],[-92.47236,42.71232],[-92.47217,42.71245],[-92.47214,42.71252],[-92.47224,42.713],[-92.47225,42.71332],[-92.47222,42.71388],[-92.47215,42.71418],[-92.4724184937466,42.71452400761283],[-92.47274,42.7146],[-92.47296401,42.71466637],[-92.47321483,42.71470267],[-92.47314,42.7147],[-92.47342,42.71471],[-92.47367,42.71476],[-92.47371545,42.71477818],[-92.4738516,42.7148096],[-92.47394,42.71482],[-92.47409,42.71486],[-92.47452012,42.71491103],[-92.47487,42.71492],[-92.47499884,42.71493741],[-92.47511,42.71494],[-92.47518205,42.71496217],[-92.47524,42.71497],[-92.47546367,42.71496791],[-92.47551603,42.71496508],[-92.47571,42.71498]]],[[[-92.47344,42.70545],[-92.47341760660248,42.70538587345254],[-92.47334,42.70553],[-92.47336,42.70592],[-92.47326,42.70695],[-92.47332,42.70765],[-92.4733,42.70827],[-92.4732,42.70883],[-92.4729,42.7095],[-92.47285,42.70972],[-92.47284,42.70998],[-92.472911,42.7102924],[-92.47285,42.70978],[-92.47336,42.70849],[-92.47344,42.70545]]]]}} as Feature<Polygon | MultiPolygon, GeoJsonProperties>

const difference1 = turf.difference(
turf.featureCollection([
    turf.truncate(poly1, { precision: 8 }), 
    turf.truncate(intersection, { precision: 8 })
]),
);
```
If I change the precision to 4 or 6 it terminates and produces a result.

as mentioned I am using
```
"@turf/turf": "^7.2.0",
```

smallsaucepan:
Addresses performance issues in these three packages introduced in v7.2.0. 

Performance improvement summary:

| function | improvement |
|-|-|
| difference | 20x to 130x |
| intersect | 55x to 60x |
|union | 10x to 120x |

Requires a custom rollup config to support es5 minified output (clipper2-ts requires bigint, which needs to be polyfilled with the JSBI library before bundling). This should only be a temporary measure until we move to a more modern target on a major version release.

Inclusion of JSBI into minified output results in approx 40% bundle size increase. Feel performance is causing more pain than bundle size though, so still consider worth while. And as mentioned above, it won't be forever.

Adds a non-published "internal" module for shared code. Can be used for other packages as a way to avoid copy-pasting code.

Resolves #2851
Resolves #2833
Resolves #2955

- [x] Summary of the changes.


bratter:
This is awesome. Thank you!! It's going to help me for sure :-). Also, I guess the build size increase is before jsts is removed from the buffer change? Which will take the size back down again... then even better after a major version!

My one question is about the defensive winding checking. It certainly might be the best thing to do for robustness, BUT running `Clipper2.AreaD` on every input polygon adds quite a lot of overhead for a defensive transformation where if the polygon is correctly wound it will just work (note that Clipper and GeoJSON use the same winding direction). Given mfedderly's comments on defensive transformation in #2949, I am wondering if it is the right way to go. I do appreciate that if the tests passed before it wouldn't be possible to break them in a minor, but maybe at least worthwhile testing if rewind is more efficient that areaD?

IMHO, but feel free to ignore, the comments in the enforceOuterRing and enforceInnerRing functions are also a little unclear. I am almost certain (Per comment here: https://github.com/Turfjs/turf/pull/2991#issuecomment-3709035936) that the winding order is the same. The reason why the rewind is needed in buffer is that the azimuthal transform in buffer flips the y which changes the winding order. The doc comment above those functions is implying the opposite.

The other question that this begs is should the test cases have improper winding? Several do at the moment, and some of them (@turf/union's not-overlapping test case for instance) even has one polygon wound one way and another the opposite in the same FeatureCollection.

smallsaucepan:
> My one question is about the defensive winding checking ... adds quite a lot of overhead for a defensive transformation

Long term I'm in the same camp. Turf will be simpler to maintain if we don't have to cater to sloppy geojson. If it's not compliant we should push back on bug reports, similar to what we're starting to do with unrealistically precise coords. That said, best to wait for a major release to put in big red letters "we're not going to be holding your hand anymore".

> maybe at least worthwhile testing if rewind is more efficient that areaD

Great idea. I'll compare them.

> The other question that this begs is should the test cases have improper winding?

Probably not (unless that's something we're testing for like in a linter). It's probably something that crept in over time and we should avoid it going forward. Would be an interesting audit!

> I am almost certain (Per comment here: https://github.com/Turfjs/turf/pull/2991#issuecomment-3709035936) that the winding order is the same.

Will double check this also. I recall reading somewhere that clipper2-ts (at least for union, etc) didn't strictly care about the order so long as you were consistent within the same invocation.

packages/internal/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@turf/internal",
+  "version": "7.3.1",
+  "private": true,
```
mfedderly: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#private

You can't depend on this from other packages if you mark it as private, they won't be able to be work because one of their dependencies will be missing.

packages/internal/clipper2/index.ts
```diff
@@ -0,0 +1,165 @@
+import { PolyPathD, PathD, PathsD, PolyTreeD, areaD } from "clipper2-ts";
+import { Polygon, MultiPolygon, Position } from "geojson";
+
+const DEFAULT_PRECISION = 8;
```
mfedderly: This probably needs a rename to make it clear that this is for coordinates in lng, lat. If you project the coords into something with meters as their unit, you probably want 2 instead.

packages/turf-difference/test/out/issue-#721-inverse.geojson
```diff
@@ -92,42 +92,6 @@
...
-          [
-            [
-              [-0.6460471709105139, 44.80564743856641],
-              [-0.6460421, 44.8056729],
-              [-0.6460471709104708, 44.80564743856663],
-              [-0.6460471709105139, 44.80564743856641]
-            ]
-          ]
-        ]
-      }
```
mfedderly, line 130: Was this supposed to disappear? Did you check all of the output for visual sameness at least?

packages/turf-intersect/test/out/Intersect1.geojson
```diff
@@ -23,10 +20,7 @@
     },
     {
       "type": "Feature",
-      "properties": {
-        "fill-opacity": 0.5,
-        "fill": "#00F"
-      },
```
mfedderly, line 29: Did you run prettier on the output geojson? I've been using `pnpm prettier --write test/out/* test/in/*` when I change test fixtures

packages/turf-intersect/test/out/issue-1394.geojson
```diff
@@ -88,24 +88,6 @@
...
-        "type": "Polygon",
-        "coordinates": [
-          [
-            [-85.42358073, 41.76885781],
-            [-85.42357713028528, 41.76878899545362],
-            [-85.42357713, 41.768789],
-            [-85.42358073, 41.76885781]
-          ]
-        ]
-      }
```
mfedderly, line 108: Another suspiciously deleted output

packages/turf/babel.config.json
```diff
@@ -3,7 +3,8 @@
     [
       "@babel/preset-env",
       {
-        "targets": "> 0.25%, last 2 versions, fully supports es5, not dead"
+        "targets": "> 0.25%, last 2 versions, fully supports es5, not dead",
+        "modules": false
```
mfedderly: https://babeljs.io/docs/babel-preset-env#modules

"Setting this to false will preserve ES modules." I think that we can't do that

packages/turf/rollup.config.js
```diff
@@ -13,10 +19,50 @@ export default [
     input,
     output: [{ file: pckg.browser, format: "umd", name: "turf" }],
     plugins: [
-      commonjs(),
+      alias({
```
mfedderly: This whole thing seems pretty rough, perhaps we'd be better off just using `pnpm patch` to directly introduce jsbi syntax just for @turf/turf's clipper2-ts build. I'll give that a shot today.

I also ran into the fact that just adding babel-plugin-transform-bigint-to-jsbi for our rollup here didn't work.

packages/internal/clipper2/index.ts
```diff
@@ -0,0 +1,165 @@
...
+
+/**
+ * Converts a multipolygon to a flattened array of clipper2 paths.
+ */
+function multiPolygonToPaths(coords: Position[][][]): PathsD {
+  const paths: PathsD = [];
+
+  for (const polygon of coords) {
+    paths.push(...polygonToPaths(polygon));
```
mfedderly: if polygonToPaths has like 1024 entries or something in it, this may break. Its safer to do a `for (const ring of polygonToPaths(polygon)) { paths.push(ring); }`

It does create an extra temporary array, but its only O(number of rings) so that's probably fine because its not O(number of coordinates)

packages/internal/src/clipper2/index.ts
```diff
@@ -0,0 +1,165 @@
...
+ * expectations. That is, counter clockwise.
+ */
+function enforceInnerRing(path: PathD): PathD {
+  if (areaD(path) > 0) {
+    // Leave original array untouched.
+    return [...path].reverse();
+  }
+
+  return path;
+}
```
mfedderly, line 69: Agreed that we should start telling the users to get the winding right before passing them in, but that's likely a major version issue.

packages/internal/clipper2/index.ts
```diff
@@ -0,0 +1,165 @@
...
+}
+
+/**
+ * Make sure this ring is wound as an outer ring, according to clipper2
+ * expectations. That is, clockwise.
+ */
+function enforceOuterRing(path: PathD): PathD {
+  if (areaD(path) < 0) {
+    // Leave original array untouched.
+    return [...path].reverse();
```
mfedderly: You should be able to mutate path instead, because ringToPath already creates your own internal representation.

packages/internal/src/clipper2/index.ts
```diff
@@ -0,0 +1,165 @@
...
+      type: "Polygon",
+      coordinates: polygons[0],
+    };
+  }
+
+  // If anything else return as MultiPolygon
+  return {
+    type: "MultiPolygon",
+    coordinates: polygons,
+  };
```
mfedderly, line 109: I kind of want to just always return MultiPolygon's even if there's only 1 polygon, but again... major version.
We still have to consider the case where we get 0 polygons back and whether *that* should be a MultiPolygon.

packages/internal/clipper2/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  for (let i = 0; i < polyPath.count; i++) {
+    const child = polyPath.child(i);
+    if (child && child.isHole) {
+      const holeRing = pathToCoordinates(child.poly);
+      if (holeRing.length > 0) {
+        rings.push(holeRing);
+      }
+
+      // Expectation is pools within islands within lakes within continents ...
+      // are handled as multipolygons. So further recursion is not required.
```
mfedderly: Have you tested this to make sure that it doesn't nest new outer rings inside holes?

packages/internal/src/clipper2/index.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  // GeoJSON requires the first and last coordinates to be identical (closed ring)
+  if (coords.length > 0) {
+    const first = coords[0];
+    const last = coords[coords.length - 1];
+    if (first[0] !== last[0] || first[1] !== last[1]) {
+      coords.push([first[0], first[1]]);
+    }
+  }
```
mfedderly, line 178: I think this is fine, as long as we never do it with a LineString, but everything going through here should be Polygon rings.

packages/turf-difference/test/out/issue-#721-inverse.geojson
smallsaucepan: I don't use REGEN, so yes visually check and try to account for the differences. In this test (around since 2017) the removed results were three polygon slivers (effectively lines) that were being returned and causing errors downstream. They're regression tests for libraries we don't use any more, so would it be clearer to remove the test entirely?

packages/turf-intersect/test/out/issue-1394.geojson
smallsaucepan: Same as 721-inverse above. A _wafer thin_ polygon being returned by a previous clipping lib that would blow up downstream turf functions.

packages/internal/package.json
smallsaucepan: Ugh. Yes you're right, we can't drop tsup _and_ keep this private. I think there's a valid need for a package where we can store shared code so will make public and document clearly it's not for direct usage, API may change at any time, etc.

packages/turf-intersect/test/out/Intersect1.geojson
smallsaucepan: Yep, at the editor level on save. And then again when lint-staged runs. Just ran manually then and didn't change anything. This is correct yes?

packages/turf/rollup.config.js
smallsaucepan: No, the pnpm patch looks like it's about 1000% cleaner. I'll get rid of all this and cherry pick that from your PR instead. Thanks!

packages/turf-intersect/test/out/Intersect1.geojson
mfedderly: Interesting. I was surprised to see it go from multi-line to single line here. 🤷 
My stuff *doesn't* run prettier on save at the moment and its driving me nuts.

packages/turf-difference/test/out/issue-#721-inverse.geojson
mfedderly: Sounds good about the extra feature removal. I guess with the isobands, isolines, and buffer changes I was just lucky enough to have the same geojson shape before and after, it didn't really change the structure just the coordinates. If you think these tests will help us regress Turf in the future I'm happy to keep them, or drop them otherwise.

I swear GitHub used to have a nice visual diff for .geojson files that we could use in PRs, but it seems to have disappeared. We desperately need better tooling for situations like this where all of the output changes but not in a meaningful way. Its so much easier to just look visually than try to reason about if two polygons switched order, or a ring got rotated around by a few points.

packages/internal/package.json
mfedderly: `@turf/clipper2-support` or similar?

packages/internal/src/clipper2/index.ts
smallsaucepan: Would make for a simpler interface. Plenty of options for what to return for 0, depending on how many liberties we're willing to take. MultiPolygon with empty coordinates array (not kosher from what I understand but used in the wild). Feature with null geometry (more correct, but back to where we started as far as return type conditional logic). Straight up _null_ as we are now?

packages/internal/package.json
smallsaucepan: Makes more sense to me to keep it general purpose so we don't have to publish a new package for every bit of common code we want to share between packages.

packages/internal/src/clipper2/index.ts
smallsaucepan: Added a parameter to at least highlight to future generations that it's closing the ring by default.

nx.json
```diff
@@ -6,7 +6,11 @@
       "{projectRoot}/package.json",
       "{projectRoot}/tsconfig.json"
     ],
-    "sources": ["{projectRoot}/index.{js,ts}", "{projectRoot}/lib/**"]
+    "sources": [
+      "{projectRoot}/index.{js,ts}",
+      "{projectRoot}/lib/**",
+      "{projectRoot}/src/**"
```
smallsaucepan, line 12: Added this under `src` rather than reusing `lib` as feel copy pasting external source into lib is an anti-pattern we may like to keep easily auditable in turf-* packages.

packages/internal/clipper2/index.ts
smallsaucepan: Altered this to `TURF_CLIPPER2_SCALE_FACTOR` but still not super happy with it. Scale factor because that's how the clipper2 docs refer to it. What are your thoughts?

When you say "if you project the coords ..." I know that's possible, though is that something we would need to do in Turf? Should we maybe abstract that detail away within a TurfClipper class that turf-* packages simply no-arg instantiate and pass decimal degrees to?

packages/internal/clipper2/index.ts
smallsaucepan: I had though clearly not well enough. Worked ok for union, but bombed on difference with nested oceans -> continents -> lakes -> islands. Should cover those cases now, with some meatier tests added as well.

smallsaucepan:
Moving to draft status while we figure out if we're using clipper2-ts.

mfedderly:
When revisiting this PR with the geoclipper2 implementation, we should also look at https://github.com/Turfjs/turf/issues/3168 as a test case

Issue #3006, opened 2026-01-26
@turf/lineOffset return type does not keep the geojson type anymore


tahini:
#2974 converted the line-offset package to typescript, but the generated return type of the function is not the same as previously and causes compilation errors in projects using it.

The function signature was previously:
```
declare function lineOffset<T extends LineString | MultiLineString>(
  line: Feature<T> | T,
  distance: number,
  options?: {
    units?: Units;
  }
): Feature<T>;
```

Which preserved the LineString or MultiLineString type of the feature in parameter.

But now it is:
```
declare function lineOffset<T extends LineString | MultiLineString>(geojson: Feature<T> | T, distance: number, options?: {
    units?: Units;
}): Feature<LineString, {
    [name: string]: any;
} | null> | Feature<MultiLineString, {
    [name: string]: any;
} | null>;
```
So when passing a LineString, the return value is not necessarily a LineString anymore.

smallsaucepan:
Thanks for raising this. We'll look at this right away.

@mfedderly would you mind advising?

PR #3009, opened 2026-01-27
Fix @turf/line-offset's return type regression

mfedderly:
The migration to TypeScript removed the explicit return type (accidentally), which broke types. This reintroduces the return annotation to fix the regression.

Fixes https://github.com/Turfjs/turf/issues/3006


mfedderly:
Thanks! Sorry for the blip 😞

PR #3010, opened 2026-01-30
@turf/boolean-contains: Add MultiPolygon support for non-polygon types

rexxars:
Extends `booleanContains` to support MultiPolygon as the containing geometry for additional geometry types. Previously, `booleanContains(MultiPolygon, X)` only worked when X was a Polygon.

New supported combinations:

- MultiPolygon contains Point
- MultiPolygon contains MultiPoint
- MultiPolygon contains LineString
- MultiPolygon contains MultiLineString
- MultiPolygon contains MultiPolygon

Follows DE-9IM "contains" semantics:

- Points must be in the interior (not on boundary) of at least one polygon
- LineStrings must be fully contained within a single polygon (cannot span the gap between polygons)
- Each component of multi-geometries must be contained within some single polygon, but different components can be in different polygons

**Questions:**

1. `booleanContains` documentation states _booleanContains(a, b) is equivalent to booleanWithin(b, a)_ - if that's true, why does `booleanWithin` not import the `booleanContains` module and swap the argument order instead of redefining all the logic?
2. Are there any rules on what should/shouldn't be exported from these modules? The contains module exports both methods without jsdoc, and ones _with_ jsdoc marked as `@private` (such as `compareCoords`). I wasn't sure if I should export the new helper methods or not, and if they should be marked as private


smallsaucepan:
Thanks heaps for putting this together @rexxars. To some of your questions:

> why does booleanWithin not import the booleanContains module and swap the argument order instead of redefining all the logic?

We asked ourselves this recently, and discovered some performance differences for what should be equivalent approaches. Before pressing ahead, we wanted to understand why that was happening first. Though yes, ideally we would have a single implementation that served both operations.

> Are there any rules on what should/shouldn't be exported from these modules? The contains module exports both methods without jsdoc, and ones with jsdoc marked as @private

Generally speaking, we try to have each package export a single function (helpers and meta being the exceptions). Any other packages that do that is doing so probably because we've never had a good place to keep common code in the monorepo. We're half way through adding an internal code only package. Suspect between your two PRs there might be some good candidates?

smallsaucepan:
And by way of confirmation this PR will fill in the blue sections below?

Screenshot 2026-01-31 at 23 58 15

rexxars:
> We asked ourselves this recently, and discovered some performance differences for what should be equivalent approaches. Before pressing ahead, we wanted to understand why that was happening first.

Curious! Not a blocker for this PR, was just curious about the context around it

> Suspect between your two PRs there might be some good candidates?

Many of the same methods are implemented in `contains` and `within`, yes - and there's a few other methods that I am sure could be moved to a helper/shared util package as well. The main reason why I was asking is that there are many functions exported from the `contains` package, that are not used by other modules. Some of them are marked as `@private`, others are not. None of them seems to appear in the API docs. Was just trying to figure out if the methods I added in this PR should also be exported:

- isLineInMultiPolygon
- isMultiLineStringInMultiPolygon
- isMultiPointInMultiPolygon
- isMultiPolygonInMultiPolygon
- isPointInMultiPolygon

For reference, these are all the (currently) exported methods from `@turf/boolean-contains`:

- booleanContains
- compareCoords
- doBBoxOverlap
- getMidpoint
- isLineInPoly
- isLineOnLine
- isMultiPointInMultiPoint
- isMultiPointInPoly
- isMultiPointOnLine
- isMultiPolyInPoly
- isPointInMultiPoint
- isPolygonInMultiPolygon
- isPolyInPoly

I could see the rationale for exporting the `is`-prefixed methods, as if you know ahead of time that you're only going to check one geometry against another, you can tree-shake away all the logic for geometries you don't use - but wasn't sure if that was the intention or just a side effect. Other methods, like `compareCoords`, `doBBoxOverlap` and `getMidpoint` seems a bit odd to export from this module however.

> And by way of confirmation this PR will fill in the blue sections below?

Yes, that appears correct.

smallsaucepan:
I would guess it's mostly historical "just in case" exporting. Can you see any drawbacks to the below as a plan:

1. merge this PR
2. get the turf-internals package published
3. hoist everything from compareCoords ... to ... isPolyInPoly into that (maybe keeping a facade in contains and within until a major release to maintain backwards compat in case someone has used those undocumented functions)
4. refactor booleanWithin to use turf-internals too

Not asking you to commit to any of this beyond what you've already done. Just making the most of your recent perusal of those two packages.

rexxars:
Yep, happy with that plan. If I could be slightly selfish, I'd love to have this merged and released and then do some follow-up PRs to address point 2-4 (happy to help with this and more, but they're not backwards incompatible changes). I'm trying to use this library in [groq-js](https://github.com/sanity-io/groq-js) (for our geo functions) and that would unblock that work :)

packages/turf-boolean-contains/index.ts
```diff
@@ -157,12 +292,21 @@ function isMultiPointOnLine(lineString: LineString, multiPoint: MultiPoint) {
...
 
 function isMultiPointInPoly(polygon: Polygon, multiPoint: MultiPoint) {
+  let oneInside = false;
   for (const coord of multiPoint.coordinates) {
-    if (!booleanPointInPolygon(coord, polygon, { ignoreBoundary: true })) {
+    // All points must be inside polygon (boundary OK)
+    if (!booleanPointInPolygon(coord, polygon)) {
       return false;
     }
+    // Track if at least one point is strictly in the interior
```
smallsaucepan, line 312: Would you mind expanding on the reasoning here? I would have expected for a MultiPoint to be contained in a poly _every_ point of that MP is contained.

rexxars: The topological definition of "contains" per the DE-9IM says
that **A** contains **B** requires two conditions:

1. No point of **B** lies in the exterior of **A** (tested on line 298)
2. At least one point of the interior of **B** lies in the interior of **A** (tested below)

If every point of the MultiPoint (B) sits exactly on the polygons boundary (eg. on an edge or vertex), then the interiors of **A** and **B** never intersect.

This is point two of what I mentioned in the PR description:
> Fixes(?) behavior of `booleanContains` to return true if every point is inside/on boundary of some polygon in the MultiPolygon, and at least one is strictly interior. This previously returned false, but differs from `booleanWithin` and doesn't match DE-9IM semantics. In my eyes this is a _bugfix_, not a breaking change

rexxars: **An example:**

Given the following square polygon:

```json
{
  "type": "Polygon",
  "coordinates": [
    [
      [0, 0],
      [1, 0],
      [1, 1],
      [0, 1],
      [0, 0]
    ]
  ]
}
```

And the following MultiPoint:

```json
{
  "type": "MultiPoint",
  "coordinates": [
    [0.5, 0.5],
    [1, 1]
  ]
}
```

Image

In this case, one point is strictly on the interior of the polygon, while one point is on the _boundary_. This previously returned `false`, but the way I see it should return `true`.

rexxars: I realize in hindsight I should have split this into two PRs - one for the bugfix and one for the new additions - let me know if you'd like me to split it up now.

smallsaucepan: Coming around to the idea that it's a bug. Key for me was to consider the interior of the MultiPoint as all points combined and regarded as a single interior. Not a collection of separate points for which the constraints need to hold true individually.

Going to guess this has been an issue for quite a while, so will discuss with the other maintainers how we want to move forward. In the meantime, yes if you could please change this PR to only expand the types of object that can be passed, we'll treat any correctness issues separately.

Well spotted 👏  I think it's going to be worth us reviewing other combos as well.

rexxars: Moved out this fix into #3024 and rebased this PR to only be about the multipolygon support for non-polygon types.

rexxars: @smallsaucepan Anything you wanted changed in this or #3024?

packages/turf-boolean-contains/index.ts
```diff
@@ -103,6 +114,122 @@ function isPolygonInMultiPolygon(multiPolygon: MultiPolygon, polygon: Polygon) {
...
+        { type: "Polygon", coordinates: polyCoords },
+        { ignoreBoundary: true }
+      )
+    );
+    if (!pointInside) {
+      return false;
+    }
+  }
+  return true;
+}
```
mfedderly, line 165: I'm not sure this implementation matches my understanding of the spec (using [wikipedia](https://en.wikipedia.org/wiki/DE-9IM) as the reference here).

`a contains b: geometry b lies in a, and the interiors intersect. Another definition: "a contains b [iff] no points of b lie in the exterior of a, and at least one point of the interior of b lies in the interior of a"`

so I think the loop would be like this (pseudocode):
```
let oneInside = false;
for (each point) {
  if (point outside multipolygon) {
    return false;
  }
  if (!oneInside && point inside multipolygon (not boundary)) {
    oneInside = true;
  }
}
return oneInside;
```

The [de9im implementation](https://github.com/dpmcmlxxvi/de9im/blob/master/src/within/point.js#L49-L63) looks like it agrees with my understanding as well.

mfedderly (commented):
A quick look makes me think there's at least one logic bug in this PR, and the fact that we've got `within` and `contains` as completely separate implementations seems a little rough. I wonder if we should revisit this from a "we should just mirror de9im's implementation" angle before continuing to maintain both of these packages in their current state. I'm not convinced that either of these implementations are actually spec-complaint at the moment.

Thoughts?

packages/turf-boolean-contains/index.ts
rexxars: Verified your reading against JSTS and you're correct - this was the same bug that was fixed for the Polygon path in #3024, and this branch predated that fix, so the new MultiPolygon code had mirrored the old pattern.

I've rebased onto master and applied the #3024 pattern to `isMultiPointInMultiPolygon`: every point must not be in the exterior (boundary-inclusive `booleanPointInPolygon` against the MultiPolygon), and at least one point must be strictly interior. The MultiPointOneOnBoundary fixture moved from false/ to true/ accordingly, and I added an all-points-on-boundary false fixture to pin the "interiors must intersect" side. 

>  the fact that we've got within and contains as completely separate implementations seems a little rough

100%. `contains(a, b) === within(b, a)` by definition, so it seems unnecessary and (as proven here) easy for bugs to slip in to one but not the other. I opened #3011 as a draft specifically because of this.

packages/turf-boolean-contains/index.ts
```diff
@@ -104,6 +115,124 @@ function isPolygonInMultiPolygon(multiPolygon: MultiPolygon, polygon: Polygon) {
...
+      isLineInPoly(
+        { type: "Polygon", coordinates: polyCoords },
+        { type: "LineString", coordinates: lineCoords }
+      )
+    );
+    if (!lineInside) {
+      return false;
+    }
+  }
+  return true;
```
mfedderly: I think that this has a similar false negative problem with a MultiLineString where the first one is on a border and the second one is at least partly inside.

As an aside I looked at the de9im package's implementation of this and they just check the line centroids which makes me a little uncomfortable.

One thing they do seem to do nicely though is upgrade to the Multi equivalents to keep the implementation a little tidier. I think part of why this implementation is tough is because you can't actually reuse the `isLineInPoly` in this way because of the nuance of a multi check actually allowing some border-only geometries along the way.

packages/turf-boolean-contains/index.ts
```diff
@@ -104,6 +115,124 @@ function isPolygonInMultiPolygon(multiPolygon: MultiPolygon, polygon: Polygon) {
...
+      isPolyInPoly(
+        { type: "Polygon", coordinates: poly1Coords },
+        { type: "Polygon", coordinates: poly2Coords }
+      )
+    );
+    if (!polyInside) {
+      return false;
+    }
+  }
+  return true;
```
mfedderly, line 249: I have a similar concern about the border-only case from the LineStrings above, but I think when we're handling Polygons this logic is actually correct.

mfedderly (commented):
Sorry, I think I missed another case that will need some attention, but this time in the MultiLineString implementation.

As an aside: these tests are fantastic (thank you!), and you should totally take credit in the `contributors` section of package.json if you'd like to.

rexxars:
> Sorry, I think I missed another case that will need some attention, but this time in the MultiLineString implementation.

Yep, you were right - my apologies. I was requiring _every_ line to have an interior segment, when the "at least one interior point" condition should apply across the MultiLineString as a whole. A line lying entirely on a boundary is fine, as long as one of its siblings "dips into" the interior.

I've pushed a fix. Since `isLineInPoly` couldn't be reused directly (as you noted, it conflates "not outside" and "has interior" into a single boolean), I split it into a `lineInPolyStatus` helper that classifies a line as `outside`/`boundary`/`interior` against a polygon. `isLineInPoly` itself just checks for `interior` now, so its behavior is unchanged. The MultiLineString check then rejects if any line touches the exterior, and accepts only if at least one line ends up strictly interior somewhere.  Let me know if you think of a better way to handle this.

Added fixtures for both the boundary-plus-interior case (true) and the all-lines-on-boundary case (false), and both agree with JSTS.

> they just check the line centroids which makes me a little uncomfortable

Same - a centroid of a line that wanders in and out of a polygon can land pretty much anywhere, so I steered away from that approach.

One thing worth noting: a line spanning two polygons that touch at a single point would still return false here, since each line is checked against individual polygons rather than the union. Can't happen for properly disjoint MultiPolygons, so I'd suggest deferring that to the de9im-alignment follow-up, maybe? If not, let me know and I can take a stab at it here.

> you should totally take credit in the contributors section of package.json if you'd like to.

Thanks! I'm already in there based on another PR I did earlier, I believe :)

mfedderly (approved):
🎉 thank you!

Issue #2975, opened 2025-12-16
Allow booleanWithin to support MultiLineString and MultiPolygon


blchelle:
Currently both MultiLineString and MultiPolygon are unsupported feature types for the booleanWithin function. However, I it should be possible to do some level of comparison for both of them.

For MultiLineString:
* Cannot be within Point, MultiPoint.
* Can be within LineString, MultiLineString. Here is an [example](https://geojson.io/#data=data:application/json,%7B%22type%22%3A%22FeatureCollection%22%2C%22features%22%3A%5B%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%23FF0000%22%2C%22stroke-width%22%3A4%2C%22stroke-opacity%22%3A1%7D%2C%22geometry%22%3A%7B%22coordinates%22%3A%5B%5B%5B0%2C0%5D%2C%5B0%2C1%5D%5D%2C%5B%5B1%2C0%5D%2C%5B1%2C1%5D%5D%5D%2C%22type%22%3A%22MultiLineString%22%7D%7D%2C%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%2306ff25%22%2C%22stroke-width%22%3A2%2C%22stroke-opacity%22%3A1%7D%2C%22geometry%22%3A%7B%22coordinates%22%3A%5B%5B0%2C0%5D%2C%5B0%2C1%5D%2C%5B1%2C1%5D%2C%5B1%2C0%5D%5D%2C%22type%22%3A%22LineString%22%7D%2C%22id%22%3A1%7D%5D%7D) of a MultiLineString being within a LineString.
    * For this I imagine a function `isMultiLineOnLine` which essentially just flattens the MultiLine into an array of Lines and then calls `isLineOnLine` for all of these lines and returns true if and only if all the flattened lines are on the comparison line.
* Can be within Polygon, MultiPolygon
    * For this I imagine a function `isMultiLineInPoly` which essentially just flattens the MultiLine into an array of Lines and then calls `isLineInPoly` for all of these lines and returns true if and only if all the flattened lines are in the comparison polygon.

For MultiPolygon
* Cannot be within Point, MultPoint
* Cannot be within LineString, MultiLineString
[...]

PR #3011, opened 2026-01-30
@turf/boolean-within: add MultiPolygon support, de-duplicate logic

### LINKED ISSUE #1444: Feature request: Support MultiPolygon booleanWithin Polygon
Hey thanks for the great work on this library.

I was wondering if it would be possible to get support for checking if a MultiPolygon is within a Polygon.

Would it be sufficient to add another case for MultiPolygon in the switch, and call `isPolyInPoly(geom1, geom2)` or is it more complex?
https://github.com/Turfjs/turf/blob/master/packages/turf-boolean-within/index.ts#L69

rexxars:
The `@turf/boolean-within` module is basically a duplicate of `@turf/boolean-contains`, with the argument order swapped. This PR removes the duplicate logic, instead importing the `@turf/boolean-contains` module and calling it directly, with swapped argument orders. This has multiple benefits:

1. Gains support for MultiPolygon now that #3010 has landed
2. Prevents drift/differences between the two. This is not a theoretical concern - the same boundary-semantics bug ended up being fixed _four separate times_ across these two packages: #3024 (`contains`, Polygon path), #3085 (`within`, same bug, independently re-fixed by another contributor months later), and twice more in #3010 for the MultiPolygon paths. With this PR that entire class of bug can only exist in one place.
3. Reduces the bundle size for anyone who should depend on _both_ modules

(This was also suggested by @bratter in [this issue](https://github.com/Turfjs/turf/issues/2975#issuecomment-3721240206) I just found)

**Behavior changes to be aware of:**

Since `booleanWithin` now supports every combination `booleanContains` does, a few combos that previously threw `geometry not supported` now return real answers - e.g. `within(Point, Point)` and the MultiLineString combinations. I'd consider these new features rather than breaking changes.

One subtler one: the old `within` implementation of the Polygon/MultiPolygon check only tested the polygon's outer-ring _vertices_, so a polygon whose vertices sit inside two disjoint polygons of a MultiPolygon - but whose middle crosses the gap between them - would incorrectly return `true`. The delegated implementation requires containment within a single polygon, which the new `MultiPolygonSpanningBothPolygons` fixtures pin down. Same as the boundary case in #3010, the way I see it this is a _bugfix_ rather than a breaking change - but flagging it since it does change observable behavior.

One thing that might be worth addressing: error messages for unsupported geometries are now phrased from `contains`' perspective, so `within(a, b)` with an unsupported `b` reports it as `feature1`. Happy to re-wrap the errors if you think it matters. A naive implementation would be something like:

```ts
try {
  return booleanContains(feature2, feature1);
} catch (err) {
  // booleanContains sees the arguments in reverse order - swap
  // feature1/feature2 in error messages to match the caller's view
  if (err instanceof Error && err.message.startsWith("feature1 ")) {
    throw new Error(err.message.replace("feature1 ", "feature2 "));
  }
  if (err instanceof Error && err.message.startsWith("feature2 ")) {
    throw new Error(err.message.replace("feature2 ", "feature1 "));
  }
  throw err;
}
```

I would argue that one could consider deprecating this entire module in favor of `@turf/boolean-contains` (or the other way around), as keeping the test suites in sync between the two is slightly annoying - but leaving this up to you as maintainers to decide :)

Fixes #1444


bratter:
Nice! My 2c is to not deprecate as they are both commonly accepted de-9im terms.

smallsaucepan:
Thanks for this PR too. Let's discuss on #3010 knowing that applies here as well.

Believe worth keeping both `contains` and `within` as sometimes one expresses the relationship more naturally, saving the user some mental gymnastics.

rexxars:
Rebased this and moved it out of draft state since #3010 landed, @mfedderly :)

mfedderly:
So here's the performance changes (m4 macbook pro, plugged in) on the tests from `master` after the code changes from this PR:
image

Some of these are pretty severe (the red ones are slower than 95% of the original speed), and we should probably figure out what's up that caused these to regress before we merge this. Is that something you'd want to tackle or should I make a PR first and you can rebase yours after? It can probably be implemented as one or more PRs for perf optimization of boolean-contains.

rexxars:
This is actually quite nice to see, as they're pre-existing `contains` costs that `within` now inherits. Meaning we can most likely optimize `contains` with similar tricks as `within` previously did, and both gets the performance win out of it. I do think that there's a couple of these that might be due to (at least partially) `within` not following spec, and `contains` now doing so.

Let me take a stab at some performance optimizations in `contains` before this lands.

rexxars:
There's some perf improvements in https://github.com/Turfjs/turf/pull/3099 - happy if anyone calls out other opportunities to improve!

packages/turf-boolean-within/package.json
```diff
@@ -69,12 +71,8 @@
     "typescript": "catalog:"
   },
   "dependencies": {
-    "@turf/bbox": "workspace:*",
-    "@turf/boolean-point-in-polygon": "workspace:*",
-    "@turf/boolean-point-on-line": "workspace:*",
+    "@turf/boolean-contains": "workspace:*",
     "@turf/helpers": "workspace:*",
```
mfedderly: I think that this can still be moved to devDependencies. I'm not seeing an error when I move it on my local version. What caused you to have to move it back?

mfedderly:
I rebased on top of #3099 and got this:

image

If you see any other obvious stuff you'd like to tackle I'm happy to check out another round of perf work. Otherwise I'm comfortable merging this with the current perf.

packages/turf-boolean-within/package.json
rexxars: Some workflow/test was failing, but I can't remember. Let me revert that part and see what we get from CI.

rexxars: "last-checks" fails:

```
not ok 1 turf-boolean-within @turf/helpers should be located in Dependencies instead of DevDependencies
```

Not sure how you want me to proceed here - given its usage, I'd say it does belong in devDependencies 🤷‍♂️

rexxars:
> If you see any other obvious stuff you'd like to tackle I'm happy to check out another round of perf work.

Yes, I see a couple of ones that could be optimized - https://github.com/Turfjs/turf/pull/3102 is one such, but I have some reservations on the approach and would love your input.

mfedderly:
https://github.com/Turfjs/turf/pull/3105 should fix the devDependencies CI thing for you

mfedderly (approved):
Thank you!

Issue #2998, opened 2026-01-12
@turf/clean-coords - behavior of degenerate polygons, addition of Collection support, and other tweaks


bratter:
There are several items that could be changed to improve the consistency and performance of @turf/clean-coords.

### 1. Degenerate Polygons

Valid (but polygons that are lines) input polygons (and multipolygons) such as the below cause clean-coords to throw due to the fewer than 4 points check [here](https://github.com/Turfjs/turf/blob/6e4c18b38f63aa98c25922621a22c5e687eb52ca/packages/turf-clean-coords/index.ts#L162).

```json
{
  "type": "Feature",
  "properties": {},
  "geometry": {
    "type": "Polygon",
    "coordinates": [
      [[-75.994645,37.95325],[-76.016553,38.95325],[-76.043938,37.95325],[-75.994645,37.95325]]
    ]
  }
}
```

While not really a bug, it isn't ideal behavior. In general, valid input should produce valid output if there is a path to do so (per other recent discussions). Also LineStrings with equivalent degenerate conditions (i.e., doubled up points), work fine, so the behavior is inconsistent between polygons and lines.

I propose making this case not throw and returning a valid polygon with a minimum viable set of points. Other solutions are possible (dropping the shape or reducing its dimension, but I think having clean coords with a contract of "same output shape as input shape" is fair.

### 2. Support FeatureCollection and GeometryCollection types

Clean-coords should support both collection types by simply operating on each contained geometry independently. Current behavior is to throw.

### 3. Multipoint treatment

This is question about consistency rather than a proposal. MultiLineStrings removes overlapping points. The other Multi- geometries don't look for equivalent conditions (identically overlapping. This is at least a minor inconsistency. Is it worth/beneficial to either (a) remove that behavior from points or (b) add it to multi lines and polys?

### 4. Improved types

The type signature of the function can be improved by adding some overloads and other refinements.

### 5. Improved consistency of structural sharing

The implementation currently uses structural sharing differently depending on the type of shape. For instance points are returned as-is, but all other types have a fresh outer wrapper when not mutating. I believe that it would be better to be consistent across all shape types. Also, the level of structural sharing can be increased to improve performance and memory use, especially in "real world" cases with larger shapes.

I propose writing in code then documenting that:

1. When not mutating, the outer object wrapper(s) will always be new objects such that `geojsonIn !== geojsonOut`. This is usually the case now, but let's make it always the case.
[...]

PR #3012, opened 2026-02-01
cleanCoords: Add collection support and typescript fixes

bratter:
Addresses #2998.

## Summary

Adds some features to clean-coords:
- Added support for `FeatureCollection` and `GeometryCollection` types through a recursive call.
- Improves consistency of structural sharing for both mutate options, and documents it (see  note below).
- Improved typscript type signature (see note below).
- Adds the ability to pass an epsilon through to the `booleanPointOnLine` calls.
- Changed the MultiPoint methodology from POJO to Set for substantial speedup.
- Made some other minor cleanups - removing redundant checks, cleaning up the benchmarks file, etc.

## Additional Notes

1. I had originally suggested also changing the behavior such that output polygons with < 4 points should not throw but instead return a valid 4 Position polygon. Changed my mind given that these, while technically valid geojson, are not OGC Simple Feature valid polygons and could lead to downstream choking.
2. Wanted to (a) maximize structural sharing to improve performance wherever possible, and (b) ensure we are clear about what gets preserved vs. not. The basic premise is that:
    - The geojson objects down to the `Geometry` are always reused when mutate=true and always new when mutate=false
    - All properties, id, bbox, are always reused irrespective of mutate value
    - Mutate=false is (as you would expect) guaranteed not to change any coordinates from the original, but otherwise, will reuse or create new or mutate if mutate=true with no specific guarantees
3. The type signature improvement aims to be maximally informative in that it passes the geometry's generic through, requiring that the function always either outputs the same type it receives or throws. If this is too constraining then can back it off, but figured more type inference was better, and it matches the behavior.
4. Benchmarks are variable but more got faster than slower, with the main impacts coming from the structural sharing changes. e.g., the biggest slowdown was in the single point test where we replace the object rather than return it. The better sharing tends to boost performance when few changes are made to larger geometries - i.e., more realistic cases.

## Future Work

I do wonder if you would be interested in a follow up PR that adds a `collapse` option? Reason being I am concerned by the workflow of input > truncate > clean throwing errors for valid input. The collapse option would remove the offending geometry at the ring level rather than throw an error. Thoughts?

---

- [x] Summary of the changes.


bratter:
After initially failed CI fixed the types to add better inference for union types and also put in a deprecated signature matching the old behavior to not break any build systems that might error on failed type resolution. This signature can be remove in v8.

smallsaucepan:
Great additions. Thank you @bratter!

Reckon we should move away from `epsilon` as a config parameter. I've used it (erroneously now I believe) in a couple of places where I'm sure there's a more relatable and accurate parameter for what we're trying to convey. `precision` or `tolerance` might be more meaningful? Thoughts?

Regarding mutation, if this model is used consistently throughout Turf, it might be worth adding a section on it to the website, so we can keep the per-function documentation as brief as possible, and cross reference instead.

Btw, our approach to mutation gives me the heebie-jeebies. To call
```
const f2 = turf.doSomething(f1);
```
and then find out changing f2.properties.whatever changes f1? Know it's a performance thing, though possibly worth documenting more clearly.

Can you please remind me about the collapse idea? Is it basically that truncating a geometry makes it more likely for points to become coincident and from there get cleaned beyond what's technically valid?

packages/turf-clean-coords/index.ts
```diff
@@ -1,18 +1,79 @@
...
+): GeometryCollection<G>;
+function cleanCoords<G extends Geometry>(
+  geojson: G,
+  options?: { mutate?: boolean; epsilon?: number }
+): G;
+function cleanCoords<T extends GeoJSON>(
+  geojson: T,
+  options?: { mutate?: boolean; epsilon?: number }
+): CleanCoordsResult<T>;
+function cleanCoords(
```
smallsaucepan, line 43: Do we need this variation if as well as the one above it?

packages/turf-clean-coords/index.ts
```diff
@@ -1,18 +1,79 @@
...
+  options?: { mutate?: boolean; epsilon?: number }
+): GeoJSON;
+/** @deprecated loosely typed version deprecated. Will be removed in next major version */
+function cleanCoords(
+  geojson: any,
+  options?: { mutate?: boolean; epsilon?: number }
+): any;
 /**
- * Removes redundant coordinates from any GeoJSON Geometry.
+ * Removes redundant coordinates from any GeoJSON Type.
```
smallsaucepan, line 53: "any GeoJSON object" might be better here.

packages/turf-clean-coords/index.ts
```diff
@@ -1,18 +1,79 @@
...
+): G;
+function cleanCoords<T extends GeoJSON>(
+  geojson: T,
+  options?: { mutate?: boolean; epsilon?: number }
+): CleanCoordsResult<T>;
+function cleanCoords(
+  geojson: GeoJSON,
+  options?: { mutate?: boolean; epsilon?: number }
+): GeoJSON;
+/** @deprecated loosely typed version deprecated. Will be removed in next major version */
```
smallsaucepan, line 47: Let's mark as deprecated now but not necessarily commit to when it will be removed.

packages/turf-clean-coords/index.ts
```diff
@@ -114,32 +211,62 @@ function cleanLine(line: Position[], type: string) {
...
+ * @returns {Position<number>} Cleaned coordinates
+ */
+function cleanRings(
+  rings: Position[][],
+  mutate: boolean,
+  epsilon?: number
+): Position[][] {
+  // Re-use the polygon's rings array when mutating to maximize sharing
+  // It would be possible to follow a similar approach to cleanLine and only
+  // create a new raings array when we know something has changed even for
```
smallsaucepan, line 263: Typo "raings"

bratter:
Thanks for the review! Will get to this when I have a sec including responding to your queries.

bratter:
> Reckon we should move away from `epsilon` as a config parameter. I've used it (erroneously now I believe) in a couple of places where I'm sure there's a more relatable and accurate parameter for what we're trying to convey. `precision` or `tolerance` might be more meaningful? Thoughts?

Hmmmm. It seems that the term is also used differently in different places too. in the context of `booleanPointOnLine` `epsilon` is specifically targeted at avoiding floating point error in cross product === 0 comparison... It does not have an intuitive numerical interpretation for coordinates in degrees like "tolerance" implies. You could also argue that a small epsilon (like 1e-15) should be the default comparison anyway given floating point. 

On reflection, the best approach might be to just drop it from the API here again (back to how it was), then consider a change in `booleanPointOnLine` as a separate issue where I might argue for a default epsilon. Will just remove from the options if you agree.

> Btw, our approach to mutation gives me the heebie-jeebies.

True, but the alternatives are really only a deep clone on every function that mutates or immutable data structures... so maybe no alternative. If it were me, I'd have all coordinate mutation functions just mutate by default so that a user would be forced to assume mutation 100% of the time and they do a deep clone if want to preserve the original. Barring that, the current approach is probably best, and my being explicit about it is really just a codification of what was generally happening with some slight consistency fixes. 

If you think this works, I can probably work through all the coordinate mutation methods and check/refine their mutation so they are all consistent, then do as you suggest and take out of the docs here. Let me know if you want to take out of docs now.

> Can you please remind me about the collapse idea?

The two cases that I'd be trying to address is (a) someone passes in a valid polygon, truncates it, then runs `cleanCoords`. This could throw if the truncation means that the cleaning degenerates the polygon. (b) even if the original polygon passed in is 0 area, it might be nice to provide a friction-free way to clean them.

What I would propose for the collapse option is that LineStrings that end with only two co-incident points get dropped instead of passed through, and  polygon rings that end up with 3 or fewer points get dropped instead of throwing. If this completely eliminates the geometry, then we can either return a null or a null geometry in the feature. 

I think this provides a nice way of capturing degenerate geometries during cleaning in a consistent way without the user having to set up try...catch. I suspect many users would just drop the whole shape in such a block, but if, say, the issue is just a small hole in a polygon, you'd likely drop the whole polygon for something very easily fixed.

That may not have made sense... lol. I can make a repro if that'll help. Would likely do as another PR anyway.

bratter:
Hi @smallsaucepan - gentle nudge here - let me know if you are still interested, have any responses to the above, want something minimal that dodges the epsilon issue. If you are still broadly interested in the simpler  fixes / extra type additions I can get rid of the epsilon addition so there is no API surface change except for the additional types. Cheers!

mfedderly:
Regarding the discussion on behavior when mutable is not set:

Does it seem reasonable to explain the (mutable: false) behavior as "If we change something, we'll make a copy but will otherwise reuse the input"? This works similarly to `redux`, `react`, and (kind of) `immer` which are all pretty widely known at this point.

I tend to agree that we don't want to be in the business of deep cloning coordinates arrays because of memory use and overall speed reasons. Someone can already perform that clone operation themselves with another npm library and pass in `mutable: true` if they want that behavior.

I have no idea if we're actually doing this consistently across every package in the codebase, but it does feel like a defensible position to take as a compromise between performance and avoiding mutation.

Thoughts?

packages/turf-clean-coords/index.ts
mfedderly: Right after merging this PR I should create a PR to remove this and add it to the v8 milestone so we don't forget to at least evaluate it then.

packages/turf-clean-coords/index.ts
```diff
@@ -1,18 +1,79 @@
...
-// To-Do => Improve Typescript GeoJSON handling
+/** Generic response type for cleanCoords */
+type CleanCoordsResult<T extends GeoJSON> =
+  T extends Feature<infer G extends Geometry, infer P extends GeoJsonProperties>
+    ? Feature<G, P>
+    : T extends Geometry
+      ? T
+      : GeoJSON;
 
+function cleanCoords<G extends Geometry, P extends GeoJsonProperties>(
```
mfedderly, line 23: Do we actually need all of these definitions? I played around with this version a bit and it may be good enough as a single definition. Maybe I missed some test cases you had in mind though?

```
function cleanCoords<I extends GeoJSON>(
  geojson: I,
  options?: { mutate?: boolean }
): I extends FeatureCollection<infer G, infer P>
  ? FeatureCollection<G, P>
  : I extends GeometryCollection<infer G>
    ? GeometryCollection<G>
    : I extends Feature<infer G, infer P>
      ? Feature<G, P>
      : I;
```

[Typescript Playground](https://www.typescriptlang.org/play/?ssl=201&ssc=1&pln=210&pc=8#code/PQKhCgAIUgVALAppAbgQwDYEsAmrMCuiAzpAGYD2ATpAC5KQBEtAngA6KORtUUdWtIFMpADiiCgCkAygHkAcpADmEgLaJaVFkIBGAK0QBjWsQB0UGPFq02xAFzBgtChQxmsGsqepLgV1RjAVGSGAOwAnAAsAGwAxMRGtFgUAHYAtACMppEWwOCIAB5s1LR07MjiUsSpleqaLLDlpAC8YmoaWgDazOWMALoA3ODgoBDQcAzoGET40yTk1HQMPRxcPHyIAtrCbVJyirL6iWYWkFY29o7Oru6e3lS+-oHBYVFxCcbJ6Vk50HmFxQEZQ4u0k1RSjQ4LVB+26rFWg2Go1OACEKAQUjgsCklJAdBQCqdzrYHE4XG5TB5aF4fH5aAEgiEIjF4okvmkAKy5fJFErA5AotEFSCtTopAiqHSbAA0kHFkplcolUqosvlKr6kAAPpAxcrFeqDfrVUqFSbDebjYiRmBTgBBSAABQoxCwSVSkCwpDQKUgaCoVDQ2xEhhcVCxKTQtBI5nGxMuZJulLutKejNeLI+7vSAGZTFkMvaA0HIMR4OiMHhQylaGhsXiNAB3RCIX20RsUP2YpZUFuQRAYRDqGsncYIZA8RAoZIEUiVGQKUscQxYMhYQxRr5+jAYCjNvCqajIAdD1smSAACkQpiUplljfg6-gkFDBErDcgs8QeGckF7k4SGtIAAWVmIhiAASmlU4dAIUp6GQQwCADM8lyMVd103D1UgwbRMF3RtSAADVlABNWUfTwC8+GzfCWAgyAAC06E7KVIBwRA1xSb9YzGGB5AoaM7CWZB4WQRssB3OVBI-SMAz3b92L3NtOwAAz1M01StbVdQtLTNNNDVVPYmZnFOVR-SUbFMAbbi13PSiRMgDBLOQLBVDYNBjCEEQdF7NAAGtsVxQx4B9FReNOaQ+0nad0VILFiCQ4hXVSUgPQQ-k5X9XgJJxOxTjSSAAG8AAFsBSAKzmsElHCs+gCB0UxQ1UYAABFOOxN0BwacocHazqUm63DIW-YA2DfQIACYMg5cIAAYtQElBICm+aMlCABfQqSvK7EqvjUl6vgRrmooVqOq44beo4frLq66MRr64BEuS1KUmIYBolCH6cy1DrDFW+appzbbxiKsqKoOmqE2O06WoGq7Hpusb7qG5HRv617Z3ez7QgySJIgyHUAaBqbIjBvjIFkbiXx9T8Ej9U5If26qLlJRtOdMMSkqoLA2FoFycXuXwcAoQxPrCzF8QoALgCm4A5Ny4LTCeWJZ2CtIxLSScsQ3aNiC1L8qDSDiuMUsTlAIf0fzC0pe1oZCPr9LLdcw6NKZY9iNE2VRsTckQ0G4F03S3L0XamtqhBoHMo+KV1s0i8ZSsKNAPMHU53MBeDyhKp0Q+zSBNvIXhVEgAByFQKD0cFy6GaBTkt6RNHXWhnQTrdRQKYS9MgFge+0nVOm7wzFX70eTQALwHs0+gsU4yAxT4PS9Zu+eMdvQ9SC9463lJhM37MIOE3fC-DtfW8PrditOSA-w0J3g471JTEHHF6BFZpWim7UdVPr5X6tiUB-L+rQcynE2vXKmg5Sj-1SAfAunddSZAyNEUw4QMiykiByUwHJIiIgbuMGBkAR4Wigc5DQfcZ4qnIcQ6eE8dIYjNv7HAUDM4iAvKvFuG9EHbzgSkCCDEb7jDvo4SAqkTCGGMh2N8ttcpdn7PJGgq5IDNjoHzC2nYOLEE0AQYwyFkBB0oMhD+nlAx1E2KcAAkLqAosoWCyknpqVo-DyHFwHIzYRMAbH0HkdxRscBygAFElEXkYFfFepAUgySDpHaOkBY7B2xLQRgEE3HcgBHyS2ETfStAtJ0a0yIxwMB0GgRm859i6AMMYXibNaqJgpFSGkDw6QMheMyd4bJUhpHAcUioEgFyKGIMuDC+styYGqNuAipBKC9iwEoX06gzTEFOBeQ6VxyS3GpCLVpzwmRvFZMvdIaCMgQVOB1FAA4NhUFIGWCseAvxMAAGRcEtvWUa0hDB8wFvEwo0ZuyZWSZsMgXlECN07GgHAtsSDIFmYgeZiyhwqlHCAf4vIgRAuCKC0E4JDjVNKMIu+kAxFtVSOXUo2JDDTA4gsOZCzIBLORexLAvZjC4U9DWTs9Bw6WzNrGIlYj6zUA4jQX8AFNiXKysQNAZANAsH5aI4AuoAqIHHjovmOI+jCR9CweuArbRErvjAaQIy1zzEypbHYFTFwUCODUqAhrUUOrvmJYS85wSjWIHq0RBrHWQDREw4KeICQ+ScqGYV1lox-nCsgHYmVbX4vLnOdoLcSCygAGKICjAY4gspFiZuzb2SAABhVwg4jmjj9eOMCsaRCZR0PiYUjLNiem9L6HKJYdhv2Ac+KaIBfQPk2MgX0PKGAWlDViYcuNnVGvvuKmsil6yZWrLWFhygU0aNzaot0z58J+gKPMONDBDw6NLOiegzYdGbHZcUZJM7xiUB3ApPAOg8JSTQAe9Kdbj1HmkgIJAZToxUBvRQZJCrZ3Vo-Ye8MLadhBwbSGx90ynJQfSjBmgOwq51E3eBuMMNSTXEaSmFpaZ2kHKzOyLkhq-gzoQwUAA-MJQUIadRMMGt+IY20bRUwAKpDWwiIa1igE3HF4mi7OWUhMAB5RD9gKP8nAybzodG0DqcUUlWi1BU7KR0IocWpEdLwfgSQSAAD4RQzp1KISzkAC2O17DJnTpmbN2YMaWncXSUiOadKZoY3HThafqFUxItT1kNK2c0x49I9kZk6UcnpGT0U5xBIFrQennTJJ0sBN8SQMtAR1AAGX9hfHEWWctYCK9xEruIdTOlwkoD0OpssYFy64FgDXfRWY3Swdz5bsxDEyUCS2qWWB4sSHpkbfmikwDy6ULDKngv2rw+zDZSYmk7NI-szMnmen5lMFNRLEnMUgsMMgWbcmFNKbBKkMb3lCUuvKMJcJoGayMG9S+MMEYowkAQc-FInGkQGpgM11rmX5tBZE0tmAdSEyEYixt6L6YOmHOzLtrIvSnWDYpQurFp2QLlfO381sin9MpFuwSmdrqmAg6wLNt7M7w3hkjT9-Of2CkA-8+MSriBqvruUxDu1tBQv4dW0R7ZqZEdke2-FvMPxDt8mO9i7nvOieYiu7iwXJVKePaYMrluOJ6eGsZ19g2v297s-AFx6b+OWsVeK-r3E4O0uQ6F0SEX4Xkzi5I5LrbcXUey9wfLjFOOTvIBp3rjVuJVck-dTdzX93+RPfD-byPhuiXG+Z-YVn5u+gW6t0D-O9WPRO+2IL4XK2PfrYl2033KP2QB+iEH7HQHQ+F-ax6aP6u4-4q14aqnz2i8pDT3fDPkZTfZ+zAUvPgOqY07q+330JfFuu+W-UuHnvIu7KR+RnbAfQhN45S37Fc+2sdYu8TrvZP4-a44Engnp-UjD4+xGsfLOclT9z4ifPVMRslrLZ5t3CvdfKvb3GvWLOvbpAPAADgP0VzxxG1608xk3PzV12Gw20E0263M071J3J17yJX7wQP-yOSfydw8Cz1EGnx5Ak2GwGXBEM2uRM2hGKl0jTkQGEnVWCi1S7G0GLjU0mimwLwdFlULVjU1wfCfA+xrDrGdiDiX0cjKWqBXG+zwHWGM3IPLzX02Q3wRzAORwo0gP21gJD2xVcwc1kxwN-34I0zQO0ydAmzoIMyM02CYOwPkwv1wOvz7x10YDMM4HeyKSrQYBEPs0QCTT53QPvSdUNRLzdQCN9SJRgAdCmFMjtk-CGgAEciB2VcAzwMJzVHwZks1QiOUmZqNLB3dgDiMos9Cd8Zd9tcNoiiVcBGNSwHcdJx1WNMR2NWEZ1AjEinRnCBByC-QUpxYsAVDt0P5uUijRDGi8hDU1CXDyCD4OdrcHRQwPMjlQ0QiDFl8UU74wsqivcaiYt9Dd9TAMdxMFcTC8c-DECjlkDLDusOjJoHD+ctAdN3jrsUgGD1CzMUCY9HCr8e8E9+97jiDswn9djews87RiwWApM-DvNHRTNfNLdhhhhF4UhtiqUs0UhS1hViApMABJQEq7fYUzC8GdKuGueBSAEk6CQ1GiL4YgVo4qVQOCb7VomWQcH0baY+Bk8k2zYotzSEr4KTbEWVGgUQWUKUltVEyAejGdCErY7MFE5zQ1YSMk54j4nrcU1ISUlIaUsQczZUw1IgtUiU0QTUolbU4U5E+UmUuU40hUs0+9EU0QjUyAAqajBk+9Fgh2R+Njc2PAMpHggYIuYYasU9W9GsBBTLVoMEnwunJk9PT7TPYSToeaWUeaOeSBaMtKe2DIPTPEn0Qk8MYgHeF7WgNJQsj6e2H+VoMsgkz7Ks5M2-JgPwxgNMu+WIpJGsXs7gIYpg4SBPRYp7DIRgGdbaTaOs8AGM+2HMUsvk1soki8Ds9grs0U3sB4qEocmElnToD0u+ccv0xPbc0Qnsk86jfsuM2gIc884cxglY-Ap8v0tgJ7KaLgcMxcm8w1MGc8wCu+fM+chcosv8SIFc-EisxTDcm-LcxgS0vrL4a8mIjdV8zoe80CoYIAA)

packages/turf-clean-coords/index.ts
mfedderly:
```suggestion
  // create a new rings array when we know something has changed even for
```

packages/turf-clean-coords/index.ts
```diff
@@ -1,18 +1,79 @@
...
+ * - Geometry(ies) and features members and their nested arrays will be reused
+ *   from the passed object or generated new based on performance and
+ *   simplicity in the implementation - and therefore will not be deep clones.
  *
  * @function
- * @param {Geometry|Feature} geojson Feature or Geometry
+ * @param {GeoJSON} geojson Any valid GeoJSON type
  * @param {Object} [options={}] Optional parameters
  * @param {boolean} [options.mutate=false] allows GeoJSON input to be mutated
+ * @param {number} [options.epsilon] Fractional number to compare with the cross product result. Useful for dealing with floating points such as lng/lat points
```
mfedderly, line 76: I need to think about the `epsilon` addition to the options some more, we already have [simplify](https://turfjs.org/docs/api/simplify) which handles nearly collinear points pretty well. I get that a user wouldn't want to have to run both of these methods though for performance reasons though.

packages/turf-clean-coords/index.ts
```diff
@@ -25,87 +86,123 @@ import { lineString } from "@turf/helpers";
...
+    case "FeatureCollection":
+      coordinates = mutate ? geojson.features : [];
+      for (let i = 0; i < geojson.features.length; i++) {
+        coordinates[i] = cleanCoords(geojson.features[i], options);
+      }
+      if (mutate) {
+        geojson.features = coordinates;
+        return geojson;
+      }
+      return featureCollection(coordinates, { bbox: geojson.bbox });
```
mfedderly, line 112: Similar to below, perhaps this is simpler?

```suggestion
      return {
        ...geojson,
        features: coordinates,
      };
```

packages/turf-clean-coords/index.ts
```diff
@@ -25,87 +86,123 @@ import { lineString } from "@turf/helpers";
...
+    case "Feature":
+      coordinates = cleanCoords(geojson.geometry, options);
+      if (mutate) {
+        geojson.geometry = coordinates;
+        return geojson;
+      }
+      return feature(coordinates, geojson.properties, {
+        bbox: geojson.bbox,
+        id: geojson.id,
+      });
```
mfedderly, line 137: I think this will explicitly add bbox and id properties set to undefined instead of leaving them completely unset if the argument had them undefined. What about a version like this?

```suggestion
      return {
         ...geojson,
         geometry: coordinates,
       };
```

packages/turf-clean-coords/index.ts
```diff
@@ -25,87 +86,123 @@ import { lineString } from "@turf/helpers";
...
   // Backwards compatible with v4.0
-  var mutate = typeof options === "object" ? options.mutate : options;
+  const mutate =
+    (typeof options === "object" ? options.mutate : options) ?? false;
   if (!geojson) throw new Error("geojson is required");
-  var type = getType(geojson);
 
   // Store new "clean" points in this Array
-  var newCoords = [];
+  let coordinates: any;
```
mfedderly, line 101: Overall implementation might be clearer if it handles FeatureCollection/GeometryCollection/Feature in a top level method, break out the individual Geometry handling into another method, and then eventually the calls to the helper methods. I'm thinking specifically about how `coordinates` is an `any` and then used for things that are not actually coordinates.

packages/turf-clean-coords/index.ts
```diff
@@ -177,9 +312,28 @@ function cleanLine(line: Position[], type: string) {
...
+ *
+ * @private
+ * @param {MultiPoint} geom Input geometry
+ * @returns {Position[]} Deduplicated coordinates
+ */
+function multipointDeduplicate(geom: MultiPoint): Position[] {
+  const seen = new Map();
+  for (const point of geom.coordinates) {
+    seen.set(point.join("-"), point);
+  }
```
mfedderly, line 330: Did you benchmark this against `Map<longitude, Map<latitude, point>>`? The tradeoff is the string creation and comparison vs more heap allocations.

mfedderly (commented):
I think that overall this PR moves Turf in a positive direction, but its hard to review with lots of changes blended together. Any chance you'd be willing to break this up into its constituent parts?

I think these changes are pretty uncontroversial:
- FeatureCollection/GeometryCollection handling
- MultiPoint speedup
- TypeScript type signature, as long as we keep compatibility for the moment

Things that I think I'd want to spend more time evaluating:
- `epsilon` option
- The exact behavior when `mutate: false`

PR #3024, opened 2026-02-11
@turf/boolean-contains: allow boundary points in MultiPoint containment checks

### LINKED ISSUE #3025: `booleanContains` does not align with DE9-IM behaviour for Polygons containing MultiPoints
When determining if a polygon contains a multipoint only one point of the multipoint needs to intersect with the interior of the polygon. The other points can be located on the boundary of the polygon.

The current implementation of booleanContains insists _every_ point of the multipoint intersects with the interior of the polygon, which is incorrect.

Example:

```json
{
  "type": "Polygon",
  "coordinates": [
    [
      [0, 0],
      [1, 0],
      [1, 1],
      [0, 1],
      [0, 0]
    ]
  ]
}
```

```json
{
  "type": "MultiPoint",
  "coordinates": [
    [0.5, 0.5],
    [1, 1]
  ]
}
```

Image

Currently returns false, even though the conditions are met by the interior of the polygon being intersected by at least one point of the multipoint's interior. References https://en.wikipedia.org/wiki/DE-9IM:

```math
\begin{bmatrix}
T & * & * \\
* & * & * \\
F & F & *
\end{bmatrix}
```

rexxars:
(Lifted from #3010)

Fixes behavior of `booleanContains` to return `true` if every point in a MultiPoint is inside/on boundary of some polygon, **and** at least one is _strictly interior._ This previously returned `false`, but differs from `booleanWithin` and doesn't match DE-9IM semantics. In my eyes this is a bugfix, not a breaking change - but will leave it to the maintainers to decide how to classify it.

**An example:**

Given the following square polygon:

```json
{
  "type": "Polygon",
  "coordinates": [
    [
      [0, 0],
      [1, 0],
      [1, 1],
      [0, 1],
      [0, 0]
    ]
  ]
}
```

And the following MultiPoint:

```json
{
  "type": "MultiPoint",
  "coordinates": [
    [0.5, 0.5],
    [1, 1]
  ]
}
```

Image

In this case, one point is strictly on the interior of the polygon, while one point is on the _boundary_. This previously returned `false`, but the way I (and DE-9IM) see it should return `true`.

Resolves #3025


mfedderly:
This seems like a pretty reasonable change, and I'm kind of torn on whether correcting behavior to match the spec is a breaking fix or not (I went back and forth on #3070 with the same debate). I think semantically they're bug fixes so 7.x is appropriate, but adding them to the v8 milestone and explicitly documenting the behavior change is probably a safer rollout strategy. Its always hard to tell on these calls until we hear back from users after a release that we broke things for them.

Thoughts?

mfedderly (approved):
Thank you!

mfedderly:
I have convinced myself that this is a bug that doesn't need a major version bump to correct. The docs and de-9im all suggest that this is a correction of buggy behavior and people should not have been depending on the broken state.

PR #3026, opened 2026-02-16
@turf/clusters-dbscan: Fix clustersDbscan not respecting units option

mradlinski:
Previously `clustersDbscan` respected the `units` option in regards to `maxDistance`. 
It seems this was overlooked when switching to `kdbush` in 6e4c18b38f63aa98c25922621a22c5e687eb52ca .

This PR fixes it by converting `maxDistance` from whatever unit was passed in the option to kilometers (which are used internally by `kdbush`).

A test was added to check if units are respected for kilometers, meters and degrees. 

I didn't find an issue for this, and thought that perhaps for such a simple issue one wasn't needed before creating a PR, sorry if that's the wrong workflow in this project.


mfedderly:
Thanks for the PR and pointing out where the broken behavior was introduced, it was really helpful as a jumping off poit to start looking into this. The PR itself has been reverted and released as 7.3.5. Is there any chance you have time to rebase this PR and have it just add the test case so we make sure not to break things going forward? If I misunderstood and we still need some form of fix in addition to the test, let me know.

packages/turf-clusters-dbscan/index.ts
```diff
@@ -142,7 +142,7 @@ function clustersDbscan(
         const neighborIndex = neighbor.index;
         const neighborPoint = points.features[neighborIndex];
         const distanceInKm = distance(point, neighborPoint, {
-          units: "kilometers",
+          units: options.units,
```
mfedderly, line 145: It turns out this still needed fixing after the revert based on the test case below.

packages/turf-helpers/README.md
```diff
@@ -15,13 +15,13 @@ depending on latitude.
 See [https://www.thoughtco.com/degree-of-latitude-and-longitude-distance-4070616][1]
 for an illustration of this behaviour.
 
-Type: (`"meters"` | `"metres"` | `"millimeters"` | `"millimetres"` | `"centimeters"` | `"centimetres"` | `"kilometers"` | `"kilometres"` | `"miles"` | `"nauticalmiles"` | `"inches"` | `"yards"` | `"feet"` | `"radians"` | `"degrees"`)
+Type: (`"meters"` | `"metres"` | `"m"` | `"millimeters"` | `"millimetres"` | `"mm"` | `"centimeters"` | `"centimetres"` | `"cm"` | `"kilometers"` | `"kilometres"` | `"km"` | `"miles"` | `"mi"` | `"nauticalmiles"` | `"nmi"` | `"inches"` | `"in"` | `"yards"` | `"yd"` | `"feet"` | `"ft"` | `"radians"` | `"rad"` | `"degrees"` | `"deg"`)
```
mfedderly, line 18: This is just coming from the automatic readme generation after https://github.com/Turfjs/turf/pull/3063 was merged. Sadly the CI is still not catching this when the readmes haven't been regenerated.

mfedderly (approved):

PR #3031, opened 2026-02-26
pnpm upgrade the world

mfedderly:
Ran `pnpm upgrade -r --latest` to yoink everything to the newest version around.

This required several manual fixes included in the 2nd commit.


eslint.config.mjs
```diff
@@ -46,6 +46,7 @@ export default tsEslint.config(
       "no-redeclare": "off",
       "no-var": "off",
       "prefer-const": "off",
+      "no-useless-assignment": "off", // the remaning errors here require more thought to clean up
```
mfedderly, line 49: The "fix" is to go from `let x = 0` to just `let x` (or similar). There's <10 errors from this, but I'm afraid to make those changes without thinking about things more closely.

mfedderly (commented):
Note I did have to revert `sweepline-intersections` back to 1.5.0.

Hopefully this sets a baseline which lets us mass upgrade the world periodically and not have to fiddle with managing upgrades one at a time. pnpm also has a setting that lets us skip taking upgrades within N minutes, if we'd like.

mfedderly:
Setting to draft for node support, perhaps we can get to this in v8 milestone instead.

packages/turf-directional-mean/test.ts
```diff
@@ -73,3 +78,18 @@ test("turf-directional-mean", (t) => {
...
+  const f = Math.pow(10, 9);
+  const props: (keyof typeof d.properties)[] = [
+    "cartesianAngle",
+    "bearingAngle",
+    "circularVariance",
+  ];
+
+  for (const k of props) {
+    d.properties[k] = Math.round(d.properties[k] * f) / f;
+  }
```
mfedderly, line 94: Really don't love this section and how I arbitrarily started passing tests with 9 digits after the decimal, but 10 didn't work.

mfedderly:
Going to do the dev dependencies for now over at https://github.com/Turfjs/turf/pull/3107

PR #3037, opened 2026-03-12
Update generate-readme script

mfedderly:
Doing some basic maintenance on the readme generation script to modernize things.

- This wasn't working properly in the per-package script, so I just removed that
- Moved from some promises to root-level await and module syntax
- Moves from `yamljs` to `yaml` (100x more popular, already exists in our devDependencies)
- Removes dependency on `glob` package (so many security alerts)
- Requires https://github.com/Turfjs/turf/pull/3036 to be merged first for some other `glob` usage
- Upgrades monorepolint so that we can use the builtin REMOVE abilities

2nd commit is automatically generated from just running monorepolint.


package.json
```diff
@@ -32,10 +32,10 @@
   "devDependencies": {
     "@eslint/eslintrc": "^3.3.1",
     "@eslint/js": "^9.31.0",
-    "@monorepolint/cli": "0.5.0",
-    "@monorepolint/config": "0.5.0",
-    "@monorepolint/core": "0.5.0",
-    "@monorepolint/rules": "0.5.0",
+    "@monorepolint/cli": "0.6.0-alpha.5",
```
mfedderly, line 35: Note that this is .5 compared to the other packages at .6. This is required to keep an older `yargs` transitive dependency so that node 18 will continue to build for us.

smallsaucepan (approved):
Great stuff. Thanks @mfedderly

PR #3038, opened 2026-03-12
Modernize test/bench runners

mfedderly:
Some of the basics for modernizing our test runner to get away from `tape`.

- Introduces support/testFixtures.mts and support/benchFixtures.mts to start to reuse code across packages. I know this won't work for every package, but it should work for most of them.
- Migrate to built-ins instead of `glob`, `load-json-file`, `tape`, `tsx`, and `write-json-file`
- Packages will have to migrate the test and bench methods at the same time


packages/turf-center/bench.ts
```diff
@@ -1,48 +1,12 @@
...
+import { benchFixtures } from "../../support/benchFixtures.mts";
+
+// Benchmark Results
+// feature-collection.geojson x 27,241,658 ops/sec ±0.34% (99 runs sampled)
+// imbalanced-polygon.geojson x 14,679,583 ops/sec ±0.27% (98 runs sampled)
+// linestring.geojson x 34,199,495 ops/sec ±0.52% (95 runs sampled)
+// point.geojson x 52,230,993 ops/sec ±0.70% (96 runs sampled)
+// points-with-weights.geojson x 24,802,237 ops/sec ±0.33% (100 runs sampled)
+// polygon-without-weights.geojson x 18,423,881 ops/sec ±0.28% (100 runs sampled)
+// polygon.geojson x 24,990,920 ops/sec ±0.45% (99 runs sampled)
```
mfedderly, line 11: Hilariously this is almost exactly 10x the original speed due to hardware progression over the years.

packages/turf-center/index.ts
```diff
@@ -1,6 +1,6 @@
-import { BBox, Feature, GeoJsonProperties, Point } from "geojson";
+import type { BBox, Feature, GeoJsonProperties, Point } from "geojson";
 import { bbox } from "@turf/bbox";
-import { point, Id, AllGeoJSON } from "@turf/helpers";
+import { point, type Id, type AllGeoJSON } from "@turf/helpers";
```
mfedderly, line 3: These `type` keywords are required when migrating from tsx to node native execution.

.monorepolint.config.mjs
```diff
@@ -38,8 +38,16 @@ for (const pk of await fs.readdir(packagesPath)) {
...
+  if (existsSync(path.join(packagesPath, pk, "test.ts"))) {
+    const testFileContents = await fs.readFile(
+      path.join(packagesPath, pk, "test.ts"),
+      "utf-8"
+    );
+    if (testFileContents.includes(`from "tape"`)) {
+      TAPE_PACKAGES.push(name);
+    } else {
+      NODE_TEST_PACKAGES.push(name);
+    }
```
mfedderly, line 45: Automatically differentiate if test.ts is `tape` or node native

.monorepolint.config.mjs
```diff
@@ -211,9 +211,21 @@ export default {
...
+      },
+      includePackages: TAPE_PACKAGES,
+    }),
+
+    packageScript({
+      options: {
+        scripts: {
+          bench: "node bench.ts",
+          "test:node": "node --test",
+          "test:tape": REMOVE,
```
mfedderly, line 221: Handles swapping bench from `tsx` execution to node as we switch to node native testing.

.monorepolint.config.mjs
```diff
@@ -211,9 +211,21 @@ export default {
         scripts: {
           bench: "tsx bench.ts",
           "test:tape": "tsx test.ts",
+          "test:node": REMOVE,
```
mfedderly, line 210: Just in case we ever migrate from node native back to tape.

.monorepolint.config.mjs
```diff
@@ -238,13 +250,49 @@ export default {
...
+          benchmark: REMOVE,
+          glob: REMOVE,
+          "load-json-file": REMOVE,
+          tape: REMOVE,
+          tsx: REMOVE,
+          "write-json-file": REMOVE,
+        },
+      },
+      includePackages: NODE_TEST_PACKAGES,
+    }),
```
mfedderly, line 283: When we switch to node native testing and use the shared fixture runner, we get to remove a lot of dependencies in favor of the builtin tooling.

package.json
```diff
@@ -36,8 +36,10 @@
     "@monorepolint/config": "0.6.0-alpha.6",
     "@monorepolint/core": "0.6.0-alpha.6",
     "@monorepolint/rules": "0.6.0-alpha.6",
+    "@types/benchmark": "^2.1.5",
     "@types/node": "22.15.3",
     "acorn": "^8.14.1",
+    "benchmark": "^2.1.4",
```
mfedderly, line 42: This is required because of shared/benchFixtures.mts, but we get to remove it from each package.

tsconfig.shared.json
```diff
@@ -8,6 +8,7 @@
     "moduleResolution": "node16",
     "importHelpers": true,
     "skipLibCheck": true,
-    "erasableSyntaxOnly": true
+    "erasableSyntaxOnly": true,
+    "rewriteRelativeImportExtensions": true
```
mfedderly, line 12: This seems to work just fine for us now. I had to do some extra stuff in geoclipper2 to have this all work with noEmit and a second tsconfig.json file for the production builds. I can revisit this situation if moving from tsup to raw tsc causes an issue.

PR #3042, opened 2026-03-23
bezier - Fix z value computation

githubhanjunjun:
Fix the issue of incorrect variable reference in the assignment of const dz on line 74 of the packages/turf-bezier-spline/lib/spline.ts file.


mfedderly (commented):
I think that this is probably a correct fix for the z values, but the package itself only spits out 2d coordinates meaning that I don't think this impacts the output of the package.

Are you trying to use this package for 3d work?

mfedderly:
Note to self: the original version of this code appears to be also broken and it lives at https://github.com/dowrow/bezier-spline-js/blob/master/bezier-spline.js

githubhanjunjun:
> I think that this is probably a correct fix for the z values, but the package itself only spits out 2d coordinates meaning that I don't think this impacts the output of the package.我认为这或许是一个解决 z 值问题的有效方法。不过，整个软件包只是输出二维坐标，因此我觉得这并不会影响到软件包的整体功能。
> 
> Are you trying to use this package for 3d work?您是想使用这个软件包来进行 3D 工作吗？

I just want the code to be correct even when the z value is not used, to avoid other potential unknown errors. I discovered this issue while studying how Bézier curves are calculated.

mfedderly (approved):
Thank you!

PR #3044, opened 2026-03-31
fix(polygonize): deduplicate segments in planar graph

Traksewt:
- [x] Summary of the changes.
- [ ] Heads up if this is a breaking change.
- [ ] Any issues this [resolves](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests).
- [ ] Inclusion of your details in the `contributors` field of `package.json` - you've earned it!

## Summary

When input geometry contains duplicate line segments (the same segment appearing more than once), `@turf/polygonize` silently returns 0 polygons instead of the expected result.

**Root cause:** `Graph.addEdge()` creates a new pair of directed edges for every segment, including duplicates. The duplicate edges corrupt the planar graph topology, causing the edge-ring traversal to fail.

**Fix:** Track edge IDs in `Graph` and skip segments that have already been added. This is a one-line guard in `addEdge()` plus cleanup in `removeEdge()`.

**Test:** Added a regression test with a simple square whose closing segment is duplicated — confirms polygonize now correctly returns 1 polygon.

This is not a breaking change. Inputs that previously worked continue to produce the same output. Only the failure case (duplicate segments -> 0 polygons) is fixed.

### DIFF
--- a/packages/turf-polygonize/lib/Graph.ts
+++ b/packages/turf-polygonize/lib/Graph.ts
@@ -44,6 +44,7 @@ function validateGeoJson(geoJson: AllGeoJSON) {
 class Graph {
   private nodes: { [id: string]: Node };
   private edges: Edge[];
+  private edgeIds: { [id: string]: true };
 
   /**
    * Creates a graph from a GeoJSON.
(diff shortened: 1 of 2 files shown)


Traksewt:
Reopening from the agriwebb org fork instead.

PR #3057, opened 2026-04-28
Move to pnpm catalogs

mfedderly:
[pnpm catalogs](https://pnpm.io/catalogs) were built to exactly solve the problem of wanting a single version of a given dependency across the workspace. We've been getting along ok with monorepolint enforcing the versions, but if we move these things into the catalog as well, it will make version bumps much simpler (just bump in 1 place, run pnpm install) and result in less churn (package.json files will no longer need updating).

Note that these will be replaced by the actual versions [when publishing](https://pnpm.io/catalogs#publishing).

First commit is the manual changes, second is just the churn from `pnpm mrl check --fix` and then `pnpm i`

### DIFF
--- a/.monorepolint.config.mjs
+++ b/.monorepolint.config.mjs
@@ -238,11 +238,11 @@ export default {
     requireDependency({
       options: {
         devDependencies: {
-          benchmark: "^2.1.4",
+          benchmark: "catalog:",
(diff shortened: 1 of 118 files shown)


smallsaucepan (approved):
Nice 😎

PR #3059, opened 2026-05-02
Migrate bench/test for area, bbox-clip, bezier-spline, and buffer

mfedderly:
Continues the path set forth in #3038 for a few more packages.

- `testFixtures` is extended to handle non-geojson return values as well
- Migrate some root package.json dependencies to the pnpm catalog version
- Limited the `node --test` execution to `test.ts` only, which avoid conflicts in packages with tstyche type testing

Side notes that I discovered while working on this:
- `colorize` gets implemented in many different test.ts files. Consolidating this is out of scope for this push because they're slightly different
- TypeScript 6 is going to have us change how we're referencing index.ts in each of these test/bench files
- TypeScript 6 is also going to force us to be explicit about when we're including @types/node
- In order to support the boolean packages, we need to handle `true` and `false` test subdirectories. Perhaps this is just a different test runner

Depends on https://github.com/Turfjs/turf/pull/3061


support/testFixtures.mts
```diff
@@ -12,9 +24,19 @@ export async function testFixtures(t: TestContext, fn: (input: any) => any) {
...
+
+      let outputPath = path.join(dirs.out, file);
+      // if we don't have geojson output, we should use .json instead
+      if (
+        result == null ||
+        typeof result != "object" ||
+        !VALID_TYPES.has(result.type)
+      ) {
+        outputPath = outputPath.replace(/\.geojson$/, ".json");
+      }
```
mfedderly, line 38: `area()` actually just returns a number, so it makes sense to have test/out/*.json and not .geojson.

mfedderly:
The build is failing here because node@18 doesn't support working against TypeScript files directly.

Issue #2038, opened 2021-02-18
Incompatibility: "metres" not allowed by the Units type in v6


tuukka:
Code such as `turfDistance(from, to, {units: "metres"})` works in 5.1.6 but Typescript fails it in 6.3.0:

`Type '"metres"' is not assignable to type '"meters" | "millimeters" | "centimeters" | "kilometers" | "acres" | "miles" | "nauticalmiles" | "inches" | "yards" | "feet" | "radians" | "degrees" | "hectares" | undefined'.`

Should `"metres"` be added in the type definition (keeping compatibility) or should "metres" be deprecated and removed from the documentation (simplifying the API)?

stebogit:
I'm quite sure we used to support "kilometre", in addition to "kilometer", so maybe we sould have both added back?

I've always thought it was a typo, but it appears it's just the common spelling outside the US. I'd suggest we support both alternatives.

rowanwins:
Support for metres should be added to the typings

PR #3063, opened 2026-05-10
Helpers - Add short forms for Units and AreaUnits

Johse:
Suggestion of adding shortform for _Units_ (eg.  "km" for "kilometers") and _AreaUnits_
 - No breaking change.


mfedderly (approved):
I went back and looked at the discussion when we last touched these to add non-US variants back (https://github.com/Turfjs/turf/issues/2038).

Its probably reasonable to add these in to align with the previous decision, and as best I can tell they are all the official SI or US Customary System shorthands.

Thanks!

PR #3064, opened 2026-05-11
ESM-only builds

mfedderly:
Migrate to esm-only packaging, which allows us to drop `tsup`.

BREAKING:
- This breaks node@18 compatibility, as it has no way to require() an ESM module (not even with a flag).

Todo:
- Add arethetypeswrong coverage
- Set package.json `engines` to node >= 22
- Specify 'evergreen' JS target in docs
- Github settings: set branch protection on master to only require 22, 24, 26 passing
- Probably needs to be re-staged as many separate smaller PRs
- Add note on esModuleInterop / allowSyntheticDefaultImports being `true` in alignment with typescript@7

[NodeJS previous releases](https://nodejs.org/en/about/previous-releases)
[NodeJS end of life](https://nodejs.org/en/about/eol) includes details on outstanding CVEs

Update 2026-05-21 - rebased after merging the consumer-style testing into master, removing it from this PR
Update 2026-05-28 - rebased to pull https://github.com/Turfjs/turf/pull/3070 into its own PR


package.json
```diff
@@ -5,9 +5,6 @@
     "docs": "node ./scripts/generate-readmes.mts",
     "lint": "pnpm run /lint:.*/",
     "lint:docs": "documentation lint packages/turf-*/index.js",
-    "lint:escheck-cjs": "es-check es8 packages/*/dist/cjs/index.cjs packages/turf/turf.min.js",
-    "lint:escheck-esm": "es-check --module es8 packages/*/dist/esm/index.js",
-    "lint:escheck-web": "es-check es5 packages/turf/turf.min.js",
```
mfedderly, line 10: escheck-cjs is no longer necessary because we're dropping cjs builds in this PR.

escheck-web is only necessary if you don't trust the Rollup step to output valid syntax. It has caught some things in the past where newer rollups ship `catch{}` syntax which isn't es5, but we're also going to move this way forward to es2022 or similar with Turf@8.

escheck-esm is no longer necessary and we are instead protected by tsconfig.shared.json's `target` of es2022 (in this PR). TypeScript will prevent us from using syntax that isn't supported at this version (which aligns with the tsconfig suggestions for node 22.

tsconfig.shared.json
```diff
@@ -1,13 +1,11 @@
 {
+  "include": ["${configDir}/index.ts", "${configDir}/lib/**/*.ts"],
   "compilerOptions": {
-    "target": "es2017",
-    "module": "node16",
+    "outDir": "${configDir}/dist",
+    "target": "es2022",
+    "module": "node18",
```
mfedderly, line 6: Just to call this out another time (it was in the commit message), node18 is actually a relatively strict setting and we would expect node >= 22 and all bundlers to be supported by the output here.

tsconfig.shared.json
```diff
@@ -1,13 +1,11 @@
 {
+  "include": ["${configDir}/index.ts", "${configDir}/lib/**/*.ts"],
   "compilerOptions": {
-    "target": "es2017",
-    "module": "node16",
+    "outDir": "${configDir}/dist",
+    "target": "es2022",
```
mfedderly, line 5: This aligns with node@22 and should limit our syntax to things that will work in all supported versions of node. It should also work in all browsers that are reasonably in use today, but we are also going to document that we intend to ship 'evergreen' js and expect users to transpile in their bundle step if they have specific needs.

tsconfig.shared.json
```diff
@@ -1,13 +1,11 @@
...
+    "module": "node18",
     "declaration": true,
-    "esModuleInterop": true,
```
mfedderly, line 6: I hope that setting this to false makes us more strict and gives our users fewer surprises when we change things. It forces us to make sure our own code works without the interop support.

tsconfig.shared.json
```diff
@@ -1,13 +1,11 @@
...
     "strict": true,
-    "moduleResolution": "node16",
-    "importHelpers": true,
-    "skipLibCheck": true,
```
mfedderly, line 10: Similar to the above, we intentionally incur the additional strictness by enabling checking of the lib types, so that anyone running full lib checking while consuming Turf will not be surprised by issues our CI missed.

mfedderly:
I pulled some of the test harness stuff out into its [own PR](https://github.com/Turfjs/turf/pull/3065)

mfedderly:
Note: this build is failing because it has changes from https://github.com/Turfjs/turf/pull/3070 as a prerequisite, but they aren't staged in this PR.

mfedderly:
Superceded by https://github.com/Turfjs/turf/pull/3135, https://github.com/Turfjs/turf/pull/3136, https://github.com/Turfjs/turf/pull/3137

PR #3070, opened 2026-05-28
Upgrade sweepline-intersections

mfedderly:
Upgrades `sweepline-intersections` to pick up fixes for `arethetypeswrong`. There is also a change https://github.com/rowanwins/sweepline-intersections/commit/e7a48be where it adds detection for endpoint intersections which previously did not exist.

## Breaking:

This changes the behavior of @turf/boolean-crosses from returning true to false when the line/line intersection is on an end point. I think this matches the DE9IM intended behavior now.

## Todo:

- [ ] ~Clean up the transitive dependencies https://github.com/rowanwins/sweepline-intersections/pull/25~ (I wound up just vendoring the code directly since the PR hasn't gotten a response yet)
- [ ] ~Investigate why we had to back this change out last time and address that~ (I thought there was an issue with these packages, but I incorrectly remembered what packages https://github.com/Turfjs/turf/pull/3017 involved. Still it may be reasonable to wait for v8 if we're going to change the results for crosses on endpoints.)

Note: tinyqueue@3 (released nearly 2 years ago) breaks node@18 compatibility. So that's another point towards only including this in turf@8.


mfedderly:
@rexxars Any chance you have a minute to verify that the change to line/line intersect behavior (under Breaking in the PR description) is the right thing to do?

mfedderly:
I've convinced myself that this can belong in 7.x because we're fixing a correctness bug to match DE-9IM and the docs as written. It also updates the sweepline-intersections code to pick up the latest changes, but sadly we had to inline it because the maintainer has been unresponsive to a PR that fixes some of its dependencies. If we took the newer versions as-is, we would be declaring an old version of esbuild as a transitive (non-dev) dependency.

mfedderly:
I backed out tinyqueue@3 and robust-predicates@3 (both to 2.x) to fix issues with node@18 compat in turf@7. I intend to bump everything to the newer versions once we start working towards turf@8 and drop support for node@18.

PR #3071, opened 2026-05-29
Fixing line-overlap missing certain horizontal/vertical lines (#2582)

### LINKED ISSUE #2582: lineOverlap tolerance is not handled in every case
The current implementation splits thepassed features into lineStrings. Then, for each segment in feature2, rbush is used to search whether there are overlapping points to a segment from feature1. In principle, at least one point must lie within the bounding box of a feature1 segment. The bug is, that the tolerance parameter passed is not taken into account here.

Even if two segments are very close to each other and run parallel, no overlapping would be detected even if a generous tolerance value was selected.

In TurfDart, I have chosen the approach of increasing the bounding box by the tolerance. This allows us to find all segments that lie within the specified tolerance. For details see: [line_overlap.dart](https://github.com/dartclub/turf_dart/blob/main/lib/src/line_overlap.dart)

In this example, I have two LineStrings that run parallel to each other. The distance is 11.4km and if you pass a tolerance of 12km, I think lineOverlap should return a result. Admittedly the distance is quite long, the example is for demonstration purposes and also works at distances of a few meters.

image
- Example on [geojson.io](http://geojson.io/#data=data:application/json,%7B%22type%22%3A%22FeatureCollection%22%2C%22features%22%3A%5B%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%230F0%22%2C%22fill%22%3A%22%230F0%22%2C%22stroke-width%22%3A25%7D%2C%22geometry%22%3A%7B%22type%22%3A%22LineString%22%2C%22coordinates%22%3A%5B%5B10%2C0%5D%2C%5B11%2C0%5D%5D%7D%7D%2C%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%23F00%22%2C%22fill%22%3A%22%23F00%22%2C%22stroke-width%22%3A10%2C%22stroke-opacity%22%3A1%2C%22fill-opacity%22%3A0.1%7D%2C%22geometry%22%3A%7B%22type%22%3A%22LineString%22%2C%22coordinates%22%3A%5B%5B10%2C0.1%5D%2C%5B11%2C0.1%5D%5D%7D%7D%2C%7B%22type%22%3A%22Feature%22%2C%22properties%22%3A%7B%22stroke%22%3A%22%2300F%22%2C%22fill%22%3A%22%2300F%22%2C%22stroke-width%22%3A3%2C%22stroke-opacity%22%3A1%2C%22fill-opacity%22%3A0.1%7D%2C%22geometry%22%3A%7B%22type%22%3A%22LineString%22%2C%22coordinates%22%3A%5B%5B10%2C0%5D%2C%5B11%2C0%5D%5D%7D%7D%5D%7D)
- Playground on [jsfiddle](https://jsfiddle.net/qe5k0638/3/)
- GeoJson:
```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "stroke": "#0F0",
        "fill": "#0F0",
        "stroke-width": 25
      },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [
            10,
            0
          ],
          [
            11,
            0
          ]
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {
        "stroke": "#F00",
        "fill": "#F00",
        "stroke-width": 10,
        "stroke-opacity": 1,
        "fill-opacity": 0.1
      },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [
            10,
            0.1
          ],
          [
            11,
            0.1
          ]
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {
        "stroke": "#00F",
        "fill": "#00F",
        "stroke-width": 3,
        "stroke-opacity": 1,
        "fill-opacity": 0.1
      },
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [
            10,
            0
          ],
          [
            11,
            0
          ]
        ]
      }
    }
  ]
}
```

jonovotny:
Line-overlap now detects overlap in horizontal and vertical segments with a parallel offset within tolerance (fixing #2582). Previously bounding boxes in the rbush tree would not overlap in these edge cases. By expanding the bbox around matching segments by the tolerance this issue should be resolved.

Added a test case of two polygons with horizontal and vertical overlap. (#2580 remains a problem and may still lead to missing segments for a different reason.)

Fixes #2582


packages/turf-line-overlap/index.ts
```diff
@@ -69,6 +70,11 @@ function lineOverlap<
       return;
     }
 
+    // Expand segment bounding box by the tolerance to create overlap in horizontal/vertical segments
+    var bb = bbox(segment);
+    bb = [bb[0]-tolerance, bb[1]-tolerance, bb[2]+tolerance, bb[3]+tolerance];
```
mfedderly: The jsdoc above implies that tolerance is in kilometers not degrees, so this needs a conversion I think.

```
import { lengthToDegrees} from "@turf/helpers";
...
const toleranceDegrees = lengthToDegrees(tolerance, "kilometers");
```

mfedderly (changes requested):
Thanks for looking at this! I think `tolerance` is in the wrong units but everything else looks good.

packages/turf-line-overlap/index.ts
mfedderly: Oh! One other thing. You may want to skip this bbox logic entirely if `tolerance === 0`

mfedderly (approved):
Thank you!

PR #3073, opened 2026-06-16
fix(clusters-dbscan): honor non-kilometer units in neighbor distance check

monad98:
## Issue
`clustersDbscan` ignores a non-kilometer `units` option. The bbox pre-filter honors `options.units` via `lengthToDegrees(maxDistance, options.units)`, but the final neighbor predicate computes the distance in kilometers and compares it directly against `maxDistance`:

```ts
const distanceInKm = distance(point, neighborPoint, { units: "kilometers" });
return distanceInKm <= maxDistance;
```

So `maxDistance` is treated as kilometers in the predicate regardless of `options.units`. For any unit smaller than a kilometer this makes the check effectively always true, and the neighborhood is no longer clipped to a `maxDistance`-radius circle — it collapses to the square bounding box, clustering points that are farther apart than `maxDistance`.

Example: with `{ units: "meters" }` and `maxDistance: 400`, two points ~566 m apart (400 m east + 400 m north) are clustered even though they are well beyond 400 m.

## Fix
Measure the neighbor distance in `options.units` so it matches the unit used by both `maxDistance` and the bbox pre-filter. When `units` is omitted, both still default to kilometers, so existing behavior is unchanged.

## Tests
Added a regression test covering a non-kilometer unit. All existing fixture tests still pass.

### DIFF
--- a/packages/turf-clusters-dbscan/index.ts
+++ b/packages/turf-clusters-dbscan/index.ts
@@ -141,10 +141,10 @@ function clustersDbscan(
       (neighbor) => {
         const neighborIndex = neighbor.index;
         const neighborPoint = points.features[neighborIndex];
-        const distanceInKm = distance(point, neighborPoint, {
-          units: "kilometers",
+        const neighborDistance = distance(point, neighborPoint, {
+          units: options.units,
         });
-        return distanceInKm <= maxDistance;
+        return neighborDistance <= maxDistance;
       }
     );
   };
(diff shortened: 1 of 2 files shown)


mfedderly:
Oof, we definitely imply that it *should* work this way in the docs, but I'm afraid to merge this change without marking it as a breaking change. I've got a bunch of prep work for v8 staged, but haven't merged any of it quite yet. I added this to the v8 milestone though so that I should be able to include it there.

Thanks!

PR #3076, opened 2026-06-22
Move turf.min.js builds to esbuild

mfedderly:
This drops rollup and babel and gives us a single modern bundler to keep up to date. The target for esbuild was taken from browserslist's 'default' setting as of 2026-06-21 and then changed to the esbuild syntax.

Babel and rollup keep hitting us with CVEs and have large npm ecosystem dependency trees. Esbuild has intentionally restricted themselves to only taking an additional dependency on a single set of APIs from Google themselves. In the event of an esbuild vulnerability the difficulty of an update should be limited.


mfedderly:
I'm curious for feedback on this if anyone else has a bundling tool that should be considered alongside esbuild. This was pretty easy to create (which makes me happy, but also makes it easy to throw away if another bundler is better).

(Some of the) Factors I'm considering when judging bundlers:
- Do they have a minimal dependency story?
- Who has ownership of the code and binaries that may be shipped into npm?
- How does their history of CVE/GHSE's look? (esbuild actually has a recent CVE, so its not disqualifying)
- What does the upgrade story look like for taking new versions?
- turf.min.js output size and time to process

mfedderly:
~Note that this PR does not deduplicate the eslint version with tsup (which I'd like to remove, which will fix the dependency situation), and~ babel is still transitively required through `documentation` which has its own dependabot struggles.

mfedderly:
Unscientific performance comparison: turf.min.js [rollup 8.9s](https://github.com/Turfjs/turf/actions/runs/31016196258/job/92340750982#step:5:523), [esbuild 171ms](https://github.com/Turfjs/turf/actions/runs/31021654434/job/92359592508?pr=3076#step:5:520)

PR #3077, opened 2026-06-22
bearing: fix final bearing returning values outside [-180, 180]

JSap0914:
## Summary

`bearing(start, end, { final: true })` can return values in the range (180, 360) instead of the documented [-180, 180].

## Root cause

`calculateFinalBearing` reverses the point pair and adds 180° to get the final bearing:

```js
let bear = bearing(end, start); // in [-180, 180]
bear = (bear + 180) % 360;      // yields [0, 360) — wrong when bear > 0
return bear;
```

When `bearing(end, start)` returns a positive value (e.g. 90°), the result is 270° — outside the documented range and inconsistent with every other bearing API in Turf.

## Fix

Add a single normalisation step after the modulo:

```js
return bear > 180 ? bear - 360 : bear;
```

## Concrete example

```js
// Traveling due west along the equator
bearing(point([10, 0]), point([0, 0]), { final: true })
// before: 270   ← wrong
// after:  -90   ← correct
```

## Changes

- `packages/turf-bearing/index.ts`: one-line normalisation in `calculateFinalBearing`
- `packages/turf-bearing/test.ts`: new test case covering the westward edge case
- `packages/turf-helpers/README.md`: auto-regenerated by the pre-commit hook (not related to this fix)

Resolves the documented contract that `bearing()` returns values between -180 and 180 degrees.


mfedderly (approved):
Thank you!

@turf/rhumb-bearing seems to already handle this case.

PR #3078, opened 2026-06-22
@turf/line-arc: always reach bearing2 (fix dropped final vertex)

chatman-media:
## Summary

`lineArc` could return an arc that stops one step short of `bearing2`.

The loop generates points with `while (alpha <= arcEndDegree)`, advancing `alpha = arcStartDegree + i * arcStep`. For some bearing/step combinations the accumulated angle overshoots `arcEndDegree` by a floating-point epsilon on the last iteration, so the loop exits early and the final vertex is dropped.

The result is an arc with `steps` vertices instead of the documented `steps + 1`, ending a full `arcStep` before `bearing2`.

### Reproduction

```js
import { lineArc } from "@turf/line-arc";
import { point } from "@turf/helpers";

const arc = lineArc(point([-75, 40]), 5, 0, 29, { steps: 7 });
arc.geometry.coordinates.length; // 7  (expected 8 = steps + 1)
// last vertex is at bearing ~24.857° instead of 29°
```

Here `0 + 7 * (29 / 7) === 29.000000000000004 > 29`, so the `alpha <= arcEndDegree` check fails on the final step.

### Fix

Iterate a fixed `steps + 1` times and pin the last vertex to `arcEndDegree`, so the arc always ends exactly on `bearing2`. A small fuzz over many `(bearing1, bearing2, steps)` combinations showed ~0.9% of cases dropped the final vertex before this change and 0 after; behaviour is unchanged for inputs that did not trigger the rounding overshoot (existing fixtures are byte-identical).

Added a regression test (`lineArc(center, 5, 0, 29, { steps: 7 })`) that asserts `steps + 1` vertices and that the last vertex lies on `bearing2`. It fails on the current code and passes with the fix; the full `@turf/line-arc` suite passes.

Note: this is orthogonal to open PR #2910 (which changes the arc radius approximation but keeps the same loop), so there is no conflict.

## Breaking change

No. Output is unchanged except that affected arcs now include the previously-missing endpoint, matching the documented `steps + 1` contract.

## Issues

No existing issue; self-discovered.

---

- [x] Summary of the changes.
- [x] Heads up if this is a breaking change. (not a breaking change)
- [x] Any issues this resolves. (none — self-discovered)
- [x] Inclusion of your details in the `contributors` field of `package.json`.
- [x] Confirmation you've read the steps for preparing a pull request.

### DIFF
--- a/packages/turf-line-arc/index.ts
+++ b/packages/turf-line-arc/index.ts
@@ -57,19 +57,21 @@ function lineArc(
   const arcStartDegree = angle1;
   const arcEndDegree = angle1 < angle2 ? angle2 : angle2 + 360;
 
-  let alpha = arcStartDegree;
   const coordinates = [];
-  let i = 0;
   // How many degrees we'll swing around between each step.
   const arcStep = (arcEndDegree - arcStartDegree) / steps;
-  // Add coords to the list, increasing the angle from our start bearing
-  // (alpha) by arcStep degrees until we reach the end bearing.
-  while (alpha <= arcEndDegree) {
+  // Add coords to the list, increasing the angle from our start bearing by
+  // arcStep degrees until we reach the end bearing. Iterate a fixed number of
+  // times (steps + 1 vertices) rather than comparing an accumulated angle
+  // against arcEndDegree: floating-point drift in arcStartDegree + i * arcStep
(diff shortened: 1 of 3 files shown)


mfedderly (approved):
Thank you!

PR #3079, opened 2026-06-23
fix(boolean-crosses): make MultiPoint/LineString crossing order-independent

spokodev:
## Problem

`doMultiPointAndLineStringCross` decides whether a MultiPoint crosses a LineString, which is a set predicate: it must be true when the point set has at least one member interior to the line and at least one member exterior to it. A set predicate has to be order-invariant, but two bugs make it depend on coordinate order and on per-segment accidents.

1. **Outer loop stops after the first point.** The `while` guard was `!foundIntPoint && !foundExtPoint`, so the loop exits as soon as either flag is set, i.e. after classifying the very first MultiPoint coordinate. Whether the predicate sees the rest of the set then depends on which point happens to come first. The sibling `doesMultiPointCrossPoly` already uses the correct guard `(!foundIntPoint || !foundExtPoint)` (keep scanning until both an interior and an exterior point are found).

2. **Inner loop conflates "not on this segment" with "exterior to the line".** Each non-matching segment set `foundExtPoint = true`, so a point that lies on the line but not on the first segment tested was wrongly counted as exterior. A single interior point could therefore report a cross.

## Fix

- Iterate the whole MultiPoint with the order-invariant guard `(!foundIntPoint || !foundExtPoint)`, matching `doesMultiPointCrossPoly`.
- Classify each point once: scan its segments, and only after the inner loop decide interior (on any segment) vs exterior (on none).

## Tests

Two fixtures added:

- `test/true/MultiPoint/LineString/MultiPointsCrossLineExteriorFirst.geojson` — exterior point listed first; must still cross.
- `test/false/MultiPoint/LineString/MultiPointAllInteriorNotCrossLine.geojson` — all points interior; must not cross.

Both fail on the current source and pass after the fix. Full package suite: 22/22.

### DIFF
--- a/packages/turf-boolean-crosses/index.ts
+++ b/packages/turf-boolean-crosses/index.ts
@@ -76,8 +76,8 @@ function doMultiPointAndLineStringCross(
   var foundIntPoint = false;
   var foundExtPoint = false;
   var pointLength = multiPoint.coordinates.length;
-  var i = 0;
-  while (i < pointLength && !foundIntPoint && !foundExtPoint) {
+  for (var i = 0; i < pointLength && (!foundIntPoint || !foundExtPoint); i++) {
+    var pointOnLine = false;
     for (var i2 = 0; i2 < lineString.coordinates.length - 1; i2++) {
       var incEndVertices = true;
       if (i2 === 0 || i2 === lineString.coordinates.length - 2) {
@@ -91,12 +91,15 @@ function doMultiPointAndLineStringCross(
           incEndVertices
         )
       ) {
-        foundIntPoint = true;
-      } else {
-        foundExtPoint = true;
+        pointOnLine = true;
+        break;
       }
     }
-    i++;
+    if (pointOnLine) {
+      foundIntPoint = true;
+    } else {
+      foundExtPoint = true;
+    }
   }
   return foundIntPoint && foundExtPoint;
 }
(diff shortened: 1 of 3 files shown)


mfedderly (approved):
This one is closer to a 30% performance impact on these, but they're still benching really fast and exit early when its known to cross. Thanks for the correctness fixes.

PostGIS comparison:
```
SELECT ST_Crosses(
    ST_GeomFromText('MULTIPOINT(1 2, 1 3)'),
    ST_GeomFromText('LINESTRING(1 1, 1 2, 1 3, 1 4)')
);
 st_crosses 
------------
 f
(1 row)

SELECT ST_Crosses(
   ST_GeomFromText('MULTIPOINT(12 12, 1 2)'),
   ST_GeomFromText('LINESTRING(1 1, 1 2, 1 3, 1 4)')
);
st_crosses 
------------
 t
(1 row)
```

```
before
MultiPointsCrossLineExteriorFirst x 71,810,879 ops/sec ±0.38% (100 runs sampled)
MultiPointsCrossLine x 65,405,630 ops/sec ±0.38% (99 runs sampled)
MultiPointNotCrossLineEnd x 65,653,389 ops/sec ±0.47% (97 runs sampled)
MultiPointNotCrossLine x 70,899,758 ops/sec ±0.24% (99 runs sampled)
MultiPointAllInteriorNotCrossLine x 64,975,633 ops/sec ±0.41% (99 runs sampled)

after
MultiPointsCrossLineExteriorFirst x 51,910,952 ops/sec ±0.45% (100 runs sampled)
MultiPointsCrossLine x 51,023,836 ops/sec ±0.89% (96 runs sampled)
MultiPointNotCrossLineEnd x 44,693,571 ops/sec ±0.96% (95 runs sampled)
MultiPointNotCrossLine x 35,712,302 ops/sec ±0.36% (99 runs sampled)
MultiPointAllInteriorNotCrossLine x 53,528,077 ops/sec ±0.42% (99 runs sampled)
```

mfedderly:
Thank you!

PR #3087, opened 2026-06-30
fix(@turf/midpoint): interpolate altitude (z-coordinate) for 3D points

### LINKED ISSUE #3086: bug(@turf/midpoint): altitude (z-coordinate) not interpolated for 3D points
## Bug Report

**Package:** `@turf/midpoint`

### Description

When both input points carry a z-coordinate (altitude/elevation), `midpoint()` returns a point whose z equals the **origin's z unchanged** instead of the linear average `(z1 + z2) / 2`.

### Reproducible example

```ts
import { midpoint } from '@turf/midpoint';
import { point } from '@turf/helpers';

const p1 = point([0, 0, 100]);   // altitude 100
const p2 = point([10, 0, 200]);  // altitude 200
const mid = midpoint(p1, p2);

console.log(mid.geometry.coordinates[2]);
// Actual:   100   ← same as p1's z
// Expected: 150   ← (100 + 200) / 2
```

### Root Cause

`midpoint()` delegates to `destination(origin, dist/2, heading)`.
`destination()` propagates only the **origin's** z-coordinate unchanged (see [source](https://github.com/Turfjs/turf/blob/master/packages/turf-destination/index.ts)):

```ts
if (coordinates1[2] !== undefined) {
  return point([lng, lat, coordinates1[2]], options.properties);
}
```

Because `midpoint` passes `point1` as the origin, the result always gets `z = z1`, discarding `z2` entirely.

### Expected Behaviour

The midpoint of two 3D points should carry `z = (z1 + z2) / 2`.

### Fix

Extract both z-values and average them when both are defined; fall through to current behaviour (no z) when either is absent.

JSap0914:
## Summary

Fixes `@turf/midpoint` returning the wrong z-coordinate when both input points have altitude (elevation).

Closes #3086

## Root Cause

`midpoint(p1, p2)` calls `destination(p1, dist/2, heading)`.
`destination()` propagates **only the origin's z** unchanged, so the result always got `z = z1`, ignoring `z2` entirely.

## Fix

After computing the geographic midpoint, extract both z-values via `getCoord()` and set:

```ts
midpoint.geometry.coordinates[2] = (coord1[2] + coord2[2]) / 2;
```

This runs only when both input points have a z-coordinate; 2D inputs are unaffected.

## Changes

- `packages/turf-midpoint/index.ts` — average z when both inputs have altitude
- `packages/turf-midpoint/package.json` — add `@turf/invariant` as explicit dependency
- `packages/turf-midpoint/test.ts` — two new tests (3D interpolation + 2D unchanged)

## Test results

**Before fix:**
```
# tests 8 / # pass 7 / # fail 1
not ok 7 z should be the average of the two input altitudes (100+200)/2 = 150
  actual:   100   expected: 150
```

**After fix:**
```
# tests 8 / # pass 8
```

### DIFF
--- a/packages/turf-midpoint/index.ts
+++ b/packages/turf-midpoint/index.ts
@@ -3,6 +3,7 @@ import { bearing } from "@turf/bearing";
 import { destination } from "@turf/destination";
 import { distance } from "@turf/distance";
 import { Coord } from "@turf/helpers";
+import { getCoord } from "@turf/invariant";
 
 /**
  * Takes two points and returns a point midway between them. The midpoint is
@@ -27,6 +28,14 @@ function midpoint(point1: Coord, point2: Coord): Feature<Point> {
   const dist = distance(point1, point2);
   const heading = bearing(point1, point2);
   const midpoint = destination(point1, dist / 2, heading);
+
+  // Interpolate altitude (z-coordinate) when both input points carry one.
+  // destination() propagates only the origin's z unchanged; average them instead.
+  const coord1 = getCoord(point1);
+  const coord2 = getCoord(point2);
+  if (coord1[2] !== undefined && coord2[2] !== undefined) {
+    midpoint.geometry.coordinates[2] = (coord1[2] + coord2[2]) / 2;
+  }
 
   return midpoint;
 }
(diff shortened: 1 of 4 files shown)


modusensus:
Hi @JSap0914! Friendly ping 👋 This fix looks good and the CI is green, but the branch has been quiet for a couple of months and has since picked up conflicts with master — just `packages/turf-midpoint/package.json` and `pnpm-lock.yaml` (likely fallout from the recent dependency/type refactors). If you don't have time to rebase, I'd be happy to resolve the conflicts and help get this over the line. Thanks again for the patch!

PR #3096, opened 2026-07-04
fix(nearest-point-on-line): segmentIndex always returns i, add segment helper functions

### LINKED ISSUE #3023: nearestPointOnLine returns invalid segmentIndex when result is at the end of the last segment of the lineString
When calling nearestPointOnLine and the result is at the end of a segment, there is legacy behaviour that sets the segmentIndex to that of the next segment (end point of one segment being the same as the start point of the next). Unsure why this is the behaviour.

Unfortunately, if the original segment was the last in the lineString the segmentIndex is still incremented meaning it now refers to a non-existent segment.

After fixing #2946 with PR #2951, it became apparent in #3008 some users were using the overflowing segmentIndex as an indicator for whether the nearest point was found at the end of the line. Although undocumented, this caused backwards compatibility issues and will be reverted.

This issue aims to fix the problem properly as a breaking change aligned with a major release. Current plan is to:

1. ensure segmentIndex always refers to a valid segment
2. clarify remaining behaviour in #3022

feliperm17:
Closes #3023
Design discussion: #3022

## Summary

`segmentIndex` now always equals the index `i` of the segment on which the nearest point was found. The legacy "favour the next segment" behaviour (returning `i + 1` when the nearest point lands exactly on an end vertex) has been removed entirely — for both intermediate vertices and the last vertex of a LineString. Three helper functions are added so callers can restore that behaviour when needed.

## Breaking change ⚠️

**Before:**
```ts
nearestPointOnLine(line, point).properties.segmentIndex
// → i + 1 when nearest point is at an end vertex (intermediate segments)
// → coords.length - 1 when at the last vertex (invalid — out of bounds)
```

**After:**
```ts
nearestPointOnLine(line, point).properties.segmentIndex
// → always i (the segment where the point was found, always valid)
```

## New helper functions

Three exported helpers allow callers to restore the previous behaviour or build their own logic (e.g. turn-by-turn navigation):

```ts
import {
  nearestPointOnLine,
  atEndOfSegment,
  atEndOfLineString,
  hasNextSegment,
} from "@turf/nearest-point-on-line";

const nearest = nearestPointOnLine(line, point);

atEndOfSegment(nearest, line);    // true if nearest point is at end vertex of its segment
atEndOfLineString(nearest, line); // true if nearest point is at the very last vertex
hasNextSegment(nearest, line);    // true if a next segment exists

// Reconstruct the old i+1 behaviour when desired:
const legacyIndex =
  nearest.properties.segmentIndex +
  (atEndOfSegment(nearest, line) && hasNextSegment(nearest, line) ? 1 : 0);
```

## Tests

- Updated the pre-existing `segmentIndex: 8 → 7` expectation (intermediate vertex case, intentional breaking change).
- Updated fixture `test/out/multiLine1.geojson`: `segmentIndex`/`index` `21 → 20` (the only real value change; full fixture regen was discarded as it produced 34k lines of pure formatting noise with no value changes).
- Added 7 new tests covering: last vertex, intermediate vertex, `atEndOfSegment`/`hasNextSegment` + legacy index reconstruction, `atEndOfLineString`, mid-segment, LineString↔MultiLineString parity via segment offsets, and helpers on bare geometry input.
- All 106 tape tests pass; `test:types` passes; eslint clean.

## Note for maintainers

`packages/turf-line-slice/index.ts:69` contains a `fixSegmentIndexBounds` workaround that was added explicitly to handle this bug. It is now a harmless no-op. Removing it is out of scope for this PR but worth a follow-up.

## Checklist

- [x] Summary of the changes.
- [x] Heads up that this is a breaking change.
- [x] Resolves #3023.
- [x] Added contributor details to `contributors` field of `package.json`.
- [x] I have read the steps for [preparing a pull request](https://github.com/Turfjs/turf/blob/master/docs/CONTRIBUTING.md#preparing-a-pull-request).


feliperm17:
@smallsaucepan when you get a chance, would appreciate your review! Happy to adjust anything based on your feedback.

mfedderly:
Putting this in the v8 bucket to re-evaluate once we start making the set of breaking changes for that release.

PR #3097, opened 2026-07-05
Fix boolean-disjoint missing overlapping collinear lines

### LINKED ISSUE #1842: boolean-disjoint on parallel lines not working (bug in boolean-disjoint)
I started using turfjs recently for calculating geospatial relations and I think i noticed a bug (correct me if I'm wrong). During a test of the boolean-disjoint function I tested if 2 parallel lines with a mutual point (so on top of each other) were disjoint. This should not be the case (since they have a lot of mutual points), but the function tells me they are disjoint. 

The lines i tried are the following: https://gist.github.com/dreeki/60d67ff4cb9e742ab9233b09c409de51

Snippet of the code is the following: 
```javascript
import { Geometries as GeometryType } from '@turf/helpers';
import booleanDisjoint from '@turf/boolean-disjoint';

const firstGeometryObject: GeometryType = giveGeoObject(first.value);
const secondGeometryObject: GeometryType = giveGeoObject(second.value);
result = booleanDisjoint(firstGeometryObject, secondGeometryObject);
```
and:
```javascript
import { geometry, Geometries as GeometryType } from '@turf/helpers';
import * as Wkt from 'terraformer-wkt-parser';
import * as Terraformer from 'terraformer';

export const giveGeoObject: (text: string) => GeometryType = (text: string) => {
    const geoJSON = Wkt.parse(text);
    if(geoJSON instanceof Terraformer.Point || 
        geoJSON instanceof Terraformer.MultiPoint ||
        geoJSON instanceof Terraformer.LineString || 
        geoJSON instanceof Terraformer.MultiLineString ||
        geoJSON instanceof Terraformer.Polygon || 
        geoJSON instanceof Terraformer.MultiPolygon){
        return geometry(geoJSON.type, geoJSON.coordinates);
    }
    return null;
}
```

cycsmail:
boolean-disjoint leans on lineIntersect to decide whether two lines touch, but lineIntersect only returns points where segments actually cross, so two collinear lines lying on top of each other come back as disjoint when they aren't.

I added a check for whether a vertex of either line falls on the other line. That covers the overlap case, since an overlapping interval always has an endpoint that is a vertex of one of the lines, and it leaves genuinely parallel, non-touching lines disjoint as before.

Added a regression fixture under test/false for two overlapping collinear lines. No breaking change.

Resolves #1842

### DIFF
--- a/packages/turf-boolean-disjoint/index.ts
+++ b/packages/turf-boolean-disjoint/index.ts
@@ -125,6 +125,19 @@ function isLineOnLine(
   if (doLinesIntersect.features.length > 0) {
     return true;
   }
+  // lineIntersect() only reports points where segments cross, so collinear
+  // (partially) overlapping lines are missed. Check whether a vertex of either
+  // line lies on the other line to also catch those overlapping cases.
+  for (const coords of lineString1.coordinates) {
+    if (isPointOnLine(lineString2, { type: "Point", coordinates: coords })) {
+      return true;
+    }
+  }
+  for (const coords of lineString2.coordinates) {
+    if (isPointOnLine(lineString1, { type: "Point", coordinates: coords })) {
+      return true;
+    }
+  }
   return false;
 }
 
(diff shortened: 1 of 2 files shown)


mfedderly (approved):
Thank you!

PR #3098, opened 2026-07-20
fix(projection) - Rework public API to allow arbitrary projections

mfedderly:
Rename `convert` to `project` and exports it as new public api.

There are plenty of projections that aren't Mercator or WGS84, and a lot of work has gone into generic projection tools like `proj4`. This lets people use the existing projection code in Turf to manage the GeoJSON wrappers without forcing Turf to continue to expand for new projections or stay up to date with these other packages.

WGS84 and Mercator are a special case because they are so often used in web mapping, so I think it makes sense to continue to ship implementations of these two methods.

Fixes #3046


mfedderly:
One possible counterexample to the `projection: (input: number[]) => number[]` API here would be trying to bulk project to something like MGRS where you wouldn't be able to return `number[]`. That also starts to get a little far from GeoJSON itself though, so maybe its a nonissue.

PR #3101, opened 2026-07-21
Use official EPSG:3857 code instead of deprecated EPSG:900913 in projection docs

### LINKED ISSUE #2891: Consider changing documentation from EPSG:900913 to EPSG:3857
EPSG:900913 is a unofficial coordinate system number which was used before Mercator had a official EPSG number. Its identical to EPSG:3857.

See: https://epsg.io/900913

Even though that site documents EPSG:900913 as being used by OpenLayers, looking at the mentions of it in their repo, it seems to mostly be listed as an alias. Most places i found have switched to calling it EPSG:3857, to the point i only learnt about this due to Turf docs using EPSG:900913.

It might make sense to switch documentation over to mentioning the official number.

cycsmail:
The projection package referred to Mercator as EPSG:900913, the old deprecated Google code. Swapped it for the official EPSG:3857 in the README, the JSDoc, and the inline comments. Also updated the proj4 calls in the tests to EPSG:3857; proj4 treats the two as aliases so the expected values are unchanged.

No behaviour change, it's just docs and comments. The projection tape tests still pass (47/47).

Resolves #2891

### DIFF
--- a/packages/turf-projection/README.md
+++ b/packages/turf-projection/README.md
@@ -4,7 +4,7 @@
 
 ## toMercator
 
-Converts a WGS84 GeoJSON object into Mercator (EPSG:900913) projection
+Converts a WGS84 GeoJSON object into Mercator (EPSG:3857) projection
 
 ### Parameters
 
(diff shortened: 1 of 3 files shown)


mfedderly (approved):
Thank you!

PR #3106, opened 2026-07-23
transform-scale, transform-translate, transform-rotate: Invalidate stale bboxes after coordinate transforms

### LINKED ISSUE #2835: Transform Translate/Rotate method do not recompute bounding box property
Turf version: 7.2.0

Use of the transformTranslate/transformRotate methods seem to simply parrot back the .bbox property of their input feature. 
I'm guessing that this generic behavior intended to preserve any/all properties that might have been on the input object.
Unfortunately, this can cause unexpected behavior when used in combination with other turf methods that depend upon or utilize the .bbox property. 

Specifically, this became a problem for turf.bbox as of the 7.0.0 release since it will now also just return the .bbox property of an input unless the `recompute` option is set.

NOTE: Oddly enough, transformScale does not seem to suffer from this behavior

Trivial example:
```
const bbox = turf.bboxPolygon([-0.1, -0.1, 0.1, 0.1]);

const translated = turf.transformTranslate(bbox, 200, 45, { mutate: false});
const scaled = turf.transformScale(bbox, 200, { mutate: false});
const rotated = turf.transformRotate(bbox, 45, { mutate: false });

console.log(turf.bbox(translated));
console.log(turf.bbox(translated, { recompute: true }));
console.log(turf.bbox(scaled));
console.log(turf.bbox(scaled, { recompute: true }));
console.log(turf.bbox(rotated));
console.log(turf.bbox(rotated, { recompute: true }));
```

Outputs: 
```
// Translated
[-0.1, -0.1, 0.1, 0.1]  
[1.1719128101790375, 1.1718310552975468, 1.371962093942102, 1.3718310552975466]

// Scaled
[-20.418973863263318, -20.000000000000007, 20.418973863263318, 20.00000000000001]
[-20.418973863263318, -20.000000000000007, 20.418973863263318, 20.00000000000001]

// Rotated
[-0.1, -0.1, 0.1, 0.1]
[-0.1414213127535504, -0.14142132033780166, 0.14142138833847184, 0.1414213203377955]
```

khalilurrrahmanridoykhan:
## What changed

- recursively removes stale `bbox` values after `transformTranslate` and `transformRotate`
- extends the same invalidation to nested objects handled by `transformScale`
- covers Features, FeatureCollections, and GeometryCollections
- preserves the existing clone-versus-mutate behavior
- adds regression coverage for both default cloning and `mutate: true`

## Why

Coordinate transforms mutate positions but previously retained cached bounding boxes at one or more levels of the GeoJSON hierarchy. A later bbox lookup could therefore return coordinates describing the pre-transform object. Invalidating those cached values lets callers explicitly recompute them when needed.

## Validation

- `pnpm --filter @turf/transform-translate test` — 29 passed
- `pnpm --filter @turf/transform-rotate test` — 25 passed
- `pnpm --filter @turf/transform-scale test` — 54 passed
- ESLint on all six changed TypeScript files
- Prettier check on all six changed TypeScript files
- `git diff --check`

Closes #2835


mfedderly:
(👍 for the recursive approach here)

packages/turf-transform-rotate/index.ts
```diff
@@ -66,8 +66,21 @@ function transformRotate<T extends GeoJSON | GeometryCollection>(
     pointCoords[0] = newCoords[0];
     pointCoords[1] = newCoords[1];
   });
+  removeBbox(geojson);
   return geojson;
 }
 
+function removeBbox(geojson: GeoJSON | GeometryCollection): void {
```
mfedderly: GeometryCollection is a Geometry is a GeoJSON :)

```suggestion
function removeBbox(geojson: GeoJSON): void {
```

packages/turf-transform-rotate/index.ts
```diff
@@ -66,8 +66,21 @@ function transformRotate<T extends GeoJSON | GeometryCollection>(
...
+  delete geojson.bbox;
+
+  if (geojson.type === "Feature") {
+    if (geojson.geometry) removeBbox(geojson.geometry);
+  } else if (geojson.type === "FeatureCollection") {
+    geojson.features.forEach(removeBbox);
+  } else if (geojson.type === "GeometryCollection") {
+    geojson.geometries.forEach(removeBbox);
+  }
+}
```
mfedderly: This seems like something that would be useful to share out of @turf/helpers.

I'd call it something like `function removeBbox(geojson: GeoJSON): void` and note that it mutates its  input recursively in the jsdoc. I wouldn't provide a `mutate` option.

khalilurrrahmanridoykhan:
Thanks for the feedback. I moved `removeBbox` to `@turf/helpers`, simplified the parameter to `GeoJSON`, documented its recursive mutation behavior, and added focused tests, type coverage, and generated documentation. All relevant local tests pass.

mfedderly (approved):
Thank you!

PR #3107, opened 2026-07-23
Upgrade devDependencies

mfedderly:
```
pnpm upgrade -D -r --latest
selectively revert some items:
- @types/* where the underlying package isn't also changed
- typescript is already on the latest v6 and v5, and we aren't moving to v7 yet
- @types/node should be the latest from 22.x
- lint-staged from 17 back down to 15, they started warning on `git add` usage, but don't commit the generated README's
pnpm install
pnpm dedupe
```


packages/turf-clusters-dbscan/package.json
```diff
@@ -54,16 +54,16 @@
     "@turf/centroid": "workspace:*",
     "@turf/clusters": "workspace:*",
     "@types/benchmark": "catalog:",
-    "@types/rbush": "^3.0.4",
+    "@types/rbush": "^4.0.0",
```
mfedderly: Needs a revert

package.json
```diff
@@ -28,29 +28,29 @@
   "devDependencies": {
     "@eslint/eslintrc": "^3.3.6",
     "@eslint/js": "^10.0.1",
-    "@monorepolint/cli": "0.6.0-alpha.5",
+    "@monorepolint/cli": "0.6.0-alpha.6",
```
mfedderly, line 31: This can finally be upgraded to match the other ones now that we dropped support for cjs 🎉

Issue #3139, opened 2026-08-05
Deduplicate types testing


mfedderly:
Right now we've got the older `types.ts` testing, and the newer `tstyche` harness.

I'd like to migrate to one or the other, and the types.ts testing is pretty janky and bespoke. Whatever we go with needs to work with typescript@7.

mrazauskas:
As it happens, TSTyche runner can be used to check all files that can be checked by `tsc`. Assertions are not mandatory. TypeScript 7 support is on the way.

mfedderly:
@mrazauskas Yep I saw the issue over on tstyche. I'm eagerly awaiting 7.1 for both tstyche and eslint 😄

PR #3140, opened 2026-08-05
Migrate to typescript@6

mfedderly:
[Typescript@6](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/) is the big breaking release before we migrate to the Go implementation in v7. Luckily no code needed migration for this step, but we did need to make a few changes to the tests.

For the types.ts testing, we're already sending a bunch of command line args, so adding [--ignoreConfig](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#specifying-command-line-files-when-tsconfig.json-exists-is-now-an-error) feels like a reasonable minimal change.

This also introduces a named catalog to hold typescript@5 for the consumer-style testing. TypeScript 6 introduces a bunch of the breaking changes, and I'd like to leave these on the latest 5.x for now as a more representative test. I think if a user is already on typescript@6 it only gives them more compatibility.

As a follow up, we need to evaluate the duplication between tstyche testing and the types.ts files we've got. I filed a [ticket](https://github.com/Turfjs/turf/issues/3139) to follow up on this later.

### DIFF
--- a/tsconfig.types.shared.json
+++ b/tsconfig.types.shared.json
@@ -1,0 +1,6 @@
+{
+  "include": ["${configDir}/types.ts"],
+  "compilerOptions": {
+    "noEmit": true
+  }
+}
--- a/tsconfig.types.shared.json
+++ b/tsconfig.types.shared.json
@@ -1,6 +1,0 @@
-{
-  "include": ["${configDir}/types.ts"],
-  "compilerOptions": {
-    "noEmit": true
-  }
-}
--- a/.monorepolint.config.mjs
+++ b/.monorepolint.config.mjs
@@ -194,7 +194,7 @@ export default {
       options: {
         scripts: {
           "test:types":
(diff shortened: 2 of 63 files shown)


mfedderly:
In addition to this needing more work, it looks like we'll be blocked on the v7 upgrade until at least v7.1 so that the linting continues to work.

mfedderly:
I'm going to think about it for a bit longer, but my gut says that Turf can go to typescript@6 now, investigate project references for further speed-up, and then block on further TypeScript upgrades until 7.1, typescript-eslint, and tstyche are all ready.

tsconfig.types.shared.json
```diff
@@ -0,0 +1,6 @@
+{
+  "include": ["${configDir}/types.ts"],
+  "compilerOptions": {
+    "noEmit": true
+  }
+}
```
mfedderly: This needs to get deleted.

PR #3141, opened 2026-08-06
Upgrade eslint and related packages

mfedderly:
Dropping old node support allows us to pick up the newest eslint as well. Hopefully this gets us closer to being able to pick up the changes whenever eslint can be used with typescript@7.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -46,6 +46,7 @@ export default tsEslint.config(
       "no-redeclare": "off",
       "no-var": "off",
       "prefer-const": "off",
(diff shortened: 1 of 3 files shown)


eslint.config.mjs
```diff
@@ -46,6 +46,7 @@ export default tsEslint.config(
       "no-redeclare": "off",
       "no-var": "off",
       "prefer-const": "off",
+      "no-useless-assignment": "off",
```
mfedderly, line 49: This seems to be a new rule added to the defaults, there's a few changes that are useful but some of its suggestions in `meta` look tricky.

PR #3150, opened 2026-08-11
Condense last-checks into @turf/turf's test script directly

mfedderly:
I don't think there is any remaining reason why the @turf/turf tests need to be special cased into `last-checks`. The tests themselves do need to be ordered, but pnpm will take care of the package dependency ordering, and the script itself runs the two steps in order.

[Originally added](https://github.com/Turfjs/turf/commit/b3b5284882fc9b23c803da1c23217629b5ed6897#diff-cef2ddf2d0e8a2b54c7bc13bae1febdcaf52d4cef3704170b74efed73198a855R20-R23) a long time ago

### DIFF
--- a/packages/turf/test.ts
+++ b/packages/turf/test.ts
@@ -335,7 +335,7 @@ test("turf -- update to newer Typescript definitions", (t) => {
 
 // File Paths
 const testFilePath = path.join(__dirname, "test.example.js");
-const turfModulesPath = path.join(__dirname, "..", "turf-*", "index.js");
+const turfModulesPath = path.join(__dirname, "..", "turf-*", "index.ts");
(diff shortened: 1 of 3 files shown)


packages/turf/test.ts
```diff
@@ -335,7 +335,7 @@ test("turf -- update to newer Typescript definitions", (t) => {
 
 // File Paths
 const testFilePath = path.join(__dirname, "test.example.js");
-const turfModulesPath = path.join(__dirname, "..", "turf-*", "index.js");
+const turfModulesPath = path.join(__dirname, "..", "turf-*", "index.ts");
```
mfedderly, line 338: Perhaps we weren't picking up the examples anymore?

PR #3151, opened 2026-08-11
Enforce README.md changes are up-to-date in CI

mfedderly:
The pre-commit hook isn't always authoritative, so rework the CI to make sure we catch any stale READMEs going forward.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -4,7 +4,6 @@ packages:
 
 allowBuilds:
   esbuild: false
-  nx: false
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -4,7 +4,6 @@ packages:
 
 allowBuilds:
   esbuild: false
-  nx: false
```
mfedderly, line 7: Unrelated change: lerna is gone so we don't need an nx entry here any more

PR #3152, opened 2026-08-11
CI enforce readmes testing

mfedderly:
Just a test PR to make sure the errors are caught for #3151


mfedderly:
Builds are failing as expected here.

PR #3161, opened 2026-09-07
Move to pnpm/setup for github workflows

mfedderly:
The previous pnpm github runners were causing [issues](https://github.com/Turfjs/turf/actions/runs/34133610212/job/101779321291?pr=3160) now that newer pnpm 11's have been released.

It appears that they have released a new simpler action, so lets see if that one works out of the box or if I need to set a specific pnpm version in package.json's devEngines.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "version": ">=11.0.0 <12.0.0",
+      "version": "11.21.0",
       "onFail": "download"
     }
   }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -58,7 +58,7 @@
   "devEngines": {
     "packageManager": {
       "name": "pnpm",
-      "version": ">=11.0.0 <12.0.0",
+      "version": "11.21.0",
```
mfedderly, line 61: Pin the specific version so that CI keeps working when newer versions get released

PR #3162, opened 2026-09-07
break(meta) BREAK Improve propEach TypeScript signature

mfedderly:
- Tighten `geojson` parameter type signatures to only Feature and FeatureCollection
- Improve inference of `currentProperties` argument in callback
- Expose `false` return from callback to TypeScript
- Remove @ts-expect-errors
- Modernize the loop syntax

Perf changes are essentially neutral from this.

### DIFF
--- a/packages/turf-meta/types.ts
+++ b/packages/turf-meta/types.ts
@@ -1,4 +1,4 @@
-import { Point, LineString } from "geojson";
+import { Point, LineString, Polygon, Feature } from "geojson";
 import * as helpers from "@turf/helpers";
 import { featureCollection, point, lineString } from "@turf/helpers";
 import * as meta from "./index.js";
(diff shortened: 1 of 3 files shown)


packages/turf-meta/types.ts
```diff
@@ -128,12 +128,24 @@ meta.propReduce(geomCollection, (previous, prop) => prop);
 /**
  * meta.propEach
  */
-const propEachValue: void = meta.propEach(poly, (prop) => prop);
-propEach(features, (prop) => prop);
-meta.propEach(features, (prop) => prop);
-meta.propEach(poly, (prop, index) => prop);
-meta.propEach<{ bar: string }>(poly, (prop) => prop.bar);
```
mfedderly, line 135: This props cast actually became an error because it knew that `poly` did not have those props. I cast the argument instead so that we can still test the inference of `prop` in the callback.

